PR #32, opened 2015-06-05
Don't add namespaced models to dashboard manifest

c-lliope:
Currently, namespaced models are unsupported.
Eventually, this should change - it would require the dashboard manifest
to return something other than a list of symbols.


ACPK:
@graysonwright - When are namespaced models going to be supported?

PR #82, opened 2015-09-14
Add support for polymorphic relationships

c-lliope:
# Problem:

Until now, polymorphic relationships were being treated as standard `belongs_to` relationships. When Administrate inferred the incorrect class name for models, it resulted in a 500 error. This was most commonly seen on the form page.

In addition, when a polymorphic relationship has a non-existant class, the dashboard was throwing errors. This was the case on Upcase, where there are several dangling references from the `Classification` model to the removed `Article` model.

More information on https://trello.com/c/dMUeObDx
# Solution:
- Detect polymorphic belongs_to relationships in the dashboard generator
- Add `Administrate::Field::PolymorphicBelongsTo`
- Rescue errors when associated object lookup fails
- Add a stopgap "not supported" message to polymorphic forms for the moment. Trello card for completing it: https://trello.com/c/D9qnFx10

Related:
- Update Changelog

Tags: #ruby

### DIFF
--- a/administrate/NEWS
+++ b/administrate/NEWS
@@ -1,5 +1,8 @@
 New in 0.1.0:
 
+* Improvement: Don't display redundant attributes for belongs_to relationships,
+  such as `post_id` or `post_type`.
+* Bug Fix: Squash several 500 errors caused by polymorphic relationships.
 * Improvement: support models that have a custom `to_param` method.
 * UI: Improve formatting of strings on `index` and `show` pages
 * Improvement: Generate a single controller to serve all resources,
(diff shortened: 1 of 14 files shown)


To1ne:
Why the stopgap "not supported" message on forms?

ACPK:
@To1ne Same question!

nickcharlton:
Hi @To1ne, @ACPK!

Looking at the discussion, polymorphic forms will add a bunch of complexity which they were trying to leave out of this PR.

Please open an issue, or better, a PR and we can start working on it!

ACPK:
@nickcharlton - What's so complex about it?

PR #103, opened 2015-10-21
Prefix HTML title with page title

danbee:
We generate the HTML title from the page heading and the application
name which is pulled from `titles.application` in the localization
files.


guillosaracco:
I'd like to customize the HTML title. Is it possible?

nickcharlton:
@guillosaracco Please open a new issue if this is still a problem for you.

PR #125, opened 2015-11-02
Fix show page alignment issues

danbee:
This fixes the alignment issues on the show page by removing the space
between the comment and the field, and adding the hyphen to the ERB tag.

We might want to remove the space from all the other partials too for
consistency.

I've also added a margin to the bottom of the label tags which improves the spacing when fields are empty.

screen shot 2015-11-02 at 14 43 08

Fixed:

screen shot 2015-11-02 at 14 43 35

### DIFF
--- a/administrate/app/assets/stylesheets/administrate/components/_attributes.scss
+++ b/administrate/app/assets/stylesheets/administrate/components/_attributes.scss
@@ -3,6 +3,8 @@
   @include span-columns(2 of 12);
   margin-top: 0;
   text-align: right;
+  margin-bottom: 1.85em;
+  clear: left;
 }
 
 .attribute-data--string {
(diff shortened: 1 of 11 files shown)


graysonwright:
ah, I didn't take this into account when I added the comments.

After seeing this regression, I'm not sure I like having the UI affected by whitespace, especially when it's leading or trailing whitespace in a partial.

What do you think about deleting [this `whitespace: pre` rule](https://github.com/thoughtbot/administrate/blob/master/administrate/app/assets/stylesheets/administrate/components/_attributes.scss#L8-L10) and adding a `<span class="preserve-whitespace">` wrapper for the string partial?

danbee:
Ah, I didn't spot that rule and removing it also fixes the problem! What's the purpose of it?

graysonwright:
@danbee `text` database columns could have long text that spans multiple lines (as shown in #79). In that case, we want to preserve the line breaks.

Funnily, we've already fixed this bug once, in #93.

Maybe it makes sense to add a `preserve_whitespace` option to `Administrate::Field::String`, so devs can turn it on or off at a per-attribute level?

danbee:
We have a `text` field type that covers that, therefore I think this rule no longer makes sense for the `string` type. Perhaps the `<span class="preserve-whitespace" />` approach would work better?

danbee:
Replaced by #126

PR #127, opened 2015-11-02
Add button to clear search field

### LINKED ISSUE #121: Feature Request: Clearing Search
I'd love to be able to easily the clear search by (1) deleting the text or (2) clicking an (X). As of right now, a user needs to click the page link in the navbar or remove the params for the URL. The former is my preferred method, but let me know what you think.

graysonwright:
Closes #121.

## Problem:

Users cannot clear their search
unless they re-submit the search with an empty field,
or click on the resource link in the navigation bar.

## Solution:

Add a button for clearing out the search field
and going back to showing all resources.

## To Do

- [ ] Only show clear icon when there is text in the search bar
- [ ] CHANGELOG


danbee:
Can I suggest actually putting a button with the word "Clear" on it? Currently it looks a little like a close button.

danbee:
On reading back I realise my last comment sounded a bit rude. Wasn't meant to be!

graysonwright:
@danbee no, you're fine! Would it be better if the X only appeared when there's text in the box? The fading in might give the user context about its meaning.

I'll try it with "Clear" and add a screenshot as well.

danbee:
Search boxes on Mac's tend to use a cross in a little circle. I'm just knocking up an SVG now to see what it looks like.

danbee:
How about this?

screen shot 2015-11-02 at 17 26 44

danbee:
I think it does make sense for the clear button to only appear when there is text in the search box. This mirrors the way I've seen search boxes work elsewhere.

ACPK:
@danbee @graysonwright This is exactly what I was looking for!

graysonwright:
@danbee I like it - did you make that icon?

I've been making custom icons for administrate so far, to avoid copyright issues. If you got that from somewhere, we should check the license or I can just make one up.

danbee:
I made it myself. Did it by scaling down the existing one and putting a circle around it. I'm not 100% happy with it though so let me make a better one.

ACPK:
@graysonwright Will this be getting pushed?

zamith:
What about using the [input type search](http://caniuse.com/#feat=input-search)?

danbee:
`input type="search"` can be a pain to style.

zamith:
@danbee And do we really need to style that?

danbee:
@zamith To fit the current layout, yes.

danbee:
Also, the default styling is very inconsistent between different browsers. Safari, for instance, has it's own magic CSS incantations to style certain aspects of it.

ACPK:
What ever happened to this?

PR #144, opened 2015-11-04
Swap app and gem

danbee:
Moves the gem to the root directory and moves the example app to a
directory within the `/spec` directory. This makes it possible to refer
to the gem directly from git or Github.

## TODO

- [x] Restore Appraisal runs in the integration test suite
- [x] Figure out how to deploy demo app to Heroku


graysonwright:
Looks like the build isn't running the appraisal builds anymore. I'm gonna start a checklist in the description.

danbee:
Yeah, I'm sure there's a boatload of stuff that I've missed/broken.

graysonwright:
but the specs are passing now - that's cool. Before the rebase, I was getting failures for missing translations.

danbee:
Ah no, that was fixed separately too. The `i18n-tasks` configuration was in the wrong place and I had to rename the `translate` method in the `application_controller.rb` file as I think it was getting overridden by i18n's built in one.

danbee:
CircleCI is now running Appraisals.

graysonwright:
:tada:

graysonwright:
@danbee this looks great! Now, what color should we paint the bikeshed? `spec/demo` or `spec/example_app`?

I was initially thinking `demo` because I've been calling it "the demo site" in most of the copy/documentation. `example_app`'s a bit more explicit though, which is nice. Thoughts?

danbee:
A lot of the docs/articles I read called it `spec/dummy_app` which I didn't think was too appropriate given the nature of it. Hence why I chose `example_app`.

`spec/demo_app` could work though.

jayroh:
This is probably completely up to the project but the convention I've seen is to house this in `spec/dummy`. That being said. That pattern is usually in cases where it's used against the specs/tests exclusively. So ... ¯\\\_(ツ)\_/¯

JoelQ:
FWIW, I believe that the generator for Rails engines creates `spec/dummy` by default.

danbee:
@JoelQ Indeed it does! I created a test engine to compare to.

graysonwright:
yeah, I'm against `dummy`. I'd vote for `example_app` or `demo_app`, both are good.

danbee:
I wondered about whether it should even live in the `spec` directory to be honest.

JoelQ:
Should there be a separate demo app (possibly in it's own repo) that gets deployed to heroku and a `spec/dummy` to use in feature specs?

graysonwright:
@JoelQ yeah, we'll probably want to break that out eventually. Right now there's test coupling, so it would be a bit painful in the near term.

Rakefile
```diff
@@ -14,4 +29,11 @@ if defined? RSpec
   end
 end
 
+desc "Deploy the example app to Heroku"
+namespace :deploy do
+  task :example do
+    exec 'git push heroku `git subtree split --prefix spec/example_app`:master --force'
```
graysonwright: Looks like this `heroku` remote differs from what gets set up in `bin/setup`? Can we change `bin/setup` to add contributors to the correct heroku app, with a remote name that matches this?

Also, I've been preferring the `$(shell_command)` format, because:
1. It differentiates the start and end of a command
2. It's copy-and-pastable into Slack code snippets :smile:

graysonwright:
@danbee one inline comment, but then this looks good to me!

Were there any outstanding tasks I forgot about?

Rakefile
danbee: Ah, then I guess we need two deploy tasks?

danbee:
I don't think so!

graysonwright:
Alright. I'm going to add a CHANGELOG entry (the only user-facing change is the ability to point bundler to the gem on github) and merge it. Thanks! :tada:

graysonwright:
I just ran `gem build administrate.gemspec` and then tested the built gem locally - everything still works!

danbee:
I'm making a note here: Huge success!

PR #148, opened 2015-11-05
Add french i18n

JulienItard:
Hi,

sorry for my failed last PR. Here is the new one.

Julien

### DIFF
--- a/administrate/config/locales/administrate.fr.yml
+++ b/administrate/config/locales/administrate.fr.yml
@@ -1,0 +1,23 @@
+---
+fr:
+  administrate:
+    actions:
+      confirm: Êtes-vous sûr ?
+      destroy: Supprimer
+      edit: Modifier
+      show: Voir
+    controller:
+      create:
+        success: "%{resource} a été correctement créé(e)."
+      destroy:
+        success: "%{resource} a été correctement supprimé(e)."
+      update:
+        success: "%{resource} a été correctement modifié(e)."
+    fields:
+      has_many:
+        more: "%{count} sur %{total_count}"
+        none: Aucun
+      polymorphic:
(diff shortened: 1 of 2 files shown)


administrate/config/locales/administrate.fr.yml
```diff
@@ -0,0 +1,23 @@
+---
+fr:
+  administrate:
+    actions:
+      confirm: Êtes-vous sûr ?
+      destroy: Supprimer
+      edit: Modifier
+      show: Voir
```
JoelQ: Thoughts on `Détails` instead of `Voir`?

JulienItard: Good idea :+1:

graysonwright:
(for reference, previous PR is #139 )

graysonwright:
@JulienItard if it's easier, we can take care of squashing the commits. It'll probably make it easier for all the contributors.

administrate/config/locales/administrate.fr.yml
graysonwright: @JulienItard once you make this change, we can merge this in. Thanks!

graysonwright: Actually, this translation file is misleading... I think we actually need the verb `to show`, because we use it alongside the resource name like `Show Product`.

Ideally, this translation would have interpolation, like: `show: Voir %{resource_name}`, but that will need to happen in a separate PR.

graysonwright:
Merged in as cf4078b. Thanks!

JulienItard:
Great ! Do you have other Thoughtbot stuff that need french translation ?

graysonwright:
@JulienItard I don't know of anything off the top of my head. I'll reach out if anything comes up. Thanks!

PR #152, opened 2015-11-05
Polish translation file

pkrefta:
As title says - my small contribution :smile:


pusewicz:
@tradziej Here's an alternative PR #171. I believe it translates better.

PR #168, opened 2015-11-06
Use local version of administrate in development

graysonwright:
## Problem:

When a contributor tries to run the example app by `cd`ing into
`spec/example_app`, running `bundle install`, and running the server,
the Gemfile downloads a separate copy of Administrate from the master
branch on Github. This prevents the contributor from testing out any
local changes to the Administrate source.

We cannot simply reference the relative path to the local Administrate
source, because that would make Heroku deploys fail.

## Solution:

Check the Rack environment to determine which source Bundler should use
for Administrate.

Contributors can now run `bundle install` and `foreman start` from their
`spec/example_app` directory to test local changes to Administrate.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,14 +3,8 @@
 *.swo
 *.swp
 /.bundle
-/.env
-/.foreman
 /coverage/*
 /db/*.sqlite3
 /log/*
-/spec/example_app/log/*
-/spec/example_app/public/system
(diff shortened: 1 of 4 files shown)


mgrachev:
:+1:

graysonwright:
I'm going to close this in favor of #172 - that PR renders this change unnecessary.

PR #169, opened 2015-11-06
Add basic validations for Product slugs

### LINKED ISSUE #162: Products index page is broken
I just randomly click on http://administrate-prototype.herokuapp.com/admin/products and it is broken. Can I fix this error and submit a patch ?   
Thanks

graysonwright:
Closes #162 

## Problem:

On the demo site, it's rather simple for users to craft product names
that don't translate into valid slugs.

When Rails tries to render links to these products, it throws a 500
error because it cannot generate a correct URL.

## Solution:

Add validators to ensure that all products have a valid and unique slug.

### DIFF
--- a/spec/example_app/app/models/product.rb
+++ b/spec/example_app/app/models/product.rb
@@ -3,6 +3,8 @@ class Product < ActiveRecord::Base
   validates :image_url, presence: true
   validates :name, presence: true
   validates :price, presence: true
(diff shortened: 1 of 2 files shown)


graysonwright:
Merging without review because:

1. Doesn't affect the library
2. Fairly straightforward
3. Fixes a security hole on the demo site

PR #170, opened 2015-11-06
Use `@example.com` for all sample email addresses

### LINKED ISSUE #149: Please change the example e-mail addresses to end in @example.com
Your demo has realistic looking e-mail addresses. A lot of those probably are real. Those will get picked up by crawlers and receive spam.

Please change the demo addresses to end in @example.com, that way nobody receives those spam mails.

P.S.: That's also a good idea for all local testing you ever do. If you ever accidentally configure your dev setup with the real e-mail server, nothing bad happens.

graysonwright:
Fixes #149

## Problem:

The demo app uses realistic email addresses, generated by [Faker].
If any of the generated emails are real,
they may be picked up by web crawlers and receive spam.

Big thanks to @lorenzk for pointing out this downside of realistic
emails.

[Faker]: https://github.com/stympy/faker

## Solution:

Use [Faker]'s `#safe_email` method, which uses `example.com` for all
email domains.

`example.com` is a reserved domain, and sending emails to it is a no-op.

### DIFF
--- a/spec/example_app/db/seeds.rb
+++ b/spec/example_app/db/seeds.rb
@@ -15,7 +15,7 @@ Product.destroy_all
   name = "#{Faker::Name.first_name} #{Faker::Name.last_name}"
   Customer.create(
     name: name,
-    email: Faker::Internet.free_email(name),
+    email: Faker::Internet.safe_email(name),
   )
 end
 


graysonwright:
Merging this in without review b/c it affects data on the demo app, and it's a straightforward change.

PR #172, opened 2015-11-06
Improve heroku deploy workflow

graysonwright:
Problem:
--------

With the git subtree strategy adopted in #144,
there are several unconventional things
surrounding the heroku deploy process:

- Pushing to heroku requires a `git subtree` command
- We need special logic in the Gemfile in order to run the demo app
  locally with our local version of Administrate.
- `rake` and `rails` commands need to be run from the
  `spec/example_app` subdirectory.

In addition, there is a dependency problem between the example app
and the master branch of administrate on github.
The example app needs to pull the master branch
as a dependency on heroku,
but the `Gemfile.lock` locks it in to a specific version.

It is impossible to point the `Gemfile.lock`
at the repo's current commit,
because that would require a new commit
before we could deploy to Heroku.

The result would be that the branch on Heroku would always differ
from the branch on GitHub by at least one commit,
and we would need to develop special tooling around pushing to heroku.

Solution:
---------

Make the `rails`, `rake`, and `bundle exec` commands work
from the root directory of the repository.
This involved:

- Move the `config.ru` script to the root of the repo
- Update `rake` and `rails` binstubs to reference nested rails app
- Move `unicorn.rb` (required by Procfile) to top-level `config` dir
- Combine both Gemfiles into a single one, at the root of the repo
- Remove spring for simplicity

I tested the following tasks.
All work from the root directory of the repo:

- deploy to heroku: `git push heroku master`
- run server locally: `rails server` or `foreman start`
- seed demo database: `rake db:seed`
- migrate demo database: `rake db:migrate`
- reset demo database: `rake db:drop db:create`
- run rails generators:
  - `rails generate administrate:field FooBar`
  - I was pleasantly surprised that this placed the files
    in the correct `spec/example_app` subdirectory
- access demo app console:: `rails console`
- view demo app routes: `rake routes`
- appraisal test suite: `bundle exec appraisal rake`

References:
-----------

http://stackoverflow.com/questions/3081699/deploy-a-subdirectory-to-heroku


spec/example_app/config/boot.rb
```diff
@@ -1,3 +1,2 @@
-ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../../Gemfile', __FILE__)
-
-require 'bundler/setup' # Set up gems listed in the Gemfile.
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../../../../Gemfile", __FILE__)
```
graysonwright: This is the important line - pointing the example app at the root `Gemfile`.

spec/example_app/Rakefile
```diff
@@ -1,9 +0,0 @@
-# Add your own tasks in files placed in lib/tasks ending in .rake,
-# for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
-
-require File.expand_path('../config/application', __FILE__)
-
-Rails.application.load_tasks
-task(:default).clear
-
-task default: "bundler:audit"
```
graysonwright: I removed this because I can't think of a reason we would need it... thoughts?

danbee:
LGTM! :+1:

PR #174, opened 2015-11-06
Be more explicit about Administrate dependencies

### LINKED ISSUE #173: Implicit dependencies on the Rails Asset Pipeline
When I added Administrate to my Rails app, things were broken out of the box. My app is designed to be an API backend only, and didn't include Sprockets or several other necessary Rails Asset Pipeline gems.

I'm not a Rails expert, so I can't speak to how this issue should be handled in general. But you might find it worthwhile to make these dependencies explicit within Administrate.

For reference, I was able to load the Administrate dashboard after adding the following to my `Gemfile`

```
gem 'sprockets-rails'
gem 'sass-rails'
gem 'uglifier'
gem 'coffee-rails'
gem 'jquery-rails'
```
(EDIT: this list is neither minimal nor exhaustive-- I just copy/pasted from the Rails Guides. As grayson mentioned below, the `'coffee-rails'` gem shouldn't be needed.)

and uncommenting the line

```
gem 'sprockets-rails'
```

in `/config/application.rb`.

graysonwright:
Fixes #173

## Problem:

Administrate depends on the asset pipeline
and several gems that are not defined in
`administrate.gemspec`.

When host applications don't have these implicit dependencies installed
(for example, if their app is a simple API server),
they'll run into errors out of the box.

## Solution:

- Move more gems into `administrate.gemspec` as explicit dependencies.
- Remove unnecessary gems from the Gemfile to cut back the noise.


tute:
:+1: :clap:

graysonwright:
Although we haven't necessarily resolved #173, this is now working for me in a plain `rails-api` app.

I think this is good enough support to release a new version, so I'll push ahead with it and try to resolve other dependency conflicts on a case-by-case basis.

PR #191, opened 2015-11-08
Improve default resource display

### LINKED ISSUE #129: Don't use `#to_s` to display models in Administrate.
Just a minor note to request the documentation of the use of the ```to_s``` method for better representing objects across the various Administrate dashboards. E.g. ```has_many``` select lists and ```belongs_to``` on index lists. This might catch out some newer Rails/Ruby users if not explicitly mentioned...

Something along the lines of:

> ```to_s``` is the standard Ruby method for converting an object to a string. You can define ```to_s``` on your model class when you want a custom string representation.

Example:

```ruby
class Tag < ActiveRecord::Base
  def to_s
    name
  end
end
```

Thus instead of displaying ```#<Tag:0x007fad07b7ad98>```, we can display the actual name of the tag. 

Thoughts?

graysonwright:
Closes #129

## Problem:

Administrate relies on `#to_s`
to display resources throughout the system.
In order to get Administrate working correctly,
developers must define `#to_s` on all models
that will be displayed in the admin dashboard.

This process is not documented,
can be confusing,
and gets in the way of a zero-configuration dashboard.

## Solution:

Add `Administrate::BaseDashboard#display_resource`,
which takes a resource and returns a sensible string representation.
Devs can overwrite this method on a per-dashboard basis
to customize how their resources are displayed.

In order to document this method,
we generate comments in each dashboard class that follow the format:

```ruby
  # Overwrite this method to customize how line items are displayed
  # across all pages of the admin dashboard.
  #
  # def display_resource(line_item)
  #   "LineItem ##{line_item.id}"
  # end
```

## TODO

- [x] Add comments to generated dashboards
- [x] CHANGELOG

### DIFF
--- a/app/views/fields/belongs_to/_form.html.erb
+++ b/app/views/fields/belongs_to/_form.html.erb
@@ -17,9 +17,6 @@ that displays all possible records to associate with.
 %>
 
 <%= f.label field.permitted_attribute %>
-<%= f.collection_select(
-  field.permitted_attribute,
-  field.candidate_records,
-  :id,
-  :to_s,
-) %>
+<%= f.select(field.permitted_attribute) do %>
+  <%= options_for_select(field.associated_resource_options) %>
+<% end %>
--- a/app/views/fields/belongs_to/_show.html.erb
+++ b/app/views/fields/belongs_to/_show.html.erb
@@ -16,5 +16,8 @@ By default, the relationship is rendered as a link to the associated object.
 %>
 
 <% if field.data %>
-  <%= link_to field.data, [Administrate::NAMESPACE, field.data] %>
+  <%= link_to(
+    field.display_associated_resource,
+    [Administrate::NAMESPACE, field.data],
+  ) %>
 <% end %>
--- a/app/views/fields/has_many/_form.html.erb
+++ b/app/views/fields/has_many/_form.html.erb
@@ -21,12 +21,6 @@ and is augmented with [Selectize].
 
 <%= f.label field.attribute_key %>
 
-<%= f.collection_select(
-  field.attribute_key,
-  field.candidate_records,
-  :id,
-  :to_s,
-  {},
-  { multiple: true },
-)
-%>
+<%= f.select(field.attribute_key, nil, {}, multiple: true) do %>
+  <%= options_for_select(field.associated_resource_options) %>
+<% end %>
--- a/app/views/fields/polymorphic/_index.html.erb
+++ b/app/views/fields/polymorphic/_index.html.erb
@@ -18,7 +18,7 @@ By default, the relationship is rendered as a link to the associated object.
 
 <% if field.data %>
   <%= link_to(
-    field.data.to_s,
+    field.display_associated_resource,
     polymorphic_path([:admin, field.data])
   ) %>
 <% end %>
--- a/app/views/fields/polymorphic/_show.html.erb
+++ b/app/views/fields/polymorphic/_show.html.erb
@@ -18,7 +18,7 @@ By default, the relationship is rendered as a link to the associated object.
 
 <% if field.data %>
   <%= link_to(
-    field.data.to_s,
-    polymorphic_path([:admin, field.data])
+    field.display_associated_resource,
+    [:admin, field.data],
   ) %>
 <% end %>
--- a/lib/generators/administrate/dashboard/templates/dashboard.rb.erb
+++ b/lib/generators/administrate/dashboard/templates/dashboard.rb.erb
@@ -40,4 +40,11 @@ class <%= class_name %>Dashboard < Administrate::BaseDashboard
   end.join("\n")
 %>
   ]
+
+  # Overwrite this method to customize how <%= file_name.pluralize.humanize.downcase %> are displayed
+  # across all pages of the admin dashboard.
+  #
+  # def display_resource(<%= file_name %>)
+  #   "<%= class_name %> ##{<%= file_name %>.id}"
+  # end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@
 
 ### Upcoming Release
 
+* [#191] [CHANGE] Improve API for specifying how resources are displayed
+  across the dashboard.
+  * Models are now displayed with a sensible default - (e.g. "User #2")
+  * Users can define `ModelDashboard#display_resource(resource)` for custom
+    display behavior
+  * Users who have generated views for the following field types
+    may need to update them to take advantage of the new API:
+    * HasOne
+    * HasMany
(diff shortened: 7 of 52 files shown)


maxkwallace:
Looks good to me. Using inheritance seems worthwhile in this case... after adding `to_s` methods to my models, I found that most of them followed a common format.

mcmire:
This seems reasonable to me. I've kind of thought overriding `to_s` felt a little dirty.

app/views/fields/belongs_to/_show.html.erb
```diff
@@ -16,5 +16,5 @@ By default, the relationship is rendered as a link to the associated object.
 %>
 
 <% if field.data %>
-  <%= link_to field.data, [Administrate::NAMESPACE, field.data] %>
+  <%= link_to field.associated_resource, [Administrate::NAMESPACE, field.data] %>
```
JoelQ: Is this change part of removing the dependency on `to_s` ?

JoelQ:
:+1:

app/views/fields/belongs_to/_show.html.erb
graysonwright: yep - Rails was taking `field.data` and turning it into `field.data.to_s`, in order to get the link text.

I'm not sure `associated_resource` is super clear. I might try changing it to `display_resource` or something.

app/views/fields/belongs_to/_form.html.erb
```diff
@@ -17,9 +17,6 @@ that displays all possible records to associate with.
...
 
 <%= f.label field.permitted_attribute %>
-<%= f.collection_select(
-  field.permitted_attribute,
-  field.candidate_records,
-  :id,
-  :to_s,
-) %>
+<%= f.select(field.permitted_attribute) do %>
+  <%= options_for_select(field.candidate_options) %>
```
graysonwright: `candidate_resource_options`, maybe?

app/views/fields/has_many/_form.html.erb
```diff
@@ -21,12 +21,6 @@ and is augmented with [Selectize].
...
-  field.attribute_key,
-  field.candidate_records,
-  :id,
-  :to_s,
-  {},
-  { multiple: true },
-)
-%>
+<%= f.select(field.attribute_key, nil, {}, multiple: true) do %>
+  <%= options_for_select(field.candidate_options) %>
```
graysonwright: same - `candidate_resource_options`?

app/views/fields/polymorphic/_index.html.erb
```diff
@@ -18,7 +18,7 @@ By default, the relationship is rendered as a link to the associated object.
 
 <% if field.data %>
   <%= link_to(
-    field.data.to_s,
+    field.associated_resource,
```
graysonwright: `associated_resource_string`, maybe?

app/views/fields/belongs_to/_show.html.erb
mcmire: If I were to describe this in English, it'd be "the human-readable representation of the resource that this field points to".

So `display_associated_resource` is *kind of* saying the same thing. But I can't think of a better name, honestly.

It would be nice if `associated_resource` were a decorator around a record instead of a record itself. Actually, every resource that needs to be `display`ed could be decorated, and then you could have your `to_s` method again, only inside the decorator class. I guess it wouldn't be easy for a user to override that, though.

app/views/fields/belongs_to/_form.html.erb
mcmire: I feel like the word `candidate` is redundant here, given that you already have `option` in the same word. In other words, something is either a `candidate` or an `option`, but being both looks strange.

What about `display_associated_resources`?

app/views/fields/polymorphic/_index.html.erb
mcmire: Should this also be `display_associated_resource`?

app/views/fields/belongs_to/_show.html.erb
mcmire: Which one looks better:

* `display_associated_resource`
* `displayed_associated_resource`

app/views/fields/polymorphic/_index.html.erb
graysonwright: yup, sounds good.

mcmire: Or `displayed_associated_resource`, in reference to the above comment. Not sure which one is better.

app/views/fields/belongs_to/_show.html.erb
graysonwright: I like `displayed_...`, because it's a noun instead of a verb. The method is returning a String, instead of taking action to display the string itself.

app/views/fields/belongs_to/_form.html.erb
graysonwright: yep - I was using `option` in the sense of the `<option>` tag for `<select>` fields, but I see what you mean. I think I like `associated_resource_options`.

app/views/fields/polymorphic/_show.html.erb
```diff
@@ -18,7 +18,7 @@ By default, the relationship is rendered as a link to the associated object.
 
 <% if field.data %>
   <%= link_to(
-    field.data.to_s,
-    polymorphic_path([:admin, field.data])
+    field.displayed_associated_resource,
```
graysonwright: After seeing this in practice, I think I prefer `display_associated_resource`. There's something about `displayed` that doesn't read quite right.

lib/generators/administrate/dashboard/templates/dashboard.rb.erb
```diff
@@ -40,4 +40,11 @@ class <%= class_name %>Dashboard < Administrate::BaseDashboard
   end.join("\n")
 %>
   ]
+
+  # Overwrite this method to customize how <%= file_name.pluralize.humanize.downcase %> are displayed
```
graysonwright: This shows up as:

> Overwrite this method to customize how widgets are displayed

tute:
:tada: :smiley:

wspurgin:
@graysonwright Any chance this change can be added to documentation somewhere? I recently updated an old system's administrate, and it took me a *long* time to find this change.

graysonwright:
@wspurgin - the change is listed in [the 0.1.1 release notes](https://github.com/thoughtbot/administrate/releases/tag/v0.1.1), but I guess it's not very visible. I'll change up the format of release notes, and add a link to them from the README.

Thanks for the feedback!

graysonwright:
See https://github.com/thoughtbot/administrate/pull/509.

PR #193, opened 2015-11-08
Assume unknown db column types are un-searchable

### LINKED ISSUE #178: Devise inet fields may be treated as Field::String in Administrate, which results in SQL errors when searching
I'm seeing an `ActiveRecord::StatementInvalid` error when searching over my Users in Administrate.

Postgres gives the following error:
```
PG::UndefinedFunction: ERROR:  function lower(inet) does not exist
```

The SQL query that's erroring looks like:
```
SELECT  "users".* FROM "users" WHERE (lower(email) LIKE '%george%' OR lower(current_sign_in_ip) LIKE '%george%' ... ) LIMIT 20 OFFSET 0
```

As far as I can tell, this is because the `current_sign_in_ip` attribute (managed by Devise) is treated as a String in Administrate:

```
    current_sign_in_ip: Field::String,
```

whereas it's actually an inet in Postgres:

```
t.inet     "current_sign_in_ip"
```

so attempting to use `lower` on it fails.

This isn't an urgent issue-- simply deleting the offending attributes from `ATTRIBUTE_TYPES` in my `user_dashboard.rb` file allows me to search just fine. The only downside is that I can't see the `current_sign_in_ip` attribute in Administrate, which really doesn't matter.  But I thought it was worth reporting.

graysonwright:
Closes #178

## Problem:

If users have a database set up with unknown column types,
Administrate treats those columns as a string.
This can cause bugs when users try to search,
because the unknown database column might not support
the string search operator.

## Solution:

- Add a `searchable` option to all field types,
  which specifies whether it is one of
  the resource's plaintext search fields.
- For unknown db column types, give the field
  a default option of `searchable: false`.

### DIFF
--- a/spec/generators/dashboard_generator_spec.rb
+++ b/spec/generators/dashboard_generator_spec.rb
@@ -36,12 +36,44 @@ describe Administrate::Generators::DashboardGenerator, :generator do
       end
 
       it "includes user-defined database columns" do
-        dashboard = file("app/dashboards/customer_dashboard.rb")
+        begin
+          ActiveRecord::Schema.define do
+            create_table(:foos) { |t| t.string :name }
+          end
 
-        run_generator ["customer"]
+          class Foo < ActiveRecord::Base
+            reset_column_information
+          end
 
-        expect(dashboard).to contain("name: Field::String,")
-        expect(dashboard).to contain("email: Field::String,")
+          run_generator ["foo"]
+          load file("app/dashboards/foo_dashboard.rb")
+          attrs = FooDashboard::ATTRIBUTE_TYPES
+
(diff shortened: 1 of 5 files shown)


spec/generators/dashboard_generator_spec.rb
```diff
@@ -36,12 +36,44 @@ class Foo < ActiveRecord::Base
       end
 
       it "includes user-defined database columns" do
```
graysonwright: Updated this spec to test that the correct classes & constants actually get defined, instead of just looking for the correct strings in the generated file.

why-el:
Yep, we faced this just a moment ago. Good work.

tute:
This LGTM. :rocket:

PR #194, opened 2015-11-08
[BUGFIX] existing date values were being cleared

### LINKED ISSUE #135: DateTime fields blank out for values that are already set
Fields set as `Field::DateTime` seem to be automatically emptied out (rather than retaining their on-load value). Assuming this has something to do with the datepicker js, as it flashes the value for a moment presumably before the javascript executes.

graysonwright:
Fixes #135

## Problem:

When a user edits a resource
with a `Date`, `Time`, or `DateTime` attribute,
and the attribute already has a value,
the date-time picker clears out the value.

Unless the user re-fills the field,
the value will be reset and will cause an error
when they hit "submit".

## Solution:

Update the `datetimepicker` date format
to match the Rails default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
 * [#148] [FEATURE] Translation: French
 * [#147] [FEATURE] Translation: German
 * [#193] [BUGFIX] Don't assume that unrecognized db column types are searchable
+* [#194] [BUGFIX] Don't clear out datetime values in form fields
 * [#156] [COMPAT] Include missing `sass-rails` dependency in gemspec
 * [COMPAT] Update repository structure so Bundler can pull the gem from github.
   (e.g. `gem "administrate", github: "thoughtbot/administrate"`)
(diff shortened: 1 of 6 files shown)


graysonwright:
The build passes locally for me - will have to debug later this week.

JoelQ:
LGTM once CI passes :package:

jayroh:
Looks good here too, man.

graysonwright:
Looks like this is an intermittently failing test. I ran `rspec --bisect -s 39930`, and got this minimal reproduction command:

```
rspec './spec/features/orders_form_spec.rb[1:2:1,1:3:1,1:3:2]' -s 39930
```

graysonwright:
damn. it looks like that "minimal reproduction" passes. Will try to narrow it down again.

graysonwright:
It looks like it's not dependent on test order or random seed - just luck. pretty strange... I wonder if master has the same problems, and we've just been lucky there so far.

graysonwright:
Switching to poltergeist seems to do the trick.

PR #198, opened 2015-11-09
Fix menu style for the sidebar

mgrachev:
Hello! 

In the sidebar for the last item in the list, there is no retreat from the bottom. Because of this, the last item in the list is sometimes not visible:

before

I modified styles to fix it:

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 ### Upcoming Release
 
+* [UI] Fix menu style for the sidebar
 * [#161] [FEATURE] Translation: Mandarin Chinese
 * [#142] [FEATURE] Translation: Brazilian Portuguese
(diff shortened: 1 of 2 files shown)


graysonwright:
@mgrachev thanks!

One thing I'm concerned about is that this only solves the problem for the specific number of models that you have in your app. If you had one more model, it would still fall off the end.

Have you tried scrolling in the sidebar? It scrolls separately from the rest of the page's content, so you should be able to access all of your models even if you have a bunch. Let me know if that's not working.

mgrachev:
Scroll to the end of the sidebar does not help, status bar hides "DotaPlayer"  - it can be seen in the screenshot:

scroll

zenati:
:+1:

graysonwright:
Merged in as 1bf79d1. Thanks!

PR #203, opened 2015-11-09
Fixed pagination styles if kaminari already used in Rails-app

mgrachev:
Hi.

If in the Rails-app already used gem `kaminari` and uses the changed views from `app/views/kaminari`, `administrate` incorrectly displays pagination:

before

After my changes:

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 
 ### Upcoming Release
 
+* [UI] Fixed pagination styles if kaminari already used in Rails-app
 * [#161] [FEATURE] Translation: Mandarin Chinese
 * [#142] [FEATURE] Translation: Brazilian Portuguese
 * [#171] [FEATURE] Translation: Polish
--- a/app/views/administrate/application/index.html.erb
+++ b/app/views/administrate/application/index.html.erb
@@ -52,4 +52,4 @@ It renders the `_table` partial to display details about the resources.
 
 <%= render "collection", collection_presenter: page, resources: resources %>
 
-<%= paginate resources %>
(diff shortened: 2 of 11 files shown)


graysonwright:
@mgrachev cool, thanks!

I'm not sure I understand what caused the original problem - you generated the kaminari views into the host application, and customized them?

I'm not sure how I feel about copying Kaminari's default views into Administrate, especially if it's only to override generated views in the host application.

It might make more sense for users who have customized Kaminari to generate the views again into their own application, under `app/views/admin/kaminari`.

Thoughts? Also, has anyone else experienced this? If it's a common problem, that might change my willingness to do this.

mgrachev:
> I'm not sure I understand what caused the original problem - you generated the kaminari views into the host application, and customized them?

Yes, it is.

I checked, this problem can be solved in another way. Necessary generate a default views of Kaminari in `app/views/administrate/kaminari`, then Administrate automatically pick up these changes because of this line of code:

```erb
<%= paginate resources, views_prefix: 'administrate' %>
```

In this case, no longer need to add the default view in the Administrate. But it is necessary to describe this process in the documentation.

graysonwright:
@mgrachev great, thanks! We'll set up a FAQ, and include this in it.

jalada:
I'm confused... the `views_prefix` hasn't been merged in, so this isn't fixed; there's still a problem if you have your own Kaminari styles in your app. What's the proposed solution?

jmeiss:
I'm also looking for a solution. Is there any?

zinovyev:
Why haven't you merged in your codebase yet? This fix helped me a lot)
@mgrachev thank you!

PR #206, opened 2015-11-10
Checkboxes should not appear to be centrally-aligned

murdoch:
## Problem

All form input elements have the following SCSS applied:

``` scss
  input,
  select,
  .selectize-control {
    @include span-columns(4 of 12);
  }

```
This causes check-boxes to appear in the middle of their columns, see the pic:

screenshot from 2015-11-10 03-37-16

## Solution

The world is just not ready for centrally-aligned check-boxes, so I have added `width: auto` to check-box inputs to ensure that they appear left aligned, see the next pic:

screenshot from 2015-11-10 03-53-10

I have read the contributing guidelines, and can't really think of a way to test this, besides visual inspection.

### DIFF
--- a/app/assets/stylesheets/administrate/components/_form.scss
+++ b/app/assets/stylesheets/administrate/components/_form.scss
@@ -36,5 +36,6 @@
   [type="checkbox"],
   [type="radio"] {
     margin: $small-spacing 0 ($base-spacing + $form-input-padding);
+    width: auto;
   }
 }


graysonwright:
@murdoch Thanks! It looks like Chrome (my primary browser) fixed this automatically, so I didn't notice it.

graysonwright:
Merged in as 2e5981c. Thanks!

PR #208, opened 2015-11-10
Update the Ruby version to 2.2.3

5minpause:
The project doesn't use the latest available Ruby version.
I think it benefits the project to use the latest Ruby version.

While updating the Gemfile and after running `appraisal install` the appraisal Gemfile for sass (`gemfiles/sass_3_4.gemfile`) was updated as well. I have never worked with Appraisal before so I have trouble judging this change. Please let me know if this is the correct behaviour.


gemfiles/sass_3_4.gemfile
```diff
@@ -16,7 +18,6 @@ gem "pg"
 gem "rails", "4.2.2"
 gem "recipient_interceptor"
 gem "refills"
-gem "sass-rails", "~> 5.0"
```
tute: Why is this gone?

5minpause: I was hoping someone could tell me. I ran the Appraisal installer and this was the result. I didn't edit it by hand.

graysonwright: yep - I think this appraisal gemfile was out of date with the one on master. It should be caught up as of #224.

Rebase if you want to, otherwise it'll come out in the wash.

circle.yml
```diff
@@ -8,3 +8,6 @@ test:
   override:
     - bundle exec rake
     - bundle exec appraisal rake
+machine:
+  ruby:
+    version: 2.2.3
```
graysonwright: is this necessary? I would have assumed CircleCI would auto-detect the ruby version based on `.ruby-version`.

5minpause:
Alright rebasing on master and just doing it again fixed both issues. This should be good to go now.
Thanks for the feedback @tute, @graysonwright

graysonwright:
Looks good!

PR #211, opened 2015-11-10
Warn about invalid tables that we skip

graysonwright:
Warnings look like:

```
WARNING: Unable to generate a dashboard for Delayed::Backend::ActiveRecord::Job.
         Administrate does not yet support namespaced models.
WARNING: Unable to generate a dashboard for Foobar.
```

### DIFF
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -30,6 +30,18 @@ module Administrate
         end
       end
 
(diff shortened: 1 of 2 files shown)


graysonwright:
This commit was originally part of #124. A rebasing mishap removed the commit from that PR, so I'm merging it in here. Since it's already gotten :+1: as part of the previous PR, I'm not waiting for a review here.

PR #214, opened 2015-11-11
Prevent white-space from appearing above header button when title is unusually long

murdoch:
When I visit the edit page for a resource with a long title I notice some unsightly white-space above the show page link that appears next to the h1 tag. See image:

flex-end

This whitespace is caused by the following CSS: 

``` css
.header {
  align-items: flex-end;
}
```

To get rid of this white-space we can change flex-end to flex-start or baseline, see below:

baseline

I will understand if you prefer flex-start over baseline.

### DIFF
--- a/app/assets/stylesheets/administrate/components/_header.scss
+++ b/app/assets/stylesheets/administrate/components/_header.scss
@@ -1,5 +1,5 @@
 .header {
-  align-items: flex-end;
+  align-items: baseline;
   display: flex;
   justify-content: space-between;
   margin-bottom: $base-spacing * 2;


graysonwright:
@murdoch this looks great! Can you do me a favor and rebase on master? We've added some visual regression tests to make sure CSS changes don't accidentally break things. I'm sure you're fine here, but just want to get into the habit.

After that, should be good to merge!

graysonwright:
Also, I think `flex-start` might be better here... it's a tough one.

@smharley and @devonharless, thoughts on `align-items: flex-start` vs `align-items: baseline`?

smharley:
`baseline` is lining up the bottom of the text, which you can see in the screenshot. I'd guess that `flex-start` would move the button up a little more. Either is probably fine, whichever you think looks better. 

My only other feedback is that it would be nice to have some `padding-right` on that `h1` so they don't bump into each other

graysonwright:
Merged in as 67d8704. Thanks!

PR #221, opened 2015-11-11
Fixes the paths for generated views

### LINKED ISSUE #202: Generating views creates wrong file paths
I ran `rails g administrate:views` and I got:
```
create  app/views/admin/applications/index.html.erb
...
```
The paths are wrong. It should be:
`app/views/administrate/application/index.html.erb` etc. 

The docs state the correct paths:
http://administrate-docs.herokuapp.com/customizing_page_views

5minpause:
Fixes #202.
 
## Problem:

The documentation correctly states that view generated via `rails g
administrate:views` should live under
`app/views/administrate/application/`. The generatator puts them under
`app/views/admin/applications/` though.

## Solution:

The methods that create the paths need to be adapted.

## Nota bene:

While running the tests, the first arg for the generator is empty. When
the generator is run inside a project, the first arg is `application`.
This leads to the view paths having `applications` instead of
`application`. This commit fixes this as well.

### DIFF
--- a/lib/administrate/view_generator.rb
+++ b/lib/administrate/view_generator.rb
@@ -21,7 +21,11 @@ module Administrate
     end
 
     def resource_path
-      args.first.try(:underscore).try(:pluralize) || "application"
+      if args.none? || args.first == "application"
+        "application"
+      else
+        args.first.underscore.pluralize
+      end
     end
   end
 end


5minpause:
How would you like me to handle the long lines?

graysonwright:
Based on conversation on #202, I think we can switch back to `admin/` instead of `administrate/`. That should get rid of most of the line length warnings.

@5minpause do you want to update this to revert to `admin/`? I can pick this up in another PR today if that's easier.

5minpause:
I gladly update this pull request. It's evening here, so you'll get this tomorrow.

5minpause:
@graysonwright I updated the commit and reverted back to `/admin`. The thing about `/applications`/`/application` remains.

lib/administrate/view_generator.rb
```diff
@@ -21,7 +21,8 @@ def copy_resource_template(template_name)
     end
 
     def resource_path
-      args.first.try(:underscore).try(:pluralize) || "application"
+      return "application" if args.none? || args.first == "application"
+      args.first.try(:underscore).try(:pluralize)
```
graysonwright: [Our style guide](https://github.com/thoughtbot/guides/tree/master/style/ruby) says to avoid conditional modifiers (lines that end with conditionals). Can we re-phrase this as:

```ruby
if args.none? || args.first == "application"
  "application"
else
  args.first.underscore.pluralize
end
```

Also, I think we can ditch the `try()` methods, because we now know that args.first is defined at this point.

graysonwright:
Left a comment - other than that, looks good!

5minpause:
Oh I wasn't aware of this rule in your style guide. Thanks for pointing this out. Let me update the commit.

5minpause:
@graysonwright I updated the pull request according to the style guide. Thanks for making me aware of it.

graysonwright:
Sorry for the hold-up here. I want to pull this branch down locally and see if I can come up with tests for it.

Implementation looks good, though.

graysonwright:
I hate to draw out the discussion on this more, but after looking at it closer I'm not sure this change is as simple as it first appeared. Here's my thinking:

## Breaking down the problem

The problem is that the generator is placing the files in `app/views/admin/applications` instead of `app/views/admin/appliction`.

This happens when the user types:

```
rails generate adminsitrate:views application
```

However, [the documentation][docs] states that if you want to generate views that apply to all resources, you should run:

```
rails generate administrate:views
```

...without the `application` at the end.

This is a subtle distinction, but it changes this issue from a bug report to a feature request. As long as users follow the docs, the generators will work fine for them. So instead of fixing broken behavior, this PR would essentially be auto-correcting people's mis-use of the generator API.

[docs]: https://administrate-docs.herokuapp.com/customizing_page_views

## The new question

Our new question is: if users mistakenly type `application` at the end of their generator string, should we assume that they want to generate the views at the application level?

There's an obvious reason to do so - we've already had at least one user confused by this, and it has the potential to help people avoid a frustrating-to-track-down error.

On the other hand, imagine that someone's writing a Rails app that processes job applications, and they have a model named `Application`. In that case, the correct place to generate the views for that model would be in `app/views/admin/application`. This PR would remove our ability to do that, and would result in another frustrating-to-track-down error in that application.

So, thoughts on the tradeoff?

graysonwright:
I'm closing due to inactivity. If we get more requests for this, I'll gladly re-open.

PR #228, opened 2015-11-12
Add uglifier to Gemfile

graysonwright:
## Problem:

The demo app could not deploy on Heroku
because of a missing dependency on uglifier.

## Solution:

Add uglifier to the `production` group in the Gemfile.


graysonwright:
Tested on staging - this works, so I'm merging.

PR #233, opened 2015-11-12
Relax Rails version requirements

alexstophel:
I just tried to install this gem on an older Rails project. It seems like Administrate only supports `~> 4.2`. Should this be added to the README?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ To do that, Administrate follows a few simple rules:
 
 ## Getting Started
 
+NOTE: Administrate depends on Rails `~> 4.2`.
+
 Add Administrate to your Gemfile:
 
 ```ruby


graysonwright:
yep, but I think we can relax the 4.2 requirement a bit first.

alexstophel:
I'm happy to help out and relax the requirement. Can you offer some guidance on what version you'd like to support?

graysonwright:
I haven't played around with compatibility much yet, but I think it would be nice to go back to 4.0 at least.

I haven't been keeping close track of changes to the Rails APIs over the minor version bumps, so I'm not sure if it'll be difficult to get compatibility back to 4.0. If you want to take a look into it, that would be great!

As a side note, we're using [Appraisal](https://github.com/thoughtbot/appraisal) on this project to test compatibility.

jcmorrow:
It seems like time has addressed this issue for us in so far as 4.2 is now the oldest version supported by rails😅. @nickcharlton do you agree that it makes sense to merge this?

nickcharlton:
@jcmorrow I wonder if slightly different wording would be better now? Something like:

> Administrate supports Rails 4.2, 5.0 and higher.

What do you think?

nickcharlton:
Closing in favour of #778, so we can suggest newer versions too. Thanks for this PR!

PR #239, opened 2015-11-14
Fix BelongsTo form partial select field not setting selected attribute

weimeng:
The selected attribute on select dropdown options is not currently being set, resulting in users having to manually re-select the already-selected option on editing a model. (What a mouthful! I could probably phrase this better, but it's 4am.)

Attempting a fix for this.

### DIFF
--- a/app/views/fields/belongs_to/_form.html.erb
+++ b/app/views/fields/belongs_to/_form.html.erb
@@ -18,5 +18,5 @@ that displays all possible records to associate with.
 
 <%= f.label field.permitted_attribute %>
 <%= f.select(field.permitted_attribute) do %>
-  <%= options_for_select(field.associated_resource_options) %>
+  <%= options_for_select(field.associated_resource_options, field.data ? field.data.id : nil) %>
 <% end %>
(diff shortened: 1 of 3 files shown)


aaronlerch:
:+1:  This problem also exists in `Field::HasMany`'s edit form as well. The simple fix (I can try to work up a PR later) is, similar to this change:

in app/views/fields/has_many/_form.html.erb on line 25:
```
<%= options_for_select(field.associated_resource_options, field.data.map(&:id)) %>
```

graysonwright:
@weimeng and @aaronlerch, good catch! Are you both using Administrate version `0.1.1`?

I've added this to the milestone for our next release.

weimeng:
@aaronlerch I took a stab at fixing the problem in `Field::HasMany` as well. Thanks!

@graysonwright Yes, I'm using `0.1.1`.

weimeng:
I'm pretty happy to close this in favour of #247, seems like a more elegant solution. 

@graysonwright Add #247 to the `0.1.2` milestone so it's not missed?

graysonwright:
@weimeng thanks! I think you're right, #247 is a little cleaner. I'll go ahead and merge that in.

Thanks for your contribution!

PR #242, opened 2015-11-16
Fixes issue #241

### LINKED ISSUE #241: Buggy BelongsTo:  Edit page for a resource shows old field data.
Dropdown boxes retain/displaying old data.

The demo app does this aswell!
http://administrate-prototype.herokuapp.com/admin

*STEPS*
1.  Go to any dashboard and view an Order (any item with dropdown selects). 
2.  Press Edit and change any dropdown box (Customer).
3.  Press update.
4.  The show page show now has the new updated values. (DB values are also updated.)
5.  Click Edit again and the old values are back.

It just seems to happen on belongs to relationships.

johnmeehan:
Fixes #241

## Problem:

The selected value for a belongs to field was not shown to the user when editing a resource.

If a belongs to field has a selected value it will now be shown when the user goes to the edit page to make changes.

## Solution:

use the selected field provided by `options_for_select`

 ` options_for_select(field.associated_resource_options, field.data.try(:id))`

### DIFF
--- a/app/views/fields/belongs_to/_form.html.erb
+++ b/app/views/fields/belongs_to/_form.html.erb
@@ -18,5 +18,5 @@ that displays all possible records to associate with.
 
 <%= f.label field.permitted_attribute %>
 <%= f.select(field.permitted_attribute) do %>
-  <%= options_for_select(field.associated_resource_options) %>
+  <%= options_for_select(field.associated_resource_options, field.data.try(:id)) %>
 <% end %>
--- a/spec/features/orders_form_spec.rb
+++ b/spec/features/orders_form_spec.rb
@@ -20,14 +20,26 @@ describe "order form" do
   end
 
   describe "belongs_to relationships" do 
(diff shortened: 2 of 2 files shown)


graysonwright:
@johnmeehan this looks great, thanks!

Can you add this test to `spec/features/orders_form_spec.rb`?

```ruby
  it "pre-fills belongs_to select boxes" do
    create(:customer)
    order = create(:order)

    visit edit_admin_order_path(order)
    value = find("select[name='order[customer_id]']").value

    expect(value).to eq(order.customer_id.to_s)
  end
```

Once you've added it, make sure that it fails without your changes, and passes with them.

Thanks! Once that's in, this looks good to merge.

graysonwright:
Hey, @johnmeehan - thanks for your work here!

We decided to merge in a fix from PR #247, since it fixed up `HasMany` relationships at the same time. Cheers!

PR #243, opened 2015-11-16
Fix up "Show" label on index page

graysonwright:
## Problem:

In fcd3fc7, we moved logic for displaying resources into dashboard
classes. Unfortunately, we missed migrating the edit page's "Show" link
over to use the new logic. It is currently displaying the resource using
Ruby's built-in `to_s` method.

## Solution:

Migrate the button to the new method of displaying resources.

### DIFF
--- a/spec/features/edit_page_spec.rb
+++ b/spec/features/edit_page_spec.rb
@@ -9,6 +9,14 @@ describe "customer edit page" do
     expect(page).to have_header("Edit #{customer.name}")
   end
 
+  it "has button to view customer details" do
+    customer = create(:customer)
+
+    visit edit_admin_customer_path(customer)
+
+    expect(page).to have_link("Show #{displayed(customer)}")
+  end
+
   it "has forms for the customer's attributes" do
     customer = create(:customer)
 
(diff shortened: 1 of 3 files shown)


spec/features/edit_page_spec.rb
```diff
@@ -9,6 +9,14 @@
     expect(page).to have_header("Edit #{customer.name}")
   end
 
+  it "has button to view customer details" do
+    customer = create(:customer)
+
+    visit edit_admin_customer_path(customer)
+
+    expect(page).to have_link("Show #{displayed(customer)}")
```
jayroh: This may be for my own reference, but - where is this mysterious `displayed` method defined? Don't see it in this spec, or rails_helper. ¯\\\_(ツ)\_/¯

graysonwright: Meant to push to this branch, but accidentally pushed to master. :confused:

This commit fixes up the include logic for the `displayed` helper.
https://github.com/thoughtbot/administrate/commit/9787489bef889e50cf0fc7f20c927b6481278a16

Issue #244, opened 2015-11-16
Has_many not populated with existing values in edit


monkbroc:
When opening the edit page for a resource with an has_many relation, none of the option tags generated for the select tag have the "selected" attribute set so existing values are lost on save.

You can reproduce this on the demo server:
1. Go to an order page with items https://administrate-prototype.herokuapp.com/admin/orders/331/
2. Click Edit
3. See that the line items ids field is blank
4. Click Save
5. See that the line items are gone.

Here's my workaround in `app/views/fields/has_many/_form.html.erb`

```ruby
<%= f.select(field.attribute_key, nil, {}, multiple: true) do %>
  <%= options_for_select(field.associated_resource_options, field.data.map(&:id)) %>
<% end %>
```

I can make a PR to fix this properly if you want.

This issue is similar to #241

monkbroc:
@graysonwright I want to make sure I understand correctly: you are OK if I write a proper fix and you will include it in 0.1.2, right?

graysonwright:
@monkbroc sorry about that.

yeah, if you want to submit a PR, I'll include it in the next release. It looks like the solution you posted above will work fine.

If you do submit a PR, can you add a test that fails without your fix, and passes with it? I gave an example on #242 .

monkbroc:
Yes, I'll include proper tests.

Nice work on the gem! The lack of DSL makes it really easy to jump in and understand the inner workings on the 1st day.

graysonwright:
Thanks! That's great to hear. :smile:

graysonwright:
Fixed by #247.

PR #247, opened 2015-11-16
Populate has_many and belongs_to in edit view

monkbroc:
Fix for #244.

I found out while implementing this PR that #239 is a fix for the same issue, but it puts the logic in the view. I put it in the field class.

I reused @weimeng's tests since they work well. Thanks.


graysonwright:
Thanks! Merged this in as 7403023.

PR #248, opened 2015-11-16
Select the correct dashboard for polymorphic fields

### LINKED ISSUE #245: Polymorphic field should override dashboard selection from Associative
Currently [`Polymorphic`](https://github.com/thoughtbot/administrate/blob/04247d78196e692e668933390f8190ba1f4f8533/lib/administrate/fields/polymorphic.rb) is only an alias for `Associative`.

`Associative#associated_dashboard` uses the attribute name to select the correct dashboard.

In `Polymorphic` this leads to a `uninitialized constant` error when calling `field.display_associated_resource` during the render of `app/views/fields/polymorphic/_index.html.erb`

For a polymorphic relation, the correct dashboard should be taken from the class of the associated object if it exists.

My workaround right now is:

```ruby
<% if field.data %>
  <%= link_to(
    "#{field.data.class.name}Dashboard".constantize.new.display_resource(field.data),
    polymorphic_path([:admin, field.data])
  ) %>
<% end %>
```

Do you agree with the approach? Is there a better way to get the associated dashboard name?

I can clean this up and make a PR for this.

monkbroc:
For a polymorphic relation, the correct dashboard should be taken from the class of the associated object if it exists, not from the name of the attribute like in the base class associative.

This also updates the default polymorphic views to use the configurable namespace.

Fixes #245

### DIFF
--- a/app/views/fields/polymorphic/_show.html.erb
+++ b/app/views/fields/polymorphic/_show.html.erb
@@ -19,6 +19,6 @@ By default, the relationship is rendered as a link to the associated object.
 <% if field.data %>
   <%= link_to(
     field.display_associated_resource,
-    [:admin, field.data],
+    [Administrate::NAMESPACE, field.data],
   ) %>
 <% end %>
--- a/spec/lib/fields/polymorphic_spec.rb
+++ b/spec/lib/fields/polymorphic_spec.rb
@@ -1,4 +1,5 @@
 require "administrate/fields/polymorphic"
+require "support/constant_helpers"
 require "support/field_matchers"
 
 describe Administrate::Field::Polymorphic do
@@ -16,4 +17,23 @@ describe Administrate::Field::Polymorphic do
   end
 
   it { should_permit_param(:foo, for_attribute: :foo) }
+
+  describe "#display_associated_resource" do
+    it "displays through the dashboard based on the polymorphic class name" do
+      begin
+        Thing = Class.new
+        ThingDashboard = Class.new do
+          def display_resource(*)
+            :success
+          end
+        end
+
+        field = Administrate::Field::Polymorphic.new(:foo, Thing.new, :show)
+        display = field.display_associated_resource
+        expect(display).to eq :success
+      ensure
+        remove_constants :Thing, :ThingDashboard
+      end
+    end
+  end
 end
--- a/spec/lib/fields/polymorphic_spec.rb
+++ b/spec/lib/fields/polymorphic_spec.rb
@@ -30,6 +30,7 @@ describe Administrate::Field::Polymorphic do
 
         field = Administrate::Field::Polymorphic.new(:foo, Thing.new, :show)
         display = field.display_associated_resource
+
         expect(display).to eq :success
       ensure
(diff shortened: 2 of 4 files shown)


spec/lib/fields/polymorphic_spec.rb
```diff
@@ -16,4 +17,23 @@
...
+        Thing = Class.new
+        ThingDashboard = Class.new do
+          def display_resource(*)
+            :success
+          end
+        end
+
+        field = Administrate::Field::Polymorphic.new(:foo, Thing.new, :show)
+        display = field.display_associated_resource
+        expect(display).to eq :success
```
graysonwright: Can you add a newline before this `expect`, to separate the test's execution phase from its expectation phase?

More at https://robots.thoughtbot.com/four-phase-test

app/views/fields/polymorphic/_show.html.erb
```diff
@@ -19,6 +19,6 @@ By default, the relationship is rendered as a link to the associated object.
 <% if field.data %>
   <%= link_to(
     field.display_associated_resource,
-    [:admin, field.data],
+    [Administrate::NAMESPACE, field.data],
```
graysonwright: :clap: good catch!

graysonwright:
This looks great! I added a small comment for style, but this is good to merge after that.

graysonwright:
Thanks! I merged this in as 476b12f.

monkbroc:
Can you please explain your unorthodox method for merging pull requests? What's the advantage over a simple merge commit? I'm curious about the reasons behind this workflow.

danbee:
@monkbroc It avoids having a merge commit and keeps the history cleaner.

graysonwright:
what @danbee said. We've been doing this for a while at thoughtbot, and it results in much cleaner logs. Instead of a complex branching history, we get something like this:

2__tmux

In projects like this, where we have a bunch of people working in parallel, that's a lot easier to parse than the history that the "Big Green Merge Button":tm: generates.

Here's the log for Rails, for comparison:

2__tmux


If you're curious, the exact rebasing process we're using is [outlined in our guides](https://github.com/thoughtbot/guides/tree/master/protocol/git).

## Other references:

- [My git alias for displaying logs in the pretty tree format](https://github.com/graysonwright/dotfiles/blob/master/gitconfig#L33)
- [Deliberate Git](http://rakeroutes.com/blog/deliberate-git/) by Stephen Ball. This blog post and talk provide a great overview of the benefits of adding rebasing to your git workflow.

monkbroc:
Thanks for the great explanation. I already do the commit message in the present and `rebase -i` to get rid of WIP commits (I would have done it here after the review if asked).

I just like seeing the actual branches merged in since it keeps the name of the feature branch in the merge commit message, rather than seeing a sequential series of commits. That's just me though.

graysonwright:
Yeah, I think it might be a matter of preference. The whole team over here is super used to this style of merging, so it'd be a bit rough to switch over.

@github, where are we at with defining custom merge strategies? :smile:

PR #249, opened 2015-11-17
No need to paginate twice

DamirSvrtan:
Pagination is already called a couple of lines before.


graysonwright:
Great, thanks! Merged in as 9350c65.

PR #251, opened 2015-11-18
Display helpful message if attribute_name missing

### LINKED ISSUE #246: Developer UX: better error for misspelled/missing attribute
If a field in `COLLECTION_ATTRIBUTES` has no key in `ATTRIBUTE_TYPES` either because I misspelled it or I forget to add it to `ATTRIBUTE_TYPES` I will get a `undefined method new for nil:NilClass` in [`Page::Base`](https://github.com/thoughtbot/administrate/blob/04247d78196e692e668933390f8190ba1f4f8533/lib/administrate/page/base.rb#L21)

It would be nicer to see a custom exception `Attribute xyz not found in dashboard AbcDashboard`

This has already happened multiple times on my first day of using Administrate so I think this is worthwhile to fix.

5minpause:
# Problem
If a field in `COLLECTION_ATTRIBUTES` has no key in `ATTRIBUTE_TYPES` either because I misspelled it or I forget to add it to `ATTRIBUTE_TYPES` I will get a undefined method new for `nil:NilClass` in `Page::Base`

# Solution
Don't access hash values directly but use `fetch`. If the attribute is not found the block raises an error with a helpful error message.

Closes #246

### DIFF
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -18,6 +18,18 @@ module Administrate
       self.class::ATTRIBUTE_TYPES
     end
 
+    def attribute_type_for(attribute_name)
+      attribute_types.fetch(attribute_name) do
+        fail "Attribute #{attribute_name} not in #{self.class}::ATTRIBUTE_TYPES"
+      end
+    end
+
+    def attribute_types_for(*attribute_names)
+      attribute_names.each_with_object({}) do |name, attributes|
+        attributes[name] = attribute_type_for(name)
+      end
+    end
+
     def form_attributes
       self.class::FORM_ATTRIBUTES
     end
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -24,7 +24,7 @@ module Administrate
       end
     end
 
-    def attribute_types_for(*attribute_names)
+    def attribute_types_for(attribute_names)
       attribute_names.each_with_object({}) do |name, attributes|
         attributes[name] = attribute_type_for(name)
       end
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -20,7 +20,7 @@ module Administrate
 
     def attribute_type_for(attribute_name)
       attribute_types.fetch(attribute_name) do
-        fail "Attribute #{attribute_name} could not be found in #{self.class}::ATTRIBUTE_TYPES"
+        fail attribute_not_found_message(attribute_name)
       end
     end
 
@@ -50,6 +50,12 @@ module Administrate
 
     def display_resource(resource)
       "#{resource.class} ##{resource.id}"
+    end
+
+    private
+
+    def attribute_not_found_message(attr)
+      "Attribute #{attr} could not be found in #{self.class}::ATTRIBUTE_TYPES"
     end
   end
 end
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -20,7 +20,7 @@ module Administrate
 
     def attribute_type_for(attribute_name)
       attribute_types.fetch(attribute_name) do
-        fail "Attribute #{attribute_name} not in #{self.class}::ATTRIBUTE_TYPES"
+        fail "Attribute #{attribute_name} could not be found in #{self.class}::ATTRIBUTE_TYPES"
       end
     end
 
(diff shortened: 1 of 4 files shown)


5minpause:
Thanks for your feedback @monkbroc and @graysonwright.
I modified the commit and opened this pull request.

Do we need tests for this? Would this be a feature test? I am asking for suggestions.

monkbroc:
@5minpause if you give me commit access to the repository https://github.com/ikuseiGmbH/administrate I'll write a test and push it to this pull request.

5minpause:
@monkbroc I added you. I am curious what you'll do. Looking forward to learning something new. :grinning:

monkbroc:
After looking at this more, it seems that the concern of attributes existing in a dashboard belongs in the dashboard class (surprise!).

It's easier to test the behavior there.

My change also decreases the knowledge collaborators like page classes have to know about the internals of the dashboard class (`attribute_types` is a hash I can use `fetch` and `slice` on).

Regarding Hound: have you thought of making your line length a little longer? 80 seems rather short for Ruby. I've had better luck with 90 or 100.

graysonwright:
Whoa, this looks great!

spec/dashboards/customer_dashboard_spec.rb
```diff
@@ -24,6 +24,45 @@
...
+          email: Administrate::Field::Email,
+        )
+      end
+    end
+
+    context "for one non-existent attribute" do
+      it "raises an exception" do
+        dashboard = CustomerDashboard.new
+        expect { dashboard.attribute_types_for(:name, :foo) }.
+          to raise_error /Attribute foo not in CustomerDashboard/
```
graysonwright: How would you feel about extracting a method here?

```ruby
expect { dashboard.attribute_types_for(:name, :foo) }.
  to raise_error(message_for("foo", "CustomerDashboard"))
  
  
def message_for(attribute, dashboard_class)
  "Attribute #{attribute} could not be found in #{dashboard_class}::ATTRIBUTE_TYPES")
end
```

Note: I also slightly changed the wording of the error message in that snippet, I think it reads a bit better.

graysonwright:
Left a small comment, but this looks good to merge after that.

spec/dashboards/customer_dashboard_spec.rb
monkbroc: I agree about the wording. That's what I had originally but the Hound was against it since it's 95 characters long :laughing:

I'll put it back

spec/dashboards/customer_dashboard_spec.rb
```diff
@@ -24,6 +24,49 @@
     end
   end
 
+  def missing_attribute_message(attribute, dashboard_class)
+    "Attribute #{attribute} could not be found in #{dashboard_class}::ATTRIBUTE_TYPES"
```
graysonwright: I'm ok ignoring hound for this line.

lib/administrate/base_dashboard.rb
```diff
@@ -18,6 +18,18 @@ def attribute_types
       self.class::ATTRIBUTE_TYPES
     end
 
+    def attribute_type_for(attribute_name)
+      attribute_types.fetch(attribute_name) do
+        fail "Attribute #{attribute_name} could not be found in #{self.class}::ATTRIBUTE_TYPES"
```
graysonwright: We can extract a method for the error message, similar to what you did in the tests. If the line's still too long after that, I'm ok ignoring it.

monkbroc: It will still be too long.

lib/administrate/page/collection.rb
```diff
@@ -14,7 +14,7 @@ def attributes_for(resource)
       end
 
       def attribute_types
-        dashboard.attribute_types.slice(*attribute_names)
+        dashboard.attribute_types_for(*attribute_names)
```
graysonwright: Let's get rid of the splat operator here (`*`), and pass the array directly.

lib/administrate/base_dashboard.rb
```diff
@@ -18,6 +18,18 @@ def attribute_types
...
+      end
+    end
+
+    def attribute_types_for(*attribute_names)
```
graysonwright: same here for the splat operator.

lib/administrate/base_dashboard.rb
graysonwright: oh, sorry - I think it's a good idea to extract the message into its own method regardless. I meant that if hound comments again, we can ignore that.

graysonwright:
Left one more small comment (in reply to hound above), then this is good to go! Thanks!

monkbroc:
Done.

graysonwright:
Thanks! Merged in as d212265.

PR #253, opened 2015-11-19
Explicitly set ruby version for CircleCI

graysonwright:
## Problem:

Since we updated to Ruby 2.2.3 in 079c17f,
CircleCI hasn't been able to detect the proper ruby version
and builds have been failing as a result.

## Solution:

Explicitly set the ruby version to 2.2.3 in the circle.yml config file.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,6 @@
+machine:
+  ruby:
+    version: 2.2.3
 dependencies:
   post:
     - bundle exec appraisal install


5minpause:
That's why I noticed as well. In my PR #208 I didn't have the problem anymore though. Strange…

graysonwright:
huh. rebuilding without the cache seems to have fixed the problem. ¯\_(ツ)_/¯

graysonwright:
Alright, I don't think this PR is needed. I'm rebuilding failing builds without the cache, and they seem to be working fine. Circle is smart enough to pick up the ruby version from `.ruby-version`, as long as the cache is cleared.

Issue #219, opened 2015-11-11
Install error


johncblandii:
We're hitting a problem installing Administrate. Any ideas would be helpful.

Steps:
- Add to `Gemfile`
- `bundle`

The error below is the output. I noticed there is a file named `app/dashboards/#<class:0x007fcf52bfe248>_dashboard.rb` it creates.

`dashboard_manifest.rb` has the following dashboards:
```
  DASHBOARDS = [
    :#<class:0x007fcf52bfe248>s,
    :#<class:0x007fcf52d7c520>s,
    :#<class:0x007fcf567ea7e8>s,
    :#<class:0x007fcf52f1f1c0>s,
    :#<class:0x007fcf560eabf0>s,
    :#<class:0x007fcf565a0898>s,
    :#<class:0x007fcf520597f8>s,
    :#<class:0x007fcf56509ba0>s,
    :#<class:0x007fcf52cf6650>s,
    :#<class:0x007fcf56202808>s,
    :users,
    :churches,
    :locations,
    :roles,
    :users_roles,
    :members,
  ]
```

Error output:
```
$: rails generate administrate:install
      create  app/controllers/admin/application_controller.rb
      create  app/dashboards/dashboard_manifest.rb
       route  namespace :admin do
    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
      resources dashboard_resource
    end

    root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
  end
      create  app/dashboards/#<class:0x007fcf52bfe248>_dashboard.rb
/Users/johncblandii/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/administrate-0.1.0/lib/generators/administrate/dashboard/dashboard_generator.rb:87:in `const_get': wrong constant name #<class:0x007fcf52bfe248> (NameError)
	from /Users/johncblandii/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/administrate-0.1.0/lib/generators/administrate/dashboard/dashboard_generator.rb:87:in `klass'
	from /Users/johncblandii/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/administrate-0.1.0/lib/generators/administrate/dashboard/dashboard_generator.rb:40:in `attributes'
	from (erb):11:in `template'
	from /Users/johncblandii/.rbenv/versions/2.2.2/lib/ruby/2.2.0/erb.rb:863:in `eval'
[...]

PR #259, opened 2015-11-20
Don't generate dashboards for non-constant models

### LINKED ISSUE #145: issues with `rake routes`
Half the time when I run my `rake routes` command I get the following error:

RailsApp/config/routes.rb:26:in `block (2 levels) in <top (required)>': uninitialized constant DashboardManifest (NameError)

Other times my routes print correctly to the console, but there's no listing for my admin namespaced routes.

I've seen the `uninitialized constant DashboardManifest (NameError)` error pop up randomly when using generators as well, but the error doesn't seem to appear consistently.

graysonwright:
Fixes #145 and #219

## Problem:

When the same `has_and_belongs_to_many` relationship is defined twice,
Rails generates an unnamed subclass of `ActiveRecord::Base`.
This is probably a bug in Rails.

The unnamed subclass shows up like this:

```ruby
> ActiveRecord::Base.descendants.last
 #<Class:0x007fddea1bdf08> (call '#<Class:0x007fddea1bdf08>.connection' to establish a connection)
```

When the install generator tries to populate the `DashboardManifest`
with all of the `ActiveRecord` models,
it trips up on unnamed models,
and generates a manifest with invalid syntax.

This results in a `NameError` for the user when they run the
`administrate:install` generator:

```
/gems/administrate-0.1.0/lib/generators/administrate/dashboard/dashboard_generator.rb:87:
in `const_get': wrong constant name #<class:0x007fcf52bfe248> (NameError)
```

This error most often pops up in applications
that are using the `spring` gem.

## Solution:

- Add a test case that reproduces the error
  by defining duplicate `has_many` associations.
- In the install generator, ignore any models that are not constants.
  We determine constants by checking that the model's `#name`
  matches its `#to_s`.
- Print a warning for the ignored models.

### DIFF
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -40,6 +40,10 @@ module Administrate
           puts "WARNING: Unable to generate a dashboard for #{invalid_model}."
           puts "         It is not connected to a database table."
         end
+
+        unnamed_constants.each do |invalid_model|
+          puts "NOTICE: Skipping dynamically generated model #{invalid_model}."
+        end
       end
 
       private
@@ -63,7 +67,7 @@ module Administrate
       end
 
       def invalid_database_models
-        models_without_tables + namespaced_models
+        models_without_tables + namespaced_models + unnamed_constants
       end
 
       def models_without_tables
@@ -72,6 +76,10 @@ module Administrate
 
       def namespaced_models
         database_models.select { |model| model.to_s.include?("::") }
+      end
+
+      def unnamed_constants
+        ActiveRecord::Base.descendants.reject { |d| d.name == d.to_s }
       end
 
       def dashboard_routes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
   `ATTRIBUTE_TYPES`
 * [#231] [UI] Fix layout issue on show page where a long label next to an empty
   value would cause following fields on the page to be mis-aligned.
+* [#259] [BUGFIX] Make installation generator more robust
+  by ignoring dynamically generated, unnamed models
 * [#243] [BUGFIX] Fix up a "Show" button on the edit page that was not using the
   `display_resource` method.
 * [#248] [BUGFIX] Improve polymorphic relationship's dashboard class detection.
--- a/spec/generators/install_generator_spec.rb
+++ b/spec/generators/install_generator_spec.rb
@@ -67,6 +67,21 @@ describe Administrate::Generators::InstallGenerator, :generator do
         remove_constants :ModelWithoutDBTable
       end
     end
+
+    it "skips models that don't have a named constant" do
+      stub_generator_dependencies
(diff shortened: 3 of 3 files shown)


lib/generators/administrate/install/install_generator.rb
```diff
@@ -40,6 +40,11 @@ def warn_about_invalid_models
           puts "WARNING: Unable to generate a dashboard for #{invalid_model}."
           puts "         It is not connected to a database table."
         end
+
+        unnamed_constants.each do |invalid_model|
+          puts "WARNING: Unable to generate a dashboard for #{invalid_model}."
+          puts "         It is not a valid Ruby constant."
```
jakecraige: Is there something that could be linked to the user to help them debug/fix this?

graysonwright: @jakecraige I... don't know.

I'm about 85% sure this is a bug in Rails, and I'm going to open a PR there to try to fix it. On the user's end, they're probably not aware they even have a non-constant class in their application, so it'll be strange for them to see a warning.

I'm considering removing this warning altogether, or changing it to something less scary-sounding, like:

```
NOTICE: Skipping dynamically generated model `#<Class:asdfasdf>`.
```

Thoughts?

graysonwright:
Ping @sgrif - figured out the issue where ActiveRecord is creating unnamed classes. Looks like it's a bug in ActiveRecord - is that right?

I have the bandwidth to open up a PR for it, if you'd like.

graysonwright:
Talked to @sgrif - since calling `has_and_belongs_to_many` multiple times for the same relationship isn't a supported use case in Rails, the bug actually lives in Spring. For some reason, it's reloading the relationship when it shouldn't be.

sgrif:
> For some reason, it's reloading the relationship when it shouldn't be.

More likely it's retaining the constant that should be getting released as part of the general constant reloading process.

graysonwright:
yep, good catch. Thanks!

lib/generators/administrate/install/install_generator.rb
sgrif: I spent a bit thinking about this. Does this even need to be a warning? I feel like we can pretty certainly say that if a class isn't assigned to a constant, you can ignore it. (Also you should look at the implementation of `descendants` if you haven't in the past, because it's hilarious)

Rodrigora:
Cool @graysonwright. :100: 

Just letting you know that I do not have duplicated `HABTM` relationships. These are my models:

```ruby
class Feature < ActiveRecord::Base
  has_and_belongs_to_many :units
end

class Unit < ActiveRecord::Base
  has_and_belongs_to_many :features
end
```

graysonwright:
@Rodrigora yep, I think there may be a bug in [spring](https://github.com/rails/spring) that is causing the equivalent effect of a duplicated HABTM relationship.

Are you using spring in your app?

Rodrigora:
Yes @graysonwright, I'm using it. 

You're right, certainly it is a Spring bug. Thanks!

PR #264, opened 2015-11-21
Test against ruby project version, 2.2.3

pisaacs:
079c17f set ruby for project to 2.2.3. Travis should test against same version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ language:
 notifications:
   email: false
 rvm:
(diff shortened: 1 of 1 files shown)


graysonwright:
Oops! We're actually not testing using Travis - We're doing our continuous integration on CircleCI. The `.travis.yml` was never actually used.

I've opened up PR #268 to remove the file.

pisaacs:
:+1:

PR #270, opened 2015-11-23
Save apologies for tragedies

mike-burns:
We do want polymorphic and hasone relationship support, but we aren't
sorry that they aren't there.

While here, remove the word "yet". It will get there when it gets there,
and won't get there if it isn't needed. "Yet" is not agile.

### DIFF
--- a/config/locales/administrate.es.yml
+++ b/config/locales/administrate.es.yml
@@ -18,6 +18,6 @@ es:
         more: Mostrando %{count} de %{total_count}
         none: Ninguno
       polymorphic:
-        not_supported: Los formularios con relaciones polimórficas no están soportados todavía. ¡Perdón!
+        not_supported: Los formularios con relaciones polimórficas no están soportados.
       has_one:
-        not_supported: Los formularios con relaciones HasOne no están soportados todavía. ¡Perdón!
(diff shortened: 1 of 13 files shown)


graysonwright:
@mike-burns as the first author of a structural change to the translation files, you now have the great honor of going through the rest of them!

mike-burns:
In some languages it makes sense to apologize here -- en-GB for example!

OK, I'll do that in a bit.

config/locales/administrate.es.yml
```diff
@@ -18,6 +18,6 @@ es:
         more: Mostrando %{count} de %{total_count}
         none: Ninguno
       polymorphic:
-        not_supported: Los formularios con relaciones polimórficas no están soportados todavía. ¡Perdón!
+        not_supported: Los formularios con relaciones polimórficas no están soportados.
       has_one:
-        not_supported: Los formularios con relaciones HasOne no están soportados todavía. ¡Perdón!
+        not_supported: Los formularios con relaciones HasOne no están soportados.
```
picandocodigo: :+1:

Tensho:
Russian (yes)

graysonwright:
@mike-burns wow, I'm impressed. :smile:

olimart:
:fr: :ok:

PR #273, opened 2015-11-26
Increases the specificity of .form-field children rules.

keichan34:
I ran in to problems styling custom elements, especially when there were multiple form elements or labels within the value column. If there's a better way to do it, let me know and I'll update the PR accordingly.

### DIFF
--- a/app/assets/stylesheets/administrate/components/_form.scss
+++ b/app/assets/stylesheets/administrate/components/_form.scss
@@ -8,34 +8,36 @@
   align-items: baseline;
   width: 100%;
 
-  label {
-    @include data-label;
-    @include span-columns(2 of 12);
-    padding: $form-input-padding 0;
-    text-align: right;
-  }
+  > {
+    label {
+      @include data-label;
+      @include span-columns(2 of 12);
+      padding: $form-input-padding 0;
+      text-align: right;
+    }
 
-  #{$all-text-inputs},
-  select,
-  textarea,
-  .selectize-control {
-    background-color: $form-field-background-color;
-    margin-bottom: $base-spacing;
-  }
+    #{$all-text-inputs},
+    select,
+    textarea,
+    .selectize-control {
+      background-color: $form-field-background-color;
+      margin-bottom: $base-spacing;
+    }
 
-  input,
-  select,
-  .selectize-control {
-    @include span-columns(4 of 12);
-  }
+    input,
+    select,
+    .selectize-control {
+      @include span-columns(4 of 12);
+    }
 
-  textarea {
-    @include span-columns(6 of 12);
-  }
+    textarea {
+      @include span-columns(6 of 12);
+    }
 
-  [type="checkbox"],
-  [type="radio"] {
-    margin: $small-spacing 0 ($base-spacing + $form-input-padding);
-    width: auto;
+    [type="checkbox"],
+    [type="radio"] {
+      margin: $small-spacing 0 ($base-spacing + $form-input-padding);
+      width: auto;
+    }
   }
 }


graysonwright:
@keichan34 I'd rather not increase the specificity of the CSS - in general, that would make it harder for people to overwrite the styles themselves, because they would need to match or exceed the specificity level we provide.

Can you give an example of the kind of CSS problem you're running into, so I can help find an alternative solution?

Thanks!

keichan34:
Sure, one specific problem I'm having with the general rules is a case where I have multiple checkboxes (with corresponding labels) within a single parent label

Something like this:

screen shot 2015-12-03 at 10 07 36 pm

I originally tried to wrap everything in a div, but the rules that are supposed to be applied to the left column labels are applied to the labels within the right side -- overwriting each style back to normal seemed like an inelegant solution to the problem.

graysonwright:
ok! I think we can fix this by adding a class of `.form-field_label` to the left-hand labels (`IMAGE GALLERY` in the above screenshot). We could then style those labels without worrying about interfering with other labels inside of `.form-field`.

@smharley @devonharless how does that sound? Are there other solutions that would work well here?

smharley:
@graysonwright I think what you're saying sounds good. `.form-field_heading` or something might also make sense

keichan34:
Keep in mind it's not only the label styles, the input styles would have to be overridden back to normal too. Instead of using the `input, select, .selectize-control { @include span-columns(4 of 12); }`, how about a class name like `.form-input, .selectize-control { @include span-columns(4 of 12); }` etc?

smharley:
@keichan34 Do you mind posting the HTML for the form that you're using? 

I'm curious about how the "label" (Image Gallery) on the left is marked up. It seems like the correct markup for it would be `<legend>`, not `<label>`. See this example: http://stackoverflow.com/questions/19457695/label-for-checkbox

Do you think changing the element used there, rather than CSS selector or specificity, could also solve your issue?

keichan34:
Sorry, I forgot to include the `<input>` in that earlier screenshot. The label is connected to that input, so I think it's appropriate to use `<label>` instead of `<legend>` in this case.

Here's the HTML I have right now:

```html
<div class="form-field form-field--refile-image-field">
  <label for="x_thumbnail_image">Thumbnail Image</label>
  <div class="value-column">
    <img class="attachment ..." src="...">
    <div class="checkbox">
      <label for="x_remove_thumbnail_image">
        <input name="x[remove_thumbnail_image]" type="hidden" value="0">
        <input type="checkbox" value="1" name="x[remove_thumbnail_image]" id="x_remove_thumbnail_image">
        Remove Thumbnail Image
      </label>
    </div>
    <input type="file" name="x[thumbnail_image]" id="x_thumbnail_image">
  </div>
</div>
```

It looks something like this: 
screen shot 2015-12-09 at 10 58 10 am

Here's the HTML for the multiple field

```html
<div class="form-field form-field--multi-refile-image-field">
  <label for="x_gallery_images_files">Gallery Images</label>
  <div class="value-column">
    <div class="thumbnail-container">
      <div class="thumbnail">
        <img class="attachment" ...>
        <div class="caption">
          <label for="x_gallery_images_attributes_0__destroy">
            <input name="x[gallery_images_attributes][0][_destroy]" type="hidden" value="0">
            <input type="checkbox" value="1" name="x[gallery_images_attributes][0][_destroy]" id="x_gallery_images_attributes_0__destroy">
            Delete
          </label>
        </div>
      </div>
      ...
    </div>
    <input multiple="multiple" type="file" name="x[gallery_images_files][]" id="x_gallery_images_files">
  </div>
</div>
```

screen shot 2015-12-09 at 11 06 53 am

Before increasing the specificity of the selectors, the labels for the checkboxes are turned in to a span-column (which would be solved by adding a class), but the `<input>` fields would break too, due to:

```scss
input,
select,
.selectize-control {
  @include span-columns(4 of 12);
}

[type="checkbox"],
[type="radio"] {
  margin: $small-spacing 0 ($base-spacing + $form-input-padding);
}
```

tysongach:
The `> {` direct child selector is usually a code smell in CSS…we should try to avoid that at all costs. We are also now mixing a lot of base HTML element styles (`input`, `label`, etc) with classed styles inside the `components` directory, which creates an unclear delineation for where styles belong and come from.

keichan34:
@tysongach I'm open to suggestions, as long as the problem of styles falling through to nested elements is solved. If the best solution is indeed to override all styles manually in my local stylesheet, I'll consider that.

tysongach:
@keichan34 Definitely.

A large problem that I see right now is that our stylesheet architecture got a bit disorganized over time. This has led to us having a `components/_form.scss` partial, as well as a `base/_forms.scss`. There is a lot of overriding and specificity issues in those.

I merged a few things in today already to start to clean this up, and am working more right now to clean up the overall architecture to help prevent this stuff.

keichan34:
Nice. I'll see what I can do after you've cleaned things up. Thanks :smiley:

graysonwright:
@tysongach - it looks like you've merged in #347, which contains a lot of stylesheet refactorings. Checking in to see if that changes the status of this PR.

tysongach:
@graysonwright It does not.

keichan34:
The refactor in #399 solved this particular problem for me. Thanks so much! :tada:

PR #275, opened 2015-11-26
Fix date_time_picker styles when choose year or month

33volcanoes:
Before fix
Before

After fix
After


graysonwright:
@33volcanoes Thanks for this fix! That CSS belongs inside the `datetime_picker_rails` gem, so I've opened up https://github.com/graysonwright/datetime_picker_rails/pull/3 to incorporate it.

Once that's in place, we can update Administrate's dependency on datetime_picker_rails to version 0.0.6, and it should be good! I'll leave this open until that point.

graysonwright:
I took care of this in #309. Thanks for the fix!

Issue #183, opened 2015-11-06
Feature Request: Add filtering


mikecx:
The ability to filter the collections on the index page would be really handy.

BGuimberteau:
:+1:

graysonwright:
@mikecx and @BGuimberteau - I'm hoping to add filtering into the search bar, if possible. Here's a discussion around it: #157. Would love your thoughts there!

nando:
Hi there,

i've just pushed this little change in my fork that lets me filter using current model scopes. Tests still missing though :-1: 

https://github.com/nando/administrate/commit/df6e8cb0a205aadc24e56a96d582c6dd7f6a0213

I'm trying to follow the CONTRIBUTING.md guide to make a PR right, but let you know now. As soon as I make the PR (hopefully tomorrow morning) i'll comment about it in #157.

Cheers,
-- nando

nando:
Finally I did the PR to filter through scopes this morning: https://github.com/thoughtbot/administrate/pull/276

fx:
+1

This has been stewing for a while, are you guys making any progress @graysonwright ?

kbighorse:
Is there a way to do arbitrary date range filtering? See "Index filters" section: http://activeadmin.info/3-index-pages.html

Issue #157, opened 2015-11-05
Smart Searching


graysonwright:
The long-term plan for the search bar is to support structured queries, to allow people to do complex searches over the data from a single field.

We'd like to support:
- Searching numeric fields for values greater than, less than, or equal to a given number
- Searching text fields for text including a given string
- Searching boolean fields by `true` or `false` value
- More than one of these ^ in the same search

Bonus points for a scheme that will let us search custom field types using dev-defined strategies.

I want to open up a discussion about the best way to structure the search queries.
- If you know of other apps or projects that do multi-field search well, please share them for inspiration.
- If you have an opinion on what the query structure should look like, chime in!

danbee:
I wonder if the whole search functionality should stay simple, but had some hooks to enable the dev to integrate more advanced search.

For example, if the app already has search implemented using something like Elasticsearch it would be better if there was an easy way to integrate that rather than having another search mechanism.

thorncp:
What do you about making the first step be a controller method that's overridable? I'm currently searching a polymorphic association, which requires some hand holding, and I have to override the action to replace Administrate's [default search](https://github.com/thoughtbot/administrate/blob/01e125c/app/controllers/administrate/application_controller.rb#L5). If this happened in a method, e.g. `search_resources`, I could override that with my custom thing.

graysonwright:
@thorncp I like that a lot. Right now there's a lot going on in the index method, and that would be a good first step towards breaking it out.

mikecx:
I agree the controller method override for search would definitely work. I currently use Elasticsearch mixed with Searchkick but I know there are probably 100 other solutions out there so there's a big need to remain very generic.

The thoughts of filtering came up with issue #183 which definitely complicate the search "chain". In my instance, I currently pass the filters as where clauses to Searchkick/Elasticsearch but that's obviously not a generic enough solution for what Administrate hopes to accomplish. 

A potential better solution might be to provide documentation and create a clean way to override the Administrate::Search class. Giving clean methods in there for both searching and filtering reduces the places to override (1 instead of per controller), gives places for query & filter (any anything else that might not have been thought of yet), and keeps the logic in one place.

5minpause:
Having the search bar support structured queries sounds like a good idea for developers. We use ActiveAdmin as a dashboard in our custom CMS. So it's user-facing for editing pages and resources. Our users like the filters/scopes that ActiveAdmin provides and which can be extended. How could a similar feature be integrated into Administrate?

[...]

PR #276, opened 2015-11-27
Search scopes

nando:
_DISCLAIMER: the original message of this PR written two months ago has been updated to represent better its current state (however the original one has been kept at the end)._

**Problem:**
If the resource managed by the dashboard has scopes defined there's no way to limit the scope of the search with them.

**Solution:**
Declare a new constant in the dashboard (COLLECTION_SCOPES) to define the valid scopes that could be used in the query while searching in the index page. The following is the current definition of the COLLECTION_SCOPES constant in the template used to generate new dashboards:

``` ruby
  # COLLECTION_SCOPES
  # an array or hash that define the valid scopes that could be used while
  # searching as part of the query string.
  COLLECTION_SCOPES = []  # Comment to use any scope, but read this text below.
  # If the above COLLECTION_SCOPES definition doesn't exist then any "scope"
  # defined could be used searching *scope:<name_of_the_scope>*. Though this
  # could be a nice feature in applications that has the dashboard access
  # properly secured **this approach is not recommended**. Administrate has
  # no way to know the scopes defined in the model and will send to the model
  # anything not included in its blacklist.
  #
  # When defined buttons will appear in the index header in order to filter the
  # resources displayed. If it's an array it will be treated internally as if
  # it were a hash with a single key called *scopes* pointing to our array. The
  # hash's keys and the scope definitions will be used to show a localized
  # caption for each button using *administrate.scopes.<model_class>* as I18n's
  # scope. If no translation the scope in that model, Administrate'll retry the
  # translation with the scope *administrate.scopes*. That will let us share
  # the same translation between different models (and be DRY!).
  #
  # Definition example with an Array:
  #
  #   COLLECTION_SCOPES = [
  #     :opened,
  #     :closed
  #   ]
  #
  # Definition example with a Hash:
  #
  #   COLLECTION_SCOPES = {
  #     status: [:opened, :closed],
  #     headquarters: [:madrid, :oviedo, :mexicodf]
  #   }
  #
  # Scopes with an argument can also be defined. An explicit value for the
  # argument can be defined adding that value after scope name between
  # parenthesis and without quotes. For example:
  #
  #   COLLECTION_SCOPES = {
  #     headquarter: ["office(madrid)", "office(oviedo)", "office(mexicodf)"]
  #   }
  #
  # Will use the scope *office(city)* using "madrid", "oviedo" and "mexicodf"
  # as arguments.
  #
  # Finally, it's possible to let the user indicate the value of the argument
  # as part of the search query adding ":*" after the scope name. For example:
  #
  #   COLLECTION_SCOPES = {
  #     headquarter: ["office(madrid)", "office(oviedo)", "office:*"]
  #   }
  #
  # Won't show any scope button for "office:*" but will let us indicate any
  # value after "office:" to use it as argument for the *office(city)* scope.
  # If our search query is "office:mexicodf" we'll get the same results than
  # clicking in the third button of the previous example (which query would be
  # "scope:office(mexicodf)").
```

**ORIGINAL MESSAGE**
I hope you like the idea (I've modified the original PR to avoid sending the scope to the model if it looks suspicious. We do not have that security problem in our project since our AdministrateController has 'before_filter :require_admin' at the beginning of its definition.


graysonwright:
@nando thanks!

I really like the idea - I think we should definitely allow the user to search by scopes.

I had imagined this as a set of buttons on the index page that correspond to scopes, where the user could click a button to filter by that scope. I think your approach removes the difficulty of figuring out which scopes are defined on models.

I'm not sure about the syntax of using `scope: other search terms`. I feel like this approach is a bit limited in that regard. For instance, it would be strange to search for multiple scopes at the same time.

Maybe it would be useful to have a discussion about what the search format should look like before moving forward with the implementation?

I think that the solution should:

- let the user search for multiple scopes at once
- indicate which scopes are available for the user to search
- have a clear syntax

nando:
@graysonwright You're very welcome!! Thank YOU for this awesome tool ;)

Of course, that discussion is *required* in order to keep this gem's good health and shape shining. In that sense we can see the code of this PR just as an idea for the final approach that we can continue discussing in https://github.com/thoughtbot/administrate/issues/183 or https://github.com/thoughtbot/administrate/issues/157

This code emerged from a particular need in our current project, but the gem solution should be developed with at least those three objectives in mind. In that sense, no problem if you feel that would be better to close this PR by now. If you do it, i'll try to keep my fork synced with upstream/master.

AFAIK current Rails version don't give us current app's scopes without patching it, so in order to "indicate which scopes are available for the user to search" we'd need to add those scopes as part of our project's Dashboard configuration, right?

If so, then my approach could be the default behaviour if no custom configuration is given in app/dashboards.

On the other side, if that configuration exists, then it could be managed as a Whitelist to allow only those scopes to be sent to the model.

What do you think?

graysonwright:
@nando I made a quick mockup to explain how I'm thinking of approaching this.

That's just a quick mockup and it would need more work, obviously. But I'd love to get your thoughts on this approach.

nando:
And you will. Thanks Grayson!

nando:
@graysonwright, does this looks good for you?:

```ruby
class Customers < Administrate::Dashboard::Base
  [...]
  COLLECTION_SCOPES = [
    :joined_last_week,
    :no_purchases,
    :inactive,
    :active
  ]
end
```

nando:
That is what i've done in this branch: https://github.com/nando/administrate/tree/collection_scopes

app/assets/stylesheets/administrate/components/_buttons.scss
```diff
@@ -34,3 +34,8 @@
     }
   }
 }
+
+.scope-buttons {
```
tysongach: We should move this into its own partial, making it a distinct component.

app/assets/stylesheets/administrate/components/_buttons.scss
```diff
@@ -34,3 +34,8 @@
...
+  margin: 3px;
+  padding-left: 1.5em;
```
tysongach: The two values above for `margin` and `padding` feel like they don’t have any meaning, or relate to anything. Where does `3px` and `1.5em` come from? Why use different units?

app/views/administrate/application/index.html.erb
```diff
@@ -55,6 +55,15 @@ It renders the `_table` partial to display details about the resources.
   </div>
 </header>
 
+<% page.scope_groups.each do |group| %>
+<div class="scope-buttons">
```
tysongach: Let’s de-couple these styles specifically from the _scope_ buttons and instead create a more generic, reusable component. Maybe something like `button-group`…? Then we can use the same styles elsewhere, when we need a grouping of buttons, again.

tysongach:
From a design perspective, I would think that these scopes should not carry the same weight as the “New” button. Meaning, in the mock above, these scope buttons are on the same level of the visual hierarchy as the “New Customer” button. The scope buttons are not the primary action and should fall back a little.

Couple thoughts:

- Can a person select multiple scopes at once, like check boxes? Or only one, like radios?
- The current `button` styles do not give us any indication that one is “selected”, or active
- A button _group_ is a common pattern that could work well here, for example: http://refills.bourbon.io/components/#er-toc-id-5
- The word “scope” feels so engineer-y…maybe “filters”?

app/assets/stylesheets/administrate/components/_buttons.scss
nando: DISCLAIMER: I'm mostly a backend developer and I've never used Sass before :( But I like it!!!

Totally agree Tyson, I put it here because in that moment the project still didn't have the components/_buttons.scss file, and i thought base/buttons.scss was the best place to put that style.

In that moment the style just was "margin: 3px", value and unit that i used to have the same as the $base-border-radius variable. Then came this merge with your (awesome;) work:

https://github.com/nando/administrate/commit/c026d414882

...and i had to put the "padding-left: 1.5em" style in order to separate them from the border using the value that the browser told me the header was using...

tysongach: Gotcha! No worries.

If you you’d like, I’m more than happy to take on the styles side.

app/views/administrate/application/index.html.erb
nando: I've tried to do what you told me with this commit:

https://github.com/nando/administrate/commit/8649b8773e132

As I told you before I'm a completely Sass newbie and also VERY bad with design and CSS, so please fix what i've done as you like. :)

And thanks!!!

tysongach: :+1:

nando:
Again Tyson, totally agree with you: they're a different type of buttons. Manage their aspect as you want, i like your style ;)

> Can a person select multiple scopes at once, like check boxes? Or only one, like radios?
Currently is like radios: only one of them can be selected. It's possible to search into more than one scope, but the user has to type them manually as part of the search query. So the code behind the scenes is ready for they to behave like check boxes, but they have to managed correctly: only has sense to search into scopes of different groups in the same query (that is, has no sense to look for resources into more than one scope of the same group).

> The current button styles do not give us any indication that one is “selected”, or active
They in fact lose the button aspect since they are not longer a link and became simple text into the div.buttons-group div. Current implementation uses the link_to_unless_current Rails helper, that let us give it a class for the possible link, but nothing otherwise. This implementation is probably too simple for something more complex and we can change it.

> A button group is a common pattern that could work well here, for example: http://refills.bourbon.io/components/#er-toc-id-5
Ohhh!!! They're awesome!!! :)

> The word “scope” feels so engineer-y…maybe “filters”?
In fact they are filters, since a scope is a type of filter. The thing is that *this particular type* of filter is well known by most of Rails programmers and I think that change it will be against one of the principles of Administrate: to avoid DSLs (i think the *filter-to-scope* mapping would be an Administrate concept) and reuse the common knowledge of the Rails and Ruby communities. But of course, that's my opinion :)

Thanks for your comment Tyson.

nando:
@tysongach i've just developed the logic needed to generate button links that let us mix more than one scope from different groups just by clicking them. Hopefully this screenshot will explain what i mean better than my English:

administrate_scopes_screenshot

When a scope is selected is into span.active. You can see the markup here: https://github.com/nando/administrate/blob/0f6ff854ce3b8375e55d4f07b20209fa47461fb5/app/views/administrate/application/_scope_buttons_group.html.erb

nando:
HELP WANTED: I've just broken the build in order to find now all missing translations for the word "Filters" that will be the default word if an array of scopes is defined in a dashboard. Current output of *i18n.tasks missing* is the following:
```
colgado@waati ~/src/administrate $ i18n-tasks missing
Missing translations (13) | i18n-tasks v0.9.2
+--------+----------------------------+----------------------------------+
| Locale | Key                        | Value in other locales or source |
+--------+----------------------------+----------------------------------+
|   da   | administrate.scopes.scopes | en Filters                       |
|   de   | administrate.scopes.scopes | en Filters                       |
|   fr   | administrate.scopes.scopes | en Filters                       |
|   it   | administrate.scopes.scopes | en Filters                       |
|   nl   | administrate.scopes.scopes | en Filters                       |
|   pl   | administrate.scopes.scopes | en Filters                       |
| pt-BR  | administrate.scopes.scopes | en Filters                       |
|   ru   | administrate.scopes.scopes | en Filters                       |
|   sv   | administrate.scopes.scopes | en Filters                       |
|   uk   | administrate.scopes.scopes | en Filters                       |
|   vi   | administrate.scopes.scopes | en Filters                       |
| zh-CN  | administrate.scopes.scopes | en Filters                       |
| zh-TW  | administrate.scopes.scopes | en Filters                       |
+--------+----------------------------+----------------------------------+
```
Thank you for your help!!!

nando:
The branch build passes upgrading to Rails 4.2.5.1

nando:
@tysongach I've used [the style code you told me about](http://refills.bourbon.io/components/#er-toc-id-5) for the "scope" buttons of my branch. Now the look much much better, thanks!
scopebuttonslookingbetter
Please, feel free to change [the button group scss code](https://github.com/nando/administrate/blob/search_scopes/app/assets/stylesheets/administrate/components/_buttons_group.scss) as you want, as i told you last week i'm a newbie with Sass.
I've also added the default translation for "scopes" as "filters" in all the languages currently present in our config/locales.
Cheers!

pablocrivella:
+1

ahmgeek:
will it ever be merged anytime soon ?

estiens:
A few months later...

nando:
We are using this branch merged with others that let use Administrate w/ Globalize and Pundit with this lines in our Gemfile:

gem 'administrate', git: 'git@github.com:nando/administrate.git', branch: 'fgs_globalize_and_search_scopes'
gem 'administrate-field-image'

We are using Administrate to manage resources in [our new website](the-cocktail.com) and are very happy and grateful to Thoughtbot and Grayson for this amazing gem (thanks again!!)

Hopefully they merge it one day. It would be a great honor for me, i admire Thoughtbot from its very beginning ;)

bastianwegge:
@graysonwright @tysongach what's up with this PR? 

It's clearly something people are requesting as in: https://github.com/thoughtbot/administrate/issues/157, https://github.com/thoughtbot/administrate/issues/183, https://github.com/thoughtbot/administrate/issues/277, 
https://github.com/thoughtbot/administrate/issues/291, https://github.com/thoughtbot/administrate/issues/563, https://github.com/thoughtbot/administrate/issues/445

I'd be glad to help solve any issues or interferences you see upcoming with this code. 🌮

pustomytnyk:
@nando I believe you unfortunately are the only one who can rebase this :)

koppen:
I'd be willing to help getting the PR in shape and already have a branch where this has been merged into a fresh master over at https://github.com/substancelab/administrate/tree/collection_scopes.

It's a huge PR, though, and there are still some issues I need to weed out (layout and a few failing specs) - I guess that is to be expected with a year old feature branch that has lived through v0.2 
to 0.7, a redesign and whatnot 😀 

That said, I have no idea what the best approach is. Just submit a new PR with what I have? Somehow rebase and clean up the branch (that seems near-impossible)? Cherry-pick the relevant commits on top of master?

Also, I'd very much prefer to not step on @nando's toes here by stealing credit or ruffling feathers, please advice.

nickcharlton:
Hi @koppen!

When I pick up old PRs and the original author hasn't replied for a while, I usually take their branch, rebase it against current `master` (this will ensure their original authorship is retained) and open a new PR. We can close this one once you do that.

As it is quite big, if you think it could be broken down I'd do that and cherry pick a few commits over a collection of different PRs.

As long as you work with @nando's original commits (even if you break them apart a bit), the credit will stay.

koppen:
So I wasn't able to cherrypick my way to a working branch - at least not in any sensible manner - so I've ended up reimplementing some of @nando's original ideas on a fresh master. Pull request is at #947 .

nickcharlton:
I'm closing this as we've got #947 now.

Issue #232, opened 2015-11-12
Generate is not idempotent for routes


ariejan:
Running `rails generate administrate:install` more than once will add extra admin namespace blocks to `config/routes.rb`. Probably a low priority thing, but people using the generate command to just "reset" their work in progress might run into this.

Marthyn:
Yeah same stuff here, whenever i run a migration i like to update my dashboard automatically too. But i have to remove this every time
```
  namespace :admin do
    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
      resources dashboard_resource
    end

    root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
  end
```

graysonwright:
I think the easiest way to fix this would be checking to see if the routes are defined before generating them again. Thanks for the feedback!

graysonwright:
Fixed by #282.

PR #282, opened 2015-12-01
[WIP] Do not add routes when they already exist

Marthyn:
Simple fix for #232

Having trouble getting a spec going, since you check per line in the other specs there's no matcher for `contain_only_once` or something like that.


Marthyn:
Rebased on #285

graysonwright:
@Marthyn this looks great! Sorry for taking a while to look over it.

I'll play around and see if I can write a test for it, but this should be good to merge!

graysonwright:
Added a test and merged this in as 353f2c2. Thanks, @Marthyn!

PR #285, opened 2015-12-02
Update Nokogiri because of security problems

Marthyn:
In the CircleCi checks on my other PR #282 i saw that there are vulnerabilities for the current Nokogiri version in the Gemfile.lock

```
Name: nokogiri
Version: 1.6.6.2
Advisory: CVE-2015-1819
Criticality: Unknown
URL: https://github.com/sparklemotion/nokogiri/issues/1374
Title: Nokogiri gem contains several vulnerabilities in libxml2 and libxslt
Solution: upgrade to ~> 1.6.6.4, >= 1.6.7.rc4
```


graysonwright:
oh, damn. I didn't see this until just now, and I created and merged in #294 to do the same thing. I actually like this a bit better because it doesn't modify the `Gemfile`, but I don't think it's a big enough issue to change.

Thanks for your help!

PR #293, opened 2015-12-04
faster conditional key fetching from a hash

humzashah:
A hash's keys were being filtered in a slow way by iterating over the hash twice; this PR collects the keys conditionally in the first iteration.

Additionally, this PR has updated the version of `nokogiri` to get tests to pass on `CircleCI`.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -26,9 +26,15 @@ module Administrate
     end
 
     def search_attributes
-      resolver.dashboard_class::ATTRIBUTE_TYPES.select do |_, type|
-        type.searchable?
-      end.keys
+      [].tap do |keys|
+        attribute_types.each do |key, type|
+          keys << key if type.searchable?
+        end
+      end
+    end
+
+    def attribute_types
+      resolver.dashboard_class::ATTRIBUTE_TYPES
     end
 
     attr_reader :resolver, :term
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -26,10 +26,9 @@ module Administrate
     end
 
     def search_attributes
-      [].tap do |keys|
-        attribute_types.each do |key, type|
-          keys << key if type.searchable?
-        end
+      attribute_types.keys.select! do |attribute|
+        type = attribute_types[attribute]
+        type.searchable?
       end
     end
 
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -26,7 +26,7 @@ module Administrate
     end
 
     def search_attributes
-      attribute_types.keys.select! do |attribute|
+      attribute_types.keys.select do |attribute|
         type = attribute_types[attribute]
         type.searchable?
       end


graysonwright:
@humzashah Thanks!

A couple thoughts - I think this is a bit less readable than the `select {}` approach. Is the speed improvement significant enough to warrant the change? I think we'll be working with relatively small hashes here.

Also, we've upgraded the nokogiri version in #294. Can you rebase on top of that? Thanks!

graysonwright:
@mcmire, thoughts on this change?

humzashah:
@graysonwright the speed difference won't of course be noticeable for the current structure, but needless to say, this paves for a faster way in case `attribute_types` grows. :)

i've now rebased the change and opted for `tap` over `each_with_object` to reduce line-length and make  the code more readable. the principle is still the same: not to iterate over the `attribute_types` hash twice.

mcmire:
I agree with Grayson on this. While replacing Enumerable methods like `select`, etc. with `each` is definitely *an* optimization, I think it's a micro-optimization in this case -- even with, say, 50 attributes, you're not going to see that much of a speed increase. Because of this, I'd favor readability over speed here.

humzashah:
@graysonwright @mcmire, i hear what you're saying. i ran some benchmarks and indeed even with 50 attributes, the speed factor wasn't high enough to compromise readability. i sat on it a while and came up with, what i believe, is a readable yet faster version.

i've replaced `hash.select { code }.keys` with `hash.keys.select! { code }`. the former creates a new hash and then a new array. the latter creates a new array and modifies the same array, which is faster. IMO the latter is arguably more readable because the code is more idiomatic. to clarify:

**before**
select only searchable key-value pairs, then fetch keys from collection
`hash.select { |_, v| v.searchable? }.keys`

**now**
select keys that have searchable values
`hash.keys.select! { |k| hash[k].searchable? }`

graysonwright:
@humzashah cool, I definitely think this is more readable. Do you have benchmark numbers for the new code?

humzashah:
@graysonwright yup, created a gist to show exactly what i benchmarked, what speed results i got, and how much each implementation is faster from the current master branch's code.

https://gist.github.com/humzashah/fb36f07de89d222067bb

graysonwright:
Great, thanks! That's really helpful.

I've verified the benchmarks (and added another test case) at https://gist.github.com/graysonwright/90017f7d74f19e80b789. That gist also bumps the number of trials up to one million per test case.

Here are the results:

## Results

```
                     user     system      total        real        percentage saved
select{}.keys        15.060000   0.160000  15.220000 ( 15.360678)  0    %
keys.select{}        11.590000   0.140000  11.730000 ( 11.951267) 22.195%
keys.select!{}       10.600000   0.070000  10.670000 ( 10.841434) 29.420% 
[].tap { x << key }  10.550000   0.050000  10.600000 ( 10.721592) 30.201% 
each_with_object([]) 14.880000   0.030000  14.910000 ( 15.078155)  1.839% 
```

At this point, I'm torn between `keys.select{}` and `keys.select!{}`. The `keys.select{}` version, despite having a slower benchmark speed, has the benefit of not mutating any objects. It makes me a bit more confident when I'm reading the code that there are no unintended side effects.

The `keys.select!{}` version is faster in benchmarks, but I don't think it would be noticeable in practice.

@mcmire would love your input on this. Thanks!

humzashah:
Because `.keys` will be generating a new array every time, I'm indifferent to it being mutated. That 7% gap between `select` and `select!` is something to consider, though.

graysonwright:
Oh, I just saw @mcmire's comment above.

I'm worried that we're still on the verge of a premature optimization.

I've found some good thoughts on the topic from across the interwebs:

> Don Knuth started the literate programming movement because he believed that the most important function of computer code is to communicate the programmer's intent to a human reader. Any coding practice that makes your code harder to understand in the name of performance is a premature optimization.
- http://stackoverflow.com/a/385529/1857539

- - -

> When deciding whether to optimize a specific part of the program, Amdahl's Law should always be considered: the impact on the overall program depends very much on how much time is actually spent in that specific part, which is not always clear from looking at the code without a performance analysis.
- https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize

- - -

The Wikipedia page above links to "Amdahl's Law", which I think plays an important role here. I'm going to summarize the law in the context of our situation:

Let's estimate that a typical request for an Administrate dashboard takes 200 ms to run, and that this snippet of code is responsible for 2 ms (1%) of that time. Suppose we are able to improve the speed of this algorithm by 90%, so on a typical request, it only runs for 0.2 ms. The request still takes the same 198 ms to complete all of its other tasks, so the new response time is 198.02 ms after we make our optimizations. Over the entire request, we've only managed to improve the response time by `1 - (198.02 / 200) = 0.99%`, which is fairly negligible.

If a 90% optimization of this algorithm results in a 1% performance gain, then a 7% improvement is surely insignificant.

If we want to make effective optimizations, we'll need to examine the effects of the optimizations at the level of entire requests, not just in isolated benchmark scripts. Tools like [Rack Mini Profiler](https://github.com/MiniProfiler/rack-mini-profiler) can help a lot with this.

## Conclusion

I'm not willing to sacrifice code readability to improve performance, unless we can show that there is a significant improvement on the level of entire requests. Request-level profiles are our best measure of how users will experience our app. Improving our users' experience should be the primary goal of any optimization.

I do think that the `keys.select{}` version is just as readable as the existing version, so I'm okay with using it. It does speed up the algorithm a little bit; that's a nice side effect, but it isn't a large factor in the decision.

I've opened up #305, which adds some of these notes about performance improvements to the contributing guide.

humzashah:
thanks @graysonwright for your research, response, and analysis. i've checked out #305 and thumbs-up'ed it.

i've replaced the `select!` with `select`. i think we should be good to merge now since the change provides a readable, faster, **and** immutable version. :)

graysonwright:
@humzashah Thanks so much for all your work here! Wouldn't have gotten close to this conclusion without your benchmarking & experimentation.

Thanks for being flexible as I figured out how to approach these kinds of optimizations. I'm glad we figured out a good middle ground, and I'm happy to have a strategy for future optimizations.

Merged this in as feb0762. Cheers! :tada:

PR #295, opened 2015-12-05
Default ActiveRecord::Enum fields to String type instead of Number

### LINKED ISSUE #290: ActiveRecord enum types incorrectly interpreted by generator
[ActiveRecord enums](http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html) are integers in the database, but ActiveRecord translates them into the associated string value before returning them to Rails application code. The problem is that the administrate generator interprets them as integers by default, which means that any generated dashboards with enums error unless you manually change the attribute type to `Field::String` in each one.

latortuga:
This inspects the class to see if it has defined enums per the ActiveRecord 4.1+ API. If the column in question is an Enum, by default interpret it as a String field.

Fixes #290.

### DIFF
--- a/lib/administrate/fields/deferred.rb
+++ b/lib/administrate/fields/deferred.rb
@@ -15,7 +15,9 @@ module Administrate
       end
 
       def ==(other)
-        deferred_class == other.deferred_class && options == other.options
+        other.respond_to?(:deferred_class) &&
+          deferred_class == other.deferred_class &&
+          options == other.options
       end
 
       def searchable?
--- a/lib/generators/administrate/dashboard/dashboard_generator.rb
+++ b/lib/generators/administrate/dashboard/dashboard_generator.rb
@@ -7,6 +7,7 @@ module Administrate
         boolean: "Field::Boolean",
         date: "Field::DateTime",
         datetime: "Field::DateTime",
+        enum: "Field::String",
         float: "Field::Number",
         integer: "Field::Number",
         time: "Field::DateTime",
@@ -15,6 +16,7 @@ module Administrate
       }
 
       ATTRIBUTE_OPTIONS_MAPPING = {
+        enum: { searchable: false },
         float: { decimals: 2 },
       }
 
@@ -61,7 +63,7 @@ module Administrate
       end
 
       def field_type(attribute)
-        type = klass.column_types[attribute.to_s].type
+        type = column_type_for_attribute(attribute.to_s)
 
         if type
           ATTRIBUTE_TYPE_MAPPING.fetch(type, DEFAULT_FIELD_TYPE) +
@@ -69,6 +71,19 @@ module Administrate
         else
           association_type(attribute)
         end
+      end
+
+      def column_type_for_attribute(attr)
+        if enum_column?(attr)
+          :enum
+        else
+          klass.column_types[attr].type
+        end
+      end
+
+      def enum_column?(attr)
+        klass.respond_to?(:defined_enums) &&
+          klass.defined_enums.keys.include?(attr)
       end
 
       def association_type(attribute)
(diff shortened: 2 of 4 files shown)


graysonwright:
@latortuga this looks great! Thanks for the detailed test to go along with it. :smile: 

One thing that might be a problem is that `String` fields are searchable by default. I suspect that when we try searching on a model with an enum, it will generate a runtime error.

I'll comment on the diff with a possible solution.

lib/generators/administrate/dashboard/dashboard_generator.rb
```diff
@@ -71,6 +71,21 @@ def field_type(attribute)
         end
       end
 
+      def column_type_for_attribute(attr)
+        type = klass.column_types[attr].type
+
+        if type == :integer && enum_type?(attr)
+          :string
```
graysonwright: How about changing this to `:enum`, so we can treat enum fields and string fields a bit differently?

graysonwright: We can then add a new `enum: "Field::String",` entry to `ATTRIBUTE_TYPE_MAPPING`, so enums still get handled by the `String` field.

To turn off searching (until it's supported), we can change `ATTRIBUTE_OPTIONS_MAPPING` to:

```ruby
ATTRIBUTE_OPTIONS_MAPPING = {
  enum: { searchable: false },
  float: { decimals: 2 },
}
```

latortuga:
@graysonwright I'm making the changes for this, thanks for the feedback! Do you prefer a single rebased commit or multiple to show the history?

latortuga:
Inspected a few other PRs and it looks like you guys go for the rebased commit style (my preference as well, hooray!) so I went ahead and did that.

There was a slight issue getting the test to pass which involved updating the expect to set the correct options. The error message without the `with_options(...)` call is not very helpful:

```
  1) Administrate::Generators::DashboardGenerator dashboard definition file #attribute_types detects enum field as `String`
     Failure/Error: expect(attrs[:status]).
     NoMethodError:
       undefined method `deferred_class' for Administrate::Field::String:Class
     # ./lib/administrate/fields/deferred.rb:18:in `=='
     # ./spec/generators/dashboard_generator_spec.rb:144:in `block (4 levels) in <top (required)>'
```

Thoughts about adding a `self.deferred_class` method to `Administrate::Field::Base` that is nil? I erred on the side of not adding it because it would really only be a testing help and I try to avoid adding code to an app solely to help testing. But the error isn't very clear =\

graysonwright:
@latortuga yep, we rebase. We typically wait until discussion has ended and a PR is ready to be merged, but it's not a big deal.

As for that expectation error message, we can solve the problem without defining `self.deferred_class` on `Administrate::Field::Base`. I would suggest changing [`Administrate::Fields::Deferred#==`][1] so it raises a more intelligent error if `other` is not a `DeferredClass`. Want to add that in to this PR?

[1]: https://github.com/thoughtbot/administrate/blob/master/lib/administrate/fields/deferred.rb#L18

latortuga:
How about the change I pushed just now? With that in place, it's a lot clearer to me what is wrong and what will make it pass:

```
  1) Administrate::Generators::DashboardGenerator dashboard definition file #attribute_types detects enum field as `String`
     Failure/Error: expect(attrs[:status]).
       
       expected: Administrate::Field::String
            got: #<Administrate::Field::Deferred:0x007fb601546848 @deferred_class=Administrate::Field::String, @options={:searchable=>false}>
       
       (compared using ==)
       
       Diff:
       @@ -1,2 +1,4 @@
       -Administrate::Field::String
       +#<Administrate::Field::Deferred:0x007fb601546848
       + @deferred_class=Administrate::Field::String,
       + @options={:searchable=>false}>
       
     # ./spec/generators/dashboard_generator_spec.rb:144:in `block (4 levels) in <top (required)>'
```

lib/administrate/fields/deferred.rb
```diff
@@ -15,7 +15,9 @@ def new(*args)
       end
 
       def ==(other)
-        deferred_class == other.deferred_class && options == other.options
+        other.respond_to?(:deferred_class) &&
```
graysonwright: :clap: :+1:

graysonwright:
@latortuga That looks great! I think that's a perfect solution.

graysonwright:
Merged this in as 8d30944. Thanks!

PR #180, opened 2015-11-06
Add translations for ActiveRecord models

mgrachev:
Hi. Added translation support for ActiveRecord models in the sidebar and on the index page (header).

File locale must be of such content:
```yml
en:
  activerecord:
    models:
      bet: Rates
```

Issue #179, opened 2015-11-06
How to use with namespaced models.


gpinkham:
I have a user model that lives in Platform::User..   I ran the generator with administrate::dashboard Platform::User and this created a user_dashboard.rb and a users_controller.rb..     when I try to go to the users page in admin it fails saying 

```
uninitialized constant Admin::Platform
```

it points to this line in the controller: 

``` ruby
class Admin::Platform::UsersController < Admin::ApplicationController
```

I tried to remove the Platform:: from the controller and dashboard file but that just produced a similar error but this time saying User was an uninitiatized constant.

thanks!
Gary

tomgatzgates:
I'm looking for something similar. 

I want to create a set of name spaced models that extend my application models so I can add in Admin specific methods and not pollute my application models.

gpinkham:
I did some digging into the code..    the Resource Resolver uses the controller_path to create the class name..   for my Platform::User it was admin_users which in turn becomes User..   making it platform_users doesn't help as that becomes admin/platform_users and in turn PlatformUser not Platform::User..

gpinkham:
feels like it's going to be quite a change to make this work. .  every where I make a small change it leads down a path to some magic somewhere else in the gem.    make a different change and some other magic fails.     there's far too many assumptions that the classes do not have a namespace..    I will continue and see if I can get it working..

tomgatzgates:
I'm at a point where I've modified the resource resolver for the namespaced class. I.e. `models/admin/user.rb` which inherits from `models/user.rb`

Currently working on getting the `has_many` and `has_one` collection tables working. 
For example: User has many Orders
My error is `undefined method `order_path'`. I need to it be looking for `admin_order_path`

I'l let you know if I get anywhere. 

I'm struggling with `lib/administrate/fields/has_many.rb` `resources` method. I'm unsure where/what `data` is and where it comers from. If I knew that, I could change it to return `Admin::Order`.

gpinkham:
BTW I had the named spaced model working though not happy with the implementation at the moment..  feels quite hackish..  (i have committed it to my fork last Friday)..    Haven't tried any has many yet..  but will today.. I'm going to end up needing to handle namspaced models with nested forms.. so I'm sure I'll tackle a few things..   hoping this works..  tired of fighting with activeadmin.. :-)

graysonwright:
@tomgatzgates I'd love to see your modified `resource_resolver`.

[...]

PR #298, opened 2015-12-06
Support translations for ActiveRecord models

graysonwright:
## Problem:

Developers are unable to display their ActiveRecord classes in the UI by
any other name than the titleized class name.

For example, if a team is building an admin dashboard that will be used in
Russian, they may have ActiveRecord models with English names, that need to
be translated before they're displayed in the UI.

## Solution:

Rails has support for [translating ActiveRecord Models][1], which relies on
defining I18n translations under the `activerecord.models` namespace.

Rails expects translations in the form:

```yaml
en:
 activerecord:
   models:
     user:
       one: Customer
       other: Customers
```

The separate `one` and `other` keys allow Rails to display the correct
plural or singular form of the word, depending on the context.

We've used Rails' built-in [`ActiveModel::Name#human` method][2] to correctly pull
the translations out of I18n, with sane fallbacks if the translation is not
defined.

[1]: http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models
[2]: http://api.rubyonrails.org/classes/ActiveModel/Name.html#method-i-human

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -4,5 +4,17 @@ module Administrate
       locals.merge!(field: field)
       render locals: locals, partial: field.to_partial_path
     end
+
+    def display_resource_name(resource_name)
+      resource_name.
+        to_s.
+        classify.
+        constantize.
+        model_name.
+        human(
+          count: 0,
+          default: resource_name.to_s.pluralize.titleize,
+        )
+    end
   end
 end
--- a/app/views/administrate/application/_sidebar.html.erb
+++ b/app/views/administrate/application/_sidebar.html.erb
@@ -11,7 +11,7 @@ as defined by the DashboardManifest.
   <% DashboardManifest::DASHBOARDS.each do |resource| %>
     <li>
       <%= link_to(
-        resource.to_s.titleize,
+        display_resource_name(resource),
         [Administrate::NAMESPACE, resource],
         class: "sidebar__link sidebar__link--#{nav_link_state(resource)}"
       ) %>
--- a/spec/features/sidebar_spec.rb
+++ b/spec/features/sidebar_spec.rb
@@ -8,4 +8,25 @@ describe "sidebar" do
 
     expect(active_link.text).to eq "Customers"
   end
+
+  it "displays translated name of model" do
+    translations = {
+      activerecord: {
+        models: {
+          customer: {
+            one: "User",
+            other: "Users",
+          },
+        },
+      },
+    }
+
+    with_translations(:en, translations) do
+      visit admin_customers_path
+
+      sidebar = find(".sidebar__list")
+      expect(sidebar).to have_link("Users")
+      expect(page).to have_header("Users")
+    end
+  end
 end
--- a/spec/helpers/administrate/application_helper_spec.rb
+++ b/spec/helpers/administrate/application_helper_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+
+RSpec.describe Administrate::ApplicationHelper do
+  describe "#display_resource_name" do
+    it "defaults to the plural of the model name" do
+      displayed = display_resource_name(:customer)
+
+      expect(displayed).to eq("Customers")
+    end
+
+    it "handles string arguments" do
+      displayed = display_resource_name("customer")
+
+      expect(displayed).to eq("Customers")
+    end
+
+    it "handles plural arguments" do
+      displayed = display_resource_name(:customers)
+
+      expect(displayed).to eq("Customers")
+    end
+
+    context "when translations are defined" do
+      it "uses the plural of the defined translation" do
+        translations = {
+          activerecord: {
+            models: {
+              customer: {
+                one: "User",
+                other: "Users",
+              },
+            },
+          },
+        }
+
+        with_translations(:en, translations) do
+          displayed = display_resource_name(:customer)
+
+          expect(displayed).to eq("Users")
+        end
+      end
+    end
+  end
+end
--- a/spec/support/i18n.rb
+++ b/spec/support/i18n.rb
@@ -1,3 +1,14 @@
 RSpec.configure do |config|
   config.include AbstractController::Translation
+
+  def with_translations(locale, translations)
+    original_backend = I18n.backend
+
+    I18n.backend = I18n::Backend::KeyValue.new({}, true)
+    I18n.backend.store_translations(locale, translations)
+
+    yield
+  ensure
+    I18n.backend = original_backend
+  end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 
 * [#251] [FEATURE] Raise a helpful error when an attribute is missing from
   `ATTRIBUTE_TYPES`
+* [#298] [FEATURE] Support ActiveRecord model I18n translations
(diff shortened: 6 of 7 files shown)


spec/support/i18n.rb
```diff
@@ -1,3 +1,14 @@
...
+  def with_translations(locale, translations)
+    original_backend = I18n.backend
+
+    I18n.backend = I18n::Backend::KeyValue.new Hash.new, true
+    I18n.backend.store_translations locale, translations
+
+    yield
+  ensure
+    I18n.backend = original_backend
+  end
```
graysonwright: Source: http://stackoverflow.com/a/15634467/1857539

graysonwright:
This was continued from the work in #180.

mgrachev:
:+1:

app/helpers/administrate/application_helper.rb
```diff
@@ -4,5 +4,17 @@ def render_field(field, locals = {})
       locals.merge!(field: field)
       render locals: locals, partial: field.to_partial_path
     end
+
+    def display_resource_name(resource_name)
+      resource_name.
```
danbee: Should this just be `resource`?

danbee:
LGTM! :+1:

app/helpers/administrate/application_helper.rb
```diff
@@ -4,5 +4,17 @@ def render_field(field, locals = {})
...
+        to_s.
+        classify.
+        constantize.
+        model_name.
+        human(
+          count: 0,
```
JoelQ: Is this always 0? how do we handle plurals?

spec/support/i18n.rb
```diff
@@ -1,3 +1,14 @@
 RSpec.configure do |config|
   config.include AbstractController::Translation
+
+  def with_translations(locale, translations)
+    original_backend = I18n.backend
+
+    I18n.backend = I18n::Backend::KeyValue.new Hash.new, true
```
JoelQ: I'm curious why you used `Hash.new` instead of `{}`? I don't see `Hash.new` used very often

spec/features/sidebar_spec.rb
```diff
@@ -8,4 +8,22 @@
...
   end
+
+  it "displays translated name of model" do
+    translations = {
+      activerecord: {
+        models: {
+          customer: {
+            one: "User",
+            other: "Users",
+          } } } }
```
mcmire: What do you feel about putting every `}` on its own line? This looks rather Lisp-y.

spec/helpers/administrate/application_helper_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    context "when translations are defined" do
+      it "uses the plural of the defined translation" do
+        translations = {
+          activerecord: {
+            models: {
+              customer: {
+                one: "User",
+                other: "Users",
+              } } } }
```
mcmire: What do you feel about putting every `}` on its own line? This looks rather Lisp-y.

app/helpers/administrate/application_helper.rb
mcmire: I would think that "resource" would refer to the whole model object, but that's just me.

app/views/administrate/application/_sidebar.html.erb
```diff
@@ -11,7 +11,7 @@ as defined by the DashboardManifest.
   <% DashboardManifest::DASHBOARDS.each do |resource| %>
     <li>
       <%= link_to(
-        resource.to_s.titleize,
+        display_resource_name(resource),
```
mcmire: I honestly think that the mapping from a resource name to a model class shouldn't be inferred and should always be explicit. This would obviate the need to do `classify` and `constantize` above. I realize that doesn't have anything to do with this PR, but it would simplify the above logic at least.

spec/support/i18n.rb
```diff
@@ -1,3 +1,14 @@
 RSpec.configure do |config|
   config.include AbstractController::Translation
+
+  def with_translations(locale, translations)
```
mcmire: Nice method here, love that it takes a block.

mcmire:
I had one comment, and I would also agree with Joël about why `count: 0` is being used above. Other than these two things I'm good with this.

spec/helpers/administrate/application_helper_spec.rb
graysonwright: @mcmire I originally had them all on their own line, couldn't decide which I liked better. If you like them on their own line, I can revert those commits.

Each on their own line: https://github.com/thoughtbot/administrate/commit/2250b3d43283737a32242dbf85302dc2957dfaf7#diff-8f8d41bf7588d283cc64bf1d2ef099dcR22

app/helpers/administrate/application_helper.rb
graysonwright: @joelq this dictates how the resource names show up in the sidebar, and in the header on the index page. I think in both of those cases, we always want the displayed name to be in the plural form.

I started to code out an argument to this function that would let you specify a custom count, but I think YAGNI applies to that. If we need to display the singular form in the future, we can adjust this method accordingly.

spec/support/i18n.rb
graysonwright: oh, this was copy-pasted from Stack Overflow, and I missed that. Good catch!

app/views/administrate/application/_sidebar.html.erb
graysonwright: @mcmire completely agreed. I'm planning to change the `DashboardManifest::DASHBOARDS` constant to an array of constants instead of an array of symbols. This would let us support namespaced models (#32, #179), and hopefully get rid of a lot of name inferences.

app/helpers/administrate/application_helper.rb
graysonwright: Yeah, I'm siding with @mcmire on this. I've been using `resource` to refer to instances of the ActiveRecord models. Here, we're referring to the class names of the ActiveRecord models.

I think if we end up going with @mcmire's suggestion in https://github.com/thoughtbot/administrate/pull/298#discussion_r46846324, then we should rename this to `resource_class` or something.

spec/features/sidebar_spec.rb
```diff
@@ -8,4 +8,25 @@
...
+            other: "Users",
+          },
+        },
+      },
+    }
+
+    with_translations(:en, translations) do
+      visit admin_customers_path
+
+      sidebar = find(".sidebar__list")
```
JoelQ: Any reason this has double underscores?

JoelQ:
One comment on a CSS class name, otherwise LGTM :rocket:

spec/features/sidebar_spec.rb
graysonwright: We're using [BEM syntax][BEM] for our CSS. In this case, the CSS class would read as: "The list element inside the sidebar block", or something like that.

CSS selector formats are something that we could have a whole other discussion on - I'm still not sure what the best solution is for Administrate. I chose BEM because many designers at thoughtbot have had pleasant experiences working with it, and it makes it a bit easier to write maintainable (S)CSS.

[BEM]: http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/.

JoelQ: :+1:

mgrachev:
:tada:

PR #299, opened 2015-12-07
Ignore `.ruby-version` file

### LINKED ISSUE #289: Set ruby version for CircleCI
Hi, excuse me for return to discuss this.

On the issue #253 it was concluded that it was not necessary to add to "circle.yml" the version of ruby.

But I think so, because the ".ruby-version" file is used by "RVM" to declare "gemset" for example, and the name of this "gemset" is user preference.

So I think it is best set to "circle.yml" version of ruby and add to ".gitignore" the file ".ruby version" like another projects.

graysonwright:
Closes #289 

## Problem:

Different ruby version managers and installers
use different formats for the `.ruby-version` file.
Contributors may need to change the contents of that file
to get their own development environment working.
They do not want to commit these changes to version control
because they would affect other contributors.

## Solution:

Don't track the `.ruby-version` file in git.
Instead, have the `bin/setup` script
generate the suggested `.ruby-version` file,
and let contributors customize it freely on their own system.

The only pain point developers might feel is that
when the project updates the ruby version,
all contributors will need to manually update
their `.ruby-version` files.

This should not be much of a pain.
Ruby version updates are relatively infrequent,
and an incorrect `.ruby-version` setup
will raise a fatal error quickly,
with a clear error message.


JoelQ:
> Different ruby version managers and installers
> use different formats for the .ruby-version file.

I thought the whole point of a `.ruby-version` file was to be manager-agnostic?

mcmire:
I agree with the point Joël is making. If I recall correctly, `.ruby-version` was invented as a common format for multiple version managers. I do realize, and have seen in the past, that different version managers have different *names* for versions. However, now that Ruby is following SemVer (or at least following a common format for Ruby versions), I would not think this is much of an issue.

Was there a particular problem that you encountered that prompted this PR?

graysonwright:
@mcmire the issue came up in #289, where someone was using a gemset with a custom name with rvm.

Perhaps their issue could be resolved by using an `.rvmrc` file locally? I'm not sure if `.rvmrc` takes precedence over `.ruby-version` or not.

calrrox:
Hi again, @graysonwright 

I tested if the .rvmrc has priority over .ruby-version and it does.

When there are these two files, rvm always takes .rvmrc file, but I think this PR is still an option as it follows the conventions.

graysonwright:
@calrrox ok. Would adding `.rvmrc` to the `.gitignore` file, and keeping `.ruby-version` in git solve your problems?

I have a slight preference for keeping `.ruby-version` in git because it will make it easier for contributors to tell when we've upgraded the ruby version. Their ruby version manager should tell them that their out of date, and they won't need to wait for an error during `bundle install`.

calrrox:
@graysonwright  Yep, adding `.rvmrc`  to `.gitignore` specifically solves my problem.

And if keep `.ruby-version` in **git** could help others to know easily when the version of ruby changes then that is the best approach.

The point is make our lives easier to keep working.

JoelQ:
I'd much rather keep `.ruby-version` in version control as that is the standard thing to do on a Ruby-based project.

graysonwright:
@JoelQ there's a bit of disagreement around that...

https://github.com/rails/rails/blob/master/.gitignore#L5
https://github.com/activeadmin/activeadmin/blob/master/.gitignore#L24

graysonwright:
That being said, I think I'm on your side as well. I'd rather keep `.ruby-version` in git, because it at least conforms to thoughtbot's workflow. Ideally, it's ruby manager-agnostic, so it _should_ make it easier for others.

mcmire:
:+1:

latortuga:
RVM has support for using `.ruby-version` alongside [additional configuration](https://rvm.io/workflow/projects) (e.g. `.ruby-gemset` and `.ruby-env`) files to allow for interop between version managers. IMHO the `.ruby-version` file should stay in source control as it serves as both a functional helper (sets the ruby version for development mode for all version managers) and documentation.

graysonwright:
Alright, sounds like we're good to close this.

PR #304, opened 2015-12-08
Redirect from the root path to the admin root

graysonwright:
## Problem:

When users visit the root URL of the demo site,
they see the default Rails welcome page
with no links or references about where to go from there.

## Solution:

Automatically redirect from the root URL to the admin dashboard.


JoelQ:
:ship:

PR #306, opened 2015-12-08
Add break-word to preserve-whitespace

rubenmoya:
Before:

screen shot 2015-12-08 at 13 21 34

After:

screen shot 2015-12-08 at 13 18 49


graysonwright:
:+1: 

I'm a little concerned that there will be cases where people don't want the content to wrap (code or CSV text, for example), but I'm ok making the change and seeing if anyone objects.

graysonwright:
Merged in as 2057804. Thanks!

PR #307, opened 2015-12-08
Fix German grammar errors

clemensg:
Fix grammar errors in the German translations.

### DIFF
--- a/config/locales/administrate.de.yml
+++ b/config/locales/administrate.de.yml
@@ -2,7 +2,7 @@
 de:
   administrate:
     actions:
-      confirm: Sind sie sicher?
+      confirm: Sind Sie sicher?
       destroy: Löschen
       edit: Editieren
       show: Anzeigen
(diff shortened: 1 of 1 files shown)


graysonwright:
Thanks!

Can someone from the community verify these changes?

mhelmetag:
In German, nouns and pronouns are capitalized so this looks good to me. Also, at least in terms of the update to `has_one`'s `not_supported` [error message](https://github.com/thoughtbot/administrate/pull/307/files#diff-2a4193ec3ebf05d4dd889ff9723a0679L21), `polymorphism`'s has Beziehungen already capitalized so this PR looks like it's just getting the nouns and pronouns all nice and uniform. Hope that helps at all.

clemensg:
I'm a native speaker, by the way.

We could also translate HasOne relationship (to "Eins-zu-eins Beziehung") but that would probably be more confusing for developers, it's often better to keep technical keywords in English, in my opinion.

graysonwright:
Sounds good.

Merged this in as b4dfd34. Thanks!

PR #311, opened 2015-12-09
Generated nested Admin::ApplicationController

seanpdoyle:
The current generator for `Admin::ApplicationController` uses the
compact style, which is a styleguide violation for our Rubopcop (and
transitively Hound) configuration.


graysonwright:
LGTM

PR #312, opened 2015-12-09
Make BelongsTo fields nullable

seanpdoyle:
Prior to this commit, there was no way to set a `belongs_to`
relationship to `nil` through the UI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * [#251] [FEATURE] Raise a helpful error when an attribute is missing from
   `ATTRIBUTE_TYPES`
 * [#298] [FEATURE] Support ActiveRecord model I18n translations
+* [#312] [FEATURE] Add a `nil` option to `belongs_to` form fields
 * [#231] [UI] Fix layout issue on show page where a long label next to an empty
   value would cause following fields on the page to be mis-aligned.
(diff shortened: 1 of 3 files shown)


graysonwright:
Good idea.

For the implementation, what do you think about leaving `associated_resource_options`, and changing `candidate_resources` to:

```ruby
  def candidate_resources
    if options[:allow_nil]
      [nil] + associated_class.all
    else
      associated_class.all
    end
  end
```

Also, can you add a test to `spec/lib/fields/has_many_spec.rb`?

seanpdoyle:
@graysonwright is this related to `Fields::HasMany`?

seanpdoyle:
@graysonwright do we need an `allow_nil` option?

If we assume that all `BelongsTo` relations are nullable, the API would stay small and focussed, and any validation failures would be surfaced like normal (through `form_for` integration with ActiveModel validations)

graysonwright:
Yeah, good point.

I'm :+1: on this - merge away.

As for how `Field::BelongsTo` is related to `Field::HasMany`, they share the same relationship as the ActiveRecord relationships - they can be used independently of one another, but they're usually used as inverses of each other. They both inherit from [`Administrate::Field::Associative`](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/fields/associative.rb).

PR #315, opened 2015-12-09
Remove has_many `_ids` suffix from form labels

### LINKED ISSUE #303: Allow Easier Editing of Labels For HasMany Field
I have a user with many Roles, and on the edit page, I see a label saying `Role ids` when I would rather see the label say `Roles` instead. The only way I see to make this minor edit is by creating my own custom field which would duplicate most of Field::HasMany. Is there an easier way, am I missing something?

graysonwright:
Fixes #303

## Problem:

When HasMany attributes are displayed in form fields,
their labels include the `_ids` suffix that Rails uses internally
(e.g. `line_item_ids`, `order_ids`).

edit_order__3___administrate_prototype

Users expect to see the pluralized form of the relationship name
(e.g. `line_items`, `orders`).

edit_order__3___administrate_prototype

## Solution:

Pass in the desired label value explicitly instead of letting Rails
guess at the correct label.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@
 * [#214] [UI] Improve header layout when there is a long page title
 * [#198] [UI] Improve spacing around bottom link in sidebar
 * [#206] [UI] Left-align checkboxes in boolean form fields
+* [#315] [UI] Remove the `IDS` suffix for `HasMany` form field labels
 * [#259] [BUGFIX] Make installation generator more robust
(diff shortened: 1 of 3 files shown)


seanpdoyle:
:clap: can we cover this through tests?

graysonwright:
Done. This is probably the largest test/implementation code ratio I've ever had.

seanpdoyle:
LGTM

PR #317, opened 2015-12-09
Add specs for Administrate::Search class

latortuga:
This class had no spec coverage and I'd like to start implementing search for Enum fields. Can't have new features without covering our existing ones :smile:

### DIFF
--- a/spec/lib/administrate/search_spec.rb
+++ b/spec/lib/administrate/search_spec.rb
@@ -15,9 +15,9 @@ end
 
 describe Administrate::Search do
   describe "#run" do
-    let(:resolver) { double(
-      resource_class: User,
-      dashboard_class: MockDashboard) }
+    let(:resolver) do
+      double(resource_class: User, dashboard_class: MockDashboard)
+    end
 
     it "returns all records when no search term" do
       begin
--- a/spec/lib/administrate/search_spec.rb
+++ b/spec/lib/administrate/search_spec.rb
@@ -22,7 +22,7 @@ describe Administrate::Search do
     it "returns all records when no search term" do
       begin
         class User; end
-
+        resolver = double(resource_class: User, dashboard_class: MockDashboard)
         search = described_class.new(resolver, nil)
         expect(User).to receive(:all)
 
@@ -35,6 +35,7 @@ describe Administrate::Search do
     it "returns all records when search is empty" do
       begin
         class User; end
+        resolver = double(resource_class: User, dashboard_class: MockDashboard)
         search = described_class.new(resolver, "   ")
         expect(User).to receive(:all)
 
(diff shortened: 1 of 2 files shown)


graysonwright:
:clap: :clap: :tada: :confetti_ball: 

This is great. Thanks!! :smile: 

I've been wanting coverage of this for a while, thanks for filling it in. I'm going to leave a couple suggestions so it fits thoughtbot's style a bit more, but overall this looks great.

spec/lib/administrate/search_spec.rb
```diff
@@ -0,0 +1,72 @@
+require "spec_helper"
```
graysonwright: Nice job not requiring `rails_helper` here, that's a big win.

spec/lib/administrate/search_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+class MockDashboardClass
+  ATTRIBUTE_TYPES = {
+    name: Administrate::Field::String,
+    email: Administrate::Field::Email,
+    phone: Administrate::Field::Number,
+  }
+end
+
+class MockResolver
```
graysonwright: Instead of defining `MockResolver`, what do you think about using a double?

```ruby
resolver = double(resource_class: User, dashboard_class: MockDashboardClass)
```

That would let us get around defining an entire `MockResolver` class, and would be more flexible if we wanted to change the behavior of the resolver on a test-by-test basis.

latortuga: 100% on board with that. I typically use Mocha as my mocking/stubbing gem of choice so when the `mock` method wasn't available, I went for a dummy class instead. I'll update this to use a double.

spec/lib/administrate/search_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  ATTRIBUTE_TYPES = {
+    name: Administrate::Field::String,
+    email: Administrate::Field::Email,
+    phone: Administrate::Field::Number,
+  }
+end
+
+describe Administrate::Search do
+  describe "#run" do
+    let(:resolver) do
```
graysonwright: [thoughtbot's guides][guides] recommend avoiding `let` - it can be easy to lose track of where variables are being defined in specs that have a lot of `let` definitions, because it removes variables from the specs in which they're used.

[Here's an example of an alternative][avoid let]. How would you feel about moving the variable definition into the individual specs?

[guides]: https://github.com/thoughtbot/guides
[avoid let]: https://github.com/thoughtbot/guides/blob/master/style/testing/avoid_let_spec.rb

latortuga: Absolutely, feel free to point out anywhere that you'd prefer changed to match your style. I'll get the hang of it.

graysonwright:
Looks great! I'll merge this in.

It doesn't have any user-facing changes, so we can skip the CHANGELOG entry for it.

graysonwright:
Merged this in as e247bd6. Looking forward to enum searching!

PR #321, opened 2015-12-11
allow new action accept params

lazing:
allow new action accept params


graysonwright:
Hey, @lazing! Thanks for the contribution. I have a couple questions:

- How will you pass values into the request?
- What will this feature look like from the user's perspective?

lazing:
Hey, @graysonwright here is my scenario:
i want create order for specific user, and i got lots of users(10000+).
so i need to search and locate a user, open the show page and click `create order`.

The link with the params `order: {user_id: xxx}`, so the order is related to the user

nickcharlton:
Hi @lazing!

I'm just coming to your PR for real now. Could you clarify how this would be used? I don't feel like I understand from your example.

Would you also be able to rebase this on top of `master`?

nickcharlton:
Closing as it's been a while since the last response.

PR #322, opened 2015-12-11
Add search support for ActiveRecord::Enum columns

latortuga:
Bit of a work in progress/hack. Map all searchable fields into a hash and
allow each field class to customize the search predicate and comparison
term based on the passed search string. If no matching enum value is
found in the search term (based on == compare), omit enum field from
search string. Otherwise, try to map it to the correct integer value.

I think this needs a new class of some kind, perhaps something like `SearchField`, instantiated by either the Field class itself or the Deferred class which would avoid having to have the ugly class methods on the fields and the arity difference for enums only. I'm seriously open to design feedback on this one because, while It Works<sup>:tm:</sup>, it's not exactly nice looking right now.

### DIFF
--- a/app/views/fields/enum/_form.html.erb
+++ b/app/views/fields/enum/_form.html.erb
@@ -1,0 +1,19 @@
+<%#
+# Enum Form Partial
+
+This partial renders an input element for a string attribute.
+By default, the input is a text field.
+
+## Local variables:
+
+- `f`:
+  A Rails form generator, used to help create the appropriate input fields.
+- `field`:
+  An instance of [Administrate::Field::Enum][1].
+  A wrapper around the Enum pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Enum
+%>
+
+<%= f.label field.attribute %>
+<%= f.select field.attribute, field.enum_options %>
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -9,6 +9,7 @@ require "administrate/fields/number"
 require "administrate/fields/polymorphic"
 require "administrate/fields/string"
 require "administrate/fields/text"
+require "administrate/fields/enum"
 
 module Administrate
   class BaseDashboard
--- a/lib/administrate/fields/base.rb
+++ b/lib/administrate/fields/base.rb
@@ -16,11 +16,11 @@ module Administrate
         false
       end
 
-      def self.search_predicate(attribute, term)
+      def self.search_predicate(attribute, _term)
         "lower(#{attribute}) LIKE ?"
       end
 
-      def self.search_term(attribute, term)
+      def self.search_term(_attribute, term)
         "%#{term}%"
       end
 
--- a/lib/administrate/fields/base.rb
+++ b/lib/administrate/fields/base.rb
@@ -16,6 +16,14 @@ module Administrate
         false
       end
 
+      def self.search_predicate(attribute, term)
+        "lower(#{attribute}) LIKE ?"
+      end
+
+      def self.search_term(attribute, term)
+        "%#{term}%"
+      end
+
       def initialize(attribute, data, page, options = {})
         @attribute = attribute
         @data = data
--- a/spec/example_app/app/dashboards/order_dashboard.rb
+++ b/spec/example_app/app/dashboards/order_dashboard.rb
@@ -14,7 +14,7 @@ class OrderDashboard < Administrate::BaseDashboard
     line_items: Field::HasMany,
     total_price: Field::Number.with_options(prefix: "$", decimals: 2),
     shipped_at: Field::DateTime,
-    status: Field::Enum,
+    status: Field::Enum.with_options(enum: {'pending' => 0, 'confirmed' => 1}),
   }
 
   READ_ONLY_ATTRIBUTES = [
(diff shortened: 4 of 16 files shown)


spec/example_app/db/schema.rb
```diff
@@ -59,9 +59,10 @@
     t.string   "address_city"
     t.string   "address_state"
     t.string   "address_zip"
-    t.datetime "created_at",       null: false
-    t.datetime "updated_at",       null: false
+    t.datetime "created_at",                   null: false
+    t.datetime "updated_at",                   null: false
     t.datetime "shipped_at"
+    t.integer  "status",           default: 0, null: false
```
latortuga: Not really sure how to fix this one, generated file?

graysonwright:
@latortuga Thanks!

You're right, I think we'll need a deeper restructuring of the Search functionality. I'll take a look at this code tomorrow and let you know if I see any low-hanging fruit.

app/views/fields/enum/_form.html.erb
```diff
@@ -0,0 +1,19 @@
...
+## Local variables:
+
+- `f`:
+  A Rails form generator, used to help create the appropriate input fields.
+- `field`:
+  An instance of [Administrate::Field::Enum][1].
+  A wrapper around the Enum pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Enum
+%>
```
graysonwright: :clap: awesome. Thanks for following the comment conventions!

lib/administrate/base_dashboard.rb
```diff
@@ -9,6 +9,7 @@
 require "administrate/fields/polymorphic"
 require "administrate/fields/string"
 require "administrate/fields/text"
+require "administrate/fields/enum"
```
graysonwright: In lists like this where there's no obvious order, thoughtbot prefers to sort things alphabetically. Would you mind doing that here?

lib/administrate/fields/base.rb
```diff
@@ -16,6 +16,14 @@ def self.searchable?
         false
       end
 
+      def self.search_predicate(attribute, _term)
+        "lower(#{attribute}) LIKE ?"
```
graysonwright: What happens if we return the string `"false"` here?

I'm just guessing, but that might mean we can get rid of the `self.searchable?` method. The `false` would get inserted into the search string as something like this:

```sql
WHERE LOWER(string_field) LIKE search_term OR false OR enum_field = 1
```

The `OR false` clause wouldn't have any effect on the query.

We could move this search predicate definition into `Administrate::Field::String`, to keep that working.

As for the search term, we could change it to `nil`, and then run `.uniq` on the array of search terms to remove nils before passing it into the search.

That's all a guess, but it would be pretty cool if it worked. What do you think of that approach? There may be downsides that I've missed.

spec/example_app/db/schema.rb
graysonwright: yeah, you can just ignore that. I opened up https://github.com/thoughtbot/administrate/pull/336 to fix this in the future.

latortuga:
So, the comment about including all fields by using the string `"false"` is interesting. In principle you're right, it won't cause any issues with the results. My biggest concern with removing `searchable?` would be the inability to opt out of search for specific columns.

graysonwright:
Hm... yeah, I think you're right. We could change it around so it would still work, but I think that would complicate things a lot.

I guess it also makes sense to keep the current search definition on `Field::Base`, because both the `String` and `Email` fields use the same search logic there.

spec/example_app/app/dashboards/order_dashboard.rb
```diff
@@ -14,6 +14,8 @@ class OrderDashboard < Administrate::BaseDashboard
     line_items: Field::HasMany,
     total_price: Field::Number.with_options(prefix: "$", decimals: 2),
     shipped_at: Field::DateTime,
+    status: Field::Enum.with_options(
+      enum: { "pending" => 0, "confirmed" => 1 }),
```
graysonwright: I'm not thrilled with needing to pass in the enum values to the field, because it's duplicating the logic between the models and the dashboard. I'm not sure I know of a better way, though.

graysonwright: Looks like `Order.statuses` returns: `{"pending"=>0, "confirmed"=>1}`. Maybe we can use that to avoid the `.with_options` call.

Here's the reference: http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

latortuga: I am also not thrilled about it but the Field class is all we have available at the time of search construction. The instantiated object of that class isn't available until we're actually showing something in the view. Another option would be to annotate the Field class with which AR class it is for, as this would give us the ability to reflect on the enum keys/values from the Field class context. I think that might be a pretty big api change though.

latortuga:
`Order.statuses` is actually the API I'm using to generate the options in the dashboard_generator because that's where we have access to the class involved:

```ruby
base_options.merge(enum: klass.send(attr.pluralize)) # e.g. Order#status => Order.statuses
```

If, instead, we instantiated Field classes with the class+attribute that they are for instead of just the attribute, we could reflect on this from the Field object without having to pass them in using `with_options`. WDYT about a possible API to pass along the associated class for an attribute? Or is there some other way that a Field can know what class it belongs to?

nickcharlton:
Hi @latortuga!

I know it's been quite a while, but would you be willing to rebase this on the current master?

thecalliopecrashed:
This needs a rebase and clean-up.

PR #324, opened 2015-12-12
Generate layout with `rails g administrate:views`

graysonwright:
## Problem:

When users run `rails generate administrate:views`,
they expect all of Administrate's views to be generated.
Currently, the layout views and partials
are not generated through that command,
so users must run `rails generate administrate:views:layout` separately.

## Solution:

Automatically generate the layout files
when the user runs `rails generate adminsitrate:views`

### DIFF
--- a/lib/generators/administrate/views/views_generator.rb
+++ b/lib/generators/administrate/views/views_generator.rb
@@ -4,6 +4,7 @@ module Administrate
   module Generators
     class ViewsGenerator < Administrate::ViewGenerator
       def copy_templates
+        Rails::Generators.invoke("administrate:views:layout")
         Rails::Generators.invoke("administrate:views:index", [resource_path])
         Rails::Generators.invoke("administrate:views:show", [resource_path])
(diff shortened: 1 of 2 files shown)


graysonwright:
From Slack:

grayson
> @danbarber, @joel, @joelq: What do you think about https://github.com/thoughtbot/administrate/pull/324 ? I could go either way.

joel
> @grayson: I would expect the layouts to be included. When I think "views" I would lump layouts in there with em

danbarber
> I’m undecided really, but it would seem right to generate everything
> You can always remove stuff you don’t need to modify

grayson
> I think I might only generate the views if the user doesn't pass in an argument.
> If I type in `rails g administrate:views User`, then I think I would only expect the User's index, show, etc. to be generated.
> but it's a bit strange to change up the behavior like that.

danbarber
> I think that’s fine. Scoping views to a model I would expect it to leave out the layout

graysonwright:
I'll let the community vote on this. I'll close it for now because it's not pressing, and if people ask for this we can re-open and merge it.

PR #325, opened 2015-12-12
create a password field

acrogenesis:
Basically is almost like the String field but has a password_field and the truncate defaults to 8 and uses `*` for displaying the text

### DIFF
--- a/app/views/fields/password/_form.html.erb
+++ b/app/views/fields/password/_form.html.erb
@@ -1,18 +1,18 @@
 <%#
-# String Form Partial
+# Password Form Partial
 
-This partial renders an input element for a string attribute.
-By default, the input is a text field.
+This partial renders a password element for an attribute.
+By default, the input is a password field.
 
 ## Local variables:
 
 - `f`:
   A Rails form generator, used to help create the appropriate input fields.
 - `field`:
-  An instance of [Administrate::Field::String][1].
-  A wrapper around the String pulled from the database.
+  An instance of [Administrate::Field::Password][1].
+  A wrapper around the Password pulled from the database.
 
-[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Password
 %>
 
 <%= f.label field.attribute %>
--- a/app/views/fields/password/_form.html.erb
+++ b/app/views/fields/password/_form.html.erb
@@ -1,0 +1,19 @@
+<%#
+# String Form Partial
+
+This partial renders an input element for a string attribute.
+By default, the input is a text field.
+
+## Local variables:
+
+- `f`:
+  A Rails form generator, used to help create the appropriate input fields.
+- `field`:
+  An instance of [Administrate::Field::String][1].
+  A wrapper around the String pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
+%>
+
+<%= f.label field.attribute %>
+<%= f.password_field field.attribute %>
--- a/app/views/fields/password/_form.html.erb
+++ b/app/views/fields/password/_form.html.erb
@@ -14,6 +14,9 @@ By default, the input is a password field.
 
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Password
 %>
-
-<%= f.label field.attribute %>
-<%= f.password_field field.attribute %>
+<div class="field-unit__label">
+  <%= f.label field.attribute %>
+</div>
+<div class="field-unit__field">
+  <%= f.password_field field.attribute %>
+</div>
--- a/app/views/fields/password/_index.html.erb
+++ b/app/views/fields/password/_index.html.erb
@@ -1,0 +1,18 @@
+<%#
+# String Index Partial
+
+This partial renders a string attribute
+to be displayed on a resource's index page.
+
+By default, the attribute is rendered as a truncated string.
+
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::String][1].
+  A wrapper around the String pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
+%>
+
+<%= field.truncate %>
--- a/app/views/fields/password/_index.html.erb
+++ b/app/views/fields/password/_index.html.erb
@@ -1,18 +1,18 @@
 <%#
-# String Index Partial
+# Password Index Partial
 
-This partial renders a string attribute
+This partial renders a password attribute
 to be displayed on a resource's index page.
 
-By default, the attribute is rendered as a truncated string.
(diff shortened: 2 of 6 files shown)


graysonwright:
Nice, thanks! I'll leave a few suggestions.

app/views/fields/password/_form.html.erb
```diff
@@ -0,0 +1,19 @@
...
+
+## Local variables:
+
+- `f`:
+  A Rails form generator, used to help create the appropriate input fields.
+- `field`:
+  An instance of [Administrate::Field::String][1].
+  A wrapper around the String pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
```
graysonwright: This comment has a number of references to `String`. Can we change them all to `Password`?

app/views/fields/password/_index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+By default, the attribute is rendered as a truncated string.
+
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::String][1].
+  A wrapper around the String pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
```
graysonwright: This comment has a number of references to `String`. Can we change them all to `Password`?

app/views/fields/password/_show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+By default, the attribute is rendered as an unformatted string.
+
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::String][1].
+  A wrapper around the String pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
```
graysonwright: This comment has a number of references to `String`. Can we change them all to `Password`?

lib/administrate/base_dashboard.rb
```diff
@@ -8,6 +8,7 @@
 require "administrate/fields/number"
 require "administrate/fields/polymorphic"
 require "administrate/fields/string"
+require "administrate/fields/password"
```
graysonwright: When a list like this doesn't have any inherent order, thoughtbot prefers to sort them alphabetically. Would you mind doing that here?

app/views/fields/password/_index.html.erb
```diff
@@ -0,0 +1,18 @@
+<%#
+# String Index Partial
+
+This partial renders a string attribute
+to be displayed on a resource's index page.
+
+By default, the attribute is rendered as a truncated string.
```
graysonwright: This description no longer applies. It should probably mention the asterisks.

app/views/fields/password/_show.html.erb
```diff
@@ -0,0 +1,18 @@
+<%#
+# String Show Partial
+
+This partial renders a string attribute,
+to be displayed on a resource's show page.
+
+By default, the attribute is rendered as an unformatted string.
```
graysonwright: This description no longer applies.

lib/administrate/fields/password.rb
```diff
@@ -0,0 +1,25 @@
+require_relative "base"
+
+module Administrate
+  module Field
+    class Password < Field::Base
+      def self.searchable?
+        false
+      end
+
+      def truncate
```
graysonwright: I'm not sure anyone would have a reason to pass in a truncation length. It seems like we could either display a fixed number of asterisks if the password exists, or an empty string if the password doesn't exist. Thoughts?

spec/lib/fields/password_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      long_pwd = Administrate::Field::Password.new(:long_pwd, lorem(30), :show)
+
+      expect(long_pwd.truncate).to eq("*" * 8)
+    end
+
+    context "with a `truncate` option" do
+      it "shortens to the given length" do
+        password = password_with_options(lorem(30), truncate: 20)
+
+        expect(password.truncate).to eq("*" * 20)
```
graysonwright: I'm not sure anyone would have a reason to pass in a truncation length. It seems like we could either display a fixed number of asterisks if the password exists, or an empty string if the password doesn't exist. Thoughts?

acrogenesis:
Oops sorry for the silly mistakes.

I think you are right but I'm not sure how many asterisks should we use. I set it to 8 because is the typical minimum length but it could be 4..

graysonwright:
@acrogenesis I'd say 6 or 8 - it probably just depends on what looks better in the UI. Can you post screenshots of both?

Also, what do you think about using the HTML entity code `&#149;` instead of asterisks? It would look like: &#149;&#149;&#149;&#149;&#149;&#149;

coneybeare:
I vote 8 and `&#149;`, both configurable via `.with_options`

zamith:
Why do you think it is important o be able to configure it? Do you think it will break someone's design? I don't think there will be a business rule saying not to use 8 dots for passwords on the admin dashboard, do you?

coneybeare:
I don't like to make any assumptions about the configuration options of others. Configurable will let people set it up the way they want.

zamith:
That is true, but we also don't want to have to handle a lot of configuration by adding options and DSLs, that's kind of the whole point with administrate. If there's an easier way to allow this to be overridden without having to add any other options that, I think, would be preferable. Do you agree?

coneybeare:
I don't entirely agree, but I see your point and how it aligns with the overall goals of the repo. I can live with sensible defaults.

acrogenesis:
Here are screenshots with 6 & 8 •




As for the truncate option I think it should be there not because of design but for consistency on the field types

graysonwright:
Whoa, I love how that looks!

I think I'd vote for 6 dots - it takes up less space, and still conveys the same meaning. I think it looks a bit less cluttered, as well.

For `truncate`, I'd vote to remove the option. We should be adding options & behavior to our fields based on the data underlying the field, not based on the options & behavior of other fields. We won't be defining `truncate` on `Number` or `Image` fields, because they don't need it. By the same logic, it's unnecessary to add that option here.

acrogenesis:
Hmm that's right better to add behavior based on the field and not on the rest of the fields

ACPK:
@graysonwright Would love to see this!

acrogenesis:
@graysonwright this is ready for merge :+1:

acrogenesis:
fyi I just updated the form field for the refactor from d0bc1c7d2fa7327d89d2e0cd50fba8f6c31a6b26

acrogenesis:
@graysonwright should I move this to a seperate gem? Maybe `administrate-field-password`?

acrogenesis:
Here's the gem https://github.com/DisruptiveAngels/administrate-field-password

graysonwright:
@acrogenesis sorry for not getting to this sooner.

Thanks for breaking it out into a separate gem! I'll take a look at adding it to the `gemspec` as a dependency. :smile:

PR #326, opened 2015-12-13
Add generators for asset files

### LINKED ISSUE #187: Custom javascript
@graysonwright How do we add javascripts to the tree for custom fields? I put a ckeditor.js file that initializes ckeditor in:

`app/assets/javascripts/administrate`

However, it's not being required by the master administrate application.js

Any ideas? Cheers!

graysonwright:
Closes #187

## Problem:

Many people want to customize the stylesheets and javascripts
that get applied to Administrate dashboards.

To do that, they must copy over the default assets
from Adminsitrate's source into their own host application,
and edit those files to add custom behavior or styles.

This approach is error-prone and not approachable.

## Solution:

Add a generator for copying over default styles and javascripts
into the user's host application.

## Usage:

```
 # Copy image assets into `app/assets/images/administrate`
rails generate administrate:assets:images

 # Copy javascript assets into `app/assets/javascripts/administrate`
rails generate administrate:assets:javascripts

 # Copy stylesheet assets into `app/assets/stylesheets/administrate`
rails generate administrate:assets:stylesheets

 # Runs all of the above generators
rails generate administrate:assets
```

## ToDo:

- [x] Actually run these generators and edit the files to verify they work.

### DIFF
--- a/lib/generators/administrate/assets/assets_generator.rb
+++ b/lib/generators/administrate/assets/assets_generator.rb
@@ -1,0 +1,13 @@
+require "administrate/view_generator"
+
+module Administrate
+  module Generators
+    class AssetsGenerator < Administrate::ViewGenerator
+      def copy_assets
+        Rails::Generators.invoke("administrate:assets:images")
+        Rails::Generators.invoke("administrate:assets:javascripts")
+        Rails::Generators.invoke("administrate:assets:stylesheets")
+      end
+    end
+  end
+end
--- a/spec/generators/assets/images_generator_spec.rb
+++ b/spec/generators/assets/images_generator_spec.rb
@@ -1,0 +1,20 @@
+require "rails_helper"
+require "generators/administrate/assets/images_generator"
+require "support/generator_spec_helpers"
+
+describe Administrate::Generators::Assets::ImagesGenerator, :generator do
+  describe "administrate:assets:images" do
+    it "copies the images directory into the `administrate` namespace" do
+      images_directory = "app/assets/images/administrate/**/*"
+
+      run_generator []
+
+      each_file_in(images_directory) do |copied_file_path|
+        expected_contents = File.read(copied_file_path)
+        contents = File.read(file(copied_file_path))
+
+        expect(contents).to eq(expected_contents)
+      end
+    end
+  end
+end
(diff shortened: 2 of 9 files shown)


graysonwright:
One drawback of this implementation is that the `administrate:assets` generator doesn't work for Rails apps where the app is not at the root of the repository.

When I run `rails generate administrate:assets`, I expect the files to be generated into `spec/example_app/app/assets/`. However, they are instead generated into `app/assets/`.

1__tmux

I tried specifying `Rails.root` (see b1f557f and 622aa1a), but that isn't supported by Ammeter, which is what we're using to test the generators.

I expect this to be an incredibly uncommon use case, so I'm OK pushing this forward.

lib/generators/administrate/assets/assets_generator.rb
```diff
@@ -0,0 +1,13 @@
+require "administrate/view_generator"
+
+module Administrate
+  module Generators
+    class AssetsGenerator < Administrate::ViewGenerator
+      def copy_templates
```
graysonwright: rename to `copy_assets`

spec/generators/assets/images_generator_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+describe Administrate::Generators::Assets::ImagesGenerator, :generator do
+  describe "administrate:assets:images" do
+    it "copies the images directory into the `administrate` namespace" do
+      files = Dir.glob("app/assets/images/administrate/**/*")
+
+      run_generator []
+
+      files.each do |copied_file_path|
+        if File.file?(copied_file_path)
```
graysonwright: maybe extract a method, `expect_identical_directories`?

spec/generators/assets_generator_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+      run_generator []
+
+      %w[images javascripts stylesheets].each do |asset|
+        expect(Rails::Generators).
+          to invoke_generator("administrate:assets:#{asset}")
+      end
+    end
+  end
+
+  def invoke_generator(*args)
```
graysonwright: Move to `support/generator_spec_helpers`

rikkipitt:
Cheers @graysonwright, I'll check this out soon. Does it help the production of custom plugins like we were talking about before too?

graysonwright:
@rikkipitt sorry for the late response.

This is a step in that direction. We're also working on #476, which will really help out with custom plugins.

PR #328, opened 2015-12-14
Add a generator for the sidebar

acrogenesis:
I needed to edit the sidebar and noticed there wasn't a generator so I created one.
Run 
```ruby
rails g administrate:views:sidebar
```
 To generate it, already verified on my app it works and tests are :white_check_mark:

### DIFF
--- a/spec/support/generator_spec_helpers.rb
+++ b/spec/support/generator_spec_helpers.rb
@@ -15,6 +15,10 @@ module GeneratorSpecHelpers
     )
   end
 
+  def invoke_generator(*args)
+    have_received(:invoke).with(*args)
+  end
+
   private
 
   def copy_to_generator_root(destination, template)
--- a/lib/generators/administrate/views/layout_generator.rb
+++ b/lib/generators/administrate/views/layout_generator.rb
@@ -12,7 +12,7 @@ module Administrate
             "app/views/layouts/admin/application.html.erb",
           )
 
-          copy_resource_template("_sidebar")
+          Rails::Generators.invoke("administrate:views:sidebar")
           copy_resource_template("_javascript")
           copy_resource_template("_flashes")
         end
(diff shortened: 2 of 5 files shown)


graysonwright:
acrogenesis, thanks!

I merged in #269 yesterday, which has a lot of overlap with this. I think there's room for both of:

```
rails generate administrate:views:layout
rails generate administrate:views:sidebar
```

I think we'll want to change the layout generator so it calls this generator, instead of duplicating the logic.

If you want to look into this, great! Otherwise I can take care of this when I merge it.

acrogenesis:
I'm not entirely sure how to invoke the generator from the LayoutGenerator I'm trying to do it calling `Rails::Generators.invoke("administrate:views:sidebar")` but the tests won't pass

graysonwright:
Yep, that's the right way to do it. The tests won't execute the sub-generator, so you'll have to stub it like I did here:

https://github.com/thoughtbot/administrate/blob/7e8943d1867539c200fdf76ac4650fa6234bba9a/spec/generators/assets_generator_spec.rb#L13-L15

(note the `invoke_generator` method defined below.

acrogenesis:
thanks! I've updated the generator and tests. I also added the `invoke_generator` method to the generator helpers.

graysonwright:
Great, thanks! I'll review this in more detail in the morning, and get it merged in as soon as I can!

spec/support/generator_spec_helpers.rb
```diff
@@ -15,6 +15,10 @@ def contents_for_application_template(view_name)
     )
   end
 
+  def invoke_generator(*args)
+    have_received(:invoke).with(*args)
+  end
```
graysonwright: Nice move.

graysonwright:
This looks good! I'll shuffle the lines around so they're < 80 characters, and merge this in.

graysonwright:
Merged in as 04ee51c. Thanks!

PR #330, opened 2015-12-14
Allows the configuration of the installation directory

zamith:
Why:

* For existing application this could be problematic, especially with
  name collisions

* Allowing the installation directory to be configured in the app, only
  once and uses that any time a generator is called. It could be
  configured like this:

```ruby
Administrate.configure do |config|
  config.admin_directory = "administrate"
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 ### Upcoming Release
 
+* [#269] [FEATURE] Add a generator for copying default layout files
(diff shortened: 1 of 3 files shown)


aaronrenner:
@zamith I like how your solution doesn't require you to set the namespace every time you run a generator. 

Have you considered what the initial install experience might be like with this approach? If a user wants to install administrate using a custom namespace, would they need to create an administrate initializer before running `rails g administrate:install`?

zamith:
@aaronrenner Yes, that would be the idea.

PR #332, opened 2015-12-14
Change admin directory

zamith:
Why:

* For existing application this could be problematic, especially with
  name collisions

* Allowing the installation directory to be configured in the app, only
  once and uses that any time a generator is called. It could be
  configured like this:

```ruby
Administrate.configure do |config|
  config.admin_directory = "administrate"
end
```

Re-opened from #330 because I changed the branches in my repo and had to close it.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -37,7 +37,7 @@ module Administrate
 
       if resource.save
         redirect_to(
-          [Administrate::NAMESPACE, resource],
+          [Administrate::Config.admin_namespace, resource],
           notice: translate_with_resource("create.success"),
         )
       else
@@ -50,7 +50,7 @@ module Administrate
     def update
       if requested_resource.update(resource_params)
         redirect_to(
-          [Administrate::NAMESPACE, requested_resource],
+          [Administrate::Config.admin_directory, requested_resource],
           notice: translate_with_resource("update.success"),
         )
       else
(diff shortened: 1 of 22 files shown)


graysonwright:
@zamith, thanks for looking into this! I've been wanting to fix this for a while, but needed a poke. Here are my thoughts:

I'd like to avoid an initializer file unless it's absolutely necessary. Changing initializers requires restarting the development server, which can be annoying and slow down the development process. It might be the case that we need an initializer to get this functionality, but I'd like to look for alternative solutions first.

I'd like to set this up in a way that allows people to set up multiple admin dashboards for a single application. For example, you could have a dashboard at `/admin` that can manage all users and their roles, and a limited dashboard at `/moderator` that lets people edit site content, but not change user permissions.

I've started #339, which I think is the first step towards removing the hard-coded namespace. It calculates the namespace based on the request's URL, so it leaves us open to support multiple dashboards in a single app.

My PR doesn't solve the problem of how the generators decide which namespace to use. If you have any thoughts on that (preferably avoiding initializers), I'd love to hear them!

Thanks for your help!

zamith:
@graysonwright Yeah, the initializer has that drawback, but on the other hand allows for someone new to the team creating a dashboard not to care about knowing where it should go by default.

Would having an initializer and the option to pass it as an attribute to the generator be an overkill? I do understand that this adds the need for the initilizer file and it might not be a strong enough feature to warrant that.

I'll look into your PR, then. Thanks.

zamith:
@graysonwright I'm closing this PR in favor of yours, but I think we should have an option to the generator that overrides the `namespace` helper method. What do you think?

Issue #143, opened 2015-11-04
Feature Request: Make it ORM-agnostic


mrbrdo:
We are using Sequel and are looking for a decent admin (which is a problem even if using ActiveRecord). Since this project still seems young and showing promise, it would be nice if you could extract ORM-specific operations into a separate module so they can more easily be changed for a different ORM.

I think the only difficult thing I found in respect to Sequel was assoc_ids getter/setter which does not exist in Sequel. But it's not that big of a deal as it is not too hard to implement (although it would be nicer if this could be handled in the controller instead of unconditionally offloading it to the model, which may not be able to handle it in some ORMs).

I got most of it working with these monkey-patches: https://gist.github.com/mrbrdo/56da0dd984b4191a7273

I would be willing to provide Sequel implementations for such a module.

Cheers

graysonwright:
@mrbrdo thanks for your interest! I don't have much experience with Sequel, so I'd need to look into it a bit. Is there a good sample Rails application that uses Sequel that I could use to test with? I'd like to try out your monkeypatches to get a feel for how they would work.

I think we can change the Administrate API a bit to make it easier to implement Sequel support, without requiring as much monkeypatching. I'll pull you in when we have those conversations.

Thanks!

mrbrdo:
Hey,

Not sure for sample apps, I was able to find this app which is using a recent version of Rails with Sequel:
https://github.com/nathanl/searchlight_demo

Other than that, https://github.com/TalentBox/sequel-rails is usually used for Rails integration (as in the above app).

Note that I just monkey-patched enough to get the thing going on our project (which is for a client so I can't disclose code, unfortunately). So I'm not claiming it makes everything work.

I really hope it will be possible to "adapterize" the ORM in administrate, as you probably know the choices in administration frameworks for Rails is not that great even when using ActiveRecord, but when using another ORM it's really a serious issue. So my wish is to have a flexible enough adapter system that people can implement adapters for various ORMs, even though of course it may not be possible to do it for all of them (if they are radically different).

As I said I will be happy to provide a Sequel adapter implementation.

PS: Initially I was going to mention https://github.com/ianwhite/orm_adapter but after taking a closer look it does not have that many features (e.g. no reflection support). One possibility, although a bit more time-consuming, would be to create a similar library with expanded functionality that is needed in Administrate. Maybe people would pick it up and cover other ORM features in the future. But this is just a tangent, it would always be possible to extract the adapter system out later if it seems like it could be used for other purposes.

[...]

PR #333, opened 2015-12-14
Configure the scope on has many fields

### LINKED ISSUE #331: Customizing `candidate_resources` per Dashboard
Is it possible to define in a per dashboard basis, the queries to be done to get the `candidate_resources`, instead of using `all` all the time.

Example: A user can have multiple roles, and depending on the role we might (or not) want to show that user in the list of users to select from.

zamith:
Why:
- Many times we want to select from a list of resources that is scope.
  For instance, a user might be scoped by role.

- Allowing an option to be passed to the `HasMany` field, with a proc to
  be run

Usage:

``` ruby
ATTRIBUTE_TYPES = {
  admins: Field::HasMany.with_options(class_name: "User", scope: -> { User.admins }),
}
```

Fixes #331


graysonwright:
I like this - I'm a little bit concerned about overlap with #143 though.

I'm going to pull in some other thoughtbotters for their thoughts on this.

zamith:
@graysonwright The default right now is coupled to AR, I guess. What would be the strategy to uncoupled it, though? Or is it deeper than that?

JoelQ:
Since Ruby doesn't differentiate between methods and attributes, could we treat the scope like just another has many collection? Something like:

```ruby
class User < ActiveRecord::Base
  def self.admins
    where(admin: true)
  end
end
```

```ruby
ATTRIBUTE_TYPES = {
  admins: Field::HasMany,
}
```

I'd rather avoid introducing DSL syntax or extra config if possible and just rely on things being methods and objects.

zamith:
@JoelQ Is that supposed to work right now? Or is it how you would envision it?

mcmire:
I honestly think that fields shouldn't have much, if any, knowledge about the ORM at all. As far as a field is concerned, it knows about an object and an attribute. There may be underlying details about the specific *type* of field you're rendering in order to display choices and such, but I'm not sure that at this level you should have to care about that.

So if we can go with what @JoelQ is proposing then I feel like that would be the better option here.

zamith:
Well, right now the data we get from the resource is coupled to that particular instance (see [here](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/page/base.rb#L23)). This means that for regular fields those are the method that will be called, instance methods on the resource instance.

As for `Associative` fields, they try to guess the associated resource class (or you can pass it one), constantize it, and then call `all` on it (which as has been mentioned, is coupled to AR).

This was just so that we are all on the same page here, because at least for me it is a new code base.

So, to make @JoelQ's code work, we would have to figure out the associated class, and call the name of the attribute as a class method instead of `all`. Is this about right? Do you see this as the default, or what would we do as a default, if the class does not have that method?

@mcmire Where you talking of something is this realm, or doing a much bigger refactor and extract [this](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/fields/has_many.rb#L49) knowledge out of the field?

zamith:
Here's a draft of what I was talking about, and works as @JoelQ mentioned. What do you think?

```ruby
def candidate_resources
  if associated_class.respond_to?(attribute)
    associated_class.public_send(attribute)
  else
    associated_class.all
  end
end
```

JoelQ:
As I think about this, I realize there are three types of collection methods that might be defined on a resource: **association**, **scope**, and **standard method**.

```ruby
class User
  # association
  has_many :posts

  # scope
  def self.admins
    where(admin: true)
  end

  # standard method
  def favorite_colors
    ["red", "green", "blue"]
  end
end
```

## Association

* *instance method* that returns a collection of objects
* can calculate all possible *associatable* objects given a class
* makes sense to show a summary (such as a count) on the "index" dashboard
* makes sense to show a list of associated objects on the "show" dashboard
* makes sense to allow users add associated records from the **associatable** list on the form.
* implementation will tend to be coupled to `ActiveRecord`

## Scoped Association (special case)

Sometimes we want to scope down an association (e.g. a user's posts awaiting moderation). This can be done in Rails by passing a lambda to the `has_many` DSL.

This association can be passed to administrate just like any other association. From the point of view of Administrate, the properties are the same as for a standard association above except:

* calculating **associatable** records may not be as straightforward
* because of the above, editing the members of the association will not be as straightforward.

## Scope

* *class method* that returns a collection of objects
* doesn't really make sense to calculate all of the objects that _could be_ in the scope if they had the correct properties
* doesn't really make sense to show scope data on an "index" dashboard as those values are not connected to a single instance
* doesn't really make sense to show scope data on a "show" dashboard as those values are not connected to a single instance
* scopes are more "read-only" data. To change their value, you need to edit the properties of the included records.

## Regular method

* *instance method* that returns a collection of objects
* can't calculate the set of _all possible_ objects as that will be wildly different for each method
* could make sense to show a summary (such as a count) on the "index" dashboard
* could make sense to show a list of associated objects on the "show" dashboard
* given that it would be really hard to figure out what the list of possible objects that could be returned from the collection and how objects get to be in the collection, it doesn't really make sense to allow this to be edited by default (requires heavy customization for each method). This is probably read-only.

## Conclusion

* We want to show/read/write **associations**
* **Scoped associations** can mostly be treated like regular associations
* **Scopes** probably don't make sense on a dashboard
* **Standard methods** are probably read-only. Could be writable with heavy user customization.

Thoughts?

zamith:
@JoelQ That's a lot to take in, but I think I agree with most of it, I just don't really get why you say

> Scopes probably don't make sense on a dashboard

What do you mean by this? Do you mean we shouldn't handle them?

JoelQ:
>  > Scopes probably don't make sense on a dashboard
>
> What do you mean by this? Do you mean we shouldn't handle them?


Because scopes such as `.admin` are _all_ of a resource type rather than a single instance, I don't really see their value in a classic dashboard use case.

For example, in the following Users "index" dashboard:

```
| id | name | admins |
-----+------+---------
| 1  | john |   3    |
-----+------+---------
| 2  | tom  |   3    |
-----+------+---------
```

we repeat the same admin count for every row. It's not something really useful to see there because there is only one value for the whole system.

Similarly, displaying the admin count seems out of place on the "show" dashboard since there isn't much of a connection between a given user and the list of admins.

> Also, by "writing" an association you mean choosing an option from a select?

Correct. I am referring to the ability to add/remove items from the association via the select.

> EDIT: Are you considering the actual scope AR method? Or does that fall under the "Scope" category as well, since it is very similar to a class method?

For most purposes, class methods that return `ActiveRecord::Relation` and the `scope` DSL are functionally equivalent.

graysonwright:
@JoelQ, @zamith - there's discussion on how to handle scopes in #276. I agree that it doesn't make sense to display scopes for individual instances - it's a class-level concept.

As for this issue, I think the root of the problem is that a `has_many` relationship is inherently a database concept. Even more specifically, it's a concept from relational databases.

@zamith, you're right - at the moment, the `HasMany` field is tied to the database. If we wanted to completely remove all database logic from the `HasMany` class, I would suggest:

- Rename `HasMany` to `Collection`
- Pass in a proc to the `Collection` object that returns all potentially associatable objects. This is used for the form field.

In all other regards, the underlying object is treated as a plain old ruby collection.

graysonwright:
So essentially, the only difference from @zamith's suggested approach is that we would rename the `scope` argument to something like `candidate_objects`, and require that option. We could get rid of the default `associated_class.all`, because that's an assumption about the underlying database.

zamith:
@JoelQ I was only thinking of scopes to be used when populating the selects to choose from in `HasMany` fields in forms. At least that's where I've found them useful. For instance, I want to select from users, but only those who are admins. Does that make sense?

@graysonwright I would definitely prefer having a more ruby like solution, but adding an argument is a possibility. I like that it would remove the coupling to the database and naming it `Collection` makes sense.

zamith:
@graysonwright What do you think of my last commit? It's just an idea, but what I was aiming at is having a ORM agnostic `Collection`, but still have the `HasMany` with the `all` default if you choose to require it.

This could even go as far as adding a `--activerecord` option to the generator or something like that.

pvande:
As an additional thought along these lines, I'd offer the following:

* Our application has Resources and Customers.
* Some Resources are free, while others are paid.
* Some Customers only want to see _some_ of the free Resources.
* When we're editing the Customer, we _only_ want to hide or show the Resources they have access to (i.e. free Resources + the paid Resources they've paid for).
  * Finding this set from within a Field requires knowledge of the Customer in question (which is not currently available without monkeypatching).

alexanderadam:
Any progress on this?

PS: isn't it funny that I need this features exactly the day the issue was created?

ryudice:
are there is still plans to implement this?

purinkle:
@zamith, is this a change you would still like to implement?

PR #335, opened 2015-12-14
Update default index view w/correct documentation

jayroh:
The object previously used for the default index view was `Administrate::Page::Table` but has since been renamed to `Administrate::Page::Collection`.

This commit makes the update to the view to point to the correct class and link to its proper docs at rubydoc.info


graysonwright:
:+1:

PR #337, opened 2015-12-15
Fix padding on input[type='file']

coneybeare:
All of the padding on the `input[type='file']` was on the bottom previously, causing the button to align above the center of the label. It was fixed by splitting the padding between the top and bottom.

Before before

After (Safari) after safari

After (Chrome) after chrome

### DIFF
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -71,7 +71,7 @@ input[type="radio"] {
 }
 
 input[type="file"] {
-  padding-bottom: $base-spacing / 2;
+  padding-bottom: ($base-spacing / 4) 0;
   width: 100%;
 }
 
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -71,7 +71,7 @@ input[type="radio"] {
 }
 
 input[type="file"] {
-  padding-bottom: ($base-spacing / 4) 0;
+  padding: ($base-spacing / 4) 0;
   width: 100%;
 }
 


app/assets/stylesheets/administrate/base/_forms.scss
```diff
@@ -71,7 +71,7 @@ input[type="radio"] {
 }
 
 input[type="file"] {
-  padding-bottom: $base-spacing / 2;
+  padding-bottom: ($base-spacing / 4) 0;
```
graysonwright: Did you mean to change this to `padding`? It seems like this might be invalid CSS.

coneybeare: Ugh, not, it should have been `padding` only. I blame late night coding.

tysongach:
Good eye! I would love to see this alignment, as well.

But this feels like a magic number…where did `$base-spacing / 4` come from? I worry that it only came about from eyeing different values until it looked right. This usually means it’s easily breakable down the road. For example, what if the font is different? Or what does this look like on other platforms (not just other browsers)?

I’d like to investigate a bit deeper into a more future-friendly way of solving this. _Maybe_ this is the only easy way forward, but I’d like to see if a better solution is out there…

coneybeare:
it was `$base-spacing / 2` before, but all on the bottom, so I split it. The css files are full of `$base-spacing / 4` and `$base-spacing / 2`, so it is not as "magic" as you think.

tysongach:
Yes, but it’s magic in the sense that if the font size of the `label` changes, things will be mis-aligned again. So it’s always a bit of a moving target.

The commit referenced above, which is part of my [PR](https://github.com/thoughtbot/administrate/pull/399), utilizes flexbox which will automatically align things without us having to introduce complex math within our stylesheets.

PR #339, opened 2015-12-15
Don't hard-code Administrate's URL namespace

### LINKED ISSUE #281: Allow administrate to be installed somewhere other than /admin
I have a project that already has an /admin namespace and I'd like to install administrate as a secondary admin interface (maybe at /administrate).This would allow me to manage CRUD operations on models that don't have a UI yet. When I run `rails g administrate:install` it would be nice if there was an option to change the namespace of where I want it installed, so it doesn't conflict with my existing controllers. Here is an example of what rails admin does: https://github.com/sferik/rails_admin/blob/master/lib/generators/rails_admin/install_generator.rb#L13

I really appreciate you building this gem and I'm looking forward to using it.

graysonwright:
Closes #281 
Related to #330, #332.

## Problem:

Administrate requires that its dashboards be mounted at `/admin`.

In applications with an `Admin` model, Administrate will cause an error because it will try to define `Admin` as a module when it's already been defined as a class.

The hard-coded namespace limits users' freedom in defining their own URL schema, and prevents users from mounting multiple Administrate dashboards at once.

## Solution:

Dynamically evaluate Administrate's namespace from the request's URL.  This dynamically calculated value is used for all links and redirects in Administrate dashboards.

The namespace is now only set by generators.  If a user wants to change their namespace away from the default of `admin`, they must run the generators and manually change `admin` to the namespace of their choice.

## Next Steps:

Change generators to accept (or read from some configuration file) an optional namespace argument instead of always using the `admin` namespace.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -37,7 +37,7 @@ module Administrate
 
       if resource.save
         redirect_to(
-          [Administrate::NAMESPACE, resource],
+          [namespace, resource],
           notice: translate_with_resource("create.success"),
         )
       else
@@ -50,7 +50,7 @@ module Administrate
     def update
       if requested_resource.update(resource_params)
         redirect_to(
-          [Administrate::NAMESPACE, requested_resource],
+          [namespace, requested_resource],
           notice: translate_with_resource("update.success"),
         )
       else
@@ -105,7 +105,8 @@ module Administrate
       dashboard.permitted_attributes
     end
 
-    delegate :resource_class, :resource_name, to: :resource_resolver
+    delegate :resource_class, :resource_name, :namespace, to: :resource_resolver
+    helper_method :namespace
 
     def resource_resolver
       @_resource_resolver ||=
--- a/lib/administrate/resource_resolver.rb
+++ b/lib/administrate/resource_resolver.rb
@@ -1,5 +1,3 @@
-require "administrate/namespace"
-
 module Administrate
   class ResourceResolver
     def initialize(controller_path)
@@ -8,6 +6,10 @@ module Administrate
 
     def dashboard_class
       Object.const_get(resource_class_name + "Dashboard")
+    end
+
+    def namespace
+      controller_path.split("/").first
     end
 
     def resource_class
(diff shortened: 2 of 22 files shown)


lib/administrate/resource_resolver.rb
```diff
@@ -22,6 +20,10 @@ def resource_title
       model_path_parts.join(" ")
     end
 
+    def namespace
+      controller_path.split("/").first
```
graysonwright: This is where we calculate the namespace. In the default setup, this returns the string `"admin"`.

app/controllers/administrate/application_controller.rb
```diff
@@ -105,7 +105,8 @@ def permitted_attributes
       dashboard.permitted_attributes
     end
 
-    delegate :resource_class, :resource_name, to: :resource_resolver
+    delegate :resource_class, :resource_name, :namespace, to: :resource_resolver
+    helper_method :namespace
```
graysonwright: This makes the namespace available in all of Administrate's views.

thewoolleyman:
@graysonwright It appears this regression was re-introduced in https://github.com/thoughtbot/administrate/commit/765d3ff0b1377fb6465df0251049b4ffff0d2965

If I already have an existing `admin/` route (which will be common, e.g. when using Devise to define admin users), versions 0.3.0 blows up like this:

```
$ rails generate administrate:install
Running via Spring preloader in process 32756
   identical  app/controllers/admin/application_controller.rb
/Users/woolley/.gem/ruby/2.3.3/gems/administrate-0.3.0/lib/generators/administrate/dashboard/dashboard_generator.rb:111:in `const_get': uninitialized constant Admins::Session (NameError)
```

How can I work around this without having to rename my existing `admin/` route?

Thanks,
-- Chad

graysonwright:
@thewoolleyman, the generators haven't been updated to accept a new route. For now, you'll have to create those controllers manually – you can see https://github.com/thoughtbot/administrate/blob/master/spec/example_app/app/controllers/admin/products_controller.rb for an example.

I opened https://github.com/thoughtbot/administrate/issues/754 to fix up the generators.

ACPK:
@graysonwright - Do you have a recommended way of controlling what items a user sees in the sidebar based on their role (i.e. moderator, admin, etc.)

nickcharlton:
@ACPK, this was merged a long time ago. I'd recommend opening a new issue for your question.

PR #345, opened 2015-12-16
SCSS-Lint Configuration

tysongach:
- Rename SCSS-Lint configuration file
  - This more easily allows the running of SCSS-Lint locally from the command line, while still setting Hound up to use this configuration, as well.
  - The reason is because when you run SCSS-Lint localy from the command line, it looks for a file in the current working directory called `.scss-lint.yml`.
  - More info: https://github.com/brigade/scss-lint#configuration
- Remove incorrect SCSS-Lint option
  - This is not an option for the SelectorFormat linter: https://github.com/brigade/scss-lint/blob/master/lib/scss_lint/linter/README.md#selectorformat
  - It was likely mistakingly taken from the NameFormat linter
- Update SCSS-Lint config
  - Align with thoughtbot's style guide: https://github.com/thoughtbot/guides/tree/master/style/sass
- Indicate which version of SCSS-Lint our config aligns to


graysonwright:
I'm good with all of this. Great work!

:ship:

PR #346, opened 2015-12-16
Allow showing a belongs to relation without a dashboard

zamith:
Why:
- We might not want to have a Dashboard for all models listed, but still be
  able to show their representation on the show page of some other
  model. For instance, we might want to show a user's role name, but not
  be able to edit them.

- Showing the resource display if rails fails to create a link to it.

### DIFF
--- a/app/views/fields/belongs_to/_show.html.erb
+++ b/app/views/fields/belongs_to/_show.html.erb
@@ -17,7 +17,7 @@ By default, the relationship is rendered as a link to the associated object.
 
 <% if field.data %>
   <% if valid_action?(:show, field.attribute) %>
-    <%= link_to(
+    <%= link_to_if_route_exists(
       field.display_associated_resource,
       [namespace, field.data],
     ) %>
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -42,5 +42,11 @@ module Administrate
     def clear_search_params
       params.except(:search, :page).permit(:order, :direction, :per_page)
     end
+
+    def link_to_if_route_exists(text, *routing_args)
+      link_to text, *routing_args
+    rescue StandardError
+      text
+    end
   end
 end
--- a/app/views/fields/belongs_to/_index.html.erb
+++ b/app/views/fields/belongs_to/_index.html.erb
@@ -17,7 +17,7 @@ By default, the relationship is rendered as a link to the associated object.
 
 <% if field.data %>
   <% if valid_action?(:show, field.attribute) %>
-    <%= link_to(
+    <%= link_to_if_route_exists(
       field.display_associated_resource,
       [namespace, field.data],
     ) %>


graysonwright:
oh, nice catch. I like this change.

app/views/fields/belongs_to/_show.html.erb
```diff
@@ -16,8 +16,12 @@ By default, the relationship is rendered as a link to the associated object.
...
-  <%= link_to(
-    field.display_associated_resource,
-    [Administrate::NAMESPACE, field.data],
-  ) %>
+  <% begin %>
+    <%= link_to(
+      field.display_associated_resource,
+      [Administrate::NAMESPACE, field.data],
+    ) %>
+  <% rescue %>
```
graysonwright: I try to avoid using `rescue` without specifying a specific type of error - there's a good chance it can rescue from unrelated errors.

How about defining a `link_if_route_exists` helper?

```ruby
    def link_if_route_exists(text, *routing_args)
      link_to(text, *routing_args)
    rescue NoMethodError
      text
    end
```

graysonwright:
I think you'll also need to update `app/views/fields/belongs_to/_index.html.erb`.

graysonwright:
for background, I tried searching for a way to do this without needing to use `rescue`. I couldn't find a way that works when you pass in the path as an array, like we're doing.

Here's the closest that I could find: http://stackoverflow.com/questions/33904557/rails-link-to-only-if-route-exists

app/views/fields/belongs_to/_show.html.erb
zamith: That was my first approach actually, but `link_to` does not always take a `text` as a first parameter,  it may take a block. Would that helper still make sense, if it does not have 100% feature parity with `link_to`?

zamith:
Right, I did that same research, and came up empty as well. Not sure if there is a reason for this to be missing in Rails, or is really just missing.

app/views/fields/belongs_to/_show.html.erb
graysonwright: @zamith I'd say yeah - the function would cover all the cases we're using it for, so it would still be valuable to have. We can always expand the function to support blocks later if we want to.

zamith:
@graysonwright I've updated the code. Not sure if I should add a test, and if so, what kind of test would that be?

zamith:
@graysonwright What do you think about this? Any thing to change?

nickcharlton:
Hi @zamith! I'd love to move towards getting this merged. Would you be able to rebase on `master`?

Did you have any thoughts on how we could go about testing this, too?

zamith:
I've rebased with master, but it seems that there is a `valid_action?` check now, is it doing the same thing?

I don't really have thoughts on testing, sorry.

pablobm:
@zamith - I believe this is now implemented. To reproduce:

1. Run the example app (as per https://github.com/thoughtbot/administrate/blob/master/CONTRIBUTING.md#start-application-in-development)
2. Open the routes file at `spec/example_app/config/routes.rb`
3. Remove the line `resources :products`
4. Visit the "Line Items" dashboard, click on a single line item.

You will see that the product for that line item appears as text, not a link. Also, there will not be a sidebar link for the products dashboard. Undoing the change to the routes file will bring back both links.

I'm going to close this. Please let us know if we are missing anything.

nickrivadeneira:
@pablobm Removing the route only works if there is a dashboard present. If you have neither route nor dashboard, it'll break. The relevant line is here: https://github.com/thoughtbot/administrate/blob/86697d6f056b6a677b015ac57785b5591d62f08d/lib/administrate/field/associative.rb#L36-L38

A simple workaround is to create a dashboard for the simple purpose of defining `#display_resource`, but it might be nice if that wasn't necessary. One option would be to render display names for records with or without a dashboard defined. If that sounds reasonable, I can make an attempt at a PR.

pablobm:
@nickrivadeneira, thank you for that report. I'm unable to reopen this thread as it's for an old PR against `master`, and the main branch is now `main`. It's a slightly different thing anyway, so I've opened a new issue at https://github.com/thoughtbot/administrate/issues/2104. I'll tag you from there.

PR #348, opened 2015-12-16
Fix partial new title

ndrluis:


graysonwright:
@ndrluis, thanks!

Unfortunately, I don't think we want to change this up. The `page_title` method displays the resource that the page is displaying. On the "New" page, there isn't a valid database model backing the page, so it will often raise an error. That's what caused the build to fail - the `page_title` method tried to access the object's `id`, but that hadn't been set in the database yet.

One potential way forward would be to rename the `page_title` method to `display_resource`, to make it more clear what that method's actually doing. I think the `page_title` name is a little misleading, because we don't always want to use it for the title.

ndrluis:
@graysonwright agree, I will close this pull request and open another to rename the `page_title` to `display_resource`

PR #350, opened 2015-12-17
Correct task for generating single model dashboard

jrnk:
Maybe also an idea to add this to the docs?


graysonwright:
@jrnk, you can use `rails generate administrate:dashboard MyModelName`. I agree that we should put this into the docs.

I'm going to close this - feel free to re-open if we can help with anything else!

PR #361, opened 2015-12-21
Update rubocop configuration file to run locally

graysonwright:
## Problem:

Hound uses Rubocop under the hood to check for Ruby style violations.
Its [default config] is different from Rubocop's defaults.
This means that people who run rubocop locally
will get different feedback than they get from Hound.

[default config]: https://raw.githubusercontent.com/thoughtbot/hound/master/config/style_guides/ruby.yml

## Solution:

Rename `.ruby-style.yml` to `.rubocop.yml`,
so Rubocop automatically picks up the configuration when run locally.

Update the rubocop config to use [thoughtbot's default].

[thoughtbot's default]: https://github.com/thoughtbot/hound/blob/master/config/style_guides/thoughtbot/ruby.yml

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,5 +1,5 @@
 ruby:
-  config_file: .ruby-style.yml
+  config_file: .rubocop.yml
 scss:
   enabled: true
   config_file: .scss-lint.yml
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,363 @@
+AllCops:
+  Exclude:
+    - spec/example_app/db/schema.rb
+
+AccessorMethodName:
+  Enabled: false
+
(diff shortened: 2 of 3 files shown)


graysonwright:
similar to https://github.com/thoughtbot/administrate/pull/345/

zamith:
What's your thoughts on having rubocop as part of the build? Is it too strict?

tysongach:
:+1:

graysonwright:
@zamith yep - we don't want style violations to break the build, and we don't want contributors to be responsible for fixing violations unrelated to their code that may have slipped in. Hound provides a nice balance, but also lets us ignore it if it makes sense on a case-by-case basis.

PR #362, opened 2015-12-22
Extract an `administrate:manifest` generator

### LINKED ISSUE #182: Dashboard Manifest generator
Rather than running `rails generate administrate:install` and going through the overwrite prompts for each model, would it be possible to have a single-model generator for when new models get added to an application?

graysonwright:
Closes #182.

## Problem:

In established applications with many models,
there is no easy way to generate Administrate dashboards
for some but not all of the models.

## Solution:

Add a generator for creating only an Administrate dashboard manifest.
Users can edit the manifest file to add or remove models,
and then run `rails g administrate:install`
to generate only the dashboards listed in the manifest.

If the user runs `rails g administrate:install`
without first generating a manifest file,
Administrate will automatically create a manifest
and use it to guide the installation.

## Small Changes:

Stub out `Rails.root` in generators during tests.
Using `Rails.root` lets us more accurately generate files
when the Rails app is not at the root of the repository,
such as in the example app.


spec/generators/install_generator_spec.rb
```diff
@@ -20,68 +20,26 @@ class ApplicationController < Administrate::ApplicationController
...
+  describe "manifest generator" do
+    it "invokes the Manifest generator if the manifest does not yet exist" do
       begin
-        class ModelWithoutDBTable < ActiveRecord::Base; end
+        remove_constants :DashboardManifest
         stub_generator_dependencies
-        manifest = file("app/dashboards/dashboard_manifest.rb")
 
-        run_generator
+        expect { run_generator }.to raise_error
```
graysonwright: This error is only raised in the test suite.

When a user runs this generator without `DashboardManifest` defined, the generator will invoke the manifest generator to fill in the missing piece.

In the test suite, we stub out the invocation of the manifest generator. The generator does not run, and `DashboardManifest` remains undefined. When the install generator tries to use `DashboardManifest` as the basis for the installation, it comes up empty and raises an error.

PR #364, opened 2015-12-22
Generate SHOW_PAGE_ATTRIBUTES as an explicit array

### LINKED ISSUE #130: Generate `SHOW_PAGE_ATTRIBUTES` as an explicit array
Right now, `COLLECTION_ATTRIBUTES`, `SHOW_PAGE_ATTRIBUTES`, and `FORM_ATTRIBUTES` use `ATTRIBUTE_TYPES.keys` as a base set of attributes. It would follow, then, that in order to re-order fields that appear in a form or table, you re-order the keys in `ATTRIBUTE_TYPES`. However, this seems a little odd, because `ATTRIBUTE_TYPES` is a hash, and so the order of its keys should be irrelevant. (It just so happens that hashes in Ruby are now ordered, but they really shouldn't be, and we shouldn't be assuming they are.)

So here's a suggestion: introduce a `BASE_ATTRIBUTES` constant that all of the other `*_ATTRIBUTES` constants use. The dashboard generator would have to set this to an explicit array based on the columns the model has, but then we'd encourage people to modify this array instead of `ATTRIBUTE_TYPES`. While we're at it, it might be useful to place `ATTRIBUTE_TYPES` first.

So a dashboard class would look like this:

``` ruby
ATTRIBUTE_TYPES = {
  author: Field::BelongsTo.with_options(class_name: "User"),
  content: Field::Text,
  id: Field::Number,
  ...
}

BASE_ATTRIBUTES = [
  :author,
  :content,
  :id,
  ...
]

READ_ONLY_ATTRIBUTES = [
  :id,
  :created_at,
  :updated_at
]

TABLE_ATTRIBUTES = BASE_ATTRIBUTES.first(4)

SHOW_PAGE_ATTRIBUTES = BASE_ATTRIBUTES

FORM_ATTRIBUTES = BASE_ATTRIBUTES - READ_ONLY_ATTRIBUTES
```

graysonwright:
Closes #130

## Problem:

The dashboard generator specifies:

```ruby
```

This means that the order of the show page attributes
is dependent on the order of keys in the `ATTRIBUTE_TYPES` hash.

Ruby hashes happen to be ordered,
but we shouldn't make an assumption that hashes are ordered
because that is not a property that hashes are designed to have.

## Solution:

Because we assume that users will be editing `SHOW_PAGE_ATTRIBUTES`
to customize the contents and order of show page attributes,
we'll generate an explicit array for them to modify.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@
 * [#269] [FEATURE] Add a generator for copying default layout files
 * [#328] [FEATURE] Add a generator for copying default sidebar partial
 * [#295] [FEATURE] Add dashboard detection for ActiveRecord::Enum fields.
+* [#364] [FEATURE] Improve dashboard generator by explicitly listing out the
+  generated `SHOW_PAGE_ATTRIBUTES` array elements.
 * [#297] [I18n] Add Italian translations
 * [#307] [I18n] Fix German grammatical errors
 
--- a/lib/generators/administrate/dashboard/templates/dashboard.rb.erb
+++ b/lib/generators/administrate/dashboard/templates/dashboard.rb.erb
@@ -28,7 +28,13 @@ class <%= class_name %>Dashboard < Administrate::BaseDashboard
 
   # SHOW_PAGE_ATTRIBUTES
   # an array of attributes that will be displayed on the model's show page.
-  SHOW_PAGE_ATTRIBUTES = ATTRIBUTE_TYPES.keys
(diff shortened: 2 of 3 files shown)


zamith:
My experience was that I ended up having to do this anyway, because I don't always want to show every attribute on the show page, plus this is much more explicit. :+1:

graysonwright:
@mcmire, thoughts?

mcmire:
Yup! That's exactly what I had in mind. :+1:

PR #365, opened 2015-12-25
Removes inline_svg dependency.

mariochavez:
inline_svg gem is currently broken to work with Sprockets 3 and the Rails
application environment set to production. So this has broken administrate as
well. Normally a fix should be implemented in broken gem but after looking at
why administrate requires inline_svg gem I realize that it is easy to replace
it with a helper method.

ApplicationHelper#svg_tag method was added to substitute the work of
inline_svg. This method creates a svg tag that will look like:

    <svg xmlns="http://www.w3.org/2000/svg"
    xmlns:xlink="http://www.w3.org/1999/xlink"><use
    xlink:href="http://localhost:3000/assets/administrate/search-360c782d91a21edd53f5754c3fd8f8984ec5d9b36389f6df1fe618a4e50405b6.svg#search" width="16" height="16"></use></svg>

This will allow to resize the svg image and to use css to change properties
like color.

But there are a couple of differences with inline_svg. First the svg file
should have an id property, for my previous example it is `search`.

The second difference is that `use` tag needs to have `height` and `width`
properties set as attributes of the tag, css properties will not affect the svg
size.

ApplicationHelper#svg_tag used to generate html shown above looks like:

```
<%= svg_tag "administrate/search.svg",
  svg_id: "search", width: "16", height: "16" %>
```

This solution will play nice with Sprockets and with modern browsers.


kesha-antonov:
+1

app/assets/images/administrate/search.svg
```diff
@@ -1,6 +1,6 @@
-<svg width="48" height="48" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
+<svg id="search" width="48" height="48" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
     <title>
         search
     </title>
-    <path d="M27.885 32.515c-2.864 1.966-6.333 3.116-10.07 3.116C7.976 35.63 0 27.656 0 17.817 0 7.976 7.976 0 17.816 0S35.63 7.976 35.63 17.816c0 3.736-1.15 7.205-3.115 10.07l14.53 14.53c1.278 1.277 1.275 3.352 0 4.628-1.28 1.278-3.353 1.278-4.63 0l-14.53-14.53zm-10.07-3.736c6.056 0 10.964-4.91 10.964-10.964 0-6.055-4.91-10.964-10.964-10.964-6.055 0-10.964 4.91-10.964 10.964 0 6.055 4.91 10.963 10.964 10.963z" fill="#D8D8D8" fill-rule="evenodd"/>
+    <path d="M27.885 32.515c-2.864 1.966-6.333 3.116-10.07 3.116C7.976 35.63 0 27.656 0 17.817 0 7.976 7.976 0 17.816 0S35.63 7.976 35.63 17.816c0 3.736-1.15 7.205-3.115 10.07l14.53 14.53c1.278 1.277 1.275 3.352 0 4.628-1.28 1.278-3.353 1.278-4.63 0l-14.53-14.53zm-10.07-3.736c6.056 0 10.964-4.91 10.964-10.964 0-6.055-4.91-10.964-10.964-10.964-6.055 0-10.964 4.91-10.964 10.964 0 6.055 4.91 10.963 10.964 10.963z" fill-rule="evenodd"/>
```
gylaz: Why did this have to change?

app/helpers/administrate/application_helper.rb
```diff
@@ -16,5 +16,25 @@ def display_resource_name(resource_name)
           default: resource_name.to_s.pluralize.titleize,
         )
     end
+
+    def svg_tag(asset, options = {})
+      svg_id = options.fetch(:svg_id)
```
gylaz: Since `svg_id` is required, what do you think about making it the second parameter instead of a hash key?

app/helpers/administrate/application_helper.rb
```diff
@@ -16,5 +16,25 @@ def display_resource_name(resource_name)
...
+
+      svg_attributes = {
+        "xlink:href".freeze => "#{asset_url(asset)}##{svg_id}",
+      }
+
+      svg_schema = "http://www.w3.org/2000/svg".freeze
+      xlink_schema = "http://www.w3.org/1999/xlink".freeze
```
gylaz: How about making these similar to `svg_attributes`:
```ruby
  xml_attributes = {
    "xmlns" => "http://www.w3.org/2000/svg",
    "xmlns:xlink" => "http://www.w3.org/1999/xlink", 
}
```
then using them:
```ruby
content_tag :svg, xml_attributes do
```

lib/administrate/engine.rb
```diff
@@ -19,5 +18,7 @@
 module Administrate
   class Engine < ::Rails::Engine
     isolate_namespace Administrate
+
+    Engine.config.assets.precompile << /\.(?:svg)\z/
```
gylaz: Why is this line needed?

spec/helpers/administrate/application_helper_spec.rb
```diff
@@ -41,4 +41,47 @@
       end
     end
   end
+
+  describe "#svg_tag" do
+    subject do
+      Nokogiri::HTML svg_tag("logo.svg",
+                             svg_id: "logo")
```
gylaz: Can these fit on one line?

spec/helpers/administrate/application_helper_spec.rb
```diff
@@ -41,4 +41,47 @@
...
+
+      expect(attribute.value).to eq("/logo.svg#logo")
+    end
+
+    context "with size attributes" do
+      subject do
+        Nokogiri::HTML svg_tag("logo.svg",
+                               svg_id: "logo",
+                               height: "15",
+                               width: "20")
```
gylaz: This project uses the following preference for formatting multi-line args:
```ruby
Nokogiri::HTML svg_tag(
  "logo.svg",
  svg_id: "logo",
  height: "15",
  width: "20",
)

File: spec/helpers/administrate/application_helper_spec.rb [hunk 7]
```diff
@@ -41,4 +41,47 @@
...
+  describe "#svg_tag" do
+    subject do
+      Nokogiri::HTML svg_tag("logo.svg",
+                             svg_id: "logo")
+    end
+
+    it "returns use tag with xlink:href attribute" do
+      use_tag = subject.xpath("//svg//use['xlink:href']")
+
+      expect(use_tag.length).to eq(1)
```
gylaz: `size` is more idiomatic Ruby.

app/assets/images/administrate/search.svg
mariochavez: CSS fill property is ignored when fill is set in the svg, at least when I was trying to implement this.

app/helpers/administrate/application_helper.rb
mariochavez: Sure!

lib/administrate/engine.rb
mariochavez: Looks like newer rails-sprocket gem requires to gems/engines to be explicit about assets. Without this Sprockets 3 crashes with:

```
Sprockets::Rails::Helper::AssetNotPrecompiled

Asset was not declared to be precompiled in production.
Add `Rails.application.config.assets.precompile += %w( administrate/search.svg )` to `config/initializers/assets.rb` and restart your server
```

atinder90:
facing same issue but only in production while deploying with Capistrano.

Gemfile.lock
```diff
@@ -62,7 +61,7 @@ GEM
       rake
       thor (>= 0.14.0)
     arel (6.0.3)
-    autoprefixer-rails (6.1.2)
+    autoprefixer-rails (6.2.1)
```
tysongach: We should handle this version bump in a separate PR.

Gemfile.lock
```diff
@@ -240,7 +235,7 @@ GEM
     rspec-support (3.1.2)
     ruby-progressbar (1.7.5)
     safe_yaml (1.0.4)
-    sass (3.4.19)
+    sass (3.4.20)
```
tysongach: We should handle this version bump in a separate PR.

app/assets/images/administrate/search.svg
```diff
@@ -1,6 +1,6 @@
-<svg width="48" height="48" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
+<svg id="search" width="48" height="48" viewBox="0 0 48 48" xmlns="http://www.w3.org/2000/svg">
```
tysongach: Can you clarify the need for this addition of an `id`? If it’s for styles, we should use a class.

mariochavez: @tysongach no, it is not for styling.

I'm not sure if you had the opportunity to run administrate with these changes, but since I'm replacing `inline_svg` with `svg`/`use` tags this ID is required by the `use` tag to select a node or group of nodes to be duplicated into the HTML DOM https://developer.mozilla.org/en-US/docs/Web/SVG/Element/use

The following is the example on how the search icon will look in the DOM.
```
<span class="search__icon">
  <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
    <use xlink:href="http://localhost:7003/assets/administrate/search-360c782d91a21edd53f5754c3fd8f8984ec5d9b36389f6df1fe618a4e50405b6.svg#search" width="16" height="16">
    </use>
  </svg>
</span>
``` 

As you can see in the attribute `xlink:href` of `use` tag, it does contains the URL of the sag file but also the ID of the nodes to duplicate. `use` tag requires this ID.

tysongach: Ah, of course! Nice.

tysongach:
With this new solution, can we pass a class to the SVG for styling purposes, too?

mariochavez:
@tysongach yes, we can pass a class.

Genkilabs:
@mariochavez Thanks for taking this on. I needed to stand up an admin tool on super short notice, but on ruby 2.3 rails 4.3.5. Your branch did the trick.
+1 for merge to master

samnang:
:+1:

lenart:
:+1:

tsrivishnu:
+1 waiting for this to me merged.

jeremyw:
Without this fix, a fresh Rails app with Administrate can't be deployed to Heroku.

gylaz:
@mariochavez Thanks a lot for this work! I've merged it as ecc6782a8b127e76ce414e0aa48962ec4711699b.

@graysonwright Can we get a new gem version released for this?

graysonwright:
Talked with greg - we're waiting to merge #424 before we release version 0.1.3. This will be included in that release.

graysonwright:
@mariochavez thanks, btw! I didn't know about linking SVGs like that, it's pretty slick. I expect I'll make that my default approach in the future. :smile:

mariochavez:
I'm glad to see this coming to master.

graysonwright:
This is packaged up in the `0.1.3` release: #428

PR #368, opened 2015-12-29
Fix sidebar__link:last-child

rubenmoya:
Hi all!

The `&:last-child` in the `sidebar__link` wasn't working properly since the link is inside an `li` so it's always the last item and the padding was always applied and the space between items was too big.

I solved it by adding a new class `sidebar__item` to the `li` and moving the `:last-child` to it. I also changed the paddings to margins, so you only can click the item when you press the link and not the space above it.

### DIFF
--- a/app/assets/stylesheets/administrate/components/_sidebar.scss
+++ b/app/assets/stylesheets/administrate/components/_sidebar.scss
@@ -6,14 +6,14 @@
 
   &__item {
     &:last-child {
-      padding-bottom: $base-spacing;
+      margin-bottom: $base-spacing;
     }
   }
 
   &__link {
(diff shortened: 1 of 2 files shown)


graysonwright:
@rubenmoya Thanks!

I somehow missed this PR before I started looking into the issue myself. I put up my own fix at https://github.com/thoughtbot/administrate/pull/408, along with a fix for a header CSS bug.

I like this solution, but I'd rather avoid introducing new HTML classes if possible. Do you mind if we close this in favor of #408?

rubenmoya:
Yeah, I think the same, it's better to avoid HTML classes :+1:

PR #369, opened 2015-12-31
use before_action instead of before_filter

BenMorganIO:
`before_filter` is deprecated and we should probably be using `before_action` to help with a Rails 5 update.

```
DEPRECATION WARNING: before_filter is deprecated and will be removed in Rails 5.1. Use before_action instead.
```


BenMorganIO:
Need this to get bumped. I'm working on a Rails 5 application and this is probably an easy fix.

PR #371, opened 2016-01-03
Updated copyright notices to 2016 [ci skip]

Atul9:


ascendbruce:
Hi @Atul9, @graysonwright, I saw this PR from CodeTriage. I am not familiar with open source legal issues. But following is my finding:

From [Is renewal of MIT license needed on github at the beginning of each year?](http://programmers.stackexchange.com/questions/210472/is-renewal-of-mit-license-needed-on-github-at-the-beginning-of-each-year)

> * You should update the year if and only if you made changes to your software in that year. 
> * **The copyright notice should include the year in which you finished preparing the release** (so if you finished it in 1998 but didn't post it until 1999, use 1998). You should add the proper year for each release; for example, “Copyright 1998, 1999 Terry Jones” if some versions were finished in 1998 and some were finished in 1999. If several people helped write the code, use all their names.
> * **For software with several releases over multiple years, it's okay to use a range (“2008-2010”) instead of listing individual years (“2008, 2009, 2010”)** if and only if every year in the range, inclusive, really is a “copyrightable” year that would be listed individually; and you make an explicit statement in your documentation about this usage.

In our case, will it be more accurate to do

``` diff
-Copyright (c) 2015 thoughtbot, inc.
+Copyright (c) 2015-2016 thoughtbot, inc.
```

:question:

Atul9:
@ascendbruce @graysonwright Done.

ascendbruce:
:+1:

PR #372, opened 2016-01-04
fix issue with appearance property not displaying correctly

almnes:
I ran into the issue of the CSS property appearance not running flawlessly on all browsers, and added general browser support to it. I also adjusted the line-height after that, to match the other input elements. 

The issue is displayed below:
screen shot 2016-01-04 at 9 19 00 am

### DIFF
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -57,6 +57,8 @@ textarea {
 }
 
 input[type="search"] {
+  -webkit-appearance: none;
+  -moz-appearance: none;
   appearance: none;
 }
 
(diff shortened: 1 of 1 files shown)


tysongach:
@almnes What browser were you using to take that screenshot?

tysongach:
We should avoid hand-writing vendor prefixes. Autoprefixer _should_ be running, and take care of this for us. So if this is a prefixing issue, maybe it’s a bug with that.

I think this needs a bit more digging before we know what the real issue is…

almnes:
I was using Google Chrome Version 47.0.2526.106 (64-bit) on Macbook Air

tysongach:
I’ve confirmed that Autoprefixer is not actually running on our CSS, which is why these prefixes are not being output in the application CSS.

I’m going to close this PR and we’ll be working on a fix to resolve that bug with Autoprefixer. It will automatically add the `-webkit-`/`-moz` prefixes to `appearance`.

carlosramireziii:
@tysongach Is there an existing issue / pull request for the fix to Autoprefixer that you described above?

tysongach:
@carlosramireziii Not that I’m aware of.

carlosramireziii:
Thanks @tysongach. I'll create one.

PR #376, opened 2016-01-04
Started Rails 5 compatibility

hayesr:
Wanted to try Administrate in a new project I'm starting with the Rails 5 beta, so I dived in to see what it would take to make it work. **This is probably not ready to merge** since I had to jump through some hoops and pull in Github versions of gems to get tests working.
- First error I ran into had to do with the `column_types` method. [See this commit](https://github.com/rails/rails/commit/70ac072976c8cc6f013f0df3777e54ccae3f4f8c).
- Next, using params in url generation is scary, and now throws an error. I like to live dangerously, and for now I just threw `.permit!` in there. [This whole thing](https://github.com/hayesr/administrate/blob/rails5_compat/app/views/administrate/application/_collection.html.erb#L28-L30) should probably be replaced by a helper method.
- The initializer in the example app to disable XML parsers causes an error, [but does not seem to be necessary](https://github.com/rails/rails/commit/c9909db9f2f81575ef2ea2ed3b4e8743c8d6f1b9). 

Finally, I had to use master versions of Capybara and RSpec-\* to get tests running. By including url_helpers in ActionView::TestCast::TestController I got tests to pass. I'm guessing this will be fixed, or there will be a more official way to do this in the future. Appraisals didn't pass, will Sass 3.4 be a concern after Rails 5?

I don't want to go much further without some feedback, but I'm happy to if it looks like I'm on the right path.


hayesr:
Looks like CI doesn't like the Github urls in the Gemfile.

spec/example_app/db/schema.rb
```diff
@@ -36,10 +36,9 @@
     t.string   "queue"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.index ["priority", "run_at"], name: "delayed_jobs_priority", using: :btree
```
BenMorganIO: This is a cool change!

hayesr: :smile: That was totally generated by Delayed Job/Rails, maybe this file should be ignored?

BenMorganIO: It looks like Rails is switch `add_index` => `t.index`. I would keep it. Its new for Rails 5 and besides, its going to have the same effect as everything else. Also, its a smaller schema.rb :p

clemensg:
Hi! What's the current status on this?
I'd really like to use administrate with my new Rails 5 apps.

hayesr:
I think there are 2 things that need to be addressed:
- [ ] How to deal with params in url generation [here](https://github.com/hayesr/administrate/blob/rails5_compat/app/views/administrate/application/_collection.html.erb#L28-L30)
- [ ] How to handle all the Gem dependencies

If one of the maintainers could provide some feedback/direction on these, I'd love to get this branch to a more merge-able state.

why-el:
@graysonwright some feedback here if you have time?

fx:
@hayesr I wouldn't merge until there are compatible releases of dependant gems, whenever Rails 5 hits RC/gets released. 

URL generation and all other changes here _should_ be backward compatible (or can be made backward compatible) for the most part.

hayesr:
I have some ideas for permitting params that I plan on tinkering on this weekend.
* I think it's fine to do a blanket `permit!` by default, as long as users know it's happening.
* The view doesn't feel like the right place to do this.
* This should be done in an override-able method in the controller or dashboard so users can lock down params if desired.

What say ye?

djsegal:
:+1:

graysonwright:
@hayesr, you rock! Thanks for getting a great start on this. I'm sorry I haven't been able to give feedback yet; I've been traveling this past month for a new job, and haven't had as much time as I'd like to devote to the project. I should have considerably more free time after March 7, when I return to San Francisco.

I agree that the view isn't the place to do `permit!`. I'll have to do a bit of reading to see if there's a reason not to use the blanket `permit!` instead of customizing it with specific acceptable params. Until then, a helper method in the controller seems like the right place for it.

As for gems, I'd be okay releasing with github-hosted versions of `rspec-rails` and `capybara`. Those gems are only used for testing, and they're only used in the context of the demo app. There won't be a user-facing effect for people using Administrate in their own projects.

hayesr:
There's definitely something at least a little unsettling about `permit!`. Am I right in thinking the only params we need in [the `collection` partial](https://github.com/hayesr/administrate/blob/rails5_compat/app/views/administrate/application/_collection.html.erb#L28-L30) is `page` and `per` for Kaminari? (so that the current page is returned with a new sort?) In that case I'll add a method that slices those params.

hayesr:
Side issue: it looks like there was a capitalization change with submit buttons, so I changed the Capybara clicks to be agnostic.

hayesr:
CI is choking on Percy, says "unauthorized".

lautis:
I've continued work on this at https://github.com/lautis/administrate/tree/rails5_compat. Mainly capybara and rspec are now installed from rubygems and there is a separate gemfile for Rails 5.

hayesr:
How about the Collection presenter handles the permitting of params? I scoped it to only the ones that came up in tests.

Typical Strong Param usage is to call `.permit(:my_param)` on params, but [this setting](https://github.com/thoughtbot/administrate/blob/master/spec/example_app/config/application.rb#L31) means that any extra params raise an exception (instead of being silently ignored). It appears to have been auto-gen'd by Suspenders; are we married to it?

Also, without a blanket permit we are waiting for [this Kaminari issue](https://github.com/amatsuda/kaminari/issues/774).

@lautis Thanks! I think I will cherry-pick the gem stuff when we settle on a method to handle params.

jalberto:
@hayesr kaminari master fix the problem

hayesr:
@jalberto Yeah, but they haven't released a gem version with the fix. We can't point to the Github repo in the gemspec.

hayesr:
Did a rebase and made a few minor changes. I'm working on the notion that this pull can be bare-minimum compatibility and we can follow-up with others to fix dep warnings and change controller testing. (I added `rails-controller-testing` for now)

One test still failing because of Kaminari. The fix has been made, just waiting for a [new release](https://github.com/amatsuda/kaminari/issues/774#issuecomment-212645037).

jalberto:
@hayesr agree, but as a workaround you can point to a commit in kaminari, so at least it will work for testing.

Thanks for the additions :)

hayesr:
@jalberto —Yes, and for the sake of noting it here, all tests pass against Kaminari's 0-17-stable branch.

imajes:
Hey-

@hayesr, @graysonwright - there are a few changes here also that need to be cleaned up (e.g. the security vuln initializer shouldn't be removed, but version scoped). I'm working on cherry picking your work @hayesr into a new branch on my fork, which hopefully should also pass validation. Perhaps you can take a look this week and see what your thoughts are with it? thanks!

onemanstartup:
@hayesr kaminari 0.17 released

hayesr:
Just coming back for an update. I think all we're waiting on now is an update to `administrate-field-image`.

nickcharlton:

PR #380, opened 2016-01-07
Add lang attribute to html element

tysongach:
Improves accessibility, among other things like assisting user agents in providing definitions using a dictionary.


tysongach:
Quite weird that just adding a `lang` attribute throws Percy off…this appears to be an issue with Percy, and not that this PR causes any visual differences.

ckundo:
<3 lgtm

PR #381, opened 2016-01-07
Associate fields with original resource

FooBarWidget:
This pull request adds a `resource` field to the Field class. This gives more flexibility to field views, allowing it to customize its display based on the original resource. For example, this will allow fields to display content based on two different columns.

My use case is as follows. My system has many Products. Each Product has a `name` (which is an internal name) and an optional `display_name`. I want to display the product name as follows:

```
product name here
(optional display name here) <--- this line is only shown if display_name is not nil
```

With this pull request, I'm able to write a view like this:

```
<%= field.resource.name %>
<% if field.resource.display_name %>
  <br>
  <small><%= field.resource.display_name %></small>
<% end %>
```

### DIFF
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -16,8 +16,9 @@ module Administrate
         false
       end
 
-      def initialize(attribute, data, page, options = {})
+      def initialize(attribute, resource, data, page, options = {})
         @attribute = attribute
+        @resource = resource
         @data = data
         @page = page
         @options = options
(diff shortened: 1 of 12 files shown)


FooBarWidget:
Hi maintainer, any updates on this?

carlosramireziii:
@FooBarWidget Could you rebase on `master` or [enable contributor pushes](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

nickcharlton:
Thank you for your PR! I'm going to close this as it's been a while since the last response.

#788 is very likely to get merged in the next few days, so this will also be solved with that.

Issue #258, opened 2015-11-20
Admin layout generation?


nextekcarl:
I'm working on integrating the Cloudinary heroku addon and I need to get that gem to include the javascript files that it normally adds by adding:

//= require cloudinary

...to the application.js manifest, but it appears that's all ignored when I'm in the administrate Rails engine. Where does the layout live for Administrate? I don't see a way listed to generate that layout that Administrate uses so that I can make changes to it the same way I can with the resource specific views, etc.

And while I can use the content_for :javascripts helper to add the javascript for cloudinary things directly, I'm pretty sure that will break when I move it to Heroku due to the SHA that added during precompile, won't it? So where do I add the require statement above to get it to get it added to the asset pipeline used in Administrate?

graysonwright:
Thanks, @nextekcarl! It looks like there's a bit of overlap with #187 here. We're hoping to add generators for SCSS and JS soon, but until then you should be able to overwrite a few files to get it working.

nextekcarl:
Thanks, I looked at that but I guess I'm just not sure how I'd translate the manifest require statement into this formt and actually expect it to work:

nextekcarl:
Sorry, comment was messed up there so I deleted to and I'll try again:

I guess I'm just not sure who I'd translate the work done by //= require cloudinary in the app manifest into a simple content for block and expect it to work wince this method doesn't add the SHAs:

<% content_for :javascript do %>
  <%= javascript_include_tag("jquery.ui.widget", "jquery.iframe-transport", 
                         "jquery.fileupload", "jquery.cloudinary") %>
  <%= cloudinary_js_config %>
<% end %>

Any help is appreciated.

graysonwright:
Hm... you may be able to copy [this file](https://github.com/thoughtbot/administrate/blob/master/app/assets/javascripts/administrate/application.js) into your app at `app/assets/javascripts/administrate/application.js`, and then add the `//= require` statement in there.

Does that solve the problem?

nextekcarl:
Yes, I think that did, thanks!

graysonwright:
Great! I'll work on a generator to make that process easier.

nextekcarl:
We can generate the files used for views, but not for the layout. How do do we modify the favicon, or change the actual layout outside of the specific content_for blocks included already? I guess you could include content for blocks for each section, but that feels kind of clunky over just generating the layouts the same way we generate the views for resources, but I don't see a way to do that.

[...]

PR #382, opened 2016-01-07
Allow customizing the layout's header

FooBarWidget:
This pull request allows one to customize the `<head>` tag on a per-page basis, e.g. to insert custom style sheets.

### DIFF
--- a/app/views/layouts/administrate/application.html.erb
+++ b/app/views/layouts/administrate/application.html.erb
@@ -22,6 +22,7 @@ By default, it renders:
   <%= stylesheet_link_tag "//fonts.googleapis.com/css?family=Lato:300,400,900", media: "all" %>
   <%= stylesheet_link_tag "administrate/application", media: "all" %>
   <%= csrf_meta_tags %>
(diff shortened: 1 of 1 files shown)


Tensho:
Actually, the similar hook will be nice to have for nav bar (display authentication info), for example. Unfortunately, as a gem user, I can't inject `content_for` inside the engine layout within Rails nested layout mechanism. Probably, the only resonable way is layout coping from gem to the application right now, because we can't predict all possible use cases.
@graysonwright, Do you have a common strategy to extend administrate `application.html.erb` layout? Particularly, it's quite reasonable to keep hook like `page_head`, but what about other stuff?

Tensho:
Sorry, found [this](https://github.com/thoughtbot/administrate/issues/258) conversation with explenations.

carlosramireziii:
@Tensho Is the application layout generator sufficient to resolve your problem? Do you prefer the solution in this PR?

nickcharlton:
Thanks for your PR! I'm going to close this as the activity has been low and there's a potential (linked) solution.

PR #383, opened 2016-01-07
Administrate::Search: ensure that column names are quoted

FooBarWidget:
One of my tables has a column named `group`, which is a keyword in PostgreSQL. This breaks search. This pull request ensures that column names in searches are quoted, which fixes this problem.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -21,7 +21,10 @@ module Administrate
     delegate :resource_class, to: :resolver
 
     def query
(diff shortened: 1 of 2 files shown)


FooBarWidget:
Hi maintainer, any updates on this?

pustomytnyk:
Quoting of column names seem to be already added.

BenMorganIO:
Yep, this is in. It's a pity that this PR came in a year ago and we only just now got this in.

PR #385, opened 2016-01-07
Ensure that collection rows are proper links

FooBarWidget:
Collection rows were fake links simulated by JavaScript. This breaks browsers that have JavaScript disabled, and also breaks middleclick and 'right click -> Open in New Tab'. This commit turns collection rows into proper links.

### DIFF
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -47,3 +47,20 @@
 .cell-label--number {
   text-align: right;
 }
+
+.cell-link {
+  color: inherit;
+  display: block;
+  overflow: hidden;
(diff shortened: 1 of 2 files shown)


FooBarWidget:
Any update on this?

nickcharlton:
Hi @FooBarWidget!

I know it's been quite a while, but would you be willing to rebase this onto current master? I'm very much a fan of removing any _requirement_ for JavaScript so I'd like to see this added.

nickcharlton:
I'm going to close this as we have approximate functionality already. I wondered about wrapping the `<tr>` in a link, but Firefox (81+) won't render that as it's invalid.

We do now link to each item in the row is a link, so we only lose out on the ability to click on the whitespace of the row.

Issue #278, opened 2015-11-30
How to sort by relationship field


Rodrigora:
Hi all!

I searched through documentation and code and didn't find a way to sort the records by a relation field.

This is related to #265.

There is someone working on it? If not, I'd be happy to provide PR.

graysonwright:
Hey, @Rodrigora! Thanks for the suggestion!

Right now this isn't supported. I'm a bit curious about how you'd sort by a relationship - it seems like it would vary depending on whether the relation is a `has_many`, `belongs_to`, `has_one`, or `polymorphic` relationship.

Can you elaborate on what behavior you'd like to see?

Rodrigora:
My models are:

```
class Product
  belongs_to :category
end

class Category
  has_many :product
end
```

I'd like to sort products by category name, since my products page has the columns **id** and **category** (Category column displays the category name).

sebbean:
I've used this before with spree: 
https://github.com/activerecord-hackery/ransack

Could be an easy win.

tomgatzgates:
I've just implemented a very hacky version. This class sits at the bottom of the user controller. 
I'm not suggesting it asa fix but might help in your case :)


```ruby
class SearchUser
  def initialize(resolver, term)
    @resolver = resolver
    @term = term
  end

  SEARCH_TABLES = %w(addresses vouchers)
  # Manually add the joins for these tables

  def run
    if @term.blank?
      resource_class.all
    else
      resource_class.uniq
        .joins(:addresses, user_vouchers: [:voucher])
        .where(query, *search_terms)
    end
  end

  private

  delegate to: :resolver

  def query
    prefixed_attributes.map { |attr| "lower(#{attr}) LIKE ?" }.join(' OR ')
  end

  def search_terms
    ["%#{term.downcase}%"] * prefixed_attributes.count
  end

  def prefixed_attributes
    [resource_class] + SEARCH_TABLES.flat_map do |table|
      attributes_for(table).map { |attr| "#{table}.#{attr}" }
    end
  end

  def attributes_for(dashboard)
    Object.const_get(dashboard.classify + 'Dashboard')::ATTRIBUTE_TYPES.select do |_, type|
      type.searchable?
    end.keys
  end

  attr_reader :resolver, :term
end
```

hughfm:
Does anyone have any updates on this type of thing? Being able to sort/search by an attribute on a belongs_to association would be a huge win for me.

carlosramireziii:
[...]

PR #386, opened 2016-01-08
Support ordering belongs to fields

pedro:
Hey folks!

Do I need to create a new class and stub the hell out of it to test this? I'd much rather have a sample AR model available to tests, but couldn't find any in the suite. Thoughts?

See #278

### DIFF
--- a/lib/administrate/field/belongs_to.rb
+++ b/lib/administrate/field/belongs_to.rb
@@ -24,7 +24,11 @@ module Administrate
       private
 
       def candidate_resources
-        associated_class.all
+        scope = associated_class
+        if options[:order]
+          scope = scope.order(options[:order])
+        end
+        scope.all
       end
 
       def display_candidate_resource(resource)


pedro:
Rebased!

lib/administrate/field/belongs_to.rb
```diff
@@ -24,7 +24,11 @@ def selected_option
       private
 
       def candidate_resources
-        associated_class.all
+        scope = associated_class
+        if options[:order]
+          scope = scope.order(options[:order])
+        end
+        scope.all
```
BenMorganIO: Refactor suggestion:

```ruby
order = options.delete(:order)
order ? associated_class.order(order) : associated_class.all
```

BenMorganIO:
@pedro Tests would be great. If you need a sample AR model to get your tests working, that should be no problem :). Or you could just take one of the existing models and add new "magic" to them.

carlosramireziii:
@pedro I know it's been a while... Are you still interested in getting this PR merged? I think we'd need to get some tests in here and maybe some documentation on the new feature. 

LMK what you think.

nickcharlton:
Closing this as #939 has since been merged.

PR #389, opened 2016-01-08
Applies a margin-bottom to attribute labels

naps62:
P.S.: Even with the known issues I present below, I still prefer this over the current version, which is why I chose to propose this. No hard feelings if it doesn't go through :smile: 

Why:
- When a `nil` attribute exists, the data element takes no height, which
  results in the row being too close to the next attribute.

- The spacing between each attribute was previously given by the
  `attribute-data` field. But when it has no height, that margin gets
  swallowed by the label. Applying the same margin to it helps
  to fix this

Known issues:
- `$spacing-base` is measured in `em`s. And since the label has a
  different font-size, this results in a slightly different margin, which
  will be noticeable in the same cases as in the original issue, but I
  guess a different margin is better than no margin.
- When the label spans multiple lines, an even larger spacing gets created

Before:
screenshot from 2016-01-08 10-46-17

After:
screenshot from 2016-01-08 16-07-41

### DIFF
--- a/app/assets/stylesheets/administrate/components/_attributes.scss
+++ b/app/assets/stylesheets/administrate/components/_attributes.scss
@@ -4,7 +4,6 @@
   clear: left;
   margin-top: 0;
   text-align: right;
-  clear: left;
   margin-bottom: $base-spacing;
 }
 
--- a/app/assets/stylesheets/administrate/components/_attributes.scss
+++ b/app/assets/stylesheets/administrate/components/_attributes.scss
@@ -4,6 +4,8 @@
   clear: left;
   margin-top: 0;
   text-align: right;
+  clear: left;
+  margin-bottom: $base-spacing;
 }
 
 .preserve-whitespace {


tysongach:
PR https://github.com/thoughtbot/administrate/pull/399 introduces a new, more consistent manner of marking up and styling forms, which should resolve this issue (because it no longer relies on the actual inputs or labels to individually apply `margin`). Though, I’ve not directly tested it yet against this case.

nickcharlton:
@tysongach Just over a year later; what are your thoughts on this now?

tysongach:
Is this still an issue after we shipped those CSS improvements in #399?

nickcharlton:
Excellent news!

@naps62 Could you rebase on `master` or [enable contributor pushes](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

naps62:
@nickcharlton rebased, and contributor pushes enabled as well (in case something else is needed)

However, I no longer have a working app where I can quickly test this, so I can't be 100% if this is still correct. I might be able to do it tomorrow if you don't have the time to.

nickcharlton:
Great, thanks! Merging in.

PR #390, opened 2016-01-08
Fix timestamp deprecation warning

danbee:
Tiny PR to fix the deprecation warning when specs are run.


graysonwright:
:+1: Merge away!

Issue #388, opened 2016-01-08
Always go to show page


samnang:
When I tried to add another column in index page to show links which contain thumbnail like:

``` ruby
<%= link_to main_app.facebook_profile_url(resource.fb_id), target: "_blank" do  %>
  <%= image_tag resource.avatar.thumbnail.url %>
<% end %>
```

So whenever users click on thumbnail, it opens the url and as well as go to show page. We don't want to user go to show page. I notice it's always go to show page if event's target is not a link:
https://github.com/thoughtbot/administrate/blob/5f444611164a0b0494ee76fa30e6b577b93decef/app/assets/javascripts/administrate/components/table.js#L9-L10

Do you have any ideas or suggestion to address it? I could work on PR on those ideas.

pablobm:
Associated PR https://github.com/thoughtbot/administrate/pull/392 was closed.

PR #392, opened 2016-01-08
Allow linkable element on row and shouldn't go to show page

samnang:
See https://github.com/thoughtbot/administrate/issues/388

The previous problem that I had when I tried to have a column which is
clickable user's thumbnail that should go to their fecebook profile.
Unfortunately, when user clicks on thumbnail it opens facebook profile
page which is correct, but it also redirects user to show page as always
when user clicks on row.

The solution, whenever the column marks with `cell-data--linkable`, then
we assume content in that column will contain linkable elements which we
don't need to have the behaviour that always go to show page when they
click on that row.

### DIFF
--- a/app/assets/javascripts/administrate/components/table.js
+++ b/app/assets/javascripts/administrate/components/table.js
@@ -6,10 +6,15 @@ $(function() {
         event.keyCode == keycodes.space ||
         event.keyCode == keycodes.enter) {
 
-      if(!event.target.href) {
+      if(!isLinkableElement(event.target)) {
         window.location = $(event.target).closest("tr").data("url");
       }
     }
+  };
+
+  var isLinkableElement = function(element) {
+    return (event.target.href ||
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/administrate/components/table.js
```diff
@@ -6,12 +6,17 @@ $(function() {
...
-      if(!event.target.href) {
+      if(!isLinkableElement(event.target)) {
         window.location = $(event.target).closest("tr").data("url");
       }
     }
   };
 
+  var isLinkableElement = function(element) {
+    return (event.target.href ||
+            $(element).closest(".cell-data--linkable").length > 0);
```
carlosramireziii: Hey there! I'm just a little unclear on how this behavior works.. Would you mind explaining it to me?

nickcharlton:
Thanks for this PR! I'm going to close as it's been a quite since the last response. I'd be more than happy to see this reopened, of course.

PR #395, opened 2016-01-09
Fix _collection.html.erb inline documentation

FooBarWidget:
Rename 'Table' to 'Collection'


graysonwright:
woot, thanks!

Merged this in as fd577b2.

PR #398, opened 2016-01-10
Upgrade to Rails 5.0

fx:
It might be a little early, but I wanted to use Rails 5 w/ Administrate.

Probably not a candidate for merging right away as it depends on a couple of unreleased gem versions (see Gemfile for notes.) But I wanted to get the process kicked off here.

This is backwards-compatible w/ Rails 4.2+, so for a merge we should swap the gemspec back to >= 4.2

No major complications to report.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
 source 'https://rubygems.org'
+git_source(:github){ |repo_name| "https://github.com/#{repo_name}" } if respond_to?(:git_source)
 
 ruby "2.2.3"
 
--- a/Gemfile
+++ b/Gemfile
@@ -23,15 +23,32 @@ group :development, :test do
   gem "faker"
   gem "i18n-tasks"
   gem "pry-rails"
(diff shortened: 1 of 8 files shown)


Gemfile
```diff
@@ -1,4 +1,5 @@
 source 'https://rubygems.org'
+git_source(:github){ |repo_name| "https://github.com/#{repo_name}" } if respond_to?(:git_source)
```
pacso: The `:github` shortcut is included with bundler (as is `:gist` and `:bitbucket`), so defining a custom source for it isn't necessary.

fx: The default is using the git protocol, not https.

pacso: Fair enough.

hayesr:
Hey, just FYI, I got things working with Rails 5 in #376

fx:
Oh thanks for the heads up @hayesr, didn't look closely enough. Looks complete, I'll close this one.

PR #379, opened 2016-01-06
CSS Refactor

tysongach:
This PR is more improvements to Administrate’s stylesheets architecture. It’s a continuation of #347.

- Administrate’s `stylesheets` directory is now more organized and clear:
  - `base` – all styles for base HTML elements (no classes, id’s, etc.)
  - `components` – the majority of styles, these are all of our components that create the Administrate design
  - `library` – any non-rendering Sass, e.g. variables and mixins
  - `application.scss` – the main Sass manifest, which imports all of the individual Sass partials
- De-variablized some CSS value that didn’t gain much value from being a variable
- More closely aligned to thoughtbot’s SCSS style guide; running `scss-lint` is down to one warning now (which is an edge case)
- All changes in this PR are mostly structural and do little to change visual design
- Plus much more!

PR #399, opened 2016-01-10
Refactor form styles

tysongach:
This PR refactors the styling of forms. It introduces a new reusable pattern for marking up and styling form elements. That pattern is a `field-unit` class, which was loosely called a `form-field` previously.

A `form-unit` is simply a grouping of elements that create a whole “field” in a form, the most common being a simple label/input pair (classed as `field-unit__label` and `field-unit__field` now). But a `form-unit` can have many children beyond `field-unit__label` and `field-unit__field`, for example, if we have a hint for the input, we could add a `field-unit__field-hint`. It offers a lot of flexibility moving forward and, overall, DRYs up a lot of the styles around forms:

- We used to have to set margins on every type of input a form had (which is a bit of a moving target), e.g. `input`, `select`, `.selectize`, etc. This gets out of hand quickly and results in duplicated margin values all over our CSS. Now that we have a consistent `field-unit` element which should wrap each whole “field” in a form, we can define the margin here, and only do it once.
- The old `components/_form.scss` partial, is gone and means we don’t have lots of base element styles _outside_ of the `base/` stylesheet directory
- This consistent structure will allow us to more easily make these form fields responsive
- Helps to eliminate specificity issues, some of which were brought up in https://github.com/thoughtbot/administrate/pull/273 and a very similar solution was discussed there.

Some look and feel has been changed in this PR (screenshots below). It wouldn’t be much effort to hit a near exact match to how it looks now, but I’m more concerned with establishing a great base for styling.

#### Before

screen shot 2016-01-10 at 13 27 29

#### After

screen shot 2016-01-10 at 13 27 34

- [x] Note: This PR depends on #379 getting merged.


app/assets/stylesheets/administrate/components/_field-unit.scss
```diff
@@ -0,0 +1,17 @@
...
+  @include clearfix;
+  @include fill-parent;
+  align-items: center;
+  display: flex;
+  margin-bottom: $base-spacing;
+  position: relative;
+}
+
+.field-unit__label {
+  @include span-columns(2 of 12);
```
creuter: Do you think some sort of flex-basis on the label would be more appropriate here, and just let the `.field-unit__field` fill with width?

app/views/fields/belongs_to/_form.html.erb
```diff
@@ -16,7 +16,11 @@ that displays all possible records to associate with.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/BelongsTo
 %>
 
-<%= f.label field.permitted_attribute %>
-<%= f.select(field.permitted_attribute) do %>
-  <%= options_for_select(field.associated_resource_options, field.selected_option) %>
-<% end %>
+<div class="field-unit__label">
```
creuter: What's the reasoning for wrapping the label, rather than putting the class on the label?

app/assets/stylesheets/administrate/components/_field-unit.scss
tysongach: Great point! I’m actually leaning towards something like; in fact, I’ve been looking at maybe just removing Neat all together, because this is really the only place it’s used.

For now, I just wanted to get this refactored and come out with minor visual adjustments at the end. Another PR could be visual improvements, which lead to some smarter flexbox usage.

app/views/fields/belongs_to/_form.html.erb
tysongach: I’m split on that, but mostly just to offer the ability to put a hint alongside the label, or a required flag, or any other element.

creuter:
:ear_of_rice:

app/assets/stylesheets/administrate/components/_field-unit.scss
```diff
@@ -0,0 +1,17 @@
+.field-unit {
```
whmii: Very high level, but I don't find this selector very descriptive

app/assets/stylesheets/administrate/components/_form.scss
```diff
@@ -1,49 +0,0 @@
-.form {
```
whmii: Yaaaayyy… :kissing_heart:

app/assets/javascripts/administrate/components/has_many_form.js
```diff
@@ -1,3 +1,3 @@
 $(function() {
-  $(".form-field--has-many select").selectize({});
+  $(".field-unit--has-many select").selectize({});
```
whmii: data attributes if you want to be :sunglasses:

whmii:
Nice clean up. looks good.

app/assets/javascripts/administrate/components/has_many_form.js
tysongach: Good point, but I didn’t originally write this JavaScript and didn’t want to change it in a CSS-focused PR…

app/assets/stylesheets/administrate/components/_field-unit.scss
tysongach: Yeah, it’s tough to find a name for this, but wanted to run with something. I see it as a collective unit of form elements, i.e. a label, input, hint, etc. Could be any number of those as children to this.

Open to suggestions of a better name!

graysonwright:
Let's ignore Percy for the time being. We've used up the quota for the month, so if you rebase you'll get a green build.

tysongach:
Merged.

app/assets/javascripts/administrate/components/has_many_form.js
graysonwright: ooh, I like that idea. Good call. :smile:

PR #408, opened 2016-01-14
Fix UI regressions

graysonwright:
When merging in #198 and #214,
the Percy.io visual regression tests weren't run
because our quota for the month had been used up.

Without the visual regression tests,
we failed to notice two visual regressions that were introduced.

The two commits in this PR fix the visual regressions that were introduced.
See the commit descriptions for more information.

### DIFF
--- a/app/assets/stylesheets/administrate/components/_sidebar.scss
+++ b/app/assets/stylesheets/administrate/components/_sidebar.scss
@@ -2,17 +2,13 @@
   flex: 0 0 auto;
   max-width: 12em;
   overflow-y: auto;
-  padding: 0 $base-spacing;
+  padding: 0 $base-spacing $base-spacing;
 
   &__link {
     color: $base-font-color;
     display: block;
     padding-top: $base-spacing;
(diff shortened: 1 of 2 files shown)


graysonwright:
Unfortunately... we've already run through our Percy quota for this month, so we can't automatically verify that these fixes are valid.

app/assets/stylesheets/administrate/components/_sidebar.scss
```diff
@@ -3,6 +3,7 @@
   max-width: 12em;
   overflow-y: auto;
   padding: 0 $base-spacing;
+  padding-bottom: $base-spacing;
```
tysongach: This addition means we are defining `padding` twice. Line 5 applies `0` to the `padding-bottom`, and Line 6 we overwrite that.

Instead, we can just update Line 5 to; `padding: 0 $base-spacing $base-spacing;`

tysongach:
One minor comment, otherwise :+1:

PR #409, opened 2016-01-16
Use System Fonts

tysongach:
Lato isn't particularly made for UI uses and I felt it was a little rough around
the edges; almost too friendly and unstructured for Administrate.

Instead, we can just use system fonts, especially now that Apple and Google have
both creating their own typefaces for UI use and are really strong typefaces.
Plus we get the benefit of not having to load web fonts!

Reference: https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/

It also means we have one less thing to rely on (Google Fonts)

#### Before (with Lato)

screen shot 2016-01-16 at 15 26 09

#### After (with system fonts, rendered as Apple’s San Francisco here)

screen shot 2016-01-16 at 15 26 20

### DIFF
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -22,7 +22,7 @@
 
 .cell-label--asc,
 .cell-label--desc {
-  font-weight: bold;
+  font-weight: 600;
 }
 
 .cell-label__sort-indicator {
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -22,7 +22,7 @@
 
 .cell-label--asc,
 .cell-label--desc {
-  font-weight: 600;
+  font-weight: $bold-font-weight;
 }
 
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/administrate/components/_cells.scss
```diff
@@ -25,7 +25,7 @@
 
 .cell-label--asc,
 .cell-label--desc {
-  font-weight: bold;
+  font-weight: 600;
```
joshuaogle: Fancy creating a `$bold-font-weight` variable for clarity?

tysongach: Yeah, probably a good idea. Updated.

creuter:
Looks good. Much more appropriate.

joshuaogle:
Lato's one of those fonts that looks great until you actually get it into a project. System fonts don't get enough respect. I like it :+1:

graysonwright:
I'm good with it! You had me at "one fewer dependency"

tysongach:
:laughing:

PR #410, opened 2016-01-16
Accessible tables

tysongach:
- Add aria-labelledby to tables
- Add scope attributes to tables

Reference: http://webaim.org/techniques/tables/data

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -18,7 +18,7 @@ to display a collection of resources in an HTML table.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Page/Table
 %>
 
-<table class="collection-data">
+<table class="collection-data" aria-labelledby="page-title">
   <thead>
     <tr>
       <% collection_presenter.attribute_types.each do |attr_name, attr_type| %>
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -24,7 +24,7 @@ to display a collection of resources in an HTML table.
       <% collection_presenter.attribute_types.each do |attr_name, attr_type| %>
         <th class="cell-label cell-label--<%= attr_type.html_class %>
           cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
-        ">
+        " scope="col">
         <%= link_to(params.merge(
           collection_presenter.order_params_for(attr_name)
         )) do %>
@@ -38,7 +38,7 @@ to display a collection of resources in an HTML table.
           <% end %>
         </th>
       <% end %>
-      <th colspan="2"></th>
+      <th colspan="2" scope="col"></th>
     </tr>
   </thead>
 
--- a/app/views/administrate/application/index.html.erb
+++ b/app/views/administrate/application/index.html.erb
@@ -44,7 +44,7 @@ It renders the `_table` partial to display details about the resources.
 <% end %>
 
 <header class="header">
-  <h1 class="header__heading"><%= content_for(:title) %></h1>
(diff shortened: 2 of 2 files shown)


app/views/administrate/application/_collection.html.erb
```diff
@@ -18,13 +18,13 @@ to display a collection of resources in an HTML table.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Page/Table
 %>
 
-<table class="collection-data">
+<table class="collection-data" aria-labelledby="page-title">
```
ckundo: Do you think a `<caption>` might work better here? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/caption

tysongach: Yeah, great point. A `caption` element represents the exact need here. The only reason I reached for `aria-labelledby` instead is because we are already outputting the title on the page and this allowed me to easily hook onto that.

If we added a `caption`, we’d render the title twice in the DOM and likely just visually hide the `caption`. Using the `caption` _instead_ of the page’s `h1` that we already have won’t work with the layout already in place…

ckundo: @tysongach do you know if `aria-labelledby` translates to the same accessibility property as caption? If so, that's great. We might need to use `aria-describedby` depending on the mapping of the `caption` element.

tysongach: Great question! I’m not sure. How would one find this out?

ckundo: >  It provides the user with a recognizable name of the object. The most common accessibility API mapping for a label is the accessible name property.

http://www.w3.org/TR/wai-aria/states_and_properties#aria-labelledby

> The accessible name is the name of a user interface element. Each platform accessibility API provides the accessible name property. The value of the accessible name may be derived from a visible (e.g., the visible text on a button) or invisible (e.g., the text alternative that describes an icon) property of the user interface element.

http://www.w3.org/TR/wai-aria/terms#def_accessible_name

ckundo: Question is, how does caption map to the accessibility API. I suspect it acts as the accessible name of the table.

ckundo: w3c suggests using `aria-describedby`:

> Approach 2: Using aria-describedby to provide a table summary

In this approach an element with an id attribute is associated as a summary by using the aria-describedby attribute of the table. Any element with a unique id attribute can be used as a summary for a table in this way.

The element containing the summary doesn’t need to be in front of the table in the document, but it helps users to discover the summary more easily if the summary is in close proximity of the table, especially if they are not using a screen reader.

https://www.w3.org/WAI/tutorials/tables/caption-summary/

ckundo: Ah, nevermind, the `aria-describedby` defines the help text, not the accessible name. 
screen shot 2016-01-19 at 4 37 42 pm

tysongach: Very interesting…

I have mixed feelings, though, because an example heading (which is what I’m tying the ARIA attribute to) for these pages is some thing like “Customers”. So it’s not really a “summary”, per se, and according to the description from the W3:

> A summary conveys information about the organization of the data in a table and helps users navigate it.

tysongach:
Thank you @ckundo, for diving into this and making sure this is the best route.

Here’s what we get, before and after, these changes:

#### Before

screen shot 2016-01-19 at 16 47 19

#### After

screen shot 2016-01-19 at 16 47 37

tysongach:
@ckundo So, do you think this is good to merge?

ckundo:
@tysongach lgtm!

tysongach:
@ckundo Thanks, buddy!

PR #411, opened 2016-01-16
Use tabular figures in table cells

tysongach:
Tab­u­lar fig­ures are set on a fixed width, so that every fig­ure oc­cu­pies the same
amount of hor­i­zon­tal space.

Reference: http://practice.typekit.com/lesson/caring-about-opentype-features/

Take a look at the “Lifetime Value” column in the GIF below and notice how after enabling this features, all of the numbers align horizontally making for a more readable experience.

table-type

### DIFF
--- a/app/assets/stylesheets/administrate/base/_tables.scss
+++ b/app/assets/stylesheets/administrate/base/_tables.scss
@@ -16,6 +16,10 @@ td {
 
 td {
   border-bottom: $base-border;
+  font-feature-settings: "kern", "liga", "clig", "calt", "lnum", "tnum";
+  font-kerning: normal;
+  font-variant-ligatures: common-ligatures, contextual;
+  font-variant-numeric: lining-nums, tabular-nums;
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
--- a/.scss-lint.yml
+++ b/.scss-lint.yml
@@ -125,7 +125,8 @@ linters:
 
   PropertySpelling:
     enabled: true
-    extra_properties: []
+    extra_properties:
+      - font-variant-numeric
     disabled_properties: []
 
   PropertyUnits:


app/assets/stylesheets/administrate/base/_tables.scss
```diff
@@ -16,6 +16,10 @@ td {
 
 td {
   border-bottom: $base-border;
+  font-feature-settings: "kern", "liga", "clig", "calt", "lnum", "tnum";
+  font-kerning: normal;
+  font-variant-ligatures: common-ligatures, contextual;
+  font-variant-numeric: lining-nums, tabular-nums;
```
tysongach: It’s real. I think Firefox uses it: https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric

I think I might be able to tell SCSS-Lint that this is an acceptable property within it’s config…

tysongach: Yep! Added in 23efbd1059901398c08b6335117aa997eb040610

tysongach:
Here’s a better representation of non-tabular versus tabular numbers:

#### Non-tabular numbers
tabular-off

#### Tabular numbers
tabular-on

joshuaogle:
Very nice :+1: 

Just to double check, we want to only use those font settings on table cells? Any benefit/harm to doing them more broadly?

tysongach:
@joshuaogle These _specific_ settings, yes, just on table cells. Tabular numbers, for example, wouldn’t feel right within a paragraph.

We can likely improve the type in many other areas, as well, though. I bet some helpful adjustments on form inputs, or other places where there’s some heavy data, could be nice. Let’s focus on those in other PR’s…

graysonwright:
:tada:

PR #416, opened 2016-01-18
Add a label for the search input

### LINKED ISSUE #415: [Accessibility] Form elements must have labels
The search input is missing a label.

I ran the [aXe](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US) Chrome extension, and it suggests to fix any of the following:

- `aria-label` attribute does not exist or is empty
- `aria-labelledby` attribute does not exist, references elements that do not exist or references elements that are empty or not visible
- Form element does not have an implicit (wrapped) `<label>`
- Element has no `title` attribute or the `title` attribute is empty

tysongach:
- Properly label the input, to improve accessibility~~, but visually hide it to
  maintain the design, which uses a placeholder to indicate use~~
~~- Uses the hide content technique from The Accessibility Project:
  http://a11yproject.com/posts/how-to-hide-content/~~
- Fixes #415

### DIFF
--- a/app/assets/stylesheets/administrate/application.scss
+++ b/app/assets/stylesheets/administrate/application.scss
@@ -10,3 +10,4 @@
 @import "library/**/*";
 @import "base/**/*";
 @import "components/**/*";
+@import "utilities/**/*";
--- a/app/assets/stylesheets/administrate/application.scss
+++ b/app/assets/stylesheets/administrate/application.scss
@@ -10,4 +10,3 @@
 @import "library/**/*";
 @import "base/**/*";
 @import "components/**/*";
-@import "utilities/**/*";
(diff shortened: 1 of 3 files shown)


joshuaogle:
Haven't seen that solution before, but it makes sense. Hooray for accessbility :+1:

tysongach:
@joshuaogle Yeah, I’m planning to add some sort of `hide-visually` mixin to Bourbon 5, which would wrap this up. Very similar to `hide-text`.

tysongach:
@ckundo Would you mind taking a look at this one, when you have a chance?

tysongach:
@LkeMitchll brings up a good point over on https://github.com/thoughtbot/bourbon/pull/818, that maybe I should just use `aria-label` (duh!). I totally glazed over that for some reason…

I imagine both approaches reach the same level of accessibility, but could be interesting to dig in to.

ckundo:
@tysongach `aria-label` seems more standards oriented if we want a non-visual label.

graysonwright:
Big :+1: for making it more accessible. I'd also lean towards `aria-label` if it's functionally the same.

tysongach:
Updated to use `aria-label`.

graysonwright:
:+1:

PR #422, opened 2016-01-19
Add support for selectable strings on edit pages

Jberlinsky:
This pull request adds support for collection select fields, much like `simple_form`'s `collection: [:foo, :bar]` option. This also provides a base field that will be helpful in implementing the `HasOne` field that is currently pending implementation.

**Example User Story**

**As a** developer
**I want** to allow administrative users to select values from a drop-down rather than enter them free-form
**So that** I don't need to worry quite as much about data integrity in my administrative interface.

### DIFF
--- a/app/views/fields/select/_index.html.erb
+++ b/app/views/fields/select/_index.html.erb
@@ -1,0 +1,16 @@
+<%#
+# Select Index Partial
+
+This partial renders a selectable text attribute,
+to be displayed on a resource's index page.
+
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::Select][1].
+  A wrapper around the attribute pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Select
+%>
+
+<%= field.data.presence %>
--- a/app/views/fields/select/_show.html.erb
+++ b/app/views/fields/select/_show.html.erb
@@ -1,0 +1,16 @@
+<%#
+# Select Show Partial
+
+This partial renders a selectable text attribute,
+to be displayed on a resource's show page.
+
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::Select][1].
+  A wrapper around the attribute pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Select
+%>
+
+<%= field.data.presence %>
--- a/spec/example_app/app/models/customer.rb
+++ b/spec/example_app/app/models/customer.rb
@@ -4,6 +4,11 @@ class Customer < ActiveRecord::Base
   validates :name, presence: true
   validates :email, presence: true
 
+  KINDS = [
+    :standard,
+    :vip,
+  ]
+
   def lifetime_value
     orders.map(&:total_price).reduce(0, :+)
   end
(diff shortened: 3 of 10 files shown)


tysongach:
Nice website.

Jberlinsky:
@tysongach Yeah...best form of flattery? There's a new one in the works.

app/views/fields/select/_index.html.erb
```diff
@@ -0,0 +1,16 @@
...
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::Select][1].
+  A wrapper around the attribute pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Select
+%>
+
+<%= field.data.presence %>
```
graysonwright: Is the `.presence` necessary? I would expect it to behave the same without it, because the result is getting rendered as a string either way.

app/views/fields/select/_show.html.erb
```diff
@@ -0,0 +1,16 @@
...
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::Select][1].
+  A wrapper around the attribute pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Select
+%>
+
+<%= field.data.presence %>
```
graysonwright: same question.

spec/features/edit_page_spec.rb
```diff
@@ -35,4 +35,26 @@
...
+    visit edit_admin_customer_path(customer)
+
+    select "vip", from: "Kind"
+
+    click_on "Update Customer"
+
+    expect(page).to have_content("KIND")
+    expect(page).to have_content("vip")
+
+    click_on "Edit"
```
graysonwright: It seems like everything below this point is testing the same functionality as above. Can we limit it to only one form submit?

spec/features/edit_page_spec.rb
```diff
@@ -35,4 +35,26 @@
...
     expect(page).to have_content("true")
   end
+
+  it "displays selectable strings as dropdowns", :js do
+    customer = create(:customer, kind: :standard)
+
+    visit edit_admin_customer_path(customer)
+
+    select "vip", from: "Kind"
+
```
graysonwright: We haven't been putting any newlines between the action lines of our test cases. Would you mind removing lines 43 and 45? Thanks!

graysonwright:
I really like this! I left a few small comments.

Jberlinsky:
@graysonwright Addressed your concerns; will squash whenever you're ready.

spec/example_app/app/models/customer.rb
```diff
@@ -4,6 +4,11 @@ class Customer < ActiveRecord::Base
   validates :name, presence: true
   validates :email, presence: true
 
+  KINDS = [
```
Jberlinsky: This array is frozen -- Hound isn't catching that it's a couple lines down.

pedantic-git:
Hm - this conflicts with my gem https://rubygems.org/gems/administrate-field-select and is missing some of the features of that gem. I'll rename the class in my gem so it doesn't conflict but perhaps we can work together to merge the two concepts into one at some point?

PR #424, opened 2016-01-20
Let users undo the `administrate:views` generator

### LINKED ISSUE #310: destroy administrate:views not working
it seems that `generate` and `destroy` do the same thing:
```
masciugo@macbookpro ~/dev/ffm $ rails generate administrate:views Course
      create  app/views/admin/courses/index.html.erb
      create  app/views/admin/courses/_collection.html.erb
      create  app/views/admin/courses/show.html.erb
      create  app/views/admin/courses/new.html.erb
      create  app/views/admin/courses/_form.html.erb
      create  app/views/admin/courses/edit.html.erb
   identical  app/views/admin/courses/_form.html.erb
masciugo@macbookpro ~/dev/ffm $ rails destroy administrate:views Course
   identical  app/views/admin/courses/index.html.erb
   identical  app/views/admin/courses/_collection.html.erb
   identical  app/views/admin/courses/show.html.erb
   identical  app/views/admin/courses/new.html.erb
   identical  app/views/admin/courses/_form.html.erb
   identical  app/views/admin/courses/edit.html.erb
   identical  app/views/admin/courses/_form.html.erb
```

graysonwright:
Fixes #310

## Problem:

Rails provides a way to undo the effects of generators,
by running the command `rails destroy GENERATOR_NAME`.

With the way that we had set up our hierarchy of generators,
Rails did not know how to undo the effects of sub-generators.

For example, with the `administrate:views` generator,
we were calling `Rails::Generators.invoke("administrate:views:index")`.

Whether the generator was run with the `generate` or `destroy` command
did not make a difference -
the index generator would always be invoked
as if it were run with `generate`.

## Solution:

Rails generators use the `@behavior` instance variable to keep track of
how the generator was run.
This variable can be either `:invoke` or `:revoke`.

Pass this variable as an option to the sub-generators
to ensure they have the same behavior as the parent.

## Minor changes:

Extract `Administrate::GeneratorHelpers` to store common
generator-related methods.

### DIFF
--- a/lib/administrate/generator_helpers.rb
+++ b/lib/administrate/generator_helpers.rb
@@ -1,0 +1,13 @@
+module Administrate
+  module GeneratorHelpers
+    def call_generator(generator, *args)
+      Rails::Generators.invoke(generator, args, generator_options)
+    end
+
+    private
+
+    def generator_options
+      { behavior: behavior }
+    end
+  end
+end
--- a/lib/administrate/view_generator.rb
+++ b/lib/administrate/view_generator.rb
@@ -1,7 +1,10 @@
 require "rails/generators/base"
+require "administrate/generator_helpers"
 
 module Administrate
   class ViewGenerator < Rails::Generators::Base
+    include Administrate::GeneratorHelpers
+
     private
 
     def self.template_source_path
--- a/spec/generators/views_generator_spec.rb
+++ b/spec/generators/views_generator_spec.rb
@@ -1,4 +1,4 @@
-require "spec_helper"
+require "rails_helper"
 require "generators/administrate/views/views_generator"
 require "support/generator_spec_helpers"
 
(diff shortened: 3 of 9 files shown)


lib/administrate/generator_helpers.rb
```diff
@@ -0,0 +1,7 @@
+module Administrate
+  module GeneratorHelpers
+    def call_generator(generator, *args)
+      Rails::Generators.invoke(generator, args, behavior: behavior)
```
gylaz: What do you think about passing `behavior` to `call_generator`? Otherwise, it seems like a mystery guest.

lib/administrate/view_generator.rb
```diff
@@ -1,7 +1,9 @@
 require "rails/generators/base"
+require "administrate/generator_helpers"
 
 module Administrate
   class ViewGenerator < Rails::Generators::Base
+    include Administrate::GeneratorHelpers
```
gylaz: Newline after this (before `private`)?

spec/generators/views_generator_spec.rb
```diff
@@ -1,4 +1,4 @@
-require "spec_helper"
+require "rails_helper"
```
gylaz: What is the reason this change is necessary?

gylaz:
Left a few suggestions/questions. Looking good!

spec/generators/views_generator_spec.rb
graysonwright: When I ran this spec in isolation, it complained about not being able to find a module that Rails provides. I think we missed it before because we were always running this spec alongside others.

lib/administrate/view_generator.rb
graysonwright: done.

lib/administrate/generator_helpers.rb
graysonwright: normally I'd agree, but that would defeat the purpose of extracting this method. I wanted to automatically include the parent's behavior, instead of passing it in explicitly.

PR #426, opened 2016-01-21
Added rails generate administrate:dashboard

benhalpern:
This is a core command and it is a bit confusing to find. I think it should be in the readme.


ACPK:
+1. I was also very surprised that it wasn't included.

carlosramireziii:
@nickcharlton A potentially more updated version of this PR is on #684

nickcharlton:
Thanks for this! I'm going to close this in favour of #684 which is a little newer.

Issue #349, opened 2015-12-17
Allow easier hooks to provide simple, custom stylesheet/javascript requires


coneybeare:
**Problem**

I made a custom Font-Awesome field that displays a stored icon name string ("fa-facebook") as its rendered icon rather than its name. After getting the appropriate gem, I need to add one line into the `application.css` file.

Generating the entire `administrate:assets:stylesheets` stack so I can plug this one value in seems a bit sledgehammerish.

**Ideas**

Not sure if any/all of these are technically possible.
- Somehow add a hook into the stylesheet and javascript manifest files that can `require` into the main app by looking in a special folder or file.
- Have main app define the root manifest files for the engine, with only a `require administrate.scss` etc... entry in it. This would allow additional requires before or after the gem scss.
- Setup some sort of config variable that could take some paths to require within `administrate.scss`

Or maybe I am just dense and am not seeing the easy way to do this without freezing the entire stylesheet stack into my main app. Ideas?

sebbean:
having this same issue right now. trying to just get something to work. curious how you accomplished injecting css

mdkirby:
One option may be to do the same thing that you guys do for javascript - have a partial that can be overridden that has a `content_for(:stylesheet)`

That way, you could use a small stylesheet and put the tag in a `content_for` block - if you only need to customize styles for certain pages. If you need a custom style to apply to everything, then you could just create your own `_stylesheets.html.erb`

bessey:
An alternative approach, one I've just taken is to generate the layout with `rails g administrate:views:layout` and either add your custom stylesheet to the head, or `@import "administrate/application";` in your custom stylesheet and completely replace the `stylesheet_link` line

carlosramireziii:
@coneybeare Wondering if [this commit](https://github.com/thoughtbot/administrate/commit/e4cc0921fba9bf713ac810161725e9b13e9102e2) solves your problem here. 

You can now call `Administrate::Engine.add_stylesheet` to pass a custom stylesheet to be included on every admin page. Does this solve the issue?

cc @sebbean @marksiemers @bessey

coneybeare:
I'm not using administrate anymore, but this does look like it would have resolved the issue.

carlosramireziii:
@coneybeare Awesome, thanks for the update! If you give Administrate another try in the future, we'd welcome your feedback. Thanks again :)

Issue #429, opened 2016-01-22
Undefined variable: "$bold-font-weight".


stevepsharpe:
Hi all

I'm using the gem directly from Github and if I run rails `generate administrate:assets:stylesheets`. I get the following error `Undefined variable: "$bold-font-weight".` when trying to load anything in /admin 

I also noticed that some of the generators seem to produce different code from the default code. For example 'rails generate administrate:views:new Client' the forms all look broken.

Any ideas how to resolve this?

tysongach:
Here’s a relevant commit, which added the `$bold-font-weight` Sass variable: https://github.com/thoughtbot/administrate/commit/28e084f10c7b65f0779ab7e1dc52f1d7b3fe16fe

`master` is always a bit in flux, and especially the past couple weeks has seen a lot of updates.

Can you look in `app/assets/stylesheets/administrate/library/_variables.scss` and see if `$bold-font-weight` is defined in there?

stevepsharpe:
Yeah, it appears to all be there. It just isn't being read correctly for some reason.

stevepsharpe:
I just set the gem to [`cf1e39`](https://github.com/thoughtbot/administrate/commit/cf1e39f92f60b2db705bfeac412ebbf808bf3198) and it seems to be working ok now. Maybe it's related to the Bourbon 5 commit?

stevepsharpe:
Ok, I spoke too soon. It's still doing it. I restarted the server and now it's doing it all of the time.

tysongach:
@stevepsharpe v0.1.3 was just released a few minutes ago and packages up everything that was on `master`. Can you swap for this official version and see if the issue persists?

stevepsharpe:
Ok gem updated, everything works as expected until I run `rails generate administrate:assets:stylesheets`, then you start to get the `Undefined variable: "$bold-font-weight"` issue.

j-dexx:
I have the same issue precompiling assets

commuun:
Ran into the same issue. There seems to be something strange going on. The full error I get on precompile is:

    Sass::SyntaxError: Undefined variable: "$bold-font-weight".
    app/assets/stylesheets/administrate/base/_forms.scss:9

As a test I removed the line `@import "base/**/*";` from `app/assets/stylesheets/administrate/application.scss` but the error persists. At that point it should not be trying to include `_forms.scss`, but it does.

Decoherence:
Same issue here.

[...]

PR #436, opened 2016-01-24
Add extension point for custom styles.

keithlayne:
This PR addresses #349.  

Notes:
- Behavior seemed inconsistent when `app/assets/stylesheets/administrate/customize.scss` was empty, but seemed to work as expected with only a comment
- Testing this was weird.  I tried to come up with something that wouldn't break the example app, but still reliably demonstrate behavior.
- Capybara gave me unintuitive results - checking visibility didn't seem to work at all, `display:none` seemed to make the div totally invisible to Capybara.

This seems too simple to screw up, but if I missed something please let me know.

### DIFF
--- a/app/assets/stylesheets/administrate/application.scss
+++ b/app/assets/stylesheets/administrate/application.scss
@@ -10,3 +10,5 @@
 @import "library/**/*";
 @import "base/**/*";
 @import "components/**/*";
+
+@import "administrate/customizations";
--- a/app/assets/stylesheets/administrate/_customizations.scss
+++ b/app/assets/stylesheets/administrate/_customizations.scss
@@ -1,0 +1,1 @@
+// Place custom css in app/assets/stylesheets/administrate/customizations.scss
--- a/spec/example_app/app/assets/stylesheets/administrate/_customizations.scss
+++ b/spec/example_app/app/assets/stylesheets/administrate/_customizations.scss
@@ -1,0 +1,3 @@
+.custom-administrate-css-hidden {
+  display: none;
+}
--- a/spec/example_app/app/views/admin/application/_flashes.html.erb
+++ b/spec/example_app/app/views/admin/application/_flashes.html.erb
@@ -1,0 +1,24 @@
+<%#
+# Flash Partial
+
+This partial renders flash messages on every page.
+
+## Relevant Helpers:
+
+- `flash`:
+  Returns a hash,
+  where the keys are the type of flash (alert, error, notice, etc)
+  and the values are the message to be displayed.
+%>
+
+<% if flash.any? %>
+  <div class="flashes">
+    <% flash.each do |key, value| -%>
+      <div class="flash flash--<%= key %>"><%= value %></div>
+    <% end -%>
+  </div>
+<% end %>
+
+<div class="custom-administrate-css-hidden">
+  If you can read this, something's wrong.
+</div>
--- a/spec/features/custom_styles_spec.rb
+++ b/spec/features/custom_styles_spec.rb
@@ -1,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe "Custom style" do
+  it "is applied", :js do
+    visit admin_customers_path
+    expect(page).to_not have_css(".custom-administrate-css-hidden")
+    page.execute_script("$('.custom-administrate-css-hidden').show()")
+    expect(page).to have_css(".custom-administrate-css-hidden")
+  end
+end


iangreenleaf:
Big :+1: on this, I think this is probably the best approach for most CSS customization needs. What can we do to get this merged?

app/assets/stylesheets/administrate/application.scss
```diff
@@ -10,3 +10,5 @@
 @import "library/**/*";
 @import "base/**/*";
 @import "components/**/*";
+
+@import "administrate/customize";
```
tysongach: Is this import working? I believe `customize.scss` is a sibling to this manifest file and so the import should be:

```scss
@import "customize";
```

Also, what about naming it `customizations`?

spec/example_app/app/assets/stylesheets/administrate/customize.scss
```diff
@@ -0,0 +1,4 @@
+.custom-administrate-css-hidden {
+  display: none;
+}
+
```
tysongach: Extra line.

keithlayne: Oops.  Seems like Hound should catch that, right?

tysongach: No, I believe the SCSS-Lint library (which Hound uses) doesn’t catch these.

spec/example_app/app/assets/stylesheets/administrate/customize.scss
```diff
@@ -0,0 +1,4 @@
+.custom-administrate-css-hidden {
```
tysongach: The name of this file should begin with an underscore, to align with standards: `_customize.scss`

app/assets/stylesheets/administrate/application.scss
keithlayne: I have no real preferences on naming, are you asking me to change it?

Make that change, and the test fails.  I think the current directory is first in the search path, so that import will just bring in the placeholder.  I want to force a lookup where the app stylesheets path comes before the gem stylesheets path, and therefore finds the override file, but only if it exists.

keithlayne: Another way testing is awkward for this - you can't test both overriding and not overriding easily - it would take two test apps.

iangreenleaf: Yeah, I think the Sass lookup order is something along the lines of:

1. Current dir (i.e. `GEM_PATH/app/assets/stylesheets/administrate`)
2. Application asset path (i.e. `RAILS_ROOT/app/assets/stylesheets`)
3. Gem asset path (i.e. `GEM_PATH/app/assets/stylesheets`)

I tested all the variations manually and can confirm that the way it is now is correct, as it will check first the app assets and then the gem assets (2 & 3) in the proper order.

spec/example_app/app/assets/stylesheets/administrate/customize.scss
iangreenleaf: Just a thought on this: since we're encouraging people to override this file, maybe it's more of a public API vs. a private API, if that's the division we want the naming scheme to signify. In that case no underscore would be best. If the naming scheme is more about top-level vs. included, then an underscore is appropriate.

keithlayne: I thought sass treated underscores as partials.  I'm not sure what that really means in this context, but I usually read it as top-level vs. included.

That said, I've never really noticed any difference in behavior between the two, but I'm sure somebody will correct me on that.

tysongach: Yes, underscores indicate a partial in Sass. By nature of declaring that `customizations` be imported into the main Sass manifest file here in Administrate, it essentially already a partial. The underscore declares to Sass that this file should not be compiled to CSS of its own.

tysongach:
Is the expectation here that you would no longer have to run `administrate:assets:stylesheets`?

I have a concern that this is very fragile and anytime any styles change in Administrate (for example, we change a class name) the custom CSS overrides that a person place in this file would fall out of sync. It would become a constant game of cat-and-mouse. It a tough nut to crack and one that really boils down to how CSS/Sass works, and that everything shares a global namespace.

I worry that this PR only solves a very short term need, but that it would cause larger pain points in the future.

app/assets/stylesheets/administrate/application.scss
tysongach: Interesting. I think this is surfacing a larger issue, which is that Administrate comes packaged with `sass-rails`, but the example app is using Sprockets in its main stylesheets manifest. Sprockets and `sass-rails` handle imports differently.

So, if we are letting Sass (`sass-rails`) handle the importing, the import statement I posted above should work. I pulled down this PR, and when I run the app locally with the current import statement, nothing I put in `_customizations.scss` is compiled into the main CSS file. Changing it to `@import "customizations";` does work.

keithlayne:
@tysongach I'd make that same argument about running the generator.  It's kind of a problem across the board with the approach of this project - any upgrade to the gem means you have to know which generators you've run, and then re-run them, examine the diffs, and deal with any conflicts.

The advantage here is a major reduction in churn, especially if you want to make only minor modifications (as in #349, which motivated this PR).  Modifications are isolated to this file and anything it imports.

I don't think overriding CSS is likely to be a short-term need.  Are you thinking that the CSS will stabilize to the point where running the generator would (likely) be a one-time deal?  That is reasonable in principle, but I'd still argue that it's a sledgehammerish approach when you want to tweak a couple of styles.

app/assets/stylesheets/administrate/application.scss
keithlayne: Not relevant, maybe?  The only CSS (other than the `customizations.scss` file) is the application manifest in the dummy app.  It does have that stupid sprockets `require_tree .` line, but I wasn't expecting the custom stuff in the main CSS file - I expect it to go in `administrate/application.css`, which I'm pretty sure it does.  After precompiling locally on my branch:

```
ag -A2 _customization /Users/keithlayne/forks/administrate/spec/example_app/public/assets/
/Users/keithlayne/forks/administrate/spec/example_app/public/assets/application-a07b8e973e3847f031ee869476ba76dedca6cc5f0979125f9f9b6330d60a210e.css
1:/* line 1, /Users/keithlayne/forks/administrate/spec/example_app/app/assets/stylesheets/administrate/_customizations.scss */
2-.custom-administrate-css-hidden {
3-  display: none; }

/Users/keithlayne/forks/administrate/spec/example_app/public/assets/administrate/application-951dc304ceab1b29c0253d06475d35bd25e97881066a6af900117a0241853739.css
1975:/* line 1, /Users/keithlayne/forks/administrate/spec/example_app/app/assets/stylesheets/administrate/_customizations.scss */
1976-.custom-administrate-css-hidden {
1977-  display: none; }
```

That disagrees with your results - could there be some other difference?

iangreenleaf:
> @tysongach I'd make that same argument about running the generator. It's kind of a problem across the board with the approach of this project - any upgrade to the gem means you have to know which generators you've run, and then re-run them, examine the diffs, and deal with any conflicts.

Agreed, the reason I bumped this PR is that I had a couple small tweaks  I want to make and I very much didn't want to generate all the stylesheets. If I did that, upgrading becomes much worse, because it involves reverse-engineering what changes I've made to the base stylesheets,  saving that somewhere, re-generating, re-applying my changes, resolving conflicts, and testing. If it's in a separate file, all I have to do is see if my stuff is still working, and if not, figure out what changed and update it.

Some people will still want to generate the stylesheets, and for some uses of this gem (like heavily customized interfaces) I'm sure that will be the most sensible way to go. But for those of us who want to do a couple small things  and not have future upgrades be an ordeal, this option seems much, much better.

tysongach:
@keithlayne Yep, I totally agree that you can make the same argument about the generator.

> Are you thinking that the CSS will stabilize to the point where running the generator would (likely) be a one-time deal?

Yes and no. As of now, Administrate is still very young and being that we haven’t even hit a 1.0 release, I think we should expect substantial changes to come in rapid iterations.

mcmire:
I'd like to vote that this be included. I think this is a better approach than `Administrate::Engine.add_stylesheet` as it means that you can use Bourbon variables in any CSS that you choose to add and would effectively solve #429 very nicely. I would also request that this  PR be extended by adding support for overriding JavaScript as well.

> I have a concern that this is very fragile and anytime any styles change in Administrate (for example, we change a class name) the custom CSS overrides that a person place in this file would fall out of sync. It would become a constant game of cat-and-mouse.

This is true, but considering that Administrate is pre-1.0, I would hope that anyone choosing to upgrade the gem in their app (especially if they have not done so in a while) should already expect things to break across the board.

mcmire:
Pinging @nickcharlton on this to see if he has any thoughts?

purinkle:
@keithlayne, is this a change you would still like to make?

keithlayne:
@purinkle Thanks for asking, but this hasn't been relevant for me for a long time...maybe @mcmire or @iangreenleaf would like to weigh in?  I haven't kept up with this project, so I have no clue if it's even relevant at this point.

iangreenleaf:
I imagine this is still an important need, but I gave up on administrate long ago.

nickcharlton:
As it's been quite a while, I'm going to close this for now. But if someone would like to pick up a similar feature in the future, it'd be welcome.

PR #449, opened 2016-02-01
Name collission with Pundit.permitted_attributes fixed

nando:
Problem
The Pundit class has a method called _permitted_attributes_ which collisions with ours.

Solution
To change the name of our method to _dashboard_attributes_.


nickcharlton:
I'm going to close this, as I recently merged #465 which should solve the same problem. Thanks for this PR!

Issue #431, opened 2016-01-22
uninitialized constant Administrate::Generators::InstallGenerator::DashboardManifest


frabrunelle:
Hi :smiley: 

When I run the installer for Administrate, I get the following error:

```
rails generate administrate:install
Running via Spring preloader in process 33128
      create  app/controllers/admin/application_controller.rb
       route  namespace :admin do
    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
      resources dashboard_resource
    end

    root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
  end
      create  app/dashboards/dashboard_manifest.rb
/usr/local/lib/ruby/gems/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:44:in `manifest': uninitialized constant Administrate::Generators::InstallGenerator::DashboardManifest (NameError)
	from /usr/local/lib/ruby/gems/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:36:in `dashboard_resources'
	from /usr/local/lib/ruby/gems/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:32:in `singular_dashboard_resources'
	from /usr/local/lib/ruby/gems/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:24:in `run_dashboard_generators'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `block in invoke_all'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `each'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `map'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `invoke_all'
	from /usr/local/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/group.rb:232:in `dispatch'
[...]

PR #462, opened 2016-02-12
Suggest spring disabling in README

xijo:
This is a small PR as followup on #431 - should save some folks some trouble :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ gem "administrate", "~> 0.1.3"
 Re-bundle, then run the installer:
 
 ```bash
-$ rails generate administrate:install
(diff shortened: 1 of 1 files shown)


graysonwright:
It looks like #464 solves the issue with spring. If we get that merged in, this should be unnecessary. Mind if we hold off?

xijo:
Sure, that'd be even better anyway.

Grayson Wright <notifications@github.com> schrieb am So., 14. Feb. 2016
01:42:

nickcharlton:
Closing because of #464. Please open again if you think it'd still be a good idea!

Issue #453, opened 2016-02-05
InstallGenerator::DashboardManifest (NameError)


Qcoin:
Hello everyone, 
I have a problem with administrate, if I install i have this:
```
boboss@boboss:~/projet/my_app$ rails generate administrate:install
Running via Spring preloader in process 9965
      create  app/controllers/admin/application_controller.rb
       route  namespace :admin do
    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
      resources dashboard_resource
    end

    root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
  end
      create  app/dashboards/dashboard_manifest.rb
/home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:44:in `manifest': uninitialized constant Administrate::Generators::InstallGenerator::DashboardManifest (NameError)
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:36:in `dashboard_resources'
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:32:in `singular_dashboard_resources'
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:24:in `run_dashboard_generators'
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `block in invoke_all'
	from /home/boboss/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `each'
[...]

Issue #448, opened 2016-01-31
0.1.3 - install generator fails uninitialized constant Administrate::Generators::InstallGenerator::DashboardManifest


yourabi:
rails generate administrate:install

```

I, [2016-01-30T17:45:36.035990 #21193]  INFO -- : ** [Raven] Raven 0.12.2 ready to catch errors
      create  app/controllers/admin/application_controller.rb
       route  namespace :admin do
    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
      resources dashboard_resource
    end

    root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
  end
      create  app/dashboards/dashboard_manifest.rb
/home/vagrant/.gem/ruby/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:44:in `manifest': uninitialized constant Administrate::Generators::InstallGenerator::DashboardManifest (NameError)
	from /home/vagrant/.gem/ruby/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:36:in `dashboard_resources'
	from /home/vagrant/.gem/ruby/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:32:in `singular_dashboard_resources'
	from /home/vagrant/.gem/ruby/2.3.0/gems/administrate-0.1.3/lib/generators/administrate/install/install_generator.rb:24:in `run_dashboard_generators'
	from /home/vagrant/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
	from /home/vagrant/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
	from /home/vagrant/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `block in invoke_all'
	from /home/vagrant/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `each'
	from /home/vagrant/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `map'
	from /home/vagrant/.gem/ruby/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:133:in `invoke_all'
[...]

PR #464, opened 2016-02-13
Replace DashboardManifest with explicit routes

graysonwright:
## Problem:

When users run the `administrate:install` generator multiple times, the generated routes from the first install prevent following installs from working correctly.

In order to get subsequent installs to run correctly, users must delete the already-generated routes and run the installer again from scratch.

This has thrown many people off, and there is discussion of the issue in:

- #431
- #448
- #453

## Solution:

We should change the generated routes so they don't rely on a `DashboardManifest` being defined.

I'm not sure what the best way to do that would be, but here are a couple options

### Defining routes directly

This option would look something like:

```ruby
namespace :admin do
  resources :customers
  resources :orders
  resources :products
  resources :line_items

  root to: "customers#index"
end
```

That would fix the error that people are getting when they run `rails generate administrate:install` multiple times, but it would be more work to update the routes if they change over time.

If we went this route, I'd want to look into getting rid of the `DashboardManifest` completely.  It would be cumbersome to update the list of dashboards in multiple locations, and we could potentially look at the routes that the user's defined to determine which sidebar links to display.

### Wrap routes in a conditional

Alternatively we could keep it how it is, but wrap the block in a conditional - that might look like:

```ruby
if defined?(DashboardManifest)
  namespace :admin do
    DashboardManifest::DASHBOARDS.each do |dashboard_resource|
      resources dashboard_resource
    end

    root controller: DashboardManifest::ROOT_DASHBOARD, action: :index
  end
end
```

### Define routes through the engine

This is the approach that RailsAdmin and Active Admin take.  It would look like:

```ruby
  mount :administrate
end
```

This approach is rather opaque, and it makes it harder for the developer to see what's going on.  Behind the scenes we could use any approach we wanted to make sure the `DashboardManifest` is defined when we need it.

## Thoughts?

What's the best approach to take here?
- Are we comfortable mounting an engine and abstracting away the complexity?
- Is it worth it to introduce a conditional check into peoples' routes file?
- Would we prefer to get rid of the `DashboardManifest` altogether, and use the user's routes to fill that role instead?


coreyward:
To be clear, this happened to me on the very first run of the generator.

sgonyea:
Likewise, I just tried installing and it blew up with this error on the first run.

(Thanks for open sourcing! :smile:)

sgonyea:
Given how painful the magic ends up being in Active Admin, I agree that it's better to just define each admin route like you define your app's routes.  The problem with most magic in the routes file is that it basically eager loads the world every time you run a basic command.

Rather than using the `resources` method, perhaps just define a `dashboard_resources` helper?  This would let you internalize definitions for use on the sidebar, while keeping a minimal footprint.

```ruby
  namespace :admin do
    dashboard_resources :orders
  end
end
```

graysonwright:
@coreyward reading back through your comments, that makes sense. I was able to reproduce it locally as well on my first run.

I believe any of the proposed solutions will resolve the first-run issue as well as the issues on subsequent runs.

I'm going to take a crack at implementing this in the routes file, and see where that gets us. Stay tuned!

graysonwright:
Alright, I restructured Administrate to get rid of the `DashboardManifest` and use the routes in its place. Need to experiment a little bit, but it seems like this approach will be simpler and more robust.

danbee:
I definitely like this solution. It keeps the route files cleaner and means one less configuration file.

graysonwright:
Sounds good, merging.

PR #465, opened 2016-02-14
Inline controller's `permitted_attributes` method

### LINKED ISSUE #417: 'permitted_attributes' method collides with pundit
if I include Pundit on my Administrate controller, it raises ArgumentError on create / update.

```
  include Pundit
  # ...
end
```

The problem is calling `permitted_attributes` actually calls the method in Pundit.
https://github.com/elabs/pundit/blob/master/lib/pundit.rb#L248
What is a better way to avoid this collision rather than monkey-patching on either of gems?

graysonwright:
Fixes #417
## Problem:

If users include the Pundit gem on an Administrate controller, the controller attempts to call the `permitted_attributes` method defined by pundit instedad of the one defined by Administrate.

Because these methods take different arguments, this results in an `ArgumentError` for the user.
## Solution:

Inline the `permitted_attributes` method to avoid the method name collision.

This narrows the public API a bit, but that's alright.  The user can accomplish all of the removed functionality by overriding the parent `resource_params` method.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -98,11 +98,7 @@ module Administrate
     end
 
     def resource_params
-      params.require(resource_name).permit(*permitted_attributes)
-    end
-
-    def permitted_attributes
-      dashboard.permitted_attributes
+      params.require(resource_name).permit(dashboard.permitted_attributes)
     end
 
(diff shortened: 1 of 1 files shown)


carlosramireziii:
@nickcharlton Take a look at this one. Seems ready to go

nickcharlton:
Agreed; going to merge this.

PR #466, opened 2016-02-14
Allow preloading has_many records

gabebw:
Let's say an `Author` has_many `Book`s.

`Field::HasMany` loads each Book record and calls `display_resource` on each one. If the `Book` dashboard defines `display_resource` to call `book.author.name`, N `Author` records will be loaded, one for each book.

This can significantly slow down load time.

To fix this, Administrate users can now tell Administrate to preload associated records using Rails' `includes` option. Example usage:

```
Field::HasMany.with_options(includes: [:author])
```

### DIFF
--- a/lib/administrate/field/has_many.rb
+++ b/lib/administrate/field/has_many.rb
@@ -47,7 +47,12 @@ module Administrate
       private
 
       def candidate_resources
-        associated_class.all
+        if options.key?(:includes)
+          includes = options.fetch(:includes)
+          associated_class.includes(*includes).all
+        else
+          associated_class.all
+        end
       end
 
       def display_candidate_resource(resource)


gabebw:
There are no tests because I couldn't figure out where tests might go. If you point me at the right place I'll happily add tests.

graysonwright:
ehhhh..... specs are tricky.

If you know how to add tests for this, put 'em in `spec/lib/fields/has_many_spec.rb`.
I guess I'm not sure what a test for this would look like.

gabebw:
> I guess I'm not sure what a test for this would look like.

Right -- the only thing I can think of is testing the SUT (which, :cry:) or testing some horrible internal ActiveRecord code, neither of which sounds maintainable.

nickcharlton:
I have no idea how to test this. So I'm going to merge it anyway and hope I don't blow everything up. Thanks Gabe!

PR #474, opened 2016-02-17
Orm adapters

mrbrdo:
#143 This works well for us with the Sequel adapter.

There is only one concern I still have. Right now I wrap models/records/relations into wrapper objects as soon as possible. This unfortunately means that if the user overrides something (e.g. a view), he will get wrapper objects instead of actual AR records. The wrapper does proxy all methods to the actual record, and it defines to_model so polymorphic_url works properly, but I'm not sure if this is the best solution.

I see 2 possible alternatives:
- include a module into the base classes that adds a method such as `#agnostic`, which returns the wrapper. then go through this method inside administrate
- use class methods ("helpers") instead of wrappers, and use that in administrate instead of calling methods on the records/models/relations

The spec failures are due to equality checks (wrapper object with wrapped object).

Or keep the current approach if it's good enough. What do you think?

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -4,7 +4,7 @@ module Administrate
       search_term = params[:search].to_s.strip
       resources = Administrate::Search.new(resource_resolver, search_term).run
       resources = order.apply(resources)
-      resources = resources.page(params[:page]).per(records_per_page)
+      resources = resources.paginate(params[:page], records_per_page)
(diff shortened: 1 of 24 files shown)


pablobm:
Thank you for this PR @mrbrdo. Apologies for not coming back to you sooner.

Unfortunately, I'm closing this one for now. ORM abstraction is desirable, but we think we should first reach v1.0 (which admittedly is a bit of a hazy target right now) before we implement it. This is so that we make sure that we are working from a stable API.

PR #476, opened 2016-02-21
Extract image field into a separate gem

graysonwright:
## Feature:

**When I** have a custom field type that I want to share with others,
**I want to** reference an official template for a field gem
**so I can** more easily package my field for sharing.

## Implementation:

Create a template for field plugins
by extracting the image field into its own gem.

This is a breaking change with an easy fix -
anybody who uses Image fields
will need to add `gem "administrate-field-image"` to their `Gemfile`.

The source for the image field gem is up at
https://github.com/graysonwright/administrate-field-image.


graysonwright:
@danbee, @mcmire - thoughts?

graysonwright:
@jferris mind if I get your eyes? I'm having some trouble figuring out the right strategy to publish this feature.

This repository contains both the `administrate` gem and [the demo application][demo]. The demo application uses images, which means that its `Gemfile` will be dependent on the new `administrate-field-image` gem.

In turn, the [`administrate-field-image` gem][image] is dependent on the `administrate` gem in this repository. This creates a little bit of a chicken-and-egg problem for publishing both of those gems.

Here's the best plan I can come up with:

- [ ] Release this branch as a new version of the `administrate` gem, as version `0.2.0.rc1`
- [ ] Make administrate-field-image dependent on `"administrate", "0.2.0.rc1", "~> 0.2.0"`
- [ ] Release `administrate-field-image` v `0.0.1`
- [ ] Make the demo app's `Gemfile` dependent on `"administrate-field-image", "~> 0.0"`
- [ ] Re-push this branch to Github to run CircleCI suite. All of the gems would be available on rubygems.org, so the suite should pass.
- [ ] Update and release this branch as version `0.2.0`
- [ ] Merge branch

What do you think? Any changes that you'd make?

I'm also not sure what would happen if we set `administrate-field-image` dependent on `"administrate", "~> 0.2.0.rc1"`. Do you know how the `~>` operator treats release candidate versions?

Thanks for your help!

[demo]: http://administrate-prototype.herokuapp.com/admin
[image]: https://github.com/graysonwright/administrate-field-image

graysonwright:
From http://guides.rubygems.org/patterns/#pessimistic-version-constraint: 
> Using ~> with prerelease versions will restrict you to prerelease versions only.

PR #482, opened 2016-02-22
Wrap table cell contents instead of CSS truncation

iangreenleaf:
I'm not sure what discussion led to this being added since the ticket referenced in ae61171e isn't public. However, I don't like the existing solution because:
- It [doesn't work](https://administrate-prototype.herokuapp.com/admin/customers?direction=desc&order=updated_at)
- The `Field::Text` type already offers truncation with ellipsis, so it duplicates logic to also perform that here. If someone sets a truncation value in their dashboard, they shouldn't have to fiddle with the CSS to have their chosen length respected.

This changes table cells to wrap normally, and to force line breaks on super long strings that don't have a natural break.

### DIFF
--- a/app/assets/stylesheets/administrate/base/_tables.scss
+++ b/app/assets/stylesheets/administrate/base/_tables.scss
@@ -20,6 +20,4 @@ td {
   font-kerning: normal;
   font-variant-ligatures: common-ligatures, contextual;
   font-variant-numeric: lining-nums, tabular-nums;
-  white-space: normal;
-  word-wrap: break-word;
 }
--- a/app/assets/stylesheets/administrate/base/_tables.scss
+++ b/app/assets/stylesheets/administrate/base/_tables.scss
@@ -20,8 +20,6 @@ td {
   font-kerning: normal;
   font-variant-ligatures: common-ligatures, contextual;
   font-variant-numeric: lining-nums, tabular-nums;
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
-  word-wrap: normal;
+  white-space: normal;
+  word-wrap: break-word;
 }


tysongach:
@iangreenleaf Can you provide a screenshot where it’s not working and indicate what browser you’re using? I’m not sure I’m seeing the issue in my environment.

Also, can you provide a screenshot of the outcome of this change?

Finally, a couple notes on the CSS:
- We shouldn’t need to set `white-space` to `normal`, as `normal` is its initial value
- Per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/word-wrap), the `word-wrap` property has been renamed `overflow-wrap`, so we need to take that into consideration. We could use [Bourbon’s `word-wrap` mixin](http://bourbon.io/docs/#word-wrap), which takes care of this for us.

iangreenleaf:
@tysongach funny, someone edited my entry in the demo app, probably because it was breaking the tables an annoying them. It's hard to show the problem on the demo app because the text fields are being truncated by Rails, but here are some screenshots at just the right width, before and after wrapping:

screenshot_2016-02-22_20-58-47

screenshot_2016-02-22_20-59-09

While testing this, I also discovered that the `word-wrap` property won't actually help break up long strings, because table layout is super weird. So I'll push an update that just removes the three lines, which will allow normal text with whitespace to wrap, at least. People with long unbroken strings are still on their own.

nickcharlton:
Ah hah! I couldn't work out why @tysongach couldn't see this. We truncate text fields by default to `50` characters, so unless you override that you probably won't ever see it.

I can confirm that if you push it up to something along the lines of `150` that it breaks the layout. As it's just removing some CSS, I'm going to merge this as it seems to work well from my testing.

Issue #451, opened 2016-02-03
Custom labels for fields


Xosmond:
Hi, i have some models with short name fields and i want to change the label of some with other more verbose, how can i change the labels of some string fields? Thanks all.

eimhin:
+1

atwoodjw:
You can do this via localization.

config/locales/en.yml
```
en:
  helpers:
    label:
      admin_user:
        password_confirmation: "Confirmation"
```

Xosmond:
I have tried but does not work :confused:. Do you have any project where you have did it to share?

graysonwright:
@Xosmond - I just added a test to verify that @atwoodjw's approach works: https://github.com/thoughtbot/administrate/pull/492

Can you post the code you're using that isn't working?

Xosmond:
Sorry for the late. This solution works only on normal attributes but on relations doesn't work:
en.yml
```
  helpers:
    label:
      control1:
        registro1s: "Registros"
        nombre_documento: "titulo"
```

Here, the "nombre_documento" is a string  attribute of control1 and It works. But "registro1s" is a relation, because control1 has many registro1s and it doesn't work.

Any solution for this?

KatherineMuedas:
I am trying this but it does not work. Do I need to create an administrate.en.yml file inside locales or can I just add it to my en.yml file ?

Xosmond:
Just in the en.yml , make sure you are not using i18n on your config/application.rb

KatherineMuedas:
Ok, In my en.yml file I added this. (school is my model and tuition fee is the attribute )
```
en:
  helpers:
    label:
      school:
        tuition_fee: "Tuition fee average"
```

Is that all I need to make it work ? I am trying to customize the label of the tuition fee attribute.

Xosmond:
Have you tried with only "Field::Number" ? maybe the kind of attribute is the key.

KatherineMuedas:
It is only working on the form page but it is not working for the show page or index page.

Xosmond:
it could be, because the way this is working. I hope they can make a better option to customize labels.

graysonwright:
@Xosmond, @KatherineMuedas - ah, I didn't realize that you were talking about the index and show pages.

I've updated #492 to work on all three types of pages. I'll try to get that merged in as soon as possible.

KatherineMuedas:
@graysonwright that would be great !! Thanks :)

graysonwright:
[...]

PR #492, opened 2016-02-29
Use I18n translations for custom labels on dashboards

graysonwright:
## Problem:

There was uncertainty about how to create custom labels for fields. In https://github.com/thoughtbot/administrate/issues/451, it was suggested that we could use I18n translations to solve the problem.

If we accept that as the standard way to set custom labels, we should have a test for it.

## Solution:

- Add a test to verify that labels can be customized through I18n.
- Add documentation about how to customize labels

## To Do:

- [x] Update view code so the same translation definition applies on the show and collection pages, in addition to the form page.

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -22,13 +22,17 @@ to display a collection of resources in an HTML table.
   <thead>
     <tr>
       <% collection_presenter.attribute_types.each do |attr_name, attr_type| %>
-        <th class="cell-label cell-label--<%= attr_type.html_class %>
-          cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
+        <th class="cell-label
+        cell-label--<%= attr_type.html_class %>
+        cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
         " scope="col">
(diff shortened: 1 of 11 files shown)


app/views/administrate/application/_collection.html.erb
```diff
@@ -22,13 +22,17 @@ to display a collection of resources in an HTML table.
   <thead>
     <tr>
       <% collection_presenter.attribute_types.each do |attr_name, attr_type| %>
-        <th class="cell-label cell-label--<%= attr_type.html_class %>
-          cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
+        <th class="cell-label
+        cell-label--<%= attr_type.html_class %>
+        cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
```
graysonwright: Unrelated change: improve spacing of a super-long HTML tag

graysonwright:
@mcmire, can I get a review? Thanks!

kuboon:
Administrate already use ``Model.human`` to translate model name.
How about using ``Model.human_attribute_name`` for attributes?
http://apidock.com/rails/ActiveModel/Translation/human_attribute_name

Issue #484, opened 2016-02-24
rails g administrative:views inflects differently than specific adminstrate generator


armahillo:
```
ruby-2.0.0-p645 armahillo@webdev:~/myapp$ rails g administrate:views:index
      create  app/views/admin/application/index.html.erb
      create  app/views/admin/application/_collection.html.erb
ruby-2.0.0-p645 armahillo@webdev:~/myapp$ rails g administrate:views:show
      create  app/views/admin/application/show.html.erb
ruby-2.0.0-p645 armahillo@webdev:~/myapp$ rails g administrate:views:edit
      create  app/views/admin/application/edit.html.erb
      create  app/views/admin/application/_form.html.erb
ruby-2.0.0-p645 armahillo@webdev:~/myapp$ rails g administrate:views:new
      create  app/views/admin/application/new.html.erb
   identical  app/views/admin/application/_form.html.erb
ruby-2.0.0-p645 armahillo@webdev:~/myapp$ rails g administrate:views
      create  app/views/admin/applications/index.html.erb
      create  app/views/admin/applications/_collection.html.erb
      create  app/views/admin/applications/show.html.erb
      create  app/views/admin/applications/new.html.erb
      create  app/views/admin/applications/_form.html.erb
      create  app/views/admin/applications/edit.html.erb
   identical  app/views/admin/applications/_form.html.erb
```

It's unclear if this is intentional or not, but if you specify the action on `rails g administrate` it uses the "application" subdirectory, whereas if you do not specify an action it pluralizes to "applications". 

If this is intentional, the documentation on http://administrate-prototype.herokuapp.com/customizing_page_views did not reflect this discrepancy.

dacur:
I had the same issue today.  I just manually renamed the `applications` folder to `application`.

carlosramireziii:
Anyone want to take a stab at a failing spec and a fix on this one?

punitcse:
@carlosramireziii I checked on my system. It seems that this issue has been fixed. Could you please close this issue?

Thanks

carlosramireziii:
Thanks for looking into this @punitcse. Closing!

PR #499, opened 2016-03-04
[#484] Fix views generator should pass args to sub-generator

tanapoln:
I think that administrate:views generator is generating files into incorrect directory /app/views/applications which should be /app/views/application instead (without s).


nickcharlton:
Thanks for your PR! I _believe_ that this came up in #645 and then was solved in #747, so I'm going to close this PR. But if that's not the case, please reopen.

Issue #505, opened 2016-03-07
Routes generator produces ill-indented code


fwolfst:
Currently (0.1.4) administrate creates routes like the following (if more than one resource is found):

```
  namespace :admin do
    resources :departments
resources :services
resources :service_types
resources :usages

    root to: "departments#index"
  end
```

I expect it to create well indented code like

```
  namespace :admin do
    resources :departments
    resources :services
    resources :service_types
    resources :usages

    root to: "departments#index"
  end
```

carlosramireziii:
This still seems to be an issue in `0.4.0`. Anyone want to take a stab at it?

fwolfst:
I am not sure anymore, but my changes might just have fixed the issue (dont know why I didnt file a PR though).

nickcharlton:
I think this was fixed in #506, so I'm going to close this. If it's not, please reopen!

PR #506, opened 2016-03-07
let routes generator indent all routes (#505).

fwolfst:
Routes generated before:

```
  namespace :admin do
    resources :departments
resources :services
resources :service_types
resources :usages

    root to: "departments#index"
  end
```

and after:

```
  namespace :admin do
    resources :departments
    resources :services
    resources :service_types
    resources :usages

    root to: "departments#index"
  end
```

Unfortunately, no spec for that (spec just checks occurence of routes).


nickcharlton:
Looks good! Merging.

Issue #164, opened 2015-11-06
undefined method `new' for nil:NilClass on custom attribute


estiens:
Every time I try to add a custom attribute to the show page, I get the following error from this code

    def attribute_field(dashboard, resource, attribute_name, page)
        value = get_attribute_value(resource, attribute_name)

        dashboard.
          attribute_types[attribute_name].
          new(attribute_name, value, page)
    end

I used the gravatar example as a base.

I am trying to display a standard paperclip `model.attachment.url(:thumb)`

graysonwright:
@estiens - it looks like you might need to add your attribute to the dashboard, under the `RESOURCE_TYPES` constant. I just realized that that part is missing from [the docs](https://administrate-docs.herokuapp.com/adding_custom_field_types).

It should look something like this - I'm assuming you're trying to add the gravatar to the `User` model's `email` field, and that your generated field class is called `GravatarField`.

```ruby
class UserDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    created_at: Field::DateTime,
    updated_at: Field::DateTime,
    name: Field::String,
    email: GravatarField,    # Update this email to use your new field class
  }
end
```

Let me know if that works!

estiens:
This works well, thanks! Maybe make a note to add to docs.

PR #508, opened 2016-03-07
Add example of how to use a custom field

graysonwright:
## Problem:

The documentation for creating custom field types does not tell the user how to use the custom field.

This gap in documentation was brought up by @estiens in https://github.com/thoughtbot/administrate/issues/164.
## Solution:

Add a note about how to use custom fields.


nickcharlton:
We'd already merged something recently like this, but I've rebased and am going to pull this in (over the top of it) as I feel it's actually clearer. Thanks!

Issue #481, opened 2016-02-22
Custom controller is assumed to have a related model


naps62:
I want to have an admin dashboard where I show some useful data. This is not necessarily tied to a single resource model

My first attempt was to create a `Admin::OverviewController` with an `index` action.
However, the sidebar calls `Administrate::Namespace.new(namespace).resources`, which, I eventually discovered, is computed by listing all controllers for which administrate has routes. So it would end up looking for an Overview model

I generated the sidebar template, and monkey patched it with this:

`(Administrate::Namespace.new(namespace).resources - [:overview]).each`
to explicitly remove the overview from the list of resources. This is ugly as hell though

Is there currently a better solution for this?

sebbean:
I'm having this issue as well. commented on the commit: https://github.com/thoughtbot/administrate/commit/765d3ff0b1377fb6465df0251049b4ffff0d2965

graysonwright:
@naps62, you're right - there should be a better way to define custom routes.

Would it make sense to add a helper method to the routes file? Something like `administrate_resources :customers`? Administrate could look for only those routes, and ignore any normal routes. It would be backwards-incompatible, though.

Any suggestions welcome.

naps62:
@graysonwright maybe I misunderstood your sugestion, but I actually think the routes are perfect as they are right now, and they're not the issue here

The thing that caught me off guard was that `administrate` assumes that, for every controller under the `Admin` namespace, I'll have a corresponding ActiveRecord model

I tried several ways to add an `Admin::OverviewController`. My original goal was to have an extra "Overview" link in the navbar, which would show me a page (using the same administrate layout), where I could show whatever I want
But `Administrate::Namespace` looks for everything under `admin`, assumes I have a model for it, a dashboard for it, and then other stuff assumes the model is an ActiveRecord one.

So, the way I see it, a possible solution would be to make `Namespace` aware of what resources I actually want him to use.
This information is already present in `DashboardManifest::DASHBOARDS`, although I'm not sure about using that, since that belongs to the app, and is not in the gem itself. But I feel that `Namespace` is trying to compute information that's already present in that constant, but it ends up finding too much

graysonwright:
Hm... I'm not sure if I understand.

I opened up https://github.com/thoughtbot/administrate/pull/514, with a potential solution to your problem. Can you tell me if that would work for you, or if you had something else in mind?

naps62:
[...]

PR #511, opened 2016-03-08
Only list resources if they're in the manifest

jayroh:
If the dashboard manifest lists out the resources `administrate` is
working with, then the links in the sidebar should only link to those
resources.

As of this commit `Administrate::Namespace` was inferring which resources
to list based on what routes had `"#{namespace}/"` in the path. If you add
a route in the namespace for a resource you do NOT want to show up, it
is possible that it would break.

### DIFF
--- a/lib/administrate/namespace.rb
+++ b/lib/administrate/namespace.rb
@@ -5,7 +5,7 @@ module Administrate
     end
 
     def resources
-      namespace_controller_paths.uniq.map do |controller|
+      whitelisted_controller_paths.map do |controller|
         controller.gsub(/^#{namespace}\//, "").to_sym
       end
     end
@@ -13,6 +13,16 @@ module Administrate
     private
 
     attr_reader :namespace
+
+    def whitelisted_controller_paths
+      if dashboard_manifest_missing?
+        return namespace_controller_paths.uniq
+      end
+
+      namespace_controller_paths.uniq.select do |path|
+        dashboard_paths.include? path
+      end
+    end
 
     def namespace_controller_paths
       all_controller_paths.select do |controller|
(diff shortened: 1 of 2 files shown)


lib/administrate/namespace.rb
```diff
@@ -14,6 +14,16 @@ def resources
...
 
+    def whitelisted_controller_paths
+      if dashboard_manifest_missing?
+        return namespace_controller_paths.uniq
+      end
+
+      namespace_controller_paths.uniq.select do |path|
+        dashboard_paths.include? path
+      end
+    end
```
jayroh: I could have gone either way here - guard clause or if/else. If everyone prefers the more explicit if/else...? Then I'm good with that. Just let me know.

graysonwright:
Looks like this is the same issue as #481.

I'd like to find a solution that doesn't rely on the DashboardManifest, if possible. The manifest is only listing out resources at the moment, and I don't think we need an entire class for that.

What do you think about adding a helper method or some identifying tag in the routes file to differentiate administrate routes from normal routes?

jayroh:
I hate to be pedantic here, but, this doesn't entirely rely on the dashboard manifest. It does, however, adhere to it if it exists.

Can you help me unpack this a little bit more? ... bc I don't think I'm following which class you're referring to :(

> The manifest is only listing out resources at the moment, and I don't think we need an entire class for that.

I agree that this is probably not the most robust solution, though. There's some pretty gnarly coupling to the naming and inferences of resource names, etc.

graysonwright:
@jayroh in #464, we removed the dependency on the DashboardManifest. I suppose I should have deprecated it in there - we should still do that.

I was thinking that we'd phase out the DashboardManfiest, and rely only on routes or other configuration. I suppose this is dependent on the routes file being flexible enough to define a dashboard in.

I opened up #514 as an alternative - would that solve the problem? What are the tradeoffs between these approaches?

jayroh:
Oh snap - I completely missed #464! Yeah, I think a deprecation warning would be very helpful. I wouldn't have known otherwise.

Will move over to #514 to give it a look. Thank you, sir :)

jayroh:
Closing this PR too, btw. It is moot.

PR #512, opened 2016-03-08
Upgrade `rails` to `4.2.5.2`

teoljungberg:
To reflect CVE:

```
Name: actionpack
Version: 4.2.5.1
Advisory: CVE-2016-2098
Criticality: Unknown
URL:
https://groups.google.com/forum/#!topic/rubyonrails-security/ly-IH-fxr_Q
Title: Possible remote code execution vulnerability in Action Pack
Solution: upgrade to ~> 3.2.22.2, ~> 4.1.14.2, ~> 4.2.5.2
```

PR #513, opened 2016-03-08
Freeze generated constants

teoljungberg:
By abiding to the thoughtbot styleguide,
We won't get any warnings from tools like Hound,
When opening a PR adding administrate to a project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
   ```
 * [#458] [BUGFIX] Update the custom field generator to match the new HTML
   structure of forms
+* [#513] [OPTIM] Freeze constants in generated dashboard classes
 
 ### 0.1.4 (February 20, 2016)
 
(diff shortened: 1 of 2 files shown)


teoljungberg:
The test failures are fixed in #512

teoljungberg:
What's your take on this @graysonwright ?

BenMorganIO:
:+1:

graysonwright:
Alright, looks good to me.

jondkinney:
This seems good on the surface. However, this removes the ability for us to add queries to the ATTRIBUTE_TYPES hash. Well, you can still add them, but only the objects present in the DB at the time of the app boot will show up, for example, in a select list. If I delete the `.freeze` then creating a new widget that I want to select on a different dashboard edit will have the full list. I realize this is "just how ruby/freeze works". Do you all have opinions on putting queries in the dashboard vs the generated form partial?

image

jondkinney:
I updated my approach here and `.freeze` is no longer a problem. Instead of storing a query directly (which only gets executed on rails boot), I'm now instead storing a reference to a proc object that I later `.call` in the `_form.html.erb` template. I even wrapped it up in a gem called [administrate-field-collection_select](https://github.com/headwayio/administrate-field-collection_select). Hope it helps someone.

PR #514, opened 2016-03-08
Ignore routes marked `administrate: false`

graysonwright:
Possibly fixes #511
## Problem:

Users often define routes in the `admin` namespace
that they don't want to show up in the sidebar.
At the moment, Administrate creates a sidebar link
for all routes in the `admin` namepsace,
without a way to ignore them.
## Solution:

Add an `administrate: false` option for dashboard routes
that will prevent Administrate
from creating a sidebar link for the resource.

### DIFF
--- a/lib/administrate/namespace.rb
+++ b/lib/administrate/namespace.rb
@@ -21,7 +21,13 @@ module Administrate
     end
 
     def all_controller_paths
-      Rails.application.routes.routes.map do |route|
+      Rails.application.routes.routes.select do |route|
+        if route.defaults[:administrate].nil?
+          true
+        else
+          route.defaults[:administrate]
+        end
+      end.map do |route|
         route.defaults[:controller].to_s
       end
     end
--- a/spec/administrate/namespace_spec.rb
+++ b/spec/administrate/namespace_spec.rb
@@ -5,12 +5,30 @@ describe Administrate::Namespace do
   describe "#resources" do
     it "searches the routes for resources in the namespace" do
       begin
-        namespace = Administrate::Namespace.new(:admin)
         Rails.application.routes.draw do
           namespace(:admin) { resources :customers }
         end
 
+        namespace = Administrate::Namespace.new(:admin)
+
         expect(namespace.resources).to eq [:customers]
+      ensure
+        reset_routes
+      end
+    end
+
+    it "leaves out resources marked administrate: false" do
+      begin
+        Rails.application.routes.draw do
+          namespace(:admin) do
+            resources :customers, administrate: false
+            resources :products, administrate: true
+          end
+        end
+
+        namespace = Administrate::Namespace.new(:admin)
+
+        expect(namespace.resources).to eq [:products]
       ensure
         reset_routes
       end
--- a/spec/example_app/config/routes.rb
+++ b/spec/example_app/config/routes.rb
@@ -3,7 +3,7 @@ Rails.application.routes.draw do
     resources :customers
     resources :line_items
     resources :orders
-    resources :products
+    resources :products, administrate: false
 
     root to: "customers#index"
   end
--- a/spec/example_app/config/routes.rb
+++ b/spec/example_app/config/routes.rb
@@ -3,7 +3,7 @@ Rails.application.routes.draw do
     resources :customers
     resources :line_items
     resources :orders
-    resources :products, administrate: false
+    resources :products
 
     root to: "customers#index"
   end


lib/administrate/namespace.rb
```diff
@@ -21,7 +21,13 @@ def namespace_controller_paths
...
 
     def all_controller_paths
-      Rails.application.routes.routes.map do |route|
+      Rails.application.routes.routes.select do |route|
+        if route.defaults[:administrate].nil?
+          true
+        else
+          route.defaults[:administrate]
+        end
+      end.map do |route|
```
jayroh: This certainly does the job but I'm trying to come up with a way to make this a little easier to unpack visually/mentally.

Pull the select out to its own method? Leave the map in here?

jayroh:
Love this solution, dude. Just that one comment above on how to tackle the work that method does.

naps62:
:+1: 
This solves the described problem, and is a big step towards the issue I opened. I'm not sure if it fully solves it though, but whatever is missing would be out of the scope of this PR.

I'll discuss that in the [original issue](https://github.com/thoughtbot/administrate/issues/481#issuecomment-193845274)

lib/administrate/namespace.rb
graysonwright: oh, good call. I wrote this quickly; I'll clean it up.

jayroh:
ping @graysonwright! Hey buddy. :)

Any chance I could get you to rebase this branch on master? I've been pointing to this branch from a project to get the benefits of the included changes and all my builds are failing bc of the rails CVE thing before I update master.

graysonwright:
@jayroh Just fixed the build, let me know if there are still problems.

swmcc:
:+1: this solved an issue for me - would be great if it could be put into master :)

lib/administrate/namespace.rb
swalberg: As a thought:

```
Rails.application.routes.routes.map do |route|
  if route.defaults[:administrate].nil? || route.defaults[:administrate]
    route.defaults[:controller].to_s
  end
end.compact
```

rob-murray:
I also have this problem but have worked around it another way -

I am a little but unsure of `administrate: false` and I would be a little confused with having this in my routes config for something that is view related. If I saw `administrate: false` I might think that the controller and action wont use any administrate code, it would go to a controller that does not inherit from the administrate app controller. It might well do so but also I suspect it in a lot of cases there will be no difference just that its not displayed in the sidebar as the intended, either way I have to check. So then what does `administrate: false` actually mean, it sounds kind of ambiguous to me?

I currently work around this by doing something like this in a helper and then using my own sidebar partial and calling `resources_for_sidebar_nav` in this view. This itself is not perfect as I need to know about the internals of your library and kind of rails route resources but maybe a better solution maybe this library providing a method in a helper to get an ordered list of resources for the sidebar, the user can then override this or use the default, and then the view calls this helper instead of doing the work itself?

```ruby
    def resources_for_sidebar_nav
      Administrate::Namespace.new(namespace).resources.reject { |resource| resources_to_ignore.include? resource }
    end

    def resources_to_ignore
      %i(foo_imports)
    end
```

I'm happy to submit a PR for something like this if you think its worthwhile or would be easier to understand by seeing the code?

graysonwright:
Yeah, I'm not thrilled with this solution. It seems like a hack to me. I wrote the code, but I still don't really want it in the project.

Here's an alternative API that I think is much cleaner and should still be easy to implement:

```ruby
# app/namespaces/admin_namespace.rb

class AdminNamespace < Administrate::Namespace
  dashboards CustomerDashboard, ProductDashboard, OrderDashboard, LineItemDashboard

  root_dashboard CustomerDashboard
end
```

```ruby
# config/routes.rb

  AdminDashboard.mount_routes(namespace: :admin)

  get "/:page", to: "docs#show"
  root to: "docs#index"
end
```

With this system, classes that inherit from `Administrate::Namespace` define which dashboards exist in the namespace, and which appear in the sidebar. Ideally they could be defined separately.

mcmire:
Is it possible to revisit this? I just ran into this and currently there's not a way to get around this without hacking Administrate. P.S. @graysonwright I like that idea, that seems really elegant.

pablobm:
Long time no activity :-( Closing, but happy if someone wants to take it up.

c4lliope:
Hey, can some maintainer delete the underlying branch here? noticed that it's still listed under the repo's stale branches.

much obliged.

PR #515, opened 2016-03-08
Wrap unsupported form field notes in proper divs

graysonwright:
## Problem:

Forms for `has_one` and `polymorphic` relationships aren't currently supported, and they show up on form pages as a note to the user.

This note is misaligned with the other form fields, because it is not wrapped in the same div that applies to the other fields.
## Solution:

Add the apropriate `field-unit__field` div to the unsupported field types.


graysonwright:
@tysongach - can I get some :eyes: ? thanks!

tysongach:
@graysonwright :rocket:

Issue #516, opened 2016-03-09
Cannot run rails generate administrate:install


nguyen7thai:
Hi All,

I'm trying gem administrate but I encounter the problem when running the command `generate administrate:install`.

As I investigated, the problem is because of this lines of code in namespace.rb:

``` ruby
def resources
    namespace_controller_paths.uniq.map do |controller|
      controller.gsub(/^#{namespace}\//, "").to_sym
    end
end
```

Its returned value includes the symbol that is created from empty string  `:""` .
Does anyone encounter this error?

Here's the backtrace:

/Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/railties-4.2.4/lib/rails/generators/named_base.rb:102:in `block in class_name': undefined method`camelize' for nil:NilClass (NoMethodError)
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/railties-4.2.4/lib/rails/generators/named_base.rb:102:in `map!'
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/railties-4.2.4/lib/rails/generators/named_base.rb:102:in`class_name'
    from (erb):3:in `template'
    from /Users/bebe/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/erb.rb:863:in`eval'
    from /Users/bebe/.rvm/rubies/ruby-2.2.3/lib/ruby/2.2.0/erb.rb:863:in `result'
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/thor-0.19.1/lib/thor/actions/file_manipulation.rb:116:in`block in template'
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/thor-0.19.1/lib/thor/actions/create_file.rb:53:in `call'
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/thor-0.19.1/lib/thor/actions/create_file.rb:53:in`render'
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/thor-0.19.1/lib/thor/actions/create_file.rb:46:in `identical?'
    from /Users/bebe/.rvm/gems/ruby-2.2.3@administrate/gems/thor-0.19.1/lib/thor/actions/create_file.rb:72:in`on_conflict_behavior'
[...]

PR #517, opened 2016-03-09
[#516] Fix error cannot run command rails generate administrate:install

nguyen7thai:
https://github.com/thoughtbot/administrate/issues/516

### DIFF
--- a/lib/administrate/namespace.rb
+++ b/lib/administrate/namespace.rb
@@ -7,6 +7,8 @@ module Administrate
     def resources
       namespace_controller_paths.uniq.map do |controller|
         controller.gsub(/^#{namespace}\//, "").to_sym
+      end.reject do |resource|
+        resource == :""
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/administrate/namespace.rb
```diff
@@ -7,6 +7,8 @@ def initialize(namespace)
     def resources
       namespace_controller_paths.uniq.map do |controller|
         controller.gsub(/^#{namespace}\//, "").to_sym
```
markmercedes: Adding this second validation to this line might be a better idea in my opinion:

`controller.starts_with?(namespace.to_s) && !controller.end_with?('/')`

It removes the complexity of having to run throughout the collection twice, also it won't allow any possible sub-route root to mess the generation up

nguyen7thai: Good idea, but in that case, the block would return a nil value (we are using method `map` here) and we have to remove the nil value in the collection thereafter?

nickcharlton:
Hi! Thanks for this PR. I'd not actually seen this when I went ahead and did #713, which purports to fix the referenced issue. I'm going to close this for now, but feel free to reopen if it doesn't.

PR #518, opened 2016-03-10
Update .rubocop.yml to support latest version of config options

OddEssay:
There are two errors with the current .rubocop.yml that are not compatible with the latest versions of Rubocop, and cause Rubocop to fail to run at all.

This PR updates the config to reflect the changes.

An alternative solution would be to fix the version of Rubocop in the Gemfile, if keeping things backwards compatible for some reason was preferable?


nickcharlton:
Thanks for this! I'm going to close in favour of #779 which syncs up completely with our guides.

PR #522, opened 2016-03-15
Modify has_many field show

### LINKED ISSUE #440: Has many field renders (potentially) wrong partial
If a dashboard has one or more has_many fields, then on that dashboard's show page the `collection` partial will be rendered for each has_many field:

https://github.com/thoughtbot/administrate/blob/master/app/views/fields/has_many/_show.html.erb#L22-L26

The problem has to do with the template lookup path.  Take for example a class `Alpha` that has many `Bravo`s.  On the `Alpha` show page, the `collection` partial will be rendered for each relation:
- If you have overridden the `collection` partial for `Bravo`, it will never be rendered for the `bravos` field.
- If you have overridden the `collection` partial for `Alpha`, it will be used to render the `bravos` collection.

Neither of these seem to be the intended behavior.  It would be easy to create a test exposing this bug.  I have successfully gotten my desired behavior by putting an ugly hack in `app/views/fields/has_many/_show.html.erb`.

It seems likely that similar bugs exist in this project in other places - I can't confirm or deny that right now.  It seems like to really fix this there needs to be support for resolving the correct administrate partial regardless of the view context while respecting template inheritance.

jasontwong:
Resolves #440 

Why:
- The template is rendering the incorrect partial.

- Create field helper `collection_partial(namespace)`.
- Update `has_many/_show.html.erb` to use new helper.

### DIFF
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -20,7 +20,7 @@ from the associated resource class's dashboard.
 
 <% if field.resources.any? %>
   <%= render(
-    collection_partial(namespace),
+    field.collection_partial(namespace),
     collection_presenter: field.associated_collection,
     resources: field.resources
   ) %>
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -20,7 +20,7 @@ from the associated resource class's dashboard.
 
 <% if field.resources.any? %>
   <%= render(
-    "collection",
+    collection_partial(namespace),
     collection_presenter: field.associated_collection,
     resources: field.resources
   ) %>
--- a/lib/administrate/field/has_many.rb
+++ b/lib/administrate/field/has_many.rb
@@ -24,6 +24,16 @@ module Administrate
         end
       end
 
+      def collection_partial(namespace)
+        partial = "collection"
+        prefix = options[:class_name].try(:underscore).try(:pluralize) || name
+        view_paths = ActionView::PathSet.new(["app/views/#{namespace}"])
+        lookup_context = ActionView::LookupContext.new(view_paths)
+        is_template = lookup_context.exists?(partial, [namespace, prefix], true)
+        partial = [namespace, prefix, partial].join("/") if is_template
+        partial
+      end
+
       def selected_options
         data && data.map(&:id)
       end


jasontwong:
Can anyone rerun the tests on CircleCI? It looks like a non-code issue cause it to fail.

wjlroe:
I've just hit this bug, is there anything that needs to be done for it to be merged?

hughfm:
Has there been a decision on whether this is the right way to fix this? I'd love to see a solution asap, and would be happy to help work towards one if there are any concerns about what has been proposed here.

hughfm:
@jasontwong thanks for your fix here, I've been using it in my application and all good so far! One thought though... 

say for example you have a `User` class that `has_many :posts`, and we are rendering a user's posts on their show page... if we have overridden the `posts/collection` partial, this will now be used correctly :+1: 

if that partial has not been overridden however, it still defaults to the `users/collection` partial, if that exists. I'm thinking it may make more sense to default straight back to `application/collection`. What do you think?

jasontwong:
@hughfm I'm glad it's been working well for you. The only reason why I left it to use the `users/collection` template was to make the default functionality relate as close to the original code as possible. This would make it easier for those already familiar with the usage to add in this override. I agree that the best way would probably be to use `application/collection` but I would rather leave it to the `administrate` team to decide what the default functionality should be.

hughfm:
Yep that's fair enough @jasontwong, would love to hear thoughts on this from the team..

allthesignals:
If you're trying to fix this bug in Rails 5, I've merged both of those branches here: https://github.com/MAPC/administrate

Include in your Gemfile like this:
`gem 'administrate', github: 'MAPC/administrate', branch: 'rails5'`

carlosramireziii:
@nickcharlton There's a question here about whether or not this is the right approach.

nickcharlton:
Hi! I'm just getting around to looking at this now. 

I think this is a reasonable approach to the problem.  Are we able to test this in both scenarios (using a default collection template, using a custom one)?

jasontwong:
It's been a while since I looked at this... I'll dig into the testing.

nickooolas:
I've noticed that after implementing this patch, the search bar for that particular model in Administrate doesn't render anymore - any thoughts on that one @jasontwong @allthesignals?

nickcharlton:
Hi @jasontwong! Did you get anywhere with looking at the testing?

josephchoe:
Has there been any progress on this?

nickcharlton:
@jasontwong Did you get a chance to look at this again?

pablobm:
Closing as there has been no activity for a while. Happy to reopen if someone wants to take over.

PR #524, opened 2016-03-18
Fix translations for show.html and index.html

pvbraga:
I found missing translations for model attributes at show.html and index.html. This pull request fix that.
I bet that could be better ways of doing that, but this will do the trick! 

While not merged I'm using a forked version from my own repo. If anyone need this translations fell free to get the gem from:

`gem 'administrate', :git => 'git://github.com/pvbraga/administrate.git'`


nickcharlton:
Hi @pvbraga! Thank you for your PR. I'm sorry we've only ended up looking at your PR more than a year after you'd opened it. A brief look makes me think we're now translating these attributes, but it's changed quite substantially since you'd opened it.

If this isn't the case, I'd love to see another PR and I'll make sure we get it in much sooner!

Issue #531, opened 2016-03-28
HasMany edit fields no longer working after upgrading from 0.1.2


michaelimstepf:
I've upgraded from 0.1.2 and have changed both the routes and `require "administrate/field/base"` on custom fields. HasMany fields are no longer a selectize field, making it very hard to select values from a long list.

``` ruby
require "administrate/base_dashboard"

class OrganizationDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    locations: Field::HasMany
  }

  FORM_ATTRIBUTES = [
    :locations
  ]
end
```

<img width="556" alt="screen shot 2016-03-28 at 3 37 04 pm" src="https://cloud.githubusercontent.com/assets/430682/14071170/ca01bfbe-f4fc-11e5-9aa9-5e517324d611.png">

carlosramireziii:
Hey @michaelimstepf! Just wanted to follow up on this one...

Is this still an issue with the latest version? If so, what kind of behavior would you like to see here? 

Let us know! Thanks

michaelimstepf:
Hey @carlosramireziii I'm not sure, I haven't used the gem for a year. I can post here if I use it again and let you know.

carlosramireziii:
Ok sounds good @michaelimstepf. Thanks for getting back to me! I'll close this for now. Cheers!

PR #540, opened 2016-04-11
example on how to use custom file types

LePhil:
"Field::" no longer necessary according to https://github.com/thoughtbot/administrate/issues/531


carlosramireziii:
@nickcharlton Documentation change

PR #541, opened 2016-04-11
bump rspec to 3.4

BenMorganIO:
This removes a deprecation warning from RSpec where we were being told to use last_description, instead of last_comment

```
[DEPRECATION] `last_comment` is deprecated.  Please use `last_description` instead.
```


graysonwright:
Thanks! Merged in as 7eb7ab9.

PR #543, opened 2016-04-12
remove the fuubar gem

### LINKED ISSUE #542: Remove the FuuBar Gem
I'm currently working through the project in my branch of `rails-5-support` and trying to get Rails 4.2 to load up RSpec 3.4 and rails 5 to load up RSpec 3.5. However, as the RSpec core version changes, this also changes the requirement of the FuuBar gem.

I'm sure that we can get the gem functional for Rails 5 and RSpec 3.5, but its extra work; especially in future upgrades of Rails and RSpec. Considering this, perhaps it would be OK to substitute a nice loading bar with smaller and easier to manage dependencies.

The FuuBar gem also has an open pull request (https://github.com/thekompanee/fuubar/pull/87) which was submitted back in February. This PR is for RSpec version 3.5.0.beta1, which will most likely conflict with 3.5.0.beta3 which was recently released last week.

FuuBar looks like a great gem though. Heck, I'm probably going to submit a few PRs to it and get it working with Rails 5 and RSpec 3.5 so that _I_ can use it my project. But, from a gem perspective that in the future will have to manage multiple versions of Rails and RSpec, it may not be the best option.

BenMorganIO:
fixes #542 

While doing a Rails 5 update, I had to appraise to a beta version of RSpec (3.5.0.beta3). While doing so, FuuBar was only set up for up to RSpec 3.4. There is currently an open PR to add 3.5.0.beta1 support, but it was back in February which sends a message that the FuuBar gem is not actively maintained. FuuBar may not be the best gem to use in a project which needs to cover multiple versions of Rails and RSpec.

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,3 +1,2 @@
 --color
 --require spec_helper
---format Fuubar
--- a/Gemfile
+++ b/Gemfile
@@ -32,7 +32,6 @@ group :test do
   gem "ammeter"
   gem "database_cleaner"
   gem "formulaic"
-  gem "fuubar"
   gem "launchy"
   gem "percy-capybara"
   gem "poltergeist"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -118,9 +118,6 @@ GEM
       activesupport
       capybara
       i18n
-    fuubar (2.0.0)
-      rspec (~> 3.0)
-      ruby-progressbar (~> 1.4)
     globalid (0.3.6)
       activesupport (>= 4.1.0)
     hashdiff (0.2.2)
@@ -218,10 +215,6 @@ GEM
     rake (11.1.2)
     rdiscount (1.6.8)
     redcarpet (3.3.3)
-    rspec (3.4.0)
-      rspec-core (~> 3.4.0)
-      rspec-expectations (~> 3.4.0)
-      rspec-mocks (~> 3.4.0)
     rspec-core (3.4.4)
       rspec-support (~> 3.4.0)
     rspec-expectations (3.4.0)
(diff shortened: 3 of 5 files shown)


BenMorganIO:
@graysonwright when you have time, I would really appreciate it if you could look at this PR. I'd love to help get Administrate Rails 5 ready.

graysonwright:
👀

graysonwright:
Looks good! Merged in as af56deb.

jfelchner:
I realize this literally just got removed, but I'm going to be releasing a version in just a minute with RSpec 3.5 beta support.  Fuubar is definitely maintained, but both it and `ruby-progressbar` are so rock solid that I rarely ever have to touch them.

The only time Fuubar ever really needs touched is when a beta version of RSpec comes out and that open issue fell through the cracks.

jfelchner:
@BenMorganIO also, for the record, `fuubar` 1.0 works all the way back to RSpec 2.0 and `fuubar` 2.0 works with everything else.  There's no dependency on Rails at all.  The formatters in RSpec are extremely stable and therefore `fuubar` is as well.  Like I said, the only reason `fuubar` even needs to be touched now is when a beta version of RSpec comes out and I would argue that it's beta and shouldn't be supported by a stable version of a gem.

I've released `2.1.0.beta1` with RSpec `3.5.0` beta support that will work until `3.5.0.beta10`.  If they do an rc (they usually do) I'll release an rc version at that time.  Feel free to ping me in an issue if I miss it.

jfelchner:
@BenMorganIO give `beta2` a whirl.  Additionally, because I know a lot of people's gems depend on `fuubar`, it (and in a few minutes `ruby-progressbar`) are now cryptographically signed using for added security.

PR #544, opened 2016-04-12
bump ruby to 2.2.4

BenMorganIO:
While running specs for the project, I noticed that we were getting these warnings:

```
>> BUNDLE_GEMFILE=/home/ubuntu/administrate/gemfiles/sass_3_4.gemfile bundle exec rake
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.4-compliant syntax, but you are running 2.2.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
warning: parser/current is loading parser/ruby22, which recognizes
warning: 2.2.4-compliant syntax, but you are running 2.2.3.
warning: please see https://github.com/whitequark/parser#compatibility-with-ruby-mri.
```

This bumps ruby to 2.2.4 which is nice for maintenance and to squash warnings :)

We should bump to ruby 2.3 in the future. If we can do so, that would be great. I'm a major user of CircleCI, but they don't support matrixes. If we want to support and test administrate on both ruby 2.2. and ruby 2.3, we may have to move to TravisCI. That's a discussion for another time though. Anywho, lets squash some warnings!


Gemfile.lock
```diff
@@ -168,6 +171,8 @@ GEM
     nokogiri (1.6.7.2)
       mini_portile2 (~> 2.0.0.rc2)
     normalize-rails (3.0.3)
+    parser (2.3.0.7)
```
BenMorganIO: Notice how the `parser` gem has its version reflect the version of ruby. It was a dependency of `i18n-tasks` which meant a bump of it was needed.

BenMorganIO:
@graysonwright Are you able to restart the specs with the cache on Travis invalidated? I'm a bit at my wits end trying to figure out why things aren't working.

PR #546, opened 2016-04-15
Support nested forms for has_many relationships

### LINKED ISSUE #192: Feature Request: Nested Forms
Hi guys,

Thanks in advance for this great project, really good job. I was looking for something like that such a long time. 

I was digging around and I have missed and option to build nested forms. I don´t know if there is an easy way to do it. IMHO its needed for the example app when you add line items to an order. Its painful to first create an order, save it then create line items and associate it. This happens on all kind of database models and It would be amazing to have this feature.

I think you can achieve this behaviour by adding custom fields types, overriding form view. Although you must allow nested attributes on your models.

Let me know what do you think about it and set a guidelines to approach this feature.

Thanks in advance and keep pushing!

graysonwright:
Closes #192

## Feature:

When I create or edit a model that has_many nested models,
I want to view and edit the attributes of the nested models
so I can set up all the relationships with a single form.

## Implementation:

Introduce a new field type, `NestedHasMany`.

I considered building the feature into the existing `HasMany` field,
but this would get in the way of `HasMany` relationships
that aren't directly nested, such as in many-to-many relationships.

The `NestedHasMany` field builds off of the [Cocoon] gem.
It renders fields from the nested model,
based on the fields defined in the nested model's dashboard class.

Cocoon provides helpers and javascript
to easily add and remove nested form fields from the page.

[Cocoon]: https://github.com/nathanvda/cocoon

## Preview

nested

## Next Steps

- [ ] Needs some design love
- [ ] Needs tests
- [ ] It would be great to have people test this on their own apps and report bugs before we merge


administrate.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rails", "~> 4.2"
   s.add_dependency "sass-rails", "~> 5.0"
   s.add_dependency "selectize-rails", "~> 0.6"
+  s.add_dependency "cocoon", "~> 1.2"
```
tysongach: :abc: this into the existing list?

app/views/fields/nested_has_many/_form.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    <%= render(
+      partial: "fields/nested_has_many/fields",
+      locals: {
+        f: nested_form,
+        field: field,
+      },
+    ) %>
+  <% end %>
+
+  <div class="links">
```
tysongach: Is this `links` class being utilized in any way? If not, can we remove for now so we don’t forget to use/remove it later?

app/views/fields/nested_has_many/_fields.html.erb
```diff
@@ -0,0 +1,12 @@
+<div class="nested-fields" style="width:100%">
+
+  <%# TODO filter out the parent relationship, without this crude array slicing %>
+  <% field.associated_form.attributes[1..-1].each do |attribute| -%>
+    <div class="field-unit field-unit--<%= attribute.html_class %>">
```
tysongach: What does `<%= attribute.html_class %>` output?

graysonwright: It gives you the type of field that is being rendered. For example, `Administrate::Field::String` attributes would return `"string"`. It's useful for styling fields based on the type of attribute.

graysonwright:
In #476, we decided on a plugin structure for new field types. Nested `has_many` forms are a good candidate for a plugin, so I extracted this code out to https://github.com/graysonwright/administrate-field-nested_has_many.

It's rough but functional. PRs are welcome on that repository.

PR #550, opened 2016-04-19
Fix `bin/release` script for pushing to rubygems

graysonwright:
CC @jberlinsky

## Problem:

There were a few bugs in the `bin/release` script
that prevented a successful push to rubygems

## Solution:

Update the script to successfully push a release to rubygems.

## TODO:

- it's still getting hung up on the `appraisal install` step.


BenMorganIO:
We should probably figure out that `appraisal install` problem before merging any new appraisal additions (think Rails 5).

PR #552, opened 2016-04-19
Remove percy-capybara

graysonwright:
A while back, we disabled Percy
because we were overusing our plan limit.
Since it's been a while since we've used it,
it makes sense to remove all dependencies on the gem.


BenMorganIO:
:+1:

PR #553, opened 2016-04-19
Feature scope option

Mrjaco12:
I don't know if the functionality is desirable but I've found myself needing it on more than one occasion, so I figured I'd put this out there for feedback.

Why:
- So a scope can be passed to an association field

- Moving candidate_resources to the associative class
- Handling when a scope is present
- Passing the current data to the scope as params

### DIFF
--- a/lib/administrate/field/associative.rb
+++ b/lib/administrate/field/associative.rb
@@ -10,7 +10,8 @@ module Administrate
       protected
 
       def candidate_resources
-        return associated_class.public_send(options[:scope], data) if options[:scope]
+        scope = options[:scope]
+        return associated_class.public_send(scope, data) if scope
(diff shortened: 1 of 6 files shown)


pablobm:
@Mrjaco12 - Thank you for this PR... and sorry that it took so long to come back to you :-(

I have a question: could this be achieved instead by adding a separate, scoped association to the model (rather than to the dashboard) and make the dashboard read from this one?

Mrjaco12:
I can take a look later this week/month. I honestly don't know if this PR is still useful or not. I haven't used Administrate in a minute.

nickcharlton:
I'm going to close this as it's been a long time since the original issue was opened and it should be solvable elsewhere.

PR #568, opened 2016-05-03
Update customizing_dashboards.md

rikkipitt:
Add missing text field type to documentation.


nickcharlton:
Good catch! Thanks; merging.

Issue #566, opened 2016-04-27
I created Administrate::Field::JSON and looking for feedback.


eddietejeda:
I created Administrate::Field::JSON and looking for feedback. I used Administrate:Field:Image created by @graysonwright as a guide.

https://github.com/eddietejeda/administrate-field-json

I use [JSONEditor](https://github.com/josdejong/jsoneditor) to view and edit the JSON. I found this library be the most complete and flexible of all the JSON editor libraries out there. It might need some styling edits to make it fit in with Administrate's look.

I've been using Administrate for few months now and testing Field::JSON for a couple days. I have not released it to Rubygems yet and would appreciate feedback before doing so.

In creating this plugin, I encountered a few problems with the plugin architecture. I can file these as separate issues, but to quickly summarize:

1) Adding assets to your plugin is tricky. application(js/css) are not automatically picked up by the asset pipeline. In Greyson's sample, he suggests that we edit the Administrate template (https://github.com/graysonwright/administrate-field-nested_has_many). I did not want to do that, so I added the plugin path to the asset pipeline and gave application(js/css) a different name. See example here: https://github.com/eddietejeda/administrate-field-json/blob/master/lib/administrate/field/json.rb. I think that's a better long term solution since it encapsulates the components within the plugin.

2)  It was unclear how I can inject my own js into `=yield :javascript` at a page level, not field level, so I have to flush `=content_for` to prevent printing out duplicates. https://github.com/eddietejeda/administrate-field-json/blob/master/app/views/fields/json/_form.html.erb Let me know if there is a better way to do this.

3) There was no `=content_for :stylesheet`, so I am currently using the  `=content_for :javascript`. I propose a simple edit to support =content_for :stylesheet. Here's how that would look like: https://github.com/thoughtbot/administrate/compare/master...eddietejeda:content_for_stylesheet?expand=1

4) For the _show.erb partial, I just display the JSON as within &lt;pre&gt; tags. I could use JSONEditor and mark it as readonly. Would like to hear preferences.

Let me know if Field::JSON works for you all and which of these other observations I should file as issues/pull requests.

Thanks.

graysonwright:
@eddietejeda, thanks for the PR! I really like a lot of your ideas, and think they would really improve the API for creating plugins. Here are some of my thoughts:

1. I agree that adding assets is tricky, and I want to improve the process. Your solution of adding a new file to the asset pipeline is a great start.
[...]

PR #573, opened 2016-05-10
Add engine class methods for registering assets

graysonwright:
See https://github.com/thoughtbot/administrate/issues/566

## Problem:

When developing a plugin for Administrate,
there is not a good way to define custom javascripts or stylesheets
to be included on admin pages.

Existing solutions either use `content_for` at the field level,
or require the end user to override Administrate's javascript partial.

## Solution:

Add methods to `Administrate::Engine` to keep track of the assets
that need to be loaded on each admin page.

Plugin developers can now use these helpers in their engine definition:

```ruby
Administrate::Engine.add_javascript "my_plugin/script"
Administrate::Engine.add_stylesheet "my_plugin/styles"
```

### DIFF
--- a/app/views/administrate/application/_javascript.html.erb
+++ b/app/views/administrate/application/_javascript.html.erb
@@ -7,7 +7,9 @@ but each page can define additional JS sources
 by providing a `content_for(:javascript)` block.
 %>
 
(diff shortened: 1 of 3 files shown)


graysonwright:
Feedback sounds positive for this, so I went ahead and merged it in.

PR #575, opened 2016-05-11
Rails 5 Compatibility PR

imajes:
There are a couple other PRs to deliver Rails 5 compat, but this one should pass specs, and be mergeable.

I've focused mostly on:
-  the few small fixes required to actually bring compat
-  supporting appraisals for rails 4 & 5
-  re-introducing fuubar, which is now rspec compatible again
-  silencing or fixing a number of deprecation notices

Regarding the strong params issue, my approach doesn't blanket permit (which is going to be a security vulnerability, due to the abstract way it's implemented and the potential for scriptable exploits). Instead it permits specific, expected params. I don't know if this is sufficient, and we may need to extend this with some of the `ATTRIBUTE` constants to encapsulate all of the expected params. Will text this further as I integrate `administrate` into my own app.

~~There is one deprec notice remaining that is particularly annoying; resolving it removes rails4 test compatibility. I considered wrapping the calls in a deprecation silencer, but felt that isn't a good idea given how it may be missed later. One option may be overriding the specific deprecation notice to silence it.~~

-- since resolved that with a wrapper def.


imajes:
Updated to handle that deprec notice which was annoying, as well as requiring ruby 2.3.0 which should be ci compatible

imajes:
@graysonwright: pr is passing; would be great to get this looked at. if you want me to base it off a different branch, lemme know.

infertux:
FYI I just tried to use your branch on an existing project and everything seems to be working perfectly :+1:

jalberto:
@imajes great work

BenMorganIO:
I would remove Fuubar. I think its a great gem for Rails apps, but for a popular open source project, the less dependencies, even in development, the easier it is to move forward.

It also adds less code complexity and helps keep things down to "stock" format. Don't get me wrong though, I think the gem is great :)

.ruby-version
```diff
@@ -1 +1 @@
-2.2.3
+2.3.0
```
BenMorganIO: I think we can remove this actually. Testing on multiple ruby versions should be up to Travis for an OSS gem. We're not a rails application, just an engine. I don't think Administrate should even be aware of the ruby version.

Appraisals
```diff
@@ -1,7 +1,32 @@
...
+appraise "rails-5" do
+  gem "rails", "5.0.0.rc1", "< 5.1"
+  gem "rspec-rails", "~> 3.5.0.beta1"
+  gem "kaminari", github: "amatsuda/kaminari", branch: "0-17-stable"
+
+  # this fork has rails at >= 4.2, allowing rails5 to work.
+  gem "administrate-field-image", github: "pedrocarmona/administrate-field-image"
+
+  # rails 5+ extracts this gem, which is need for shoulda-matchers:
+  gem "rails-controller-testing"
```
BenMorganIO: Hmmm... Can you provide some examples of where its needed? Rails did remove it for the right reasons.

Since this is a rails engine though, it may be appropriate to test some of the internals, since we are setting up some internals inside of a users Rails app...

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-ruby "2.2.3"
+ruby "2.3.0"
```
BenMorganIO: We should probably remove this altogether.

Gemfile
```diff
@@ -26,7 +26,7 @@ group :development, :test do
   gem "faker"
   gem "i18n-tasks"
   gem "pry-rails"
-  gem "rspec-rails", "~> 3.4.2"
+  gem "fuubar", "~> 2.1.0.beta2"
```
BenMorganIO: I'm in favour of removing this to help make RSpec changes in the future easier, but I know some other people have some different opinions :)

Gemfile.lock
```diff
@@ -305,13 +314,15 @@ DEPENDENCIES
...
-  rspec-rails (~> 3.4.2)
   shoulda-matchers (~> 2.8.0)
   timecop
   uglifier
   unicorn
   web-console (>= 2.1.3)
   webmock
 
+RUBY VERSION
+   ruby 2.3.1p112
```
BenMorganIO: Well, this is interesting...

app/views/administrate/application/_collection.html.erb
```diff
@@ -25,8 +25,8 @@ to display a collection of resources in an HTML table.
         <th class="cell-label
         cell-label--<%= attr_type.html_class %>
         cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
-        " scope="col">
-        <%= link_to(params.merge(
+      " scope="col">
```
BenMorganIO: Whitespace?

gemfiles/rails_4.gemfile
```diff
@@ -0,0 +1,51 @@
...
+  gem "webmock"
+end
+
+group :staging, :production do
+  gem "rack-timeout"
+  gem "rails_stdout_logging"
+  gem "uglifier"
+end
+
+gemspec :path => "../"
```
BenMorganIO: Do we need to commit these? Aren't these auto-generated via Appraisal?

lib/administrate/version.rb
```diff
@@ -1,3 +1,3 @@
 module Administrate
-  VERSION = "0.2.0".freeze
+  VERSION = "0.2.1".freeze
```
BenMorganIO: Maybe save the version bump for its own PR :)

lib/generators/administrate/install/templates/application_controller.rb
```diff
@@ -6,7 +6,7 @@
 # you're free to overwrite the RESTful controller actions.
 module Admin
   class ApplicationController < Administrate::ApplicationController
-    before_filter :authenticate_admin
+    before_action :authenticate_admin
```
BenMorganIO: :+1:

Appraisals
imajes: `shoulda-matchers` uses `assert_template`, which was extracted into that gem. I'm not wildly in love with that either, especially since `shoulda-matchers` is quite out of date/unmaintained... however I didn't feel this was a great time to make those changes.

Gemfile
imajes: [Rails 5 requires Ruby 2.2.2+](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-2-2-2). We will need to find a way to signpost that in for devs. I agree that the way it's implemented in gemfiles etc is poor, but without it you end up locking in CI configs etc which is also non optimal.

Gemfile
imajes: It's not removed, it's a varying requirement between Rails 4/5. It's listed in appraisals.

Gemfile.lock
imajes: That's just the version used to build the lock (i.e. the one i am running)

gemfiles/rails_4.gemfile
imajes: They are autogenerated, but we are taught to commit our gemfiles to share expectations of compatibility. We also want to version to manage potential regressions.

lib/administrate/version.rb
imajes: actually it is a requirement to get this to work, otherwise you get compatibility issues since version 0.2.0 has dependencies as defined previously to this work. Upgrading rails with cross deps is pretty annoying, and bundler is pretty dumb to the change, hence the version bump. This PR is the review; i'm sure things will get edited out for merge.

imajes:
re fuubar. I did these changes without it, and added near the end. It reminded me exactly how much time it saves running tests, by surfacing issues faster. Gems often have poor coverage as PRs etc don't typically spend as much time as they might considering testing, therefore I think tools which reduce overhead for testing are valuable.

This gem is already beleaguered, as all Rails engines are, with needing to support multiple Rails versions, and other compatibility issues-- thoughtbot's own initial setup tooling is also more complex than average. I see no reason to lose developer performance for the sake of a lighter weight gem development environment. 

NB: i would agree for actual gem deps; they should be as minimal as possible.

lukeredpath:
I've just setting up this gem from this branch in a fresh Rails 5 RC1 app. The generator fails with this error:

```
WARNING: Unable to generate a dashboard for ApplicationRecord.
         Make sure your database migrations are up to date.
      create  app/controllers/admin/application_controller.rb
      create  app/dashboards/_dashboard.rb
/Users/luke/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/railties-5.0.0.rc1/lib/rails/generators/named_base.rb:102:in `map!': undefined method `camelize' for nil:NilClass (NoMethodError)
```

imajes:
@lukeredpath: hey Luke! is your app not hooked up to any db at all? It does appear to use `AR::B` for the routes generator-- i think that could get switched to just the new `AR`-- but i didn't have this issue, possibly because i had already setup a DB for my app.

Gemfile
BenMorganIO: Well, we're supporting Rails 4 as well. So, we should be able to support older versions of ruby if we can.

gemfiles/rails_4.gemfile
BenMorganIO: Cool. KK

lib/administrate/version.rb
BenMorganIO: Understood. We should probably move to a beta model then. Such as `0.3.0.beta` :)

BenMorganIO:
@lukeredpath can you open a new issue for this? I have a feeling its because administrate may not be respecting the `abstract_class` setting.

Gemfile
imajes: Yeah, like i said, the correct way to define this isn't exactly clear, and its a dev dependency. Since ruby (should) have appropriate BC, it seems best to suggest the minimum required version for all features, hence 2.2+

lib/administrate/version.rb
imajes: Yeah. i didn't really want to jump too far until we had some of @graysonwright's guidance... however i have definitely noticed a trend in thoughtbot repos to sit with PRs for, well, a long time.

Appraisals
BenMorganIO: Yep, just took a dive into Shoulda Matchers. Latest update was in January and there's tons of low hanging fruit when it comes to issues. I'll open a PR to remove the gem. The gem is great for making sure you didn't leave any typos around when you make a new association, but beyond that, I don't really find that much use for it.

Anywho, going to remove because the gem isn't up to date and we don't really need it. We don't use Shoulda with Solidus.

Gemfile
BenMorganIO: Cool. Maybe we could switch from CircleCI to Travis and utilize the ruby matrix? Support Ruby 2.2, 2.3, and head?

imajes: Sounds great, but i don't see that as a blocker to getting this landed. Seems like it's another PR.

Appraisals
imajes: great.

Gemfile
BenMorganIO: Yep. thinking of a making a PR for it right now.

BenMorganIO:
Linking this PR here for historical reasons: https://github.com/BenMorganIO/administrate/pull/1

lukeredpath:
@BenMorganIO sure, done

BenMorganIO:
Alright, so I just happen to be working on a Rails 5.0.0.rc1 app right now that is getting nicely sizable. @imajes I'll be using your branch this week. I totally want to get your work in. Let's aim for Monday next week after some battle testing :).

In the mean time, how do you feel about splitting you +237 lines of code out into smaller PRs? The `before_filter` => `before_action` is something we can definitely get in pretty fast. I'm sure there's other hunks of code that can get reviewed and merged on the fast lane as well.

imajes:
@BenMorganIO: do you have write access to this repo?

BenMorganIO:
@images that's on the way from what I hear :) cc @graysonwright

acrogenesis:
If you are using this branch with ruby 2.3.1 it won't build, had to fork this branch and change the Gemfile ruby version

imajes:
@graysonwright: so.... any particular reason you haven't responded to this PR?

ACPK:
@graysonwright @imajes  +1

netmask:
This will be merge any time soon ?

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -164,19 +164,27 @@ def expect_locals
...
       expect(response).to redirect_to(admin_customers_url)
     end
   end
 
+  def kwarg_params(args)
+    if Rails::VERSION::MAJOR >= 5
+      {params: args}
+    else
+      args
+    end
```
BenMorganIO: You could probably change this to a ternary operator here:

```ruby
Rails::VERSION::MAJOR >= 5 ? { params: args } : args
```

acrogenesis: more than ternary it should be a guard clause, no?
```ruby
def kwarg_params(args)
  return { params: args } if Rails::VERSION::MAJOR >= 5
  args
end
```

BenMorganIO: Yep, that's another solution. I would opt for the ternary mainly because I think the purpose of the method is to format the params if its greater than or equal to Rails 4... If we had to use guard clauses, I think a more optimal outcome would be:

```ruby
return args if Rails::VERSION::MAJOR <= 4
{ params: args }
```

🚲 🏠

imajes: Actually this is the worst place to put in any kind of optimal code. This is a yak shaving method to bring about compatibility in a situation where the api has broken. 

It will eventually be removed, but till then it should be as self documenting as possible and as plain spoken as possible lest it turn into an actual valuable feature. 

347.433.0567 | @imajes

acrogenesis:
I made a PR to [greetpoint/rails5#2](https://github.com/greetpoint/administrate/pull/2) updating kaminari, because if your dashboard needs pagination it will crash with the old kaminari version

fx:
@graysonwright +1

onemanstartup:
Maybe update kaminari to 1.0.0.alpha?

ID25:
Hey guys, how is your progress?

chamnap:
How is it going, guys?

thomascchen:
Can this PR be closed? It looks like Rails 5 support has been merged in with https://github.com/thoughtbot/administrate/pull/636

pedantic-git:
I don't think the master branch is ready for Rails 5 yet. For example, the app/views/administrate/application/_search.html.erb line 17 still generates this error:

"Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure."

I've been using @v-kumar's branch from #656 which resolves all my Rails 5 issues, but the current master does not.

pcantrell:
I’ve been using the master branch with Rails 5 for a few days of development now, and am not having the sanitized params problem. I was on commit 84096c for a while, now e533a6:

```
$ bundle show administrate
/Users/paul/.rvm/gems/ruby-2.3.0/bundler/gems/administrate-e533a6088521
```

@pedantic-git, what commit are you showing? Do you perhaps have a customized `_search.html.erb` that still uses `params` instead of `sanitized_params`?

pcantrell:
My bad — the fix isn’t in `_search.html.erb`; it’s in kaminari. All it takes to fix it is `bundle update kaminari` to get it to version 0.17. An update to this administrate’s gemfile should fix it. I’d already done that in my app, which is why it worked for me.

nickcharlton:

PR #577, opened 2016-05-14
Use correct inflection for has-many counts of 1

Billiam:
This fixes the issue where a has-many association with one record will be
displayed with a plural on index pages, ex: "1 orders", instead of "1 order".

Rails expects the second argument to the pluralize helper to be the
singular form of the word.


BenMorganIO:
👍 great work @Billiam

Billiam:
On second thought, this should be passed through 18n helpers instead.

PR #579, opened 2016-05-14
Change dashboard USAGE to README

DanAndreasson:
Changed `USAGE` to `README.md` so that Github can pick up the file when browsing the dashboard generators and present the content.

649989e2f74d0773ce9266ed3be0a900

### DIFF
--- a/lib/generators/administrate/dashboard/README.md
+++ b/lib/generators/administrate/dashboard/README.md
@@ -1,0 +1,11 @@
+# Dashboard generators
+
+## Description:
+Generates a Dashboard object for a model,
+pulling the attributes from database columns.
+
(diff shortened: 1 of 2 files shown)


BenMorganIO:
I would fence the `rails g ...` in the example, but that's just bikeshedding...

This is a great change!

BenMorganIO:
Err have to close. I think this actually by design. When you ask for help about a generation command, Rails will use the `USAGE` file to display the text.

Example: https://github.com/rails/rails/blob/52ce6ece8c8f74064bb64e0a0b1ddd83092718e1/railties/lib/rails/generators/rails/plugin/USAGE

Issue #536, opened 2016-04-06
Hide resources in sidebar | Pagination in ::HasMany


LePhil:
Is it possible to customize which resources are being shown in the sidebar? We have some assignment groups (many-to-many relation) which should not be accessible through the sidebar.

Also, there is an automatic limit on "sub-views" (HasMany-Fields), with "Showing X out of Y" - how would one disable/override that to show all related entries, or even better, offer pagination there as well as in the main (index) views?

OddEssay:
You can generate a customizable version of the sidebar with the undocumented `administrate:views:sidebar` and modify that however you wish.

Depending on how you modify it, you might need to in future manually add new resources, but it'd be a solid starting point.

LePhil:
Immediately after posting we found the generator you mentioned :smiley: 

Any ideas on the Pagination in :HasMany?

LePhil:
Adding an example screenshot:
![Example](https://cloud.githubusercontent.com/assets/1488366/14427315/6830e398-fff4-11e5-9f81-bacac8385a41.png)

It's nice to see how many are being displayed out of how many total, but it would be even nicer if the pagination would work here too. I created a custom field for that (HasManyPaginated) but I can't seem to figure out how to work the pagination via Kaminari out.

Xosmond:
That would be good to make a limit number of items showing, also there would be good to select what attributes should be show it there (parent view) and what attributes should be show it on the general view (model view). But that will break a lot of things.

micapam:
I know this is not my repo so it's not exactly my business, but would be handy to keep issues in separate threads rather than combining ;)

jasonswett:
You can hide resources by modifying your route configuration. For example, let's say your route file looks like this:

```ruby
Rails.application.routes.draw do
  namespace :admin do
    resources :users
    resources :authors
    resources :authors_books
    resources :books
    resources :books_tags
    resources :tags

    root to: "users#index"
  end

  resources :tags
  devise_for :users
  resources :authors
  resources :books
  root "books#index"
end
```

And let's say you want to remove `authors_books` and `books_tags` from your sidebar. You can simply remove those resources from the `:admin` namespace like this:

```ruby
Rails.application.routes.draw do
  namespace :admin do
    resources :users
    resources :authors
    resources :books
    resources :tags

    root to: "users#index"
  end

  resources :tags
  devise_for :users
  resources :authors
  resources :books
  root "books#index"
end
```

And those items will disappear from the sidebar.

[...]

PR #580, opened 2016-05-16
Allow links to resources to be hidden from the sidebar

micapam:
As per #536, I've found that the default behaviour of showing links to all resources' index pages in the sidebar is rarely desirable. 

I understand that it's possible to generate the sidebar and then put some logic in there - exclusions and the like. However, my experience from all use cases thus far (current project is the third time I'm using Administrate) tells me that this should be immediately available and obvious functionality. Thus this PR.

### DIFF
--- a/app/views/administrate/application/_sidebar.html.erb
+++ b/app/views/administrate/application/_sidebar.html.erb
@@ -8,7 +8,9 @@ as defined by the routes in the `admin/` namespace
 %>
 
 <ul class="sidebar__list">
-  <% Administrate::Namespace.new(namespace).resources.each do |resource| %>
+  <% Administrate::Namespace.new(namespace).resources.select {|resource|
+  	  include_resource_in_sidebar?(resource)
+  	}.each do |resource| %>
     <li>
       <%= link_to(
         display_resource_name(resource),
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -5,6 +5,10 @@ module Administrate
       render locals: locals, partial: field.to_partial_path
     end
 
+    def include_resource_in_sidebar?(resource_name)
+      "#{resource_name.to_s.classify}Dashboard".constantize.new.show_in_sidebar?
+    end
+
     def display_resource_name(resource_name)
       resource_name.
         to_s.
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -53,6 +53,10 @@ module Administrate
       "#{resource.class} ##{resource.id}"
     end
 
+    def show_in_sidebar?
+      true
+    end
+
     private
 
     def attribute_not_found_message(attr)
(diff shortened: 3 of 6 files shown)


BenMorganIO:
:+1: for giving users the option to turn sidebar presence off.

wkurosawa:
What if we give more control on sidebar list? 

I was trying to remove some resources and also use a specific order rather than alphabetical, then I redefined `_sidebar.html.erb` using my own array of resources as symbols, sorted and filtered.

Something like select/filter/sort resources using a priority integer? (ActiveAdmin-like?)

BenMorganIO:
I do love the idea of a priority for sorting. We should also allow people to "hardcode" the sorting as well. that could prove useful when you want "Dashboard" at the top and the rest alphabetized.

micapam:
I know the committers here aren't keen on ActiveAdmin :) Maybe the best way would be to shift the sidebar logic out of the view and into the main controller: something like a sidebar_resources method that by default just shows them all (as per default behaviour) and people can override as they see fit?

micapam:
Happy to make those changes but would be good to hear opinions from Thoughtbot folks...

micapam:
@graysonwright, thoughts ? :)

app/views/administrate/application/_sidebar.html.erb
```diff
@@ -8,7 +8,9 @@ as defined by the routes in the `admin/` namespace
 %>
 
 <ul class="sidebar__list">
-  <% Administrate::Namespace.new(namespace).resources.each do |resource| %>
+  <% Administrate::Namespace.new(namespace).resources.select {|resource|
+  	  include_resource_in_sidebar?(resource)
+  	}.each do |resource| %>
```
graysonwright: Can we extract this logic out of the view into `Administrate::Namespace`?

graysonwright:
@micapam I think the best place for defining the resources that appear in the sidebar would be as arguments to `Administrate::Namespace`. That would let users define an `/admin` namespace that doesn't show line items, and a separate `/super_admin` namespace that does.

I'm not sure what the logic would look like for that yet. Right now namespaces are pretty small helper objects, and they don't carry a lot of state. If we wanted to let them keep track of the sidebar resources that show up, we'd probably have to define them in a single place and pass them around the application.

micapam:
@graysonwright, just saw your message, sorry for delay. 

Do you mean arguments to the Namespace constructor method (#initialize)? Or to the resources method?

ACPK:
@micapam +1.

nickcharlton:
Hi @micapam! I'd love to try and get this merged. Could you rebase on `master` or [enable contributor pushes](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

pablobm:
Closing as:

1. This has been inactive for a long time.
2. This can be achieved by removing the `:index` routes for undesired links, since https://github.com/thoughtbot/administrate/pull/1591.

Happy to re-examine this though!

Issue #581, opened 2016-05-16
Generator not respecting abstract_class


lukeredpath:
I'm trying to set up Administrate with a new Rails 5 project, using the as yet unmerged compatibility branch.

The administrate:install generator fails with the error:

```
WARNING: Unable to generate a dashboard for ApplicationRecord.
         It is not connected to a database table.
         Make sure your database migrations are up to date.
      create  app/controllers/admin/application_controller.rb
      create  app/dashboards/_dashboard.rb
/Users/luke/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/railties-5.0.0.rc1/lib/rails/generators/named_base.rb:102:in `map!': undefined method `camelize' for nil:NilClass (NoMethodError)
```

Its possible this isn't a Rails 5 compatibility issue but something that has surfaced due to Rails including ApplicationRecord, an abstract sub-class of ActiveRecord::Base, by default.

Also mentioned on #575 

Not sure if this is related, but I also ended up with an empty "_dashboard.rb" file. After looking at the generator code I notice that if I run `Administrate::Namespace.new(:admin).resources` on the console it returns `[:""]`.

lukeredpath:
On further inspection I think there's two issues here: the warning generated (which is just a warning so its not the reason the generator is failing) and the attempt to try and generate dashboards crashing with the error above.

Here's the full backtrace to the first line of administrate code:

```
     create  app/dashboards/_dashboard.rb
/Users/luke/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/railties-5.0.0.rc1/lib/rails/generators/named_base.rb:102:in `map!': undefined method `camelize' for nil:NilClass (NoMethodError)
	from /Users/luke/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/railties-5.0.0.rc1/lib/rails/generators/named_base.rb:102:in `class_name'
	from (erb):3:in `template'
	from /Users/luke/.rbenv/versions/2.2.3/lib/ruby/2.2.0/erb.rb:863:in `eval'
	from /Users/luke/.rbenv/versions/2.2.3/lib/ruby/2.2.0/erb.rb:863:in `result'
	from /Users/luke/.rbenv/versions/2.2.3/lib/ruby/gems/2.2.0/gems/thor-0.19.1/lib/thor/actions/file_manipulation.rb:116:in `block in template'
[...]

PR #582, opened 2016-05-16
Abstract model classes should be skipped without warning.

lukeredpath:
I’ve also clarified the existing specs by including an assertion against the presence of a warning, so the name of the spec matches what is actually being tested.

I'm not 100% happy with checking against an exact match on the warning text but it seemed important to check for the presence (or absence) of a warning.

This is related to #581 but I'm not sure its the only issue I'm having with this generator on Rails 5.

### DIFF
--- a/lib/generators/administrate/routes/routes_generator.rb
+++ b/lib/generators/administrate/routes/routes_generator.rb
@@ -43,7 +43,7 @@ module Administrate
       end
 
       def database_models
-        ActiveRecord::Base.descendants.reject { |klass| klass.abstract_class? }
+        ActiveRecord::Base.descendants.reject(&:abstract_class?)
       end
 
       def invalid_database_models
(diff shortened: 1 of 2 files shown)


BenMorganIO:
@lukeredpath totally want to get this merged. Can you rebase and squash 8e945d6 and 4551dc6 into a single commit? You could also use heredocs to help shorten the strings; or multilines with `\`. :)

carlosramireziii:
@lukeredpath Just wanted to check in on this one.. Would you be able to take care of the feedback from above? We'd love to get this one merged. Thanks!

nickcharlton:
@lukeredpath Could I get you to fix up the Hound warnings on this and rebase? Or alternatively, [enable contributor pushes](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)?

nickcharlton:
I'm going to close this as I've rebased myself and opened #857.

PR #583, opened 2016-05-16
Handle running install generator where no admin resources defined

lukeredpath:
This should resolve the issue I was seeing in #581


BenMorganIO:
This is great :+1: @lukeredpath

lukeredpath:
Possible dupe of #516 though I just realised.

nickcharlton:
Closing as this should be fixed with #713. Please open an issue if that's not the case!

PR #591, opened 2016-05-20
only specify ruby version for ci

BenMorganIO:
This PR removes the requirement for the developer to have to specify a specific version of ruby. If we have a hard setting of ruby in development, we could end up causing regressions in Rails projects. (Such as bumping ruby to 2.3 and using `&.` syntax, and unknowingly causing regressions for 2.2 and 2.1 users.)

Also, this is a first step in making sure that we can improve the ruby version coverage for our CI.


graysonwright:
This looks good. Once the build's passing, merge it on in.

PR #594, opened 2016-05-25
Fix I18n translation for attributes in has_many collections

hughfm:
# Problem:

Following on from changes in #492... 

When collection partials are rendered in the context of a `has_many` field, translations for the attribute labels are currently being looked up under the parent model's scope. For example, label for `shipped_at` attribute on `Order` model is looking in `en: helpers: label: customer: shipped_at:`, rather than `en: helpers: label: order: shipped_at:`
# Solution:

Use `collection_presenter.resource_name`, which is scoped to the field being rendered, instead of `resource_name`.

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -32,7 +32,7 @@ to display a collection of resources in an HTML table.
           collection_presenter.order_params_for(attr_name)
         )) do %>
         <%= t(
-          "helpers.label.#{resource_name}.#{attr_name}",
+          "helpers.label.#{collection_presenter.resource_name}.#{attr_name}",
           default: attr_name.to_s,
         ).titleize %>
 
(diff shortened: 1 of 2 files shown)


BenMorganIO:
Thanks @hughfm don't forget to fix the hound warnings :)

hughfm:
Thanks @BenMorganIO I'm still a bit new to this :) I've just fixed those up, should I just push an extra commit onto this branch?

BenMorganIO:
Cool. Can you rebase and squish them into a single commit? I'll get them merged on in after :)

hughfm:
thanks @BenMorganIO that's all done, but not sure why CI has failed now - let me know if there's an issue, cheers.

BenMorganIO:
@hughfm just an old cache that needs to get expired due to #591.

marcusg:
Hi guys, any updates here? It seems to be still an issue in current master. Thanks!

nickcharlton:
Hi @hughfm, @marcusg! This is looking good and has no conflicts so I'm going to go ahead and merge. Thanks!

PR #614, opened 2016-06-21
Add Bourbon as a dependency

tysongach:
Neat 1.8 was just released, which removed Bourbon as a dependency.

Up until now, Neat brought in Bourbon (which Administrate needs). But
that was fragile and now that Neat doesn't explicitly declare Bourbon as
a dependency, this nested dependency has fallen apart.


tysongach:
Looks like the appraisal for Bourbon 5 is breaking the build.

I’m unsure and confused why we have the appraisal; not sure it’s providing any value? Upgrading to Bourbon 5 will force some CSS refactoring and supporting both Bourbon 4 _and_ 5 at the same time will not be possible.

jalada:
My workaround for now is just to add Bourbon as a dependency in my Gemfile.

teoljungberg:
@tysongach I agree, I removed the Appraisal for Bourbon 5 in this PR.

Gemfile
```diff
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem "administrate-field-image"
+gem "bourbon"
```
tysongach: @teoljungberg Shall we lock this down to `4.x`, `, "~> 4.2"`? Once 5 ships, it will bust.

teoljungberg: Ah, good point! Done in 273373b8290c7780ed284eb65c7b207cb339ac8e

tysongach:
One more comment…

Otherwise 👍 👍 👍 👍 👍 👍 👍

gemfiles/sass_3_4.gemfile
```diff
@@ -2,9 +2,8 @@
 
 source "https://rubygems.org"
 
-ruby "2.2.3"
-
 gem "administrate-field-image"
+gem "bourbon"
```
teoljungberg: Gah, I'm sloppy today. Thanks for being my eyes, Tyson!

tysongach:
Thanks, @teoljungberg!

PR #631, opened 2016-07-11
Rails 5

lyonsv:
I tried to run the generator rails generate administrate:install, however the process failed as ActiveRecord no longer supports the column_types method on a model.

I've updated that to attribute_types.


BillyParadise:
As a friendly heads up, you might be reinventing the wheel

I'm running administrate successfully on rails 5 now using

gem 'administrate', github: 'greetpoint/administrate', branch: 'rails5'

nickcharlton:

Issue #563, opened 2016-04-22
Specify scope with model in admin page?


jdell64:
I have a user model with a `role enum: [:student, :teacher, :admin]`

In my administrate classroom dashboard, a classroom belongs to a teacher and has many students:

```
require "administrate/base_dashboard"

class ClassroomDashboard < Administrate::BaseDashboard

ATTRIBUTE_TYPES = {
    teacher: Field::BelongsTo.with_options(class_name: "User"),
    classroom_students: Field::HasMany,
    students: Field::HasMany.with_options(class_name: "User"),
    id: Field::Number,
    subject: Field::String,
    teacher_id: Field::Number,
    created_at: Field::DateTime,
    updated_at: Field::DateTime,
  }

....

FORM_ATTRIBUTES = [
    :subject,
    :teacher,
    :students,

  ]

  def display_resource(classroom)
    "#{classroom.subject} - #{classroom.teacher.first_name} #{classroom.teacher.last_name}"
  end
```

Everything works fine, but I'm wondering if I can make it to where my BelongsTo dropdown on the form can be limited using a scope I have in my User model `User.teachers`:

```
  # scopes
  scope :students, -> { where(role: User.roles[:student]) }
  scope :teachers, -> { where(role: User.roles[:teacher]) }
  scope :admins, -> { where(role: User.roles[:admin]) }
```

As it stands, the admin could make a student a teacher of a class (undesired). Whether or not it uses scopes, could I specify "This drop down contains model Users, where :type => :teacher"?

mattscilipoti:
I have a similar issue.  Leaving this comment to +1 and to subscribe to new comments.

adriandelarco:
I have the same issue.

carlosramireziii:
@jdell64 Could you solve this with a custom field? [Here's an example](https://github.com/thoughtbot/administrate/issues/131#issuecomment-283523064) which builds a custom field off of the existing `BelongsTo` in order to add an order to the collection. Could you use the same approach to add a scope?

Spharian:
I'm late to the game but I got a decent solution, for the ones still using this gem in 2019.

I needed to apply a scope on the `HasMany` field and the cleaner solution I found was to create a field, extend from the default `HasMany` field and allow a dynamic scope using fields `options`. All the code is commented below.

**app/fields/has_many_scoped_field.rb** (you have nothing to change in this file, just create it)
```
class HasManyScopedField < Administrate::Field::HasMany
  # apply a dynamic scope
[...]

PR #632, opened 2016-07-22
Allow specifying scope in with_options for a belongs_to field.

k3rni:
Addresses #563. 

It's now possible to do:

``` ruby
class WidgetDashboard < Administrate::Dashboard
  ATTRIBUTE_TYPES = [
    manager: Field::BelongsTo.with_options(class_name: "User", scope: :managers)
  ]
end
```

where the `User` model has a scope named `managers`. If not specified, defaults to `all`, so no change in behaviour.

### DIFF
--- a/lib/administrate/field/belongs_to.rb
+++ b/lib/administrate/field/belongs_to.rb
@@ -24,7 +24,11 @@ module Administrate
       private
 
       def candidate_resources
-        associated_class.all
+        associated_class.public_send candidate_scope
+      end
+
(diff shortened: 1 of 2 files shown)


seanpdoyle:
@k3rni @nickcharlton is this covered by https://github.com/thoughtbot/administrate/pull/1104?

k3rni:
@seanpdoyle indeed, and I think it's even more flexible than my attempt

Issue #625, opened 2016-07-04
I18N: Attributes labels translation issue


michaelkl:
Hello.

First of all, here's my `locale/ru.yml`:

```
ru:
  activerecord:
    attributes:
      user:
        email: Email
        password: Пароль
        password_confirmation: Пароль ещё раз
        name: Имя
        role: Роль
        provider: OAuth-провайдер
```

Now in my User dashboard on _new/edit_ page (one which got the edit form) labels of attributes are displayed using thier translation from locale/ru.yml, but on _index_ and _show_ pages they are don't!

I've dived in sources and found out that on index page the translation is taken from `helper.label` section, not from the `activerecord.attributes` one.

I think translation sources should be unified all over Administrate pages and using `activerecord.attribues` section looks more straightforward to me.

Thank you.

FerPerales:
@michaelkl I ran into this problem. So far, what I've done to avoid duplicated translations is this:

```ruby

common: &commons
  patient:
    address: 'Direccion'
    weight: 'Peso (kg)'
    height: 'Altura (cm)'
    allergies: 'Alergias'
    patient_antecedents: 'Antecedentes del paciente'
```

And then, include _commons in 

```ruby
es
  activerecord:
    models:
      attributes:
         <<: *commons

```

Hope it helps

michaelkl:
@FerPerales Good workaround until issue will have been fixed! Thank you!

fikrikarim:
Any update on this? Thanks :)

nickcharlton:
I'm closing this as the original issue is quite old and the translations have changed quite a bit since this was opened, please open another if this is still an issue.

kaelumania:
This issue still exists!

pablobm:
Yup; I can reproduce.

I agree this is odd. I wonder what the original intent was, using `helpers.label.RESOURCE_NAME.ATTR_NAME` instead of `activerecord.attributes.RESOURCE_NAME.ATTR_NAME`.

Hm, thinking a bit more, I can see an edge case. In forms, fields will tend to match ActiveRecord attributes. On the other hand, in lists/shows there's a higher chance of showing other things, such as constructed attributes (eg: `Customer#lifetime_value`). I'm not sure if these make sense under `activerecord.attributes.*` (honestly don't know; perhaps it makes sense).

[...]

PR #633, opened 2016-07-29
Better use of i18n for field names

k3rni:
Addresses #625 by adding the possibility to use `activerecord.attributes` locale namespace for attribute name translations. Unifies behavior across all places where fields were displayed - index, show and form pages.

As a special fallback, `has_many` fields also lookup their name with `_ids` suffix, if the short one is not present.  This was the original motivation for change, as their render naively used the field name and ignored all translations present.

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -29,10 +29,7 @@ to display a collection of resources in an HTML table.
         <%= link_to(params.merge(
           collection_presenter.order_params_for(attr_name)
         )) do %>
-        <%= t(
-          "helpers.label.#{resource_name}.#{attr_name}",
-          default: attr_name.to_s,
-        ).titleize %>
+            <%= field_label_text(resource_name, attr_name).titleize %>
 
             <% if collection_presenter.ordered_by?(attr_name) %>
               <span class="cell-label__sort-indicator cell-label__sort-indicator--<%= collection_presenter.ordered_html_class(attr_name) %>">
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -5,6 +5,18 @@ module Administrate
       render locals: locals, partial: field.to_partial_path
     end
 
+    def field_label_text(resource_name, field_name, alternate = nil)
+      resource_class = resource_name.to_s.classify.constantize
+
+      I18n.t(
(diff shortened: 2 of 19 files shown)


michaelkl:
There is also an issue with Field::HasMany's form view (_fields/has_many/_form.html.erb_) where labels not actually get translated.
It can be fixed by replacing
```
<%= f.label field.attribute_key, field.attribute %>
```
with
```
<%= f.label field.attribute_key, t(
        "helpers.label.#{resource_name}.#{field.attribute}",
        default: field.attribute.to_s,
      ).titleize %>
```
but this pull request seams to resolve this even better!

k3rni:
@michaelkl this was exactly my original motivation for writing this code, but as you can see, I didn't stop there.

michaelkl:
@k3rni good job! Hope it will be accepted soon. I need it in production ASAP. :-)

mcmire:
This makes a lot of sense, I think.

Having the same needs, I have two suggestions:

* What do we feel about translating the attribute name the responsibility of the Field class, since it already has access to the attribute name? I'm wondering if we can change the `name` method, specifically.
* What do we feel about allowing developers to use a different i18n namespace for Administrate field names than are used for ActionView form labels? I don't think it makes sense to always share translations between public-facing views and admin views.

app/views/administrate/application/_collection.html.erb
```diff
@@ -29,10 +29,7 @@ to display a collection of resources in an HTML table.
         <%= link_to(params.merge(
           collection_presenter.order_params_for(attr_name)
         )) do %>
-        <%= t(
-          "helpers.label.#{resource_name}.#{attr_name}",
-          default: attr_name.to_s,
-        ).titleize %>
+            <%= field_label_text(resource_name, attr_name).titleize %>
```
pustomytnyk: I think you should move `titleize` into `field_label_text`, because it can broke text retrieved from `human_attribute_name`

nickcharlton:
@k3rni Could I get you to rebase this?

jondkinney:
Nice work @k3rni. Need a hand rebasing? Happy to help.

k3rni:
Feel free to take over @jondkinney! I haven't used administrate in a while, so this is not high priority for me.

pablobm:
Closing due to lack of activity. Happy to see this reopen if someone wants to take over.

PR #634, opened 2016-07-29
Improve model plural on index pages

k3rni:
This uses the same method and motivation as activeadmin/activeadmin#2791. Using `2.1` for the count improves translation quality for many non-English locales.


nickcharlton:
Thanks! Merging.

Issue #441, opened 2016-01-28
Empty fields submit blank values (integers get zeros)


geoffharcourt:
I noticed when submitting empty fields that my models were failing validation. It appears that a blank value in the edit view leads to an empty string being submitted to the model. Should we `nil` these out?

geoffharcourt:
Update: submitting a blank value for an integer field results in a 0 when it should be `nil`.

graysonwright:
Good catch. It makes sense to me to `nil` those values out. If you want to take a crack at that, go ahead. Otherwise it's going to sit a bit - I'm a bit backed up with issues at the moment.

jakesen:
Ah, this is causing trouble for me to. I'm going to see if I can solve it.

mnort9:
Any update on this?

akestner:
I've also run into this issue recently and am curious if there are any updates.

jakesen:
Finally getting around to looking at this again. My use case is actually for a string field that needs to save null to the database if the submitted value is an empty string. I would be happy to do a PR that only takes care of number fields, but should the new behavior be configurable or should blank input values always be converted to `nil`?

mnort9:
I think empty strings should be converted to nil by default. Can't really think of a use case for why you'd want empty strings, but could be configurable in the field settings, something like:

 ```Field::String.with_options(allow_empty_strings: true)```

jakesen:
There are probably cases where a database has a string field where null is not allowed but empty strings are an acceptable value. If we change the current behavior of administrate it could break people's admin until they make schema changes or patch their admin dashboards.

mnort9:
Good point. Maybe allowing nil should be opt-in instead for now? Still seems like allowing empty strings is more of the edge case, but I understand it's a breaking change.

nickcharlton:
We are pre-1.0, so I'm currently happy with making a breaking change (we'll make a point of communicating this the best we can), so I'd not rule that out.

Adding an option first though might be a nice idea (which we can later flip) as then people can get the fix sooner.

lucasdavid:
@geoffharcourt did you manage to find a solution for this? I'm having this problem right now. :|

passuf:
I also run into the same problem, is there now an option as mentioned by @nickcharlton ?

mian-bilal:
@nickcharlton Any update on this?

@geoffharcourt Should we open this issue?

gjrevans:
[...]

PR #637, opened 2016-08-02
Normalize resource params

jakesen:
This provides a solution for issues like these: https://github.com/thoughtbot/administrate/issues/441

The solution is provided by allowing each field class to decide how to handle empty values via a normalize class method. A substitute value can be specified with an `empty_value` option when initializing a field using `with_options`.

This PR should probably have a little more test coverage, but first I want to see what adjustments might need to be made to the implementation to make it acceptable.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -100,11 +100,15 @@ module Administrate
     end
 
     def resource_params
-      params.require(resource_name).permit(*permitted_attributes)
+      normalize params.require(resource_name).permit(*permitted_attributes)
     end
 
     def permitted_attributes
       dashboard.permitted_attributes
+    end
+
+    def normalize(params)
+      params.map { |attr, value| [attr, dashboard.normalize(attr, value)] }.to_h
     end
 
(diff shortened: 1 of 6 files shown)


jakesen:
Hmmm. Hound isn't leaving any comments to indicate why it is failing.

carlosramireziii:
@jakesen Would you consider doing a narrower PR which just focused on the solution suggested in #441?

> It makes sense to me to nil those values out. If you want to take a crack at that, go ahead. Otherwise it's going to sit a bit - I'm a bit backed up with issues at the moment.

You would be able to add the necessary test coverage and perhaps that would make it an easier merge.

jakesen:
Yeah. I think that will work for what we're doing. I'll close this and do a new PR with the narrower scope.

akestner:
@jakesen You ever get around to a more narrowly-scoped PR?  I'd consider taking a stab at it if you haven't.

jakesen:
@akestner I haven't got around to it but I would be happy to collaborate on a good solution. You might want to weigh in on the conversation over at https://github.com/thoughtbot/administrate/issues/441 WRT which types of fields and what default behavior is desired.

PR #642, opened 2016-08-12
Add missing multiplier option for number fields.

### LINKED ISSUE #520: No multiplier option for Field::Number.
The [customizing dashboards documentation](https://github.com/thoughtbot/administrate/blob/e9a713439bbe6ac1bb0263872de85d9b91191a58/docs/customizing_dashboards.md) mentions using a `multiplier` to display a currency value stored in cents. As far as I can tell that's not actually possible.

I'm guessing it should be something close to the code below. Unsure if it should attempt any casting or type checking though.

``` ruby
module Administrate
  module Field
    class Number < Field::Base
      def to_s
        if data.nil?
          "-"
        else
          format_string % value
        end
      end

      protected

      def value
        data * options.fetch(:multiplier, 1)
      end
    end
  end
end
```

tristandunn:
Fixes #520.

This option is mentioned at [the bottom of the Customization Dashboards](https://github.com/thoughtbot/administrate/blob/c86a9f3677645b5b4b1e7d2bbc9f9b9da661a2ad/docs/customizing_dashboards.md) documentation, but is not currently available.

### DIFF
--- a/lib/administrate/field/number.rb
+++ b/lib/administrate/field/number.rb
@@ -7,7 +7,7 @@ module Administrate
         if data.nil?
           "-"
         else
-          format_string % data
+          format_string % value
         end
       end
 
@@ -23,6 +23,10 @@ module Administrate
 
       def decimals
         options.fetch(:decimals, 0)
+      end
+
+      def value
+        data * options.fetch(:multiplier, 1)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


kevinelliott:
What is the timeline for this PR? I would like to have this functionality, but it appears 4-5 months have passed with no activity here.

carlosramireziii:
@nickcharlton WDYT, is this one ready to go?

nickcharlton:
I like it. Going to merge.

@kevinelliott It'll go out in the next release, which should be Friday (2017-03-10).

PR #646, opened 2016-08-17
Update Field::BelongsTo to accept limit option

alzapiedi:
I am using Field::BelongsTo.with_options and noticed on the edit page that it will always populate the drop down field with every model from the whole DB.  In my application there are certain cases where we want the drop down to only display some models, so I made a slight tweak to the gem that I think could be helpful to others.

`Field::BelongsTo.with_options(class_name: "Model", limit: { foo: "bar" })`
will now populate the dropdown with the result of Model.where(foo: "bar")

### DIFF
--- a/lib/administrate/field/belongs_to.rb
+++ b/lib/administrate/field/belongs_to.rb
@@ -24,7 +24,11 @@ module Administrate
       private
 
       def candidate_resources
-        associated_class.all
+        if options[:limit]
+          associated_class.where(options[:limit])
+        else
+          associated_class.all
+        end
       end
 
       def display_candidate_resource(resource)


pustomytnyk:
@alzapiedi 1) I would expect the option named `limit` to be some number. I'd rather rename the option to `scope` etc. 2) I think it not flexible enough. What if I want to use existing scope from my model?

pablobm:
Closing due to lack of activity. Happy to see this reopen if someone wants to take over.

pablobm:
Related: `Field::BelongsTo` accepts an `:scope` option. Perhaps this would be useful here?

PR #649, opened 2016-08-31
bug:  rails g  administrate:install -  shouldn't fill in config/route…

roadt:
rails g  administrate:install -  shouldn't write garbage in config/routes.rb when  no valid resources


nickcharlton:
Closing as this should be fixed with #713. Please open an issue if that's not the case!

PR #663, opened 2016-10-05
permit! params

zshannon:
ActionPack 5 throws up `Attempting to generate a URL from non-sanitized request parameters! An attacker can inject malicious data into the generated URL, such as changing the host. Whitelist and sanitize passed parameters to be secure.` if we don't permit params prior to `url_for` (called via `link_to` here).

I just threw in `.permit!` here because:
- this is a show stopping error that breaks rendering `:index`, and
- "attackers" shouldn't have access to admin


purinkle:
@zshannon, is this a change you would still like to make? There seem to be changes in your pull request that are unrelated to your description. Would we be able to pull them into separate pull requests?

zshannon:
Ack, I don't know anymore it's been too long since a) I wrote this fix, and b) I stopped using administrate (went back to active admin). Closing

PR #667, opened 2016-10-12
Default all labels to human_attribute_name instead of raw attr_name

pcantrell:
Administrate is inconsistent about the default human-visible attribute name. It uses Rails’s attr name localization for form labels, but defaults to showing the raw Ruby attribute name in other views.

This PR brings the collection & show views into line with what the form shows.

There is also a collection/show discrepancy in how `titleize` gets applied — to the localized version in collection, only the default in show — but I’ve left that in place for now.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -110,6 +110,7 @@ module Administrate
     delegate :resource_class, :resource_name, :namespace, to: :resource_resolver
     helper_method :namespace
     helper_method :resource_name
+    helper_method :resource_class
 
     def resource_resolver
       @_resource_resolver ||=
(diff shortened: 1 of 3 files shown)


pcantrell:
CI is failing on appraisal:

    You have already activated bundler 1.12.5, but your Gemfile requires bundler 1.13.3

Don’t think that’s my changes…?

kuboon:
I agree with this pull request.
I don't like #492.

nickcharlton:
Hi @pcantrell! Thanks for your PR!

I think I prefer how this standardises things. It's been a while, but would you be able to rebase on the current `master` (to lose those CI failures)?

Are we also able to test this change?

nickcharlton:
Closing in favour of #1064, which I hope we'll merge soon.

Issue #647, opened 2016-08-19
with_options usage documentation


giuseb:
The "Customizing Dashboard" documentation page says:

> Each of the Field types take a different set of options, which are specified through the `.with_options` class method.

I seem unable to figure out what options are available for each field type. Any pointers? Reference documentation and/or relevant source code to inspect?

I have detailed my specific problem [here](http://stackoverflow.com/questions/39052135/customizing-fields-with-the-administrate-gem-for-rails).

erose:
+1, I'm also looking for this documentation.

ianagne:
@giuseb I just responded to your StackOverflow post.  It doesn't use the `.with_options` class method, but it does expose another area where we could expand the documentation (overwriting the BaseDashboard class to customize appearance).

ianagne:
@giuseb @erose If you're still looking for documentation on .with_options, check out the PR referenced right above this comment.  I added all of the currently active options that I could find in the source.  If you are using an option that isn't listed, let me know what it is and what it does and I'll add it to the PR.

erose:
@ianagne: Thank you!

ianagne:
@JoelQ With #670 merged, it should be safe to close this issue.

Issue #292, opened 2015-12-04
with_options(title: '...') not setting field titles in dashboards


alexkrolick:
It seems like the default stringifier cuts off anything that ends with `_id` when it generates the labels for fields. I tried making a custom title for the fields like this:

``` ruby
thing_id: Field::String.with_options(title: "Thing ID")
```

But the labels throughout the dashboard are still the same.

graysonwright:
@whokilledtheelectricmonk, thanks for pointing this out.

We'd absolutely like to support custom labels, but this feature isn't in place yet. I'm going to assign this to the `0.1.3` milestone, and we'll take a look at it as soon as we've ironed out more pressing bug fixes.

janklimo:
As a temporary solution, see how the title is being set [right now](https://github.com/thoughtbot/administrate/blob/master/app/views/administrate/application/show.html.erb#L36). For instance, I wanted to rename `item_total` attribute of `Order` to `Subtotal`, so I have the following in my `locales/en.yml`:

``` ruby
en:
  helpers:
    label:
      order:
        item_total: Subtotal
```

pablobm:
I wasn't aware of that `title` option. I don't think it exists anymore? I can't find it in the code.

In any case @janklimo's approach is preferred: use the locale file to provide names for fields.

PR #670, opened 2016-10-14
Update Field .with_options Documentation

ianagne:
Per issue #647, documentation does not contain possible options to set
for fields via .with_options.

* Add possible options for Field::HasMany, Field::Number, Field::Select,
Field::String, and Field::Text
* Remove :multiplier and :title options from Field::Number.with_options
example, as they are not yet implemented (see issue #292)

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -62,19 +62,57 @@ specify, including:
 - `Field::Select`
 - `Field::String`
 
+## Customizing Fields
+
+### Setting Options
+
 Each of the `Field` types take a different set of options,
-which are specified through the `.with_options` class method.
-For example, you might use the following to display currency,
-if the value is stored by the number of cents:
+which are specified through the `.with_options` class method:
+
+**Field::HasMany**
+
+`:limit` - Set the number of resources to display in the show view. Default is
+5.
+
+**Field::Number**
+
+`:decimals` - Set the number of decimals to display. Defaults to 0.
+
+`:prefix` - Prefixes the number with a string. Defaults to "".
+
+For example, you might use the following to display U.S. currency:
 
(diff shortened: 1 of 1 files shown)


JoelQ:
@ianagne I like this! Thanks for contributing! 👍 

I'm going to merge this once we get `master` green again

JoelQ:
I rebased this and merged it into master as 6ae59506b5e75af02c9875e0512b38f34c6ecead. Thanks for contributing!

j-dexx:
@ianagne @JoelQ  I'm not sure this documentation is correct with regards to being searchable.

[String](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/string.rb#L6)
[Select](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/select.rb#L6)
[Text](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/string.rb#L6)

Options is a hash passed to the field on initializing, every searchable thing I can find is a class method.

This could be changed fairly easily by switching [this](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/search.rb#L33) to initialize the field, and switching them all to instance methods, which get from the options hash with defaults.

Happy to submit a PR, but figured I'd check that's what you want @JoelQ first.

ianagne:
@j-dexx The option for searchable is in the [Deferred](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/deferred.rb#L24) class.

j-dexx:
Ah, cool. Thanks for setting me straight @ianagne

ianagne:
Thanks for using that trick to highlight lines on GitHub, @j-dexx! I didn't know that was a thing I could do.

PR #671, opened 2016-10-17
Fix the specs with the Rails 5 branch and the clear search button

danbee:


danbee:
This is failing on CI because of [this issue](https://github.com/thoughtbot/appraisal/issues/114) with Appraisal.

PR #673, opened 2016-10-20
Engine requires bourbon, fixes #615

### LINKED ISSUE #615: Dependency on bourbon not listed
After upgrading to administrate 0.2.2 (from 0.2.0) my app started throwing:

```
File to import not found or unreadable: bourbon.
[snip]
./app/views/layouts/application.html.haml:8:in `_app_views_layouts_application_html_haml___2930206960703904785_47038691733580'
```

This was fixed by adding the 'bourbon' gem to my Gemfile, but I feel it should properly be included in administrate's dependencies instead.

kpheasey:


JoelQ:
I'd love to get some testing around this. Currently the test suite is green (if you rebase the latest `master`) without this change. Can we make the `sample_app` fail the same way as everyone else and then this fix makes it go green?

.gitignore
```diff
@@ -14,3 +14,4 @@
 /tags
 /tmp/*
 /spec/example_app/tmp/*
+.idea
```
kpheasey: Never knew there was global .gitignore, thanks for the knowledge!

I'll revert it

kpheasey:
Rebased with master and removed the .gitignore stuff.

Not sure how to test the require statement in the sample app.  Since the sample app uses the engine's gemfile bourbon gets included.  When bourbon is not in the gemfile, it is not required by the engine.

JoelQ:
Ah, I see what you're saying.

I think we can add `require: false` to bourbon in the `Gemfile`. This will prevent it from being auto-required. Doing this on my machine caused all the feature specs to fail with:

```
ActionView::Template::Error:
        File to import not found or unreadable: bourbon.
```

Adding in your require to `engine.rb` then caused the test suite to go green.

Alternatively, we could try to go down the path of having a separate `Gemfile` for the sample app.

Thoughts?

cc @danbee @BenMorganIO

kpheasey:
I believe dependencies should be defined in the gemspec, http://guides.rubyonrails.org/engines.html#other-gem-dependencies <http://guides.rubyonrails.org/engines.html#other-gem-dependencies>

If you define a dependency in the gemspec, it will need to be explicitly required upon initialization.  This would avoid further confusion and not require testing.

BenMorganIO:
@kpheasey could you squash your three commits? I think they could all be just one in this instance.

JoelQ:
> I believe dependencies should be defined in the gemspec

Good point. It looks like we have some dependencies just in the Gemfile, some just in the gemspec, and some in both. Not sure why that happened.

Do you want turn this PR into a migration from using the Gemfile to using the gemspec? Doing that should both fix the bourbon problem and fail the specs when not requiring he proper files.

kpheasey:
I've completely screwed this up trying to squash, I'm going to open a new pull request with a separate branch, sorry about that.

JoelQ:
No worries @kpheasey 😃

PR #679, opened 2016-10-27
Update customizing_dashboards.md

abunsen:
Added a little section to docs that covers how to add a select box because I spent 10 min searching the internet when I should have spent 30 seconds. Now someone won't have to, hopefully.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -76,6 +76,14 @@ if the value is stored by the number of cents:
   )
 ```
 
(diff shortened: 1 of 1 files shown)


JoelQ:
@abunsen thanks for contributing! I just merged #670 which adds documentation of how to configure the options and which options are available for each dashboard. Does you think that fulfills the purpose? Or do you think there's more to add?

nickcharlton:
I'm going to close this as it's a little old now and we've already got a similar (but more detailed) version.

Please open again if you think our docs could be better!

PR #683, opened 2016-10-28
administrate engine requires bourbon. fixes #615

### LINKED ISSUE #615: Dependency on bourbon not listed

kpheasey:
Continued from https://github.com/thoughtbot/administrate/pull/673

### DIFF
--- a/lib/administrate/engine.rb
+++ b/lib/administrate/engine.rb
@@ -1,3 +1,4 @@
+require "bourbon"
 require "datetime_picker_rails"
 require "jquery-rails"
 require "kaminari"


kpheasey:
> Good point. It looks like we have some dependencies just in the Gemfile, some just in the gemspec, and some in both. Not sure why that happened.
> 
> Do you want turn this PR into a migration from using the Gemfile to using the gemspec? Doing that should both fix the bourbon problem and fail the specs when not requiring he proper files.

@JoelQ I don't know enough about administrate to turn the gemfile into a gemspec.  The dependencies need to be explicit version ranges.  Additionally, I wouldn't know what to do about the staging/production requirements.

nickcharlton:
Thanks for this! I'm electing to pick just the `bourbon` for now, we'll target others as they happen.

jipiboily:
Seems like it's still an issue, right? Or was just never shipped?

klaseskilson:
@jipiboily This fix hasn't been released yet. It will once `> 0.3.0` is released. If you need the fix earlier, point your Gemfile entry towards Github, like so: 

```ruby
gem 'administrate', git: 'https://github.com/thoughtbot/administrate.git'
```

jipiboily:
Thanks for the confirmation. That's what I just did a couple of minutes ago. It felt weird that it was not released 2 months later, so I wanted to double check :) Thanks!

pedantic-git:
@kpheasey @nickcharlton - there are issues with moving to Bourbon 5 because it no longer ships with an Engine and instead has a generator to add the SCSS files into your asset pipeline.

Which means that if you're using administrate in a project that doesn't use Bourbon (and therefore has not run the generator) you get "File to import not found or unreadable: bourbon".

I think either Administrate needs to ship with the Bourbon 5 SCSS already in its vendor/stylesheets or it needs to remove its dependency on Bourbon, which shouldn't be too hard?

Issue #438, opened 2016-01-26
Can fields have reference to resource instead of just the data?


jeffreyguenther:
For example, I'd like to add a password field.

``` ruby
class PasswordField < Administrate::Field::Base
  def to_s
    if data.present?
        "*" * data.size
    else
        "Not set"
    end
  end
end
```

This makes sense until you realize that `data` is referring to the string provided by `password`. If you're using a library like `clearance` you don't have a password attribute on your user model. You have `encrypted_password`. I want to detect if `encrypted_password` has any content and format the string accordingly.

straight-shoota:
I'd definitly like this as well and I don't think, there would be any harm in that. Every instance of a field class is already specific to a single resource by holding its distinct attribute value. Why not make that resource directly available from the instance?
There are a lot of good reasons for access to the resource or other attributes.
The autosafe feature for [SimpleMDE](https://github.com/NextStepWebs/simplemde-markdown-editor) for example needs a unique identifier, for which I'd want to use the resource class and record id. In the template, there is already access to the record through the FormBuilder (`f.object`). But this is only available in `form` views, not in index or show. Genereally it would be better to put domain logic into the field class, not the template.
It should be quite simple to extend `Page::Base#attribute_field` and `Field::Base#initialize` to deliver the current resource to the field instance.

mcmire:
Also see #267.

nickcharlton:
I _think_ this issue was solved with #706, so I'm going to close this. But please open a new one if that's not the case.

Issue #267, opened 2015-11-22
Model object in the field


Tensho:
I've found that resource (record) disappears, while administrate converts show page attributes to the fields.

```
# administrate-0.1.1/lib/administrate/page/base.rb
…

protected

def attribute_field(dashboard, resource, attribute_name, page)
  value = get_attribute_value(resource, attribute_name)

  dashboard.
    attribute_types[attribute_name].
    new(attribute_name, value, page)
end
```

In some cases it’s really necessary to take reference to the object, some helpers require model. For example, Administrate::Page::Show instance looses such public reference during the rendering. That’s why I have to use some hack, when I want to provide Refile attachment url (which requires record object) during the template rendering:

```
# app/views/fields/file_field/_show.html.erb
<% record = assigns['_requested_resource'] %>
<%= image_tag attachment_url(record, field.attribute, :fill, 300, 300, format: "jpg") %>
```

What do you think about the record exposing, making it public via field object? Does it violate SRP (or any other good principles) of the field class design?

graysonwright:
@Tensho - let me make sure I understand:

You're creating a custom field type called `FileField`, and the `show.html.erb` partial for the field needs a reference to the resource itself, and not just the attribute value. Is that right?

If that's the case, there's a workaround:

First, run `rails generate administrate:views:show`. This will generate a file at `app/views/admin/application/show.html.erb`.

On line 35 of the generated file, change the line that looks like:

```eruby
<%= render_field attribute %>
```

to:

```eruby
<%= render_field attribute, resource: page.resource %>
```

Now, inside of `app/views/fields/file_field/_show.html.erb`, you'll have access to the `resource` variable, which is the ActiveRecord object.

Let me know if that works for you, or if I made a mistake somewhere. Cheers!

Tensho:
Yeah, you're right. Actually, I've already found this solution by myself for `index` partial:
```
# /app/helpers/administrate/application_helper.rb
module Administrate
  module ApplicationHelper
    def render_field(field, resource = nil, locals = {})
      locals.merge!(resource: resource, field: field)
      render locals: locals, partial: field.to_partial_path
    end
  end
end

# /app/views/administrate/application/_collection.html.erb
<td class="cell-data cell-data--<%= attribute.html_class %>">
  <%= render_field attribute, resource %>
</td>
```
My main question is related to design concerns. Are you going to hide resource/record/model further and substitute it with the field object during the rendering process intentionally?
[...]

PR #706, opened 2016-12-12
Add support for referencing the resource from fields.

dts:
First off - Administrate solves this problem really nicely!  I am so excited someone finally did it right!

Second, my specific use case motivating this PR is having a Refile image field, and Refile prefers access to the resource & attribute name, as opposed to just the data.  It seemed like a minor change based on the code, but I'm not sure where else in the code I should edit things (`.new` isn't exactly that unique...).

Others apparently wanted this as well:

https://github.com/thoughtbot/administrate/issues/438

Lastly, I'm afraid I couldn't figure out how to get tests to even run on my system, and ran out of time to figure it out.  Things work fine for me in my rails 5 test project, fwiw.  Since there wasn't much feedback on the above issue I figured I'd just shoot in the dark and hope I don't hit anyone.  Thanks/sorry!


dts:
I adjusted the tests to match the new method signature.

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -11,7 +11,7 @@
     it "returns a partial based on the page being rendered" do
       page = :show
       owner = double
-      field = Administrate::Field::BelongsTo.new(:owner, owner, page)
+      field = Administrate::Field::BelongsTo.new(Class.new.new,:owner, owner, page)
```
mcmire: Why are you having to say `Class.new.new`? This seems a little strange.

mcmire:
Also see #267.

spec/lib/fields/belongs_to_spec.rb
dts: In other places I inserted nil.  It doesn't really matter what goes in that slot, because no existing form field types actually use it.  I'm 100% open to other architectures, method signatures, whatever, I just want to be able to get to the resource from the fields.

dts:
Good reference, I didn't see that before just jumping in!

nickcharlton:
I've had a quick look at both this and #788. It's been a little while, the other implementation is a little simpler and so the other is looking easier to merge now. Thanks for your hard work on this!

PR #712, opened 2016-12-16
Docker-based development environment

dts:
I don't like to wrangle the dependencies of any given applications on my computer, so I use Docker-compose to create ad-hoc development environments as I need them.  I added a simple one so that I could run the tests, and I figured I'd post it back in case it's useful.  Documentation is in CONTRIBUTING.rb on how to run tests and work on the code with an more straightforward setup.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,19 @@ This is a time for discussion and improvements,
 and making the necessary changes will be required before we can
 merge the contribution.
 
+## Docker Development environment:
+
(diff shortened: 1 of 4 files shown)


nickcharlton:
Thanks for this PR!

I've given it a bunch of thought, but I'd like to avoid adding complexity where possible. I personally don't really use it (except in a few circumstances) and I feel that it's unnecessary except with complex dependency trees (be it other gems or services or whatever), that we don't have here.

Issue #502, opened 2016-03-06
Undefined method 'camelize' for nil:NilClass


coisnepe:
Hey,

I tried adding administrate to my project but I kept getting this error. Just to make sure it wasn't because of the project, I started a fresh new app, only added administrate to the Gemfile, bundled and ran `rails generate administrate:install`, to no avail: keep getting the same error.

```
rails generate administrate:install
Running via Spring preloader in process 30057
       route  namespace :admin do

    root to: "#index"
  end
      create  app/controllers/admin/application_controller.rb
      create  app/dashboards/_dashboard.rb
/Users/.../.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-4.2.5.2/lib/rails/generators/named_base.rb:102:in `block in class_name': undefined method `camelize' for nil:NilClass (NoMethodError)
    from /Users/.../.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-4.2.5.2/lib/rails/generators/named_base.rb:102:in `map!'
    from /Users/.../.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/railties-4.2.5.2/lib/rails/generators/named_base.rb:102:in `class_name'
    from (erb):3:in `template'
```

Edit: I had installed administrate in another project earlier that day without running into any issues. Not sure what happened in between... I tried `gem uninstall administrate`, created a new app, added `gem "administrate", "~> 0.1.4"`, bundled etc, but no luck...

Edit2: Tried bundling with Ruby 2.3 / 2.2.4 / 2.2.3 and Rails 4.2.5 / 4.2.5.2 (every possible combination) , still no luck

eirvandelden:
+1 I have the exact same issue for a fresh application

Amandeepsinghghai:
+1 have the same issue here as well

neves:
+1 same here

neves:
The problem happens when it tries to install without any model.
It generates the follow route:
```ruby
Rails.application.routes.draw do
  namespace :admin do

    root to: "#index"
  end
end
```

The correct behaviour would be to alert the missing model.
But a quick fix is to replace this line with this code:
`~/.rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/administrate-0.1.4/lib/generators/administrate/install/install_generator.rb:34`

```ruby
dashboard_resources.map(&:to_s).reject(&:empty?).map(&:singularize)
```

[...]

Issue #613, opened 2016-06-16
rails g install error


civcastillo:
I haven't been able to install administrate 

Already took a look at #516 but still no luck

---

Running via Spring preloader in process 8585
   identical  app/controllers/admin/application_controller.rb
   identical  app/dashboards/user_dashboard.rb
   identical  app/controllers/admin/users_controller.rb
/.gem/ruby/2.3.0/gems/activerecord-4.2.6/lib/active_record/reflection.rb:883:in `derive_class_name': undefined method`class_name' for nil:NilClass (NoMethodError)
Did you mean?  class_eval
    from /.gem/ruby/2.3.0/gems/activerecord-4.2.6/lib/active_record/reflection.rb:157:in `class_name'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:111:in`relationship_options_string'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:98:in `association_type'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:61:in`redundant_attributes_for'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:56:in `block in redundant_attributes'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:55:in`each'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:55:in `flat_map'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:55:in`redundant_attributes'
    from /.gem/ruby/2.3.0/gems/administrate-0.2.2/lib/generators/administrate/dashboard/dashboard_generator.rb:47:in `attributes'
    from (erb):11:in`template'
    from /.rubies/ruby-2.3.0/lib/ruby/2.3.0/erb.rb:864:in `eval'
    from /.rubies/ruby-2.3.0/lib/ruby/2.3.0/erb.rb:864:in`result'
[...]

Issue #601, opened 2016-06-06
rails g administrate:install error


tengju:
Hi its my first time trying to use this gem but when i genereate it this error keeps popping up.

/gems/ruby-2.3.1@global/gems/railties-4.2.6/lib/rails/generators/named_base.rb:102:in `block in class_name': undefined method`camelize' for nil:NilClass (NoMethodError)

How can i fix this? or what is exactly the problem?

p.s im using rails 4.2.6 with ruby 2.3.1

tengju:
fixed by making models first

see: https://github.com/thoughtbot/administrate/issues/516

nickcharlton:
Closing as this should be fixed with #713. Please open an issue if that's not the case!

PR #555, opened 2016-04-20
Avoid generating invalid root route when there's no model

yujinakayama:
This fixes #502, and fixes #516.

When there's no model, `rails generate administrate:install` invokes `administrate:routes` generator first, and it generates the following invalid routing:

``` ruby
Rails.application.routes.draw do
  namespace :admin do

    root to: "#index"
  end

  ...
```

... and that causes the error in `administrate:dashboard` generator.

PR #713, opened 2016-12-16
Avoid writing to routes.rb when no models exist.

nickcharlton:
This closes: #502, #516, #555, #649, #583, #601, #613.

### DIFF
--- a/lib/generators/administrate/routes/routes_generator.rb
+++ b/lib/generators/administrate/routes/routes_generator.rb
@@ -8,7 +8,7 @@ module Administrate
       source_root File.expand_path("../templates", __FILE__)
 
       def insert_dashboard_routes
-        unless File.read(rails_routes_file_path).include?(dashboard_routes)
+        if should_route_dashboard?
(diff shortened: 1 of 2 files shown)


lib/generators/administrate/routes/routes_generator.rb
```diff
@@ -8,7 +8,7 @@ class RoutesGenerator < Rails::Generators::Base
       source_root File.expand_path("../templates", __FILE__)
 
       def insert_dashboard_routes
-        unless File.read(rails_routes_file_path).include?(dashboard_routes)
+        unless routes_includes_resources? || valid_dashboard_models.empty?
```
mike-burns: Do you mind moving these conditions into a `should_route_dashboard?` method? I struggle to read `unless`, especially with a longer logical expression.

nickcharlton: Yes, I like this idea! Done in a `fixup` if you'd take a look!

mike-burns:
LGTM.

PR #719, opened 2016-12-27
Fix format for datetimepicker

straight-shoota:
`format` option uses the same syntax as PHP date function.
http://xdsoft.net/jqplugins/datetimepicker/#format

### DIFF
--- a/app/assets/javascripts/administrate/components/date_time_picker.js
+++ b/app/assets/javascripts/administrate/components/date_time_picker.js
@@ -1,6 +1,6 @@
 $(function () {
   $(".datetimepicker").datetimepicker({
     debug: false,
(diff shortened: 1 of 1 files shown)


nickcharlton:
Hi @straight-shoota!

Just so I understand your PR fully:

1. Was this not working beforehand?
2. Is there a way we can go about testing this change?
3. Do you feel like the PHP style syntax is clearer?

Thanks!

straight-shoota:
I'm sorry, nevermind. I've just realised that the problem I was facing occured because a [different datetime_picker js-library](http://xdsoft.net/jqplugins/datetimepicker/) had been included in my test. With similar syntax but different format option from [raysonwright/datetime_picker_rails](https://github.com/graysonwright/datetime_picker_rails).

PR #721, opened 2016-12-30
Use `count(:all)` for associations in HasMany field

klaseskilson:
Problem: Calling `count` on an association can cause invalid SQL queries to be created where the `SELECT COUNT(a, b, c)` function receives multiple  columns. This will cause a `StatementInvalid` exception later on.

Solution: Use `count(:all)`, which generates a `SELECT COUNT(*)...` query independently of the association. The test's `MockRelation` helper is also patched to make it follow how the actual association works.

Background: This appears to be an known problem, see https://github.com/rails/rails/issues/15138 and https://github.com/rails/rails/issues/13648.

Hit me with your feedback! 👍

### DIFF
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -30,7 +30,7 @@ from the associated resource class's dashboard.
       <%= t(
         'administrate.fields.has_many.more',
         count: field.limit,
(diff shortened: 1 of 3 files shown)


nickcharlton:
Looks good to me!

PR #727, opened 2017-01-06
Add link to show resource from collection table

odlp:
Adds a link to show the resource within each cell of the collection table, as an alternative to the `data-url` on the `tr`. The link is styled with the default text colour so there's no visual change.

This allows a resource to be shown without javascript enabled (and also facilitates faster tests using RackTest).

On my machine running the same test comparing:

- `click_show_link_for` (+ no JS)
- `click_row_for` (+ with JS) 

is a lot faster, as expected:

| Run     | no JS  | JS     |
|---------|--------|--------|
| 1       | 0.88   | 2.04   |
| 2       | 0.85   | 1.93   |
| 3       | 0.85   | 1.96   |
| 4       | 0.84   | 2.02   |
| 5       | 0.82   | 2.04   |
| *Avg*     | *0.85* | *2.00* |
| *Std.Dev* | *0.02* | *0.05* |

(in seconds, running each test independently, excluding file load times)

### DIFF
--- a/spec/features/index_page_spec.rb
+++ b/spec/features/index_page_spec.rb
@@ -22,6 +22,15 @@ describe "customer index page" do
     expect(page).to have_content(customer.email)
   end
 
+  it "links to the customer show page without javascript", js: false do
+    customer = create(:customer)
+
+    visit admin_customers_path
+    click_show_link_for(customer)
+
+    expect(page).to have_header(displayed(customer))
+  end
+
   it "links to the edit page" do
     customer = create(:customer)
 
--- a/app/assets/stylesheets/administrate/components/_table.scss
+++ b/app/assets/stylesheets/administrate/components/_table.scss
@@ -21,3 +21,11 @@
 .table__action--destroy {
   color: $light-red;
 }
+
+.table__link-plain {
+  color: inherit;
+
+  &:hover {
+    color: inherit;
+  }
+}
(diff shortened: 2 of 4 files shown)


spec/features/index_page_spec.rb
```diff
@@ -22,6 +22,15 @@
     expect(page).to have_content(customer.email)
   end
 
+  it "links to the customer show page without javascript", js: false do
```
nickcharlton: I wonder if this means that now we'll be able to disable JavaScript by default and _enable_ JS instead?

nickcharlton:
I like it! :shipit:

spec/features/index_page_spec.rb
odlp: I think the default driver is currently RackTest for most of the feature specs.

11 / 56 feature specs enable javascript with the `:js` flag:

```
bundle exec rspec spec/features
# 56 examples, 0 failures

bundle exec rspec spec/features --tag js
# 11 examples, 0 failures
```

We could convert these to use the hyperlink, but it's quite handy to have the coverage & confidence the `data-url` / JS option still works.

What do you think @nickcharlton?

odlp: Oh sorry... re-reading your comment. Do you mean invert the test names? So `links to the customer show page` is the vanilla version, then `links to the customer show page with javascript` or similar?

PR #729, opened 2017-01-12
Allow primary_key option on relationships instead of hard coding id

npezza93:
Currently BelongsTo and HasMany default to using #id of the
related records. This will allow you to pass the option of primary_key
and use that fields' value in the form.


BenMorganIO:
This is awesome. Thank-you <3

PR #730, opened 2017-01-13
Search Bar Display Logic

akestner:
Currently the search bar is shown regardless of whether or not a resource contains searchable attributes.  This PR adds a method to the `Administrate::ApplicationController` that checks whether any attributes are searchable and passes a new local variable (`show_search_bar`) to the view which is then used to determine whether or not to render the search partial.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -125,17 +125,9 @@ module Administrate
     end
 
     def show_search_bar?
-      show_search_bar = false
       dashboard.attribute_types_for(
         dashboard.collection_attributes
-      ).each do |name, attribute|
-        if attribute.searchable?
-          show_search_bar = true
-          break
-        end
-      end
-
-      show_search_bar
+      ).any? { |_name, attribute| attribute.searchable? }
     end
   end
 end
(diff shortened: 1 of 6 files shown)


app/controllers/administrate/application_controller.rb
```diff
@@ -122,5 +123,19 @@ def translate_with_resource(key)
...
+        dashboard.collection_attributes
+      ).each do |name, attribute|
+        if attribute.searchable?
+          show_search_bar = true
+          break
+        end
+      end
+
+      show_search_bar
+    end
```
BenMorganIO: Refactor suggestion:

```ruby
dashboard.attribute_types_for(
).any? do |_name, attribute|
  attribute.searchable?
end
```

BenMorganIO:
I really like this PR. Left some comments. Great work @akestner :+1:

app/controllers/administrate/application_controller.rb
akestner: Nice, I'll update.

akestner:
Thanks @BenMorganIO!  I initially wrote it for a client and thought it seemed like a good change to submit back upstream.

I've updated the `show_search_bar?` method to make use of `.any?` like you suggested, though I left out the documentation as I didn't see many other documented methods in the codebase.

PR #733, opened 2017-01-16
Update administrate-field-image to release

nickcharlton:


deivid-rodriguez:
@nickcharlton What if you update bundler to `1.13.7` in `circle.yml` and `Gemfile.lock`? The error message seems like some bundler version conflict, no?

nickcharlton:
@deivid-rodriguez Yep! That's the solution. I've just run out of time for now.

It'd be nice to not have it break on each new release of `bundler`, but that's a problem for another time.

nickcharlton:
And fixed. It was the Circle cache, but updating bundler was a good idea anyway.

deivid-rodriguez:
:tada:

PR #736, opened 2017-01-19
Paginate has_many show view

acrogenesis:
This PR paginates the has_many show and takes into account that the show view can have several has_many fields.

Here you can see how the Attendee model is on page 1 and the User model is on page 2
screen shot 2017-01-19 at 10 21 19 am

PS. I blurred some personal information.

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -34,7 +34,7 @@ module Administrate
     end
 
     def sanitized_order_params
-      params.permit(:search, :id, :order, :page, :per_page, :direction)
+      params.permit(:search, :id, :order, :page, :per_page, :direction, :orders)
     end
 
     def clear_search_params
--- a/spec/features/show_page_spec.rb
+++ b/spec/features/show_page_spec.rb
@@ -4,18 +4,19 @@ RSpec.describe "customer show page" do
 
   it "displays the customers orders paginated" do
     customer = create(:customer)
-    orders = []
-    10.times do
-      orders << create(:order, customer: customer)
-    end
+    first_order, _a, _b, last_order = Array.new(4) { create(:order, customer: customer) }
 
     visit admin_customer_path(customer)
+    within('.attribute-data--has-many') do
+      expect(page).to have_content(first_order.id)
+      expect(page).to have_no_content(last_order.id)
+    end
 
-    expect(page).to have_content(orders[3].id)
-    expect(page).to have_no_content(orders[8].id)
     click_on("Next ›")
-    expect(page).to have_content(orders[8].id)
-    expect(page).to have_no_content(orders[3].id)
+    within('.attribute-data--has-many') do
+      expect(page).to have_content(last_order.id)
+      expect(page).to have_no_content(first_order.id)
+    end
   end
 
   it "displays the customer's title attribute as the header" do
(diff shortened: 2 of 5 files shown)


BenMorganIO:
Woot! Awesome! Thanks @acrogenesis!

BenMorganIO:
Could you add a simple feature spec please?

acrogenesis:
@BenMorganIO I've added the test, I'm not entirely sure if it's the best way to make it. What do you think?

spec/features/show_page_spec.rb
```diff
@@ -1,6 +1,23 @@
 require "rails_helper"
 
 RSpec.describe "customer show page" do
+
+  it "displays the customers orders paginated" do
+    customer = create(:customer)
+    orders = []
+    10.times do
```
BenMorganIO: Refactor suggestion:

```ruby
orders = 10.times.each_with_object([]) do |_index, arr|
  arr << create(:order, customer: customer)
end
```

BenMorganIO: I don't think we need to do this 10 times. That could slow the tests down a bit. You could do 2 orders and set `Kaminari.per_page = 2`. Something like:

```ruby
first_order, second_order = Array.new(2) { create :order }

visit admin_customer_path customer

expect(page).to have_content(first_order.id)
expect(page).to have_no_content(second_order.id)
click_on 'Next >'
expect(page).to have_content(second_order.id)
expect(page).to have_no_content(first_order.id)
```

acrogenesis:
Hey @BenMorganIO I've done some improvements on the tests, sorry for taking so long I've had a lot of work.

nickcharlton:
I like it! Going to squash and merge.

acrogenesis:
Thanks!

BenMorganIO:
No, thank you @acrogenesis!

app/helpers/administrate/application_helper.rb
```diff
@@ -34,7 +34,7 @@ def svg_tag(asset, svg_id, options = {})
     end
 
     def sanitized_order_params
-      params.permit(:search, :id, :order, :page, :per_page, :direction)
+      params.permit(:search, :id, :order, :page, :per_page, :direction, :orders)
```
pablobm: Sorry, a bit late to the party but: was this change intended? Looks to me like it comes from "orders" being one of the examples, and it slipped through.

acrogenesis: 🤔 I think you are right. I'll runt the tests without that and report back.

pablobm: OK, don't spend too long on this though! Sorry, I was looking at something else and this confused me and thought of commenting here.

acrogenesis: It fails:
```
1) customer show page displays the customers orders paginated when the second page is specified displays the second page of results
     Failure/Error: params.permit(:search, :id, :order, :page, :per_page, :direction)
     
     ActionView::Template::Error:
       found unpermitted parameter: :orders
```

pablobm: Thank you for looking into that 👍 

After your finding, I dug a bit deeper: it's a bug! `orders` is only relevant in that very specific example, where a "customer" has many "orders". The `orders` param tells the page number for the `orders` has_many list.

If I try to paginate through any other relationship, it breaks because the param has a different name.

No matter, I'll fix this.

PR #737, opened 2017-01-20
Update the Prototype App to Rails 5

nickcharlton:
This intends to update the prototype app to use Rails 5 (the one we deploy), the process of this will allow us to upgrade a few gems which are broken with Ruby 2.4.0, which in turn will allow us to start developing on that.

* [x] Add `rails-controller-testing` to the main `Gemfile`.
* [ ] Rails 4.2 tests fail with the `link_to` helper.


jcmorrow:
@nickcharlton do you understand how to fix the `link_to` failures here? I've dug in a little bit and I've drawn some conclusions but I don't have a fix. 😕

jcmorrow:
My research:

minimal reproduction spec:

```
  it "minimal reproduction case" do
    product = create(:product)

    expect{ url_for([:admin, product]) }.not_to raise_error
  end
```

In `5.0` `method(:url_for).source_location` returns 
```
.../rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/actionview-5.0.1/lib/action_view/routing_url_for.rb:78
```

while in `4.2` it returns

```
.../rbenv/versions/2.3.3/lib/ruby/gems/2.3.0/gems/actionpack-4.2.8/lib/action_dispatch/routing/url_for.rb:150
```

The latter cannot handle an array as an argument, hence the spec failures, but it seems like it must be some interaction of Appraisal and the rails auto loader that is causing the wrong method lookup, because it doesn't happen when we are running `4.2` outside of Appraisal.

nickcharlton:
@jcmorrow Oh that is fun! It sounds like you've gotten a little further on than I did. I wonder if we should raise this up on Appraisal and see if it's occurred before?

nickcharlton:
Closing as this was solved in a previous PR.

PR #739, opened 2017-01-20
Dependent destroy demo app models.

nickcharlton:
This fixes #738, where foreign key constrains were violated trying out
the deletion in the demo app.

Issue #738, opened 2017-01-20
Deleting Customers and/or Orders from the Demo App throws an exception


nickcharlton:
https://sentry.io/thoughtbot-6l/administrate/issues/207094470/

```
PG::ForeignKeyViolation: ERROR:  update or delete on table "customers" violates foreign key constraint "fk_rails_99d97e3ae6" on table "orders"
DETAIL:  Key (id)=(14627) is still referenced from table "orders".

  active_record/connection_adapters/postgresql_adapter.rb:598:in `async_exec'
    log(sql, name, binds) { @connection.async_exec(sql, type_casted_binds) }
  active_record/connection_adapters/postgresql_adapter.rb:598:in `block in exec_no_cache'
    log(sql, name, binds) { @connection.async_exec(sql, type_casted_binds) }
  active_record/connection_adapters/abstract_adapter.rb:589:in `block in log'
    :binds          => binds) { yield }
  active_support/notifications/instrumenter.rb:21:in `instrument'
    yield payload
  active_record/connection_adapters/abstract_adapter.rb:583:in `log'
    @instrumenter.instrument(
...
(110 additional frame(s) were not displayed)

ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "customers" violates foreign key constraint "fk_rails_99d97e3ae6" on table "orders"
DETAIL:  Key (id)=(14627) is still referenced from table "orders".
: DELETE FROM "customers" WHERE "customers"."id" = $1

ActiveRecord::InvalidForeignKey: PG::ForeignKeyViolation: ERROR:  update or delete on table "customers" violates foreign key constraint "fk_rails_99d97e3ae6" on table "orders"
DETAIL:  Key (id)=(14627) is still referenced from table "orders".
: DELETE FROM "customers" WHERE "customers"."id" = $1
```

PR #743, opened 2017-01-25
Products also need to destroy dependent models.

nickcharlton:
This is related to #739 and #738.

### DIFF
--- a/spec/example_app/app/models/product.rb
+++ b/spec/example_app/app/models/product.rb
@@ -1,4 +1,6 @@
 class Product < ActiveRecord::Base
+  has_many :line_items, dependent: :destroy
+
   validates :description, presence: true
   validates :image_url, presence: true
   validates :name, presence: true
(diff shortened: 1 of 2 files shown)


danbee:
Is this actually desired though? In a normal online store you'd want historic orders to maintain a record of the products that were actually ordered.

nickcharlton:
In an actual store, no.

So we could either do this to avoid getting a 500 when someone tries deleting, adjust the demo data model to take that into account or have an error message. As it's just a demo, this is an easier way of solving that.

danbee:
Given that this is a demo this is a reasonable approach. Although we could mark the products as deleted instead which would solve the `line_item` association problem in a way that would be more in line with how a real store would do it.

nickcharlton:
Yeah, I'm wondering now if we shouldn't use it being a demo as an excuse and instead do it properly.

BenMorganIO:
In theory, line items should be "destroyed" when orders are destroyed; although orders should never be destroyed (especially completed ones). Paranoia would be great for archiving the products.

But this is just a demo and I feel it'd be a tad bit too much. I'm :+1: for this. Another alternative could be to nullify the foreign key, rather than cascade.

PR #746, opened 2017-02-03
Add suffix/postfix option to number field

klaseskilson:
This introduces a `suffix` option to append to the display of numbers after the number. Much like the `prefix` option. This could for instance be `hours`, `miles`, `mph` or similar.


nickcharlton:
Looks great! Merging.

Issue #645, opened 2016-08-17
Generate administrate:views creates incorrect path


tehfailsafe:
Took a bit to figure out why customizing views wasn't working, but looks like it's appending an `s` to the end of `application` in the path.

![image](https://cloud.githubusercontent.com/assets/1388454/17723182/efeb43a8-63ec-11e6-96a3-c74e39305b31.png)

ashchan:
Hit by this too. Thanks for the finding.

jeffreyguenther:
@nickcharlton we need to fix this one in the next release.

PR #747, opened 2017-02-03
Redefine base resource path

purinkle:
Before, the base resource path for view generators was a hard-coded string. This was causing issues for some generators.  When called without an argument, their sub-generators had the wrong resource path. Replaced the base resource path string with a value object.

* Fixes https://github.com/thoughtbot/administrate/issues/645


nickcharlton:
Love it!

Can you squash into one commit and merge?

PR #750, opened 2017-02-07
Sort by default_sort on association class.

lunks:
This allows us to sort by these associations using a given attribute that uses name by default but can be overwritten by the static method default_sort on the associated class.

This is an initial implementation, feedback is more than welcome. I'd like to know if this is something that is desirable and/or if this approach is ok for the given problem.

Relates to #278.


hudsonsferreira:
hi @graysonwright could you pls take a look on this PR? thanks

nickcharlton:
Thanks for your PR!  I'm going to close this, as I've merged #810 instead.

lunks:
Great to have the feature in, thanks!

PR #755, opened 2017-02-22
Add support for changing form field labels.

dts:
I found myself wanting to change the labels associated with form fields, which are often good for programmers and maybe not so good for administrators.  The syntax looks like so:

```
subject: Field::String(label: "Model or Vehicle"),
```

### DIFF
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -27,6 +27,10 @@ module Administrate
         attr
       end
 
+      def label
+        options[:label] || attribute.to_s.humanize.capitalize
+      end
+
       def html_class
         self.class.html_class
       end
--- a/app/views/fields/belongs_to/_form.html.erb
+++ b/app/views/fields/belongs_to/_form.html.erb
@@ -17,7 +17,7 @@ that displays all possible records to associate with.
 %>
 
 <div class="field-unit__label">
-  <%= f.label field.permitted_attribute %>
+  <%= f.label field.permitted_attribute, field.label %>
 </div>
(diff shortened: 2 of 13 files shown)


lib/administrate/field/base.rb
```diff
@@ -27,6 +27,10 @@ def self.permitted_attribute(attr)
         attr
       end
 
+      def label
+        options[:label] || attribute.to_s.humanize.capitalize
```
jcmorrow: What if we allowed the overriding but didn't attempt to figure out the label name from the attribute if the option was missing and instead just returned `nil`? Then we can continue to let the form builder handle the unspecified case.

dts: Sure, that would work as well - I don't think there's any reason to do it this way (other than that it is explicit).

dts:
What kinds of specs would you think would be useful for this change?

jondkinney:
I'd like this ability, too. However, I'd prefer if we used the existing i18n paradigm so that if we specify it in en.yml for a field to override a table header, that same override would get applied to the form field. Like this PR, it's a trivial change and I've done it in a fork that I'm maintaining. Happy to provide the code if that's helpful, but it's just a copy and paste of the i18n query from the collection partial's table header lookup.

cpytel:
Thank you for the help. Since this PR was suggested, we've added the ability to customize the field labels with I18n. You can find the documentation for this at https://github.com/thoughtbot/administrate/blob/master/docs/customizing_dashboards.md#defining-labels

Issue #741, opened 2017-01-23
Date Field?


gregblass:
I defined an attribute in my schema with type "Date", but it is being defaulted to "DateTime". I tried to update this in ATTRIBUTE_TYPES and change it to Field::Date, but I'm getting: uninitialized constant Administrate::Field::Date.

Are date's without times not supported or am I doing something wrong? I would expect the 00:00:00 to not show up on the form and for the datepicker to not include the time control?

gregblass:
I tried generating a field for type Date:

`rails generate administrate:views:field date`

The generator seemed to run fine, but I'm getting:

`uninitialized constant Administrate::Field::Date`

Anything I'm doing wrong here? Shouldn't there be fields for Date, Time, and DateTime individually? I use all 3 types typically depending if I need to store one, the other, or both.

BenMorganIO:
Sorry, we've only got a field for `DateTime` ATM. If you'd like to do a PR and get a `Date` and `Time` field into this project, that'd be super appreciated! 👍

gregblass:
Got it, thanks for the info!

pablobm:
Implemented at https://github.com/thoughtbot/administrate/issues/741

pablobm:
Implemented at https://github.com/thoughtbot/administrate/pull/1530

Issue #716, opened 2016-12-19
Format for datetimepickers should be customizable


mcmire:
Right now the format for all datetimepickers is hardcoded to include time elements:

``` javascript
$(function () {
  $(".datetimepicker").datetimepicker({
    debug: false,
    format: "YYYY-MM-DD HH:mm:ss",
  });
});
```

However, if you want to create a date field, then having a format that includes time elements doesn't make any sense.

Here is what I have in my app:

``` javascript
(function () {
  $("[data-datetimepicker]").each(function (i, el) {
    var field = $(el);
    var format = field.data("format") || "YYYY-MM-DD HH:mm:ss";

    field.datetimepicker({
      debug: false,
      format: field.data("format"),
    });
  });
})();
```

This would, of course, necessitate changing the views to use `data` attributes instead of a class.

carlosramireziii:
Would #757 help here?

mcmire:
Yup, that would work just as well. 👍

pablobm:
Two things here:

1. Picker should differentiate between date and date+time.
2. Format should be customisable.

The first is now fixed. Since https://github.com/thoughtbot/administrate/pull/903, the current `DateTime` and `Time` fields render with data attributes that work as a hint for the JS widget.

The second is not yet possible. My first idea would be to add a new option to both fields, `format` (or similar), which then will be used for a new data attribute that the JS will pick up and use for its own `format` option. Ideally there should be some sort of locale integration too, but I think this can come later.

This second one is a potential easy contribution, if someone wants to give it a go. Still, for now I'm going to close this as it's been a while. Feel free to keep commenting if you have more input.

PR #757, opened 2017-02-24
Date and time fields

schpet:
This adds date and time fields. #741 :v:

### DIFF
--- a/lib/administrate/field/time.rb
+++ b/lib/administrate/field/time.rb
@@ -1,0 +1,11 @@
+require_relative "base"
+
+module Administrate
+  module Field
+    class Time < Base
+      def default
+        data && data.strftime("%H:%M:%S")
+      end
+    end
+  end
+end
--- a/lib/administrate/field/time.rb
+++ b/lib/administrate/field/time.rb
@@ -3,9 +3,6 @@ require_relative "base"
 module Administrate
   module Field
     class Time < Base
-      def default
-        data && data.strftime("%H:%M:%S")
-      end
     end
   end
 end
--- a/spec/lib/fields/date_spec.rb
+++ b/spec/lib/fields/date_spec.rb
@@ -1,0 +1,6 @@
+require "spec_helper"
+require "administrate/field/date"
+require "support/field_matchers"
+
+describe Administrate::Field::Date do
+end
--- a/app/assets/javascripts/administrate/components/date_time_picker.js
+++ b/app/assets/javascripts/administrate/components/date_time_picker.js
@@ -8,4 +8,9 @@ $(function () {
     debug: false,
     format: "YYYY-MM-DD",
   });
+
+  $(".timepicker").datetimepicker({
+    debug: false,
+    format: "HH:mm:ss",
+  });
 });
--- a/app/assets/javascripts/administrate/components/date_time_picker.js
+++ b/app/assets/javascripts/administrate/components/date_time_picker.js
@@ -3,4 +3,9 @@ $(function () {
     debug: false,
     format: "YYYY-MM-DD HH:mm:ss",
   });
+
+  $(".datepicker").datetimepicker({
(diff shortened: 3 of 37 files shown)


lib/administrate/field/time.rb
```diff
@@ -0,0 +1,11 @@
+require_relative "base"
+
+module Administrate
+  module Field
+    class Time < Base
+      def default
+        data && data.strftime("%H:%M:%S")
```
nickcharlton: Is there a default Rails format we can use here?

schpet: @nickcharlton unfortunately I wasn't able to find any. I checked out here:
https://github.com/rails/rails/blob/master/activesupport/lib/active_support/locale/en.yml

jcmorrow: What about adding this format as described in http://guides.rubyonrails.org/i18n.html#adding-date-time-formats?

schpet: @jcmorrow that sounds good, like adding:

```
  time:
    formats:
      time: "%H:%M:%S"
```

to the [administrate locale file](https://github.com/thoughtbot/administrate/blob/master/config/locales/administrate.en.yml) right?

jcmorrow: Exactly, although you may want to choose a name other than `time`. Perhaps `military` or `military_with_seconds` or something along those lines?

carlosramireziii:
Also addresses #716

lib/administrate/field/time.rb
nickcharlton: Ah, right. This is because Rails doesn't represent time without a date (this makes sense if you think about the concept of a specific time). So yes, we'd need a custom format. I'd suggest `time_without_date` or something.

@jcmorrow: "military time" is an amusing Americanism to me. 😛

jcmorrow: @nickcharlton I realized as I was typing it that probably only Americans called it that, and that shamefully I knew no other name for it 😓 . `twenty_four_hour_time` just doesn't roll off the keys.

schpet: haha. i've added some commits + rebased. i've added the `time_without_date` to each locale file, which seems dumb, but adequate, so please let me know if you see room for improvement 👍

fyi i'm getting a few failed tests on master, and they remain failing on this branch:

```
rspec ./spec/features/sidebar_spec.rb:12 # sidebar displays translated name of model
rspec ./spec/features/show_page_spec.rb:105 # customer show page displays translated labels
rspec ./spec/i18n_spec.rb:9 # I18n does not have missing keys
rspec ./spec/features/index_page_spec.rb:50 # customer index page displays translated labels
rspec ./spec/features/form_spec.rb:12 # edit form displays translated labels
```

spec/lib/fields/date_spec.rb
```diff
@@ -0,0 +1,6 @@
+require "spec_helper"
+require "administrate/field/date"
+require "support/field_matchers"
+
+describe Administrate::Field::Date do
```
nickcharlton: These are touched on in a previous spec, but I feel uncomfortable having these empty. Would you be able to add something for this and the `time_spec.rb` file?

bronislav:
Any progress with it?

I would like to see this merge :)

schpet:
@bronislav if you would like to take over this work and get it merged as a new PR, i would be very grateful!

i'm not sure how much has changed since i opened it up in feb. i haven't made any changes other than what's up here. sorry for not seeing this one through to the end!

bronislav:
@schpet Got it. I will look into it tomorrow.

bronislav:
@nickcharlton @schpet I've created a new branch and PR #931

schpet:
@bronislav appreciate it! i'm going to close this PR down in favor of #931.

PR #104, opened 2015-10-21
WIP: Support Bourbon 5

graysonwright:
https://trello.com/c/ylP4P7vJ/88-bourbon-5-support

Failing specs courtesy of [Appraisal]

Might want to wait until we open-source this and get Percy.io set up
before merging, in order to make sure there are no breaking CSS changes.

[Appraisal]: https://github.com/thoughtbot/appraisal

PR #759, opened 2017-03-03
Update instances of bourbon to 5.0.0.beta.7; sass to ~>3.4

vcavallo:
As requested in [this issue comment](https://github.com/thoughtbot/administrate/pull/104#issuecomment-274096057) This updates `administrate` to use bourbon 5 properly, I hope. Along the way it seemed the `sass` gem dependency needed to be updated as well.

I have to admit these layered dependencies and Appraisal tests are slightly out of my depth, but as far as I can tell all tests pass and a skeleton sample app I set up using Rails 5.0.2 and Ruby 2.3.1 and including these two gems, verbatim:

```ruby
gem 'bourbon'
gem "administrate", :git => 'https://github.com/vcavallo/administrate.git', :branch => 'bourbon-5-beta-7'
```

does work just fine, whereas when trying to use `administrate` on its own I got bundler errors like so:

screen shot 2017-03-03 at 11 53 55 am

Also, I'm not sure of the larger implications of requiring all Administrate users to upgrade to bourbon 5, especially since it seems there may be [other issues](https://github.com/thoughtbot/bourbon/issues/972#issuecomment-283997731) with bourbon 5 dependencies?

Let me know if anything needs to change or be updated.
Thanks!

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,5 +1,5 @@
 appraise "sass-3-4" do
-  gem "sass", "3.4.0"
+  gem "sass", "~> 3.4"
   gem "rails-controller-testing"
 end
 
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem "administrate-field-image"
-gem "bourbon", "~> 4.2"
+gem "bourbon", "~> 5.0.0.beta.7"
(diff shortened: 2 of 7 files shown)


nickcharlton:
Wonderful! I was curious that Circle didn't run with this, but it's running fine locally. So I'm going to merge!

PR #762, opened 2017-03-03
Fix pagination specs

jcmorrow:
The `MockRelation` class did not implement `page` or `per` so some specs
which rely on those methods were failing. Also, some content matchers
were not specific enough, so I've made those more specific and added
some helper methods.

Changes:

- Add some methods to `MockRelation`
- Make content matchers more specific to make spec more resilient to different ID's of related items

### DIFF
--- a/spec/features/show_page_spec.rb
+++ b/spec/features/show_page_spec.rb
@@ -1,21 +1,46 @@
 require "rails_helper"
 
 RSpec.describe "customer show page" do
+  describe "displays the customers orders paginated" do
+    context "when a page is not specified" do
+      it "displays the first page of results" do
+        customer = create(:customer)
+        orders = create_list(:order, 4, customer: customer)
+        first_page = orders.first(2)
+        second_page = orders.last(2)
 
-  it "displays the customers orders paginated" do
-    customer = create(:customer)
-    first_order, _a, _b, last_order = Array.new(4) { create(:order, customer: customer) }
+        visit admin_customer_path(customer)
 
-    visit admin_customer_path(customer)
-    within('.attribute-data--has-many') do
-      expect(page).to have_content(first_order.id)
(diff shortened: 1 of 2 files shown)


spec/features/show_page_spec.rb
```diff
@@ -1,21 +1,20 @@
...
-
     click_on("Next ›")
+
     within('.attribute-data--has-many') do
-      expect(page).to have_content(last_order.id)
-      expect(page).to have_no_content(first_order.id)
+      expect(page).to have_order_row(orders.last.id)
+      expect(page).not_to have_order_row(orders.first.id)
     end
   end
```
jayroh: @jcmorrow What do you think about adding a context around those pagination specs to give a little further ... well, context ... as to why you would see some on one page versus the next?  Right now it's not clear to me when I first come to this index page that there's a limit per page, how many that limit is, how many I should be seeing, etc.

jcmorrow: @jayroh I agree that the limit of two is extremely unclear here. I can try and clear that up.

jcmorrow: @jayroh I tried to address this in 48b0e2f38e669d5230f3480f5c1600bbf0566395, let me know what you think!

spec/features/show_page_spec.rb
```diff
@@ -1,21 +1,46 @@
...
+        end
+      end
     end
 
-    click_on("Next ›")
-    within('.attribute-data--has-many') do
-      expect(page).to have_content(last_order.id)
-      expect(page).to have_no_content(first_order.id)
+    context "when the second page is specified" do
+      it "displays the first page of results" do
```
thorncp: Should this be "displays the second page of results?"

jcmorrow: yes.

Issue #758, opened 2017-03-03
Administrate::Namespace#namespace_controller_paths


artificis:
I am trying to add devise for admin authentication. Here's a `routes.rb` file
```ruby
...
namespace :admin do
  resources :subscriptions
  root to: 'subscriptions#index'
end

devise_for :admin_users, controllers: {
  sessions: 'admin_users/sessions'
}
...
```

Now administrate throws a weird error saying `uninitialized constant AdminUsers::Session`.
So I decided to dig this gem source code and turns out that [this line](https://github.com/thoughtbot/administrate/blob/master/lib/administrate/namespace.rb#L19) causes this problem.
```ruby
def namespace_controller_paths
  all_controller_paths.select do |controller|
    controller.starts_with?(namespace.to_s)
  end
end
```

It should be `controller.starts_with?("#{namespace.to_s}/")` instead, no?

omake43:
👍

PR #763, opened 2017-03-04
Namespace Regex Refactor

jcmorrow:
I'm having trouble coming up with a spec that reproduces your error @clthck, maybe you could contribute one?

Changes:

- Factor out a `namespace_regex` method to use consistently across the `Namespace` class.

Addresses https://github.com/thoughtbot/administrate/issues/758

### DIFF
--- a/lib/administrate/namespace.rb
+++ b/lib/administrate/namespace.rb
@@ -12,7 +12,7 @@ module Administrate
       @routes ||= all_routes.select do |controller, _action|
         controller.starts_with?(namespace.to_s)
       end.map do |controller, action|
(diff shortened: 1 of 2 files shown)


punitcse:
@jcmorrow Could you please rebase the branch with master?

BenMorganIO:
I've rebased this. Looking at the implementation now, I genuinely wonder if this solves #758 anymore.

jcmorrow:
@BenMorganIO thanks for the rebase, I agree, it's hard to figure out if it fixes #758 without a good example spec 😞

pablobm:
Looks like this was solved elsewhere...?

Issue #766, opened 2017-03-06
Ensure that Autoprefixer is running on Administrate's CSS


carlosramireziii:
As per @tysongach 

> I’ve confirmed that Autoprefixer is not actually running on our CSS...

Related: 
- #372 
- #366

nickcharlton:
@carlosramireziii Now we've mered in #767, do you think we can confirm and close this?

carlosramireziii:
Is there any way to confirm that it's running? I checked the CSS on the demo app and it seems like there are _some_ prefixes, but I'm not sure if that's 100% guarantee that Autoprefixer is working. 

I'm fine closing this up and re-opening if we discover an issue.

nickcharlton:
I've just updated the demo app. It's been a bit out of sync, so what if you try again?

tysongach:
It appears to be running successfully now. I’m looking at the demo app, and given this input:

```scss
.app-container {
  @include margin(null auto);
  align-items: stretch;
  display: flex;
  max-width: 100em;
  min-height: 100vh;
}
```

…we now get this output:

```css
.app-container {
  margin-right: auto;
  margin-left: auto;
  -webkit-box-align: stretch;
  -ms-flex-align: stretch;
  align-items: stretch;
  display: -webkit-box;
  display: -ms-flexbox;
  display: flex;
  max-width: 100em;
  min-height: 100vh;
}
```

nickcharlton:
Wonderful, thanks @tysongach! Closing.

Issue #366, opened 2015-12-25
[SCSS] FIX - appearence none


nicovak:
There is an issue with select and `appearance` css property. The dropdown 'icon' is displayed twice (tested on chrome).
A mixin should be used (It's SASS), Bourbon is imported, you could used it like this:

Eg: administrate/app/assets/stylesheets/administrate/base/_forms.scss

```
select {
@include appearance(none);
}
```

It will generates all browser's prefix for the property.

IE is the bad kid but Opera seems to be also : [W3C](http://www.w3schools.com/cssref/css3_pr_appearance.asp)

The best way to hanlde It would be to include a javascript library for a better customization.

tysongach:
Definitely an issue many people are seeing. There’s some chatter and a possible fix over on #372, but Autoprefixer _should_ be running and take care of this for us (using Bourbon for prefixing is no longer the best solution). So if this is a prefixing issue, maybe it’s a bug with Autoprefixer.

tysongach:
I’ve confirmed that Autoprefixer is not actually running and outputting any prefixes, including the ones required here for these styles.

carlosramireziii:
As per #372, the solution to this issue is to fix Autoprefixer to make sure it's running on the CSS. I'm going to close this ticket in favor of that and will link the associated pull request when it's created.

PR #767, opened 2017-03-07
Add autoprefixer-rails

jcmorrow:
As mentioned in https://github.com/thoughtbot/administrate/issues/766, `autoprefixer` isn't currently running on this project, and since we aren't prefixing things by hand this means that selectors that need prefixing aren't getting it.

Changes:

- Add the `autoprefixer-rails` gem

Also related:

- #372 
- #366

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -289,6 +289,7 @@ DEPENDENCIES
   administrate-field-image
   ammeter
   appraisal
+  autoprefixer-rails
   awesome_print
   bourbon (~> 5.0.0.beta.7)
   bundler-audit
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 gem "administrate-field-image"
+gem "autoprefixer-rails"
 gem "bourbon", "~> 5.0.0.beta.7"
 gem "delayed_job_active_record"
 gem "faker"


Gemfile.lock
```diff
@@ -319,4 +320,4 @@ DEPENDENCIES
   webmock
 
 BUNDLED WITH
-   1.14.4
+   1.14.6
```
nickcharlton: Does adjusting the bundle version have any side-effects with Circle or Appraisal? This is something I've seen before (you might want to look at the run commands).

jcmorrow: As far as I know this line only matters when it comes to diagnosing problems with bundler, but I am only mildly certain of that. I am happy to revert the line, though I think that may cause even more confusion if there *were* a problem with bundler, since our lock file would be inaccurate. Thoughts?

nickcharlton: Something I've been seeing (with dep versions shifting ahead of us here) are loading issues with the version of bundler. Because of the CI setup, it'll end up loading multiple versions and then fails. What do you think of reverting this line on the assumption that we'll re-run before the next release?

jcmorrow: 👍  done in 42b0c31b338de424fab615621277300cfe9e4358

carlosramireziii:
Quick question on this one-- will Autoprefixer automatically be running on the assets coming from the project? Will this exclude assets from the main app or will it force users to auto prefix everything, even their own project's assets?

nickcharlton:
@carlosramireziii I'm actually not sure. Do you know, @jcmorrow?

jcmorrow:
@carlosramireziii I also do not know. 🤷‍♀️ I suppose it could be easy to find out though.

Issue #599, opened 2016-06-01
Viewing User Record: ActionView::Template::Error: undefined method `limit' for nil:NilClass


ACPK:
When I try to "view" an individual user record, I get the following error:

MESSAGE:

```
ActionView::Template::Error: undefined method `limit' for nil:NilClass
```

BACKTRACE:

```
show.html.erb  37  block in _app_views_admin_vendors_show_html_erb(...)
[PROJECT_ROOT]/app/views/admin/vendors/show.html.erb:37:in `block in _app_views_admin_vendors_show_html_erb'
```

LINE CAUSING THE ISSUE:

```
# app/views/admin/vendors/show.html.erb:37 :in `block in _app_views_admin_vendors_show_html_erb`

<dd class="attribute-data attribute-data--<%=attribute.html_class%>">
  <%= render_field attribute %></dd>
<% end %>
```

ACPK:
Has anyone else his this issue?

ID25:
I got the same error when i have an association without any records.

ACPK:
@ID25 @simmerer If you don't mind me asking, what's the recommended fix?

ACPK:
@ID25 What's the best way to find the issue?

haseeb-ahmad:
I am having the same issue here, cannot view a single record as well as cannot view collection of records..
`<%= render_field attribute %>` 
if i comment this out it runs but does not show attributes.. Does any one has any leads on how to solve this..I have been on this for a couple of days and cannot seem to find the solution..

nickcharlton:
I've just merged #771, so this should now be fixed. I'm going to close this issue, but feel free to open another if this doesn't solve it for you.

PR #771, opened 2017-03-09
Handle case where `Field::HasMany` data is nil

carlosramireziii:
When an association has no records, the field is being created with no
data, and when methods are called on that data, undefined method errors
occur.

To fix, use the Null Object Pattern to ensure the data is always
present. Use Active Record's NullRelation as the null object.

Addresses #599

### DIFF
--- a/lib/administrate/field/has_many.rb
+++ b/lib/administrate/field/has_many.rb
@@ -25,7 +25,9 @@ module Administrate
       end
 
       def selected_options
-        data && data.map { |object| object.send(primary_key) }
+        return if data.empty?
+
+        data.map { |object| object.send(primary_key) }
       end
 
       def limit
@@ -42,6 +44,10 @@ module Administrate
 
       def more_than_limit?
         data.count(:all) > limit
+      end
+
+      def data
+        @data ||= associated_class.none 
       end
 
       private
--- a/spec/lib/fields/has_many_spec.rb
+++ b/spec/lib/fields/has_many_spec.rb
@@ -101,6 +101,17 @@ describe Administrate::Field::HasMany do
 
       expect(field.more_than_limit?).to eq(false)
     end
+
+    context "when there are no records" do
+      it "returns false" do
+        resources = nil
+
+        association = Administrate::Field::HasMany
+        field = association.new(:customers, resources, :show)
+
+        expect(field.more_than_limit?).to eq(false)
+      end
+    end
   end
 
   describe "#resources" do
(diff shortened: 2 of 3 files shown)


lib/administrate/field/has_many.rb
```diff
@@ -25,7 +25,9 @@ def associated_resource_options
       end
 
       def selected_options
-        data && data.map { |object| object.send(primary_key) }
+        return if data.empty?
```
carlosramireziii: In order to keep the same behavior as before, I added this guard clause. 

Now that `data` won't be `nil`, we can let the `map` method run and it will return an empty array.

So in the case of an empty collection, do we want this method to return `nil` or `[]`?

lib/administrate/field/has_many.rb
```diff
@@ -44,6 +46,10 @@ def more_than_limit?
         data.count(:all) > limit
       end
 
+      def data
+        @data ||= associated_class.none 
+      end
```
carlosramireziii: Is this the best way to implement the null object pattern for the `data` method?

lib/administrate/field/has_many.rb
nickcharlton: Are there any side effects of providing one or the other?

carlosramireziii: None that I can see or think of...

The only place `selected_options` is used is as the 2nd argument in a call to [`options_for_select`](https://github.com/thoughtbot/administrate/blob/master/app/views/fields/has_many/_form.html.erb#L27). Tracing that method, when the 2nd argument is `nil` it seems to get turned into an empty array anyway (see [here](https://github.com/rails/rails/blob/66a7cfa91045e05f134efc9ac0e226e66161e2e6/actionview/lib/action_view/helpers/form_options_helper.rb#L354) and [here](https://github.com/rails/rails/blob/66a7cfa91045e05f134efc9ac0e226e66161e2e6/actionview/lib/action_view/helpers/form_options_helper.rb#L781)), however I'm wondering if it's just better to leave the previous behavior and not introduce any potential breaking changes. 

I don't think there's any downside to leaving it as `nil` except for the presence of the guard clause there.

lib/administrate/field/has_many.rb
nickcharlton: Yeah, I think so.

nickcharlton:
Thanks @carlosramireziii! I think we're good on this, so I'm going to merge it.

PR #774, opened 2017-03-10
Update Kaminari `~> 0.16` -> `>= 1.0`

klaseskilson:
As the title says, let's get that Kaminari 1 support going!


administrate.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
   s.add_dependency "bourbon", "~> 5.0.0.beta.7"
   s.add_dependency "datetime_picker_rails", "~> 0.0.7"
   s.add_dependency "jquery-rails", "~> 4.0"
-  s.add_dependency "kaminari", "~> 0.16"
+  s.add_dependency "kaminari", "~> 1.0"
```
nickcharlton: What do you think about doing something like `>= 1.0` so we'll get 1.0 and above and no lower?

klaseskilson: That would absolutely work for me, I just assumed you'd prefer this way due to any potential future breaking changes.

klaseskilson: Let me know which one you prefer and I'll fix it straight away.

nickcharlton: Let's go for it and see what ends up happening!

Kaminari is pretty well used and stable, so my feeling is that it's best for us to be more liberal with our versioned dependencies.

klaseskilson: All right! Fixing. 👌

nickcharlton:
Wonderful, thanks!

PR #776, opened 2017-03-10
Include docs in gem (via gemspec file listing).

### LINKED ISSUE #775: Shipped gem should include documentation (guides)
Continuation from #503 , @nickcharlton seems to be positive about it too (https://github.com/thoughtbot/administrate/issues/503#issuecomment-285750224).

I opt for including the guides (in `/doc`) in the shipped gem. Although they are not necessary for usage, they provide means to access a proper high-level documentation independently from external services (like github, heroku, thoughbot) and internet connectivity.

While optimally the documentation might be included in the documentation that can automatically be built for the gem (rdoc), imho the markdown files are good enough.

fwolfst:
Closes #775.

### DIFF
--- a/administrate.gemspec
+++ b/administrate.gemspec
@@ -11,7 +11,7 @@ Gem::Specification.new do |s|
   s.summary = "A Rails engine for creating super-flexible admin dashboards"
   s.license = "MIT"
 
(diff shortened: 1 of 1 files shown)


nickcharlton:
This is great, thanks! Merging.

PR #779, opened 2017-03-15
Update .rubocop.yml to follow thoughtbot/guides.

nickcharlton:
This goes a little further than #518 and syncs up completely.


BenMorganIO:
Very :+1:

PR #780, opened 2017-03-16
Add Sentry for error tracking on the prototype app.

nickcharlton:
We've actually been running this for real for a while, but I've been rebasing manually every so often.

This introduces it fully.


nickcharlton:
The diff on `Gemfile.lock` is an exciting mess. I'll fix that up.

PR #781, opened 2017-03-16
Update readme screenshot

tysongach:
The current screenshot is a bit out of date from what Administrate looks like today:

screen shot 2017-03-16 at 09 18 25


BenMorganIO:
Thanks!

PR #782, opened 2017-03-16
Rails 5.1

BenMorganIO:
On the journey to reach Rails 5.1 support, we've got to take care of some Deprecation Warnings. There was only one that I could find in 2cc3d42.

I've also updated the example_app to use `ApplicationRecord`. I don't expect this to do anything significant, but why not...


config/locales/administrate.de.yml
```diff
@@ -6,6 +6,8 @@ de:
       destroy: Löschen
       edit: Editieren
       show: Anzeigen
+      new: Nieuwe
```
cdegroot: "Neu"

config/locales/administrate.de.yml
```diff
@@ -6,6 +6,8 @@ de:
...
+      back: Terug
```
cdegroot: "Zurück"

config/locales/administrate.nl.yml
```diff
@@ -6,6 +6,8 @@ nl:
       destroy: Verwijder
       edit: Bewerk
       show: Toon
+      new: Ny
```
cdegroot: "Nieuw"

config/locales/administrate.nl.yml
```diff
@@ -6,6 +6,8 @@ nl:
...
+      back: Tilbake
```
cdegroot: "Terug"

config/locales/administrate.fr.yml
```diff
@@ -6,6 +6,8 @@ fr:
       destroy: Supprimer
       edit: Modifier
       show: Détails
+      new: Nouveau
+      back: Arrière
     controller:
       create:
         success: "%{resource} a été correctement créé(e)."
```
hickscorp: Correct here, because the feminine `(e)` form is put into parenthesis, which is fairly comon practice when the genre is unknown.

config/locales/administrate.fr.yml
```diff
@@ -6,6 +6,8 @@ fr:
       destroy: Supprimer
       edit: Modifier
       show: Détails
+      new: Nouveau
```
hickscorp: This is tricky. In French there is the notion of feminine / masculine for objects. Nouveau only applies to masculine, making the button wrong. I would go for "Création" instead.

config/locales/administrate.fr.yml
```diff
@@ -6,6 +6,8 @@ fr:
...
+      back: Arrière
```
hickscorp: This doesn't make much sense in the context of a flow. I would go for `Précédent` if the goal is `Previous`, or `Retour` if in a modal.

nickcharlton:
What do you think about splitting the translations out into a separate PR? I feel like it's a bit separate from Rails 5.1 support and might take a bit longer to get just right.

BenMorganIO:
@nickcharlton agreed!

BenMorganIO:
See #786 for where I placed the translation commits.

BenMorganIO:
OK, I'm going to split this guy into a lot of PRs. Why? Well, look at the commits. There's some deprecations and what not in here with a lot of changes for Rails 5.1 support by gem version.

BenMorganIO:
Note: To bump to Rails 5.1.0, need to wait till https://github.com/rails/sass-rails/pull/400 is merged.

PR #784, opened 2017-03-17
Update the Appraisal Gemfiles.

nickcharlton:
* Reduce the bourbon requirement slightly (7 to 6) so we can avoid updating Sass.
* Adds refills, autoprefixer-rails, sentry-raven which were previously merged.

### DIFF
--- a/gemfiles/rails42.gemfile
+++ b/gemfiles/rails42.gemfile
@@ -16,7 +16,6 @@ gem "unicorn"
 gem "rails", "~> 4.2.0"
 
 group :development do
-  gem "refills"
   gem "web-console", ">= 2.1.3"
 end
 
--- a/gemfiles/rails42.gemfile
+++ b/gemfiles/rails42.gemfile
@@ -3,6 +3,7 @@
 source "https://rubygems.org"
 
 gem "administrate-field-image"
+gem "autoprefixer-rails"
 gem "bourbon", "~> 5.0.0.beta.7"
 gem "delayed_job_active_record"
 gem "faker"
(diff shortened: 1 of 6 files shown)


Appraisals
```diff
@@ -1,5 +1,5 @@
 appraise "sass-3-4" do
-  gem "sass", "~> 3.4"
+  gem "sass", "~> 3.4.22"
```
jayroh: Would this level of specificity cause issues for projects that have versions `3.4.0..3.4.21`?

nickcharlton: Ooh, I hadn't thought of that. What do you think would be a better way of specifying it?

jayroh: Depends, I think. What does administrate need from .22 that .0 doesn't provide? Looks like it's Bourbon? Can the bourbon version be bumped backwards a little bit to accommodate some flexibility in that realm.

If possible, I'd suggest changing the Administrate sass to NOT use any features that are locked into later/younger versions.

tysongach: > Can the bourbon version be bumped backwards a little bit to accommodate some flexibility in that realm.

Yes, we can and probably should do that.

nickcharlton: I've switched the Bourbon requirement to be `5.0.0.beta.6` or greater and this seems to solve this.

gemfiles/rails42.gemfile
```diff
@@ -3,17 +3,20 @@
...
 gem "high_voltage"
 gem "markdown-rails"
 gem "pg"
 gem "redcarpet"
+gem "sentry-raven"
 gem "unicorn"
 gem "rails", "~> 4.2.0"
 
 group :development do
+  gem "refills"
```
tysongach: I suggest removing Refills because it is going through a major rewrite. The gem is likely to go away (or dormant).

The gem is just a generator anyway, so we don’t _need_ to have it to use some of the components.

nickcharlton: Okay, cool! I'll do that in the same PR and merge. Thank you!

Appraisals
tysongach: FYI, I opened https://github.com/thoughtbot/bourbon/pull/1015 on Bourbon to loosen these requirements.

PR #786, opened 2017-03-17
Translations

BenMorganIO:

PR #788, opened 2017-03-22
Allow resource to be accessed from field (alternative implementation)

pedantic-git:
I was taking a look at @dts's excellent #706, and hoping something similar would be accepted upstream soon as I needed support for accessing the resource too.

This is a much more lightweight implementation that achieves the same effect - the resource is available as `field.resource`. However, this one degrades gracefully in case people have overridden the constructor in their own field implementations, and as a consequence the tests (hopefully) don't need changing.

I wasn't able to run the tests because they seem to depend on a running Postgres instance. Let me know if anything fails and I can take a look!

### DIFF
--- a/lib/administrate/field/deferred.rb
+++ b/lib/administrate/field/deferred.rb
@@ -11,7 +11,8 @@ module Administrate
       attr_reader :deferred_class, :options
 
       def new(*args)
-        deferred_class.new(*args, options)
+        new_options = if args.last.is_a? Hash then args.last else {} end
+        deferred_class.new(*args, options.merge(new_options))
       end
 
       def ==(other)
--- a/lib/administrate/field/deferred.rb
+++ b/lib/administrate/field/deferred.rb
@@ -11,7 +11,7 @@ module Administrate
       attr_reader :deferred_class, :options
 
       def new(*args)
-        new_options = if args.last.is_a? Hash then args.last else {} end
+        new_options = if args.last.is_a? Hash then args.pop else {} end
         deferred_class.new(*args, options.merge(new_options))
       end
 
--- a/lib/administrate/field/deferred.rb
+++ b/lib/administrate/field/deferred.rb
@@ -11,7 +11,6 @@ module Administrate
       attr_reader :deferred_class, :options
 
       def new(*args)
-        # If the args already end with a "keyword hash", merge it with options
         new_options = args.last.respond_to?(:merge) ? args.pop : {}
         deferred_class.new(*args, options.merge(new_options))
       end
--- a/lib/administrate/field/deferred.rb
+++ b/lib/administrate/field/deferred.rb
@@ -11,7 +11,8 @@ module Administrate
       attr_reader :deferred_class, :options
 
       def new(*args)
-        new_options = if args.last.is_a? Hash then args.pop else {} end
+        # If the args already end with a "keyword hash", merge it with options
+        new_options = args.last.respond_to?(:merge) ? args.pop : {}
         deferred_class.new(*args, options.merge(new_options))
       end
 
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -20,6 +20,7 @@ module Administrate
         @attribute = attribute
         @data = data
         @page = page
+        @resource = options.delete(:resource)
         @options = options
       end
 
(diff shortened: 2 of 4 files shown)


lib/administrate/field/deferred.rb
```diff
@@ -11,7 +11,8 @@ def initialize(deferred_class, options = {})
       attr_reader :deferred_class, :options
 
       def new(*args)
-        deferred_class.new(*args, options)
+        new_options = if args.last.is_a? Hash then args.pop else {} end
```
BenMorganIO: Could you use a ternary and a `respond_to?` as that's more ruby-esque?

```ruby
new_options = agrs.last.respond_to?(:pop) ? args.pop : {}
```

BenMorganIO:
Thanks for all the work you're doing @pedantic-git! Are you able to take a look at my comments and we can get this guy merged in :)

We can probably add a spec where you pass a resource as an option and then make sure it returns the correct resource. Simple input/output test :)

lib/administrate/field/deferred.rb
pedantic-git: Heh - I actually usually use ternaries but I recently discovered the "verbose ternary with `then`" and I like it a lot more. I was happy to switch though.

However, moving to respond_to? caused tests to fail, presumably because other objects that are not hashes have a `pop` method. It really needs to be a Hash here because the whole point of this is to trap Ruby's Hash-as-final-argument syntax which was used for positional params prior to 2.2. 

I could *probably* write `args.respond_to?(:pop) && args.respond_to?(:merge)` but it seems clumsy and actually not saying what we're trying to say here.

pedantic-git:
Hi @BenMorganIO - apologies for all the commits (you may want to squash them!)

I was working on this blind as I didn't have a way of running the specs on my own environment. It should address your concerns and pass everything now, although it's failing CircleCI due to a Nokogiri vulnerability which is unrelated.

pedantic-git:
OK - I think this is all sorted now. I've still got a test failure to do with I18n but I haven't changed the frontend - any idea why?

nickcharlton:
`i18n` has been causing failures for a little while. We're working on it in #786.

pedantic-git:
@BenMorganIO Do let me know if there's anything else you need from me before merging this PR.

kesha-antonov:
Need this +1

lib/administrate/field/deferred.rb
```diff
@@ -11,7 +11,9 @@ def initialize(deferred_class, options = {})
       attr_reader :deferred_class, :options
 
       def new(*args)
-        deferred_class.new(*args, options)
+        # If the args already end with a "keyword hash", merge it with options
```
nickcharlton: Could we lose the comment here?

nickcharlton:
One tiny thing, then I'm going to go ahead and merge this!

lib/administrate/field/deferred.rb
pedantic-git: I'm happy to do that - just wondering why you'd like it to go? To me what's happening on the following line is not intuitive at all, hence the comment.

pedantic-git:
Hi @nickcharlton - I've removed that comment but could you please help me understand why you wanted it removed? I feel like the two lines following it are not very intuitive and benefit from a small comment.

nickcharlton:
Hi @pedantic-git!

Sorry for taking a while to reply to you. I like to try and avoid comments in the code as they often drift over time and get forgotten about. Putting the same thing in the commit message which introduces the change allows us to (hopefully) solve the same problem in a manner which does get stuck in time.

On this however, now you mention it again I'm not sure, the method implementation is quite complex. What do you think?

pedantic-git:
I'm easy either way! Whatever it takes to get this merged before 0.7 is released. :smile: 

Also, if there's a more idiomatic way to say "merge kwargs hashes" in Ruby we can use that.

nickcharlton:
I feel like I don't have a good answer for this, so I'm going to let it drop out of 0.7.0 and pass it over to 0.8.0. We'll get it in there for sure. Thanks for holding in!

pedantic-git:
@nickcharlton OK! I hope 0.8 is soon! Would you like me to put the comment back in?

nickcharlton:
Having looked at this a while later, I think we're good to go now. So I'm going to merge.

pedantic-git:
@nickcharlton I can't click the :tada: button enough!

morellan:
@pedantic-git We have been using your solution on a private fork of administrate for a while now, now it's on the new release!, congrats!

PR #789, opened 2017-03-22
Remove bourbon

pedantic-git:
This might be controversial, but I want to remove bourbon from the administrate codebase. It's not used in many places and inlining the CSS is quite straightforward.

Whenever I update or make changes to administrate, there always seems to be some bourbon-related issue. Either it's the removal of the Railtie in Bourbon 5.x, or the fact that my app won't load in Cucumber any more, everything seems to be related to Bourbon.

What do you think?

### DIFF
--- a/app/assets/stylesheets/administrate/library/_variables.scss
+++ b/app/assets/stylesheets/administrate/library/_variables.scss
@@ -12,7 +12,6 @@ $base-line-height: 1.5;
 $heading-line-height: 1.2;
 
 // Other Sizes
-$modular-scale-ratio: $minor-third;
 $base-border-radius: 4px;
 $base-spacing: $base-line-height * 1em;
 $small-spacing: $base-spacing / 2;
@@ -50,7 +49,7 @@ $focus-outline-offset: 1px;
 $flash-colors: (
   alert: $light-yellow,
   error: $red,
-  notice: tint($blue, 50%),
+  notice: mix($white, $blue, 50%),
   success: $light-green
 );
 
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,6 @@ gemspec
 
 gem "administrate-field-image"
 gem "autoprefixer-rails"
-gem "bourbon", "~> 5.0.0.beta.7"
 gem "faker"
 gem "pg"
 gem "redcarpet"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -6,7 +6,6 @@ PATH
       actionview (>= 4.2, < 5.1)
       activerecord (>= 4.2, < 5.1)
       autoprefixer-rails (~> 6.0)
-      bourbon (>= 5.0.0.beta.6)
       datetime_picker_rails (~> 0.0.7)
       jquery-rails (>= 4.0)
       kaminari (>= 1.0)
@@ -59,9 +58,6 @@ GEM
     autoprefixer-rails (6.7.7.1)
       execjs
     awesome_print (1.6.1)
-    bourbon (5.0.0.beta.7)
-      sass (~> 3.4.22)
-      thor (~> 0.19.1)
     builder (3.2.3)
     bundler-audit (0.4.0)
       bundler (~> 1.2)
@@ -260,7 +256,6 @@ DEPENDENCIES
   appraisal
   autoprefixer-rails
   awesome_print
-  bourbon (~> 5.0.0.beta.7)
   bundler-audit
   byebug
   database_cleaner
--- a/administrate.gemspec
+++ b/administrate.gemspec
@@ -19,7 +19,6 @@ Gem::Specification.new do |s|
   s.add_dependency "activerecord", ">= 4.2", "< 5.1"
 
   s.add_dependency "autoprefixer-rails", "~> 6.0"
-  s.add_dependency "bourbon", ">= 5.0.0.beta.6"
   s.add_dependency "datetime_picker_rails", "~> 0.0.7"
   s.add_dependency "jquery-rails", ">= 4.0"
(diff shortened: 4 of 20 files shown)


BenMorganIO:
@pedantic-git this is timely, as we've had discussions on the administrate slack about this. You've taken the time to show us what the code can look like with one less dependency. I'm :+1: for this. I was going to make a PR to _attempt_ fixing some of the bugs with Bourbon, but if you think we can remove it, then I'm all for it!

app/assets/stylesheets/administrate/library/_variables.scss
```diff
@@ -12,7 +12,7 @@ $base-line-height: 1.5;
 $heading-line-height: 1.2;
 
 // Other Sizes
-$modular-scale-ratio: $minor-third;
+$modular-scale-ratio: 1.2;
```
tysongach: This is a Bourbon setting for it’s `modular-scale` function. If that function is not used anywhere, this variable should be able to be removed.

app/assets/stylesheets/administrate/library/_variables.scss
```diff
@@ -54,4 +54,4 @@ $base-border: 1px solid $base-border-color;
 
 // Transitions
 $base-duration: 250ms;
-$base-timing: $ease-out-cubic;
+$base-timing: cubic-bezier(0.215,  0.610, 0.355, 1.000);
```
tysongach: We could probably just simplify this to something like `ease-in-out`.

tysongach:
Oh, one realization: it looks like Administrate requires Neat 1.1+? If that’s true, by removing Bourbon we would need to require Neat 1.8, because Neat has a Bourbon dependency up until 1.8.

So maybe it’s worth removing Neat, as well. I don’t _think_ it’s used much more than as we’ve seen with Bourbon, but I’ve not dug into it. Honestly, the view don’t need much of a grid system.

Also, Neat 2.0 is a complete re-write, so we’d eventually start running into conflict issues with people using 2.0 in their project and Administrate require the 1.x cycle.

pedantic-git:
Thanks very much for saying you agree in principle with this! I'm happy to address the review comments above but I'm afraid I'm a bit out of my depth where removing Neat is concerned - it seems like the CSS it provides is much more complex than that provided by Bourbon, but maybe it's just my fear of the unfamiliar?

tysongach:
@pedantic-git No worries. I can investigate the removal of Neat. Let’s keep this PR Bourbon-specific.

**Note to maintainers about my [comment above](https://github.com/thoughtbot/administrate/pull/789#issuecomment-288828794) regarding the Bourbon/Neat dependency; might not want to merge this until that is sorted out.**

BenMorganIO:
@pedantic-git that's a pretty reasonable concern 😛 

However, if Neat depends on Bourbon, then yeah, we'll have to pull it in, BUT we can still limit and lower the usage of Bourbon. Step 2 for updating Neat to 1.9 would mean just removing a few requires :)

Side question: If we're removing Bourbon and Neat, does that mean Administrate is going cold turkey and we'll become sober?

pedantic-git:
Hi @tysongach / @BenMorganIO - I think I've made all the requested changes.

I've got a CI failure but I can't seem to load the results to see what it is. I also can't easily address the one remaining Hound violation because the code is trying to affect _all_ buttons, so I think this violation is required.

I think you're right that the Neat removal (or upgrade to 1.8) needs to happen first. I noticed there was still one Bourbon function in there when I came to address your comments and there might be more too that are hiding away.

app/assets/stylesheets/administrate/components/_header.scss
```diff
@@ -6,7 +6,7 @@
 }
 
 .header__heading {
-  font-size: modular-scale(3);
+  font-size: 1.728em; // 1.2 ** 3
```
tysongach: We can simplify this to `1.75em`; it doesn’t have to match the output of the old `modular-scale` value exactly. We can also remove the comment.

app/assets/stylesheets/administrate/library/_data_label.scss
```diff
@@ -1,6 +1,6 @@
 @mixin data-label {
   color: $hint-grey;
-  font-size: modular-scale(-1);
+  font-size: 1em / 1.2em;
```
tysongach: Can we put a raw value here that closely matches what the output of `modular-scale` was, instead of math?

app/assets/stylesheets/administrate/library/_mixins.scss
```diff
@@ -0,0 +1,7 @@
+@mixin administrate-clearfix {
+  &::after {
+    clear: both;
+    content: "";
+    display: block;
+  }
+}
```
tysongach: No newline at end of file.

app/assets/stylesheets/administrate/library/_mixins.scss
```diff
@@ -0,0 +1,7 @@
+@mixin administrate-clearfix {
```
tysongach: Can we name this file `_clearfix.scss` to match the name of the mixin it houses? It makes finding things super duper easy.

pedantic-git:
@tysongach Sure! Done.

tysongach:
Thanks for all of this great work, @pedantic-git!

tysongach:
Awaiting removal of Neat before merging this. I intend to work on that today.

tysongach:
This can be merged once https://github.com/thoughtbot/administrate/pull/798 is.

tysongach:
I’m going to rebase and fix the conflicts.

tysongach:
Rebased and squashed. I think this is ready to merge, @nickcharlton.

PR #793, opened 2017-03-24
update ruby on circleci

BenMorganIO:


nickcharlton:
Does Circle support a `.ruby-version`? Maybe we can do that instead and then we'll get that locally too (and less Circle specific configuration).

BenMorganIO:
> We infer your Ruby version from your .rvmrc file, .ruby-version file, or Gemfile.

https://circleci.com/docs/1.0/unrecognized-ruby-version/

I would add a `ruby '2.4.1'` to the Gemfile that way rvm recognizes it, although rbenv is different...

nickcharlton:
Cool! I usually do something like `ruby-2.4.1`, as that works with `chruby`.

BenMorganIO:
I feel like we should use Travis though and test this on multiple ruby versions.

PR #796, opened 2017-03-24
Added latvian locale to the I18n

DuckForQuack:


BenMorganIO:
@DuckForQuack can you rebase against master and add `back` and `new`?

nickcharlton:
I'm going to close this for the time being as it's been a month since the last response, and adding this would have a small impact, but large cost. I would love to support Latvian properly, but I don't think we can do it justice without external help.

PR #799, opened 2017-03-26
fix css pagination

kuntoaji:
this PR will fix this style:
administrate-pagination

### DIFF
--- a/app/assets/stylesheets/administrate/components/_pagination.scss
+++ b/app/assets/stylesheets/administrate/components/_pagination.scss
@@ -3,11 +3,8 @@
   @include padding(null $base-spacing);
   text-align: center;
 
-  .first,
-  .prev,
-  .page,
-  .next,
-  .last {
+  li {
+    display: inline-block;
     margin: $small-spacing;
   }
 


tysongach:
Hi @kuntoaji! Welcome.

1. What page is your screenshot from?
1. What browser and OS are you on?
1. From what I can tell, our pagination is output in `span` elements, not `li`’s. Do you have a screenshot from the changes introduced in this PR?

nickcharlton:
I'm not able to replicate this on Firefox 45.0.1, Safari 10.0.3, Chrome 57. But all of these are on macOS. Are you using something else?

janklimo:
This looks familiar. It happens when you override default kaminari pagination partials so they match Bootstrap pagination markup. I.e. `nav > ul > li` style instead of `nav > span`. @kuntoaji is that the case?

kuntoaji:
@janklimo thanks, that's the case. I thought administrate use their own style for pagination, not partial view from kaminari.

davidbenton:
I'll share how I dealt with this, in case it's useful to anyone else. This was because I overrode the default kaminari theme, as @janklimo said. The solution was to move themes into their own directory and set the theme explicitly in my non-administrate views:

Move files (from the bootstrap4 kaminari theme, in my case):
```bash
% mkdir app/views/kaminari/bootstrap4
% mv app/views/kaminari/*.html.erb app/views/kaminari/bootstrap4/
```

And call `paginate` like:
```erb
<%= paginate @resources, theme: 'bootstrap4' %>
```

PR #803, opened 2017-03-29
improve translation

FunnyQ:
improve translation in `administrate.zh-TW.yml`


BenMorganIO:
@FunnyQ that was me... Sorry about that!

PR #807, opened 2017-03-31
Add foreign key option to Fields

iarie:
#777 
- Allows to specify `foreign_key` option in associative fields
- Updates example_app to demonstrate this feature


spec/example_app/app/models/country.rb
```diff
@@ -0,0 +1,4 @@
+class Country < ActiveRecord::Base
+  validates :code, presence: true
+  validates :name, presence: true
```
BenMorganIO: Maybe do:

```ruby
validates :code, :name, presence: true
```

spec/example_app/db/migrate/20170331144723_create_countries.rb
```diff
@@ -0,0 +1,12 @@
+class CreateCountries < ActiveRecord::Migration
+  def change
+    create_table :countries do |t|
+      t.string :code
```
BenMorganIO:
```ruby
t.string :code, null: false, index: { unique: true }
```

That way you don't need the `add_index` at the bottom. Also, since you're validating at the application level, these should be `null: false` as well.

spec/example_app/db/seeds.rb
```diff
@@ -10,12 +10,20 @@
 Order.destroy_all
 Customer.destroy_all
 Product.destroy_all
+Country.destroy_all
+
+Country.create(code: "US", name: "USA")
+Country.create(code: "CA", name: "Canada")
+Country.create(code: "CN", name: "China")
+Country.create(code: "RU", name: "Russia")
+Country.create(code: "AU", name: "Australia")
```
BenMorganIO: These could fail silently for a developer. Why not:

```ruby
Country.create! [
  { code: "US", name: "Trump Temptations: The Billionaire & The Bellboy" },
  { code: "CA", name: "Canada" },
  { code: "CN", name: "China" },
  { code: "RU", name: "Russia" },
  { code: "AU", name: "Australia" }
]
```

spec/example_app/db/seeds.rb
```diff
@@ -10,12 +10,20 @@
...
 
 100.times do
   name = "#{Faker::Name.first_name} #{Faker::Name.last_name}"
   Customer.create(
     name: name,
     email: Faker::Internet.safe_email(name),
+    country: Country.first,
```
BenMorganIO: This iterates 100 times. Maybe do:

```ruby
country = Country.first
customers = Array.new(100) do
  {
    name: Faker::Name.name,
    email: Faker::Internet.safe_email(name),
    country: country
  }
end

Customer.create!(customers)
```

`Array.new` is faster than `Integer#times`; best practice to always use `Array.new`. Also, you don't have to call `Customer.create` 100 times, just once with an array for it to iterate over. We don't need this blazing fast. It's not inserting 2,000 records. However, we nearly made it do an extra 100 queries with `Customer.first` instead of just one. We could actually make it even simpler with taking the data returned from our `Country.create!` and sampling from there.

spec/example_app/db/seeds.rb
iarie: Ok, good point, but should we also add create-bang to other models in seeds.rb, i did `create` to preserve uniformity. Also i'll better stick with old definition of US 🤣

iarie:
@BenMorganIO let me know if there is something we need to make/improve

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -66,4 +66,15 @@
...
   end
+
+  describe "foreign_key option" do
+    it "determines what foreign key is used on the relationship for the form" do
+      association =
+        Administrate::Field::BelongsTo.with_options(
+          foreign_key: "customers_uuid",
+        )
+      field = association.new(:customers, [], :show)
+      expect(field.permitted_attribute).to eq("customers_uuid")
```
jcmorrow: Could we break up the phases of this test with newlines? (https://robots.thoughtbot.com/four-phase-test)

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -66,4 +66,15 @@
       end
     end
   end
+
+  describe "foreign_key option" do
+    it "determines what foreign key is used on the relationship for the form" do
+      association =
```
jcmorrow: I think this line spacing is a bit awkward, what about 

```
      association = Administrate::Field::BelongsTo.with_options(
        foreign_key: "customers_uuid",
      )
```

lib/administrate/field/belongs_to.rb
```diff
@@ -8,7 +8,7 @@ def self.permitted_attribute(attr)
       end
 
       def permitted_attribute
-        self.class.permitted_attribute(attribute)
+        foreign_key || self.class.permitted_attribute(attribute)
```
jcmorrow: So, I found this whole `permitted_attribute` instance method calling a class method thing to be hard to follow and reason about. How would you feel about renaming class level implementations of `permitted_attribute` to like `default_attribute` or something like that? Or we could add an instance level `default_attribute` method and put that here instead of calling a class method as a default?

lib/administrate/field/associative.rb
```diff
@@ -24,6 +24,10 @@ def associated_class_name
       def primary_key
         options.fetch(:primary_key, :id)
       end
+
+      def foreign_key
+        options.fetch(:foreign_key, nil)
```
jcmorrow: I'm not a huge fan of the way this reads in combination with line 25 above. Line 25 is handing you a default if not specified (which is what I would expect), but `foreign_key` is instead returning `nil`, and then allowing `permitted_attribute` to implement the default behavior for a `foreign_key`, but it's not clear from this line that that is what is going on. Does that make sense?

lib/administrate/field/belongs_to.rb
iarie: @jcmorrow i ended up with 
```ruby 
def foreign_key
  options.fetch(:foreign_key, :"#{attribute}_id")
end
```
on the instance level of BelongsTo, i found dangerous to rewrite `permitted_attributes` because it's used by BaseDashobard and Deffered classes.

iarie:
CircleCI fails with:
`>> BUNDLE_GEMFILE=/home/ubuntu/administrate/gemfiles/sass_3_4.gemfile bundle exec rake
bundler: failed to load command: rake (/home/ubuntu/administrate/vendor/bundle/ruby/2.4.0/bin/rake)
Gem::LoadError: You have already activated bundler 1.13.7, but your Gemfile requires bundler 1.14.6. Prepending `bundle exec` to your command may solve this.
`
i'm not quite understand why this happening...

nickcharlton:
Hi @iarie!

I think this is almost good to go. Do you think you'd be able to document this in the `docs` directory?

spec/example_app/app/dashboards/country_dashboard.rb
```diff
@@ -0,0 +1,23 @@
+require "administrate/base_dashboard"
+
+class CountryDashboard < Administrate::BaseDashboard
+  ATTRIBUTES = [
```
iarie: @nickcharlton done! And should i fix that dashboard attributes definition according to Hound?

iarie:
rebased against master

nickcharlton:
This looks great! Could I get you to fix up those Hound warnings? I'd done that on a recent one I'd merged in, so it'd be good to start doing that.

iarie:
@nickcharlton fixing that syntax in example_app i thought should we also remake dashboard generators to rely on hound's syntax( add symbols %i array and freeze stuff)? Not here, in another PR?

nickcharlton:
Mmm. As they're not related to this PR, we should fix them up in another.

spec/example_app/app/dashboards/country_dashboard.rb
```diff
@@ -0,0 +1,23 @@
+require "administrate/base_dashboard"
+
+class CountryDashboard < Administrate::BaseDashboard
+  ATTRIBUTES = %i(
+    code
+    name
+  ).freeze
```
nickcharlton: Could we put this all on one line?

spec/example_app/db/seeds.rb
```diff
@@ -12,28 +12,40 @@
 Product.destroy_all
 ProductMetaTag.destroy_all
 Series.destroy_all
+Country.destroy_all
 
-100.times do
+countries = Country.create! [
```
nickcharlton: Should this be indented inside this block?

iarie: @nickcharlton Sorry, i don't understand what do you mean.
Something wrong with brackets?
I can separate attributes declaration from `.create!` like this:
```ruby
country_attributes = [
  { code: "US", name: "USA" },
  { code: "CA", name: "Canada" },
  { code: "CN", name: "China" },
  { code: "RU", name: "Russia" },
  { code: "AU", name: "Australia" },
]

countries = Country.create!(country_attributes)
```

nickcharlton: Oh, oops! Apparently I didn't read the diff correctly.

gemfiles/.bundle/config
```diff
@@ -0,0 +1 @@
+
```
nickcharlton: Hmm, git still seems to think this file exists (but with just a blank line). Could you remove it completely?

nickcharlton:
Excellent, thank you! Merging now.

dan-ding:
curious - why not use reflections to get the foreign key of a relation?

pablobm:
@dan-ding - I think you are right, and we just didn't think of it at the time.

PR #809, opened 2017-04-03
[Docs] Add missing `suffix` documentation for number field

klaseskilson:
In #746, a `suffix` option was added to the number field. However, it was merged without documentation. My bad, here it is!


BenMorganIO:
Thanks!

PR #817, opened 2017-04-04
improve performance for n+1s

BenMorganIO:
This guy solves some n+1 problems in the application. The solution is global. ~I need to take time though to scope the n+1s down to just what attribute constants we're actually using.~

Tried using `COLLECTION_ATTRIBUTES` but that didn't work out very well. Didn't remove a single n+1. As per n+1 issues, I think sticking to `ATTRIBUTE_TYPES` is a good first step. I think we should integrate this with `Administate::Page::Collection`.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -5,6 +5,7 @@ module Administrate
     def index
       search_term = params[:search].to_s.strip
       resources = Administrate::Search.new(resource_resolver, search_term).run
+      resources = resources.includes(*resource_includes) if resource_includes.any?
       resources = order.apply(resources)
       resources = resources.page(params[:page]).per(records_per_page)
       page = Administrate::Page::Collection.new(dashboard, order: order)
@@ -73,11 +74,7 @@ module Administrate
 
     helper_method :nav_link_state
     def nav_link_state(resource)
-      if resource_name.to_s.pluralize == resource.to_s
-        :active
-      else
-        :inactive
-      end
+      resource_name.to_s.pluralize == resource.to_s ? :active : :inactive
     end
 
     helper_method :valid_action?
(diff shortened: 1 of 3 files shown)


kevinelliott:
I would love to see this make it in.

app/controllers/administrate/application_controller.rb
```diff
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
     def index
       search_term = params[:search].to_s.strip
       resources = Administrate::Search.new(resource_resolver, search_term).run
+      resources = resources.includes(*resource_includes) if resource_includes.any?
```
nickcharlton: Let's try that and see how it goes?

app/controllers/administrate/application_controller.rb
```diff
@@ -111,6 +108,18 @@ def find_resource(param)
...
 
+    def resource_includes
+      association_classes = [
+        Administrate::Field::HasMany, Administrate::Field::HasOne,
+        Administrate::Field::BelongsTo
+      ]
+
+      dashboard.class::ATTRIBUTE_TYPES.map do |key, value|
+        key if association_classes.include?(value) ||
+               association_classes.include?(value.try :deferred_class)
```
nickcharlton: This feels a little messy. Maybe we can split the checking here into one line, the `try` to another?

BenMorganIO: Yeah, I think was thinking of doing that. I'm going to take another stab at this.

app/controllers/administrate/application_controller.rb
BenMorganIO: Well, I thought about moving it to the collection class, but that didn't work out the way I thought it would. Instead, it's been moved to the dashboard and we pass the includes from there.

PR #822, opened 2017-04-10
Upgrade webmock to make feature spec compatible with Ruby 2.4

punitcse:


punitcse:
I understand that the gem is set up to run with ruby 2.3.1  [doc](https://github.com/thoughtbot/administrate/blob/master/CONTRIBUTING.md) . But can we still consider this PR as most people want to run with the latest ruby version(2.4).

thanks :bowing_man: .

Gemfile.lock
```diff
@@ -188,6 +189,7 @@ GEM
       slop (~> 3.4)
     pry-rails (0.3.4)
       pry (>= 0.9.10)
+    public_suffix (2.0.5)
```
BenMorganIO: Not a new gem... :p

BenMorganIO:
@punitcase see #793 for us bumping ruby version

Gemfile.lock
punitcse: @BenMorganIO `public_suffix` gem is the dependency for addressable gem. Please refer https://github.com/thoughtbot/administrate/pull/822/files#diff-e79a60dc6b85309ae70a6ea8261eaf95R55 . Do we still want to remove it?

BenMorganIO: Nah, just some tongue and cheek humor from moi. I've been working a lot on making our Gemfile smaller such as removing some rails dependencies we don't use (such as actionmailer) to getting rid of some development dependencies (such as web-console).

PR #823, opened 2017-04-10
Do not redirect to show page when selecting text

### LINKED ISSUE #770: Suggestion: Disable click event on text selection in index tables
Version: 0.4.0

On the index page of a give resource, it is currently not possible to select some information from the table without navigating away from the index page. 

Mouse down+drag to select text+mouse up causes navigation to occur when the intent was only to copy text, and not click. [Examples](https://www.google.ca/search?q=jquery+disable+click+event+when+selecting+text)

One fix would be to change the code in https://github.com/thoughtbot/administrate/blob/v0.4.0/app/assets/javascripts/administrate/components/table.js to:

```javascript
$(function() {
  var keycodes = { space: 32, enter: 13 };

  var visitDataUrl = function(event) {
    if (event.type=="click" ||
        event.keyCode == keycodes.space ||
        event.keyCode == keycodes.enter) {
      var selection = window.getSelection().toString(); // code added
      if (selection.length == 0) {  // code added
        if(!event.target.href) {
          window.location = $(event.target).closest("tr").data("url");
        }
      }
    }
  };

  $("table").on("click", ".table__row", visitDataUrl);
  $("table").on("keydown", ".table__row", visitDataUrl);
});

```

dsolomadin:
This PR fixes the problem when user wants to select text on index page but instead we redirect them to show page of particular record. 

Here is how it works:

screencast 2017-04-10 15-23-43

I've also tried to write a test on this, but looks like Capybara (Poltergeist?) does not support `getSelection` API.

Fixes #770

/r @carlosramireziii
/cc @stevenmilton58

### DIFF
--- a/app/assets/javascripts/administrate/components/table.js
+++ b/app/assets/javascripts/administrate/components/table.js
@@ -11,7 +11,8 @@ $(function() {
       }
 
       var dataUrl = $(event.target).closest("tr").data("url");
-      if (dataUrl) {
+      var selection = window.getSelection().toString();
+      if (selection.length === 0 && dataUrl) {
         window.location = dataUrl;
       }
     }


dsolomadin:
FYI build is broken not due to these changes.

BenMorganIO:
OK, I just tested this. I can confirm, it will transition you to the show page if you select text and then finish on the row. If you select text and navigate the cursor off the row, you won't be transitioned.

Yes, binding of caller has a native extension that's breaking things. Let me see to removing it and then you can rebase.

nickcharlton:
Thanks! Merging.

Issue #829, opened 2017-04-11
Column name must be quoted


sanami:
Search fails in postgres if any column name is not all downcased/

https://github.com/thoughtbot/administrate/blob/master/lib/administrate/search.rb#L23

it should be

```ruby
def query
  search_attributes.map do |attr|
    "lower(#{ ActiveRecord::Base.connection.quote_column_name(attr) }) LIKE ?"
  end.join(" OR ")
end
```

https://www.postgresql.org/docs/current/static/sql-syntax-lexical.html#SQL-SYNTAX-IDENTIFIERS

> Quoting an identifier also makes it case-sensitive, whereas unquoted names are always folded to lower case. For example, the identifiers FOO, foo, and "foo" are considered the same by PostgreSQL, but "Foo" and "FOO" are different from these three and each other. (The folding of unquoted names to lower case in PostgreSQL is incompatible with the SQL standard, which says that unquoted names should be folded to upper case. Thus, foo should be equivalent to "FOO" not "foo" according to the standard. If you want to write portable applications you are advised to always quote a particular name or never quote it.)

BenMorganIO:
@sanami would you like to get a PR in with this fix?

klaseskilson:
On it in #830

PR #830, opened 2017-04-12
Add table name to search query

klaseskilson:
Hi! As always, thank you for a great gem. Ran into a problem with a pretty straightforward fix. Hit me with your feedback! 👍 

## Problem

When default scopes with joins are used, the generated search query can lead to ambiguity. For instance, the column `id` may exist in two tables, which will lead to a SQL error when it is impossible to decide which table to use.

## Solution

Add `resource_class`'s `table_name` to the generated search query. This ensures that the generated query produces the format `table_name.id` instead of `id` in the where clause generated for the search.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -19,9 +19,10 @@ module Administrate
     private
 
     delegate :resource_class, to: :resolver
+    delegate :table_name, to: :resource_class
 
     def query
-      search_attributes.map { |attr| "lower(#{attr}) LIKE ?" }.join(" OR ")
+      search_attributes.map { |attr| "lower(#{table_name}.#{attr}) LIKE ?" }.join(" OR ")
     end
 
     def search_terms
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -19,10 +19,14 @@ module Administrate
     private
 
     delegate :resource_class, to: :resolver
-    delegate :table_name, to: :resource_class
 
     def query
-      search_attributes.map { |attr| "lower(#{table_name}.#{attr}) LIKE ?" }.join(" OR ")
+      search_attributes.map do |attr|
+        table_name = ActiveRecord::Base.connection.
+          quote_table_name(resource_class.table_name)
+        attr_name = ActiveRecord::Base.connection.quote_column_name(attr)
+        "lower(#{table_name}.#{attr_name}) LIKE ?"
+      end.join(" OR ")
     end
 
     def search_terms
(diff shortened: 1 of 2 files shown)


lib/administrate/search.rb
```diff
@@ -19,9 +19,10 @@ def run
     private
 
     delegate :resource_class, to: :resolver
+    delegate :table_name, to: :resource_class
```
klaseskilson: kind of thinking this was a bad idea, and that using `resource_class.table_name` might be a bit more clear. let me know what you think!

lib/administrate/search.rb
```diff
@@ -19,9 +19,10 @@ def run
...
 
     def query
-      search_attributes.map { |attr| "lower(#{attr}) LIKE ?" }.join(" OR ")
+      search_attributes.map { |attr| "lower(#{table_name}.#{attr}) LIKE ?" }.join(" OR ")
```
BenMorganIO: @klaseskilson someone just opened a PR showing that this code doesn't align with the SQL standard in #829. Are you able to quote the table name and, just to leave things in better shape than they were before, the column name as well?

spec/lib/administrate/search_spec.rb
```diff
@@ -43,11 +43,11 @@ class User; end
 
     it "searches using lower() + LIKE for all searchable fields" do
       begin
-        class User; end
+        class User < ActiveRecord::Base; end
         resolver = double(resource_class: User, dashboard_class: MockDashboard)
         search = Administrate::Search.new(resolver, "test")
         expected_query = [
-          "lower(name) LIKE ? OR lower(email) LIKE ?",
+          "lower(users.name) LIKE ? OR lower(users.email) LIKE ?",
```
BenMorganIO: The SQL should probably be:

```sql
lower("users"."name") LIKE ? OR lower("users"."email") LIKE ?
```

lib/administrate/search.rb
BenMorganIO: I would use `resource_class.table_name` because you aren't' frequently changing data types as you chain and you don't expect `resource_class` to become possibly `nil` at any point.

lib/administrate/search.rb
klaseskilson: Sure thing, on it.

klaseskilson:
@BenMorganIO Pushed 63d01d0 with quoting. 👍

BenMorganIO:
fixes #829

BenMorganIO:
Note: fixes https://github.com/thoughtbot/administrate/issues/829

nickcharlton:
Looks great! Thanks!

PR #832, opened 2017-04-13
Fix(column name):Column name must be quoted

### LINKED ISSUE #829: Column name must be quoted

01navneet:
Fixes #829

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -21,7 +21,9 @@ module Administrate
     delegate :resource_class, to: :resolver
 
     def query
-      search_attributes.map { |attr| "lower(#{attr}) LIKE ?" }.join(" OR ")
+  search_attributes.map do |attr|
+    "lower(#{ ActiveRecord::Base.connection.quote_column_name(attr) }) LIKE ?"
+  end.join(" OR ")
     end
 
     def search_terms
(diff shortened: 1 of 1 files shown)


BenMorganIO:
Unfortunately, @klaseskilson already beat you to it in #830.

PR #839, opened 2017-04-22
Add instructions on how to add dashboards

arcticbarra:
This part was already on the docs/getting_started file but it should be on the readme as well to make it easier to find.


BenMorganIO:
Thanks!

PR #841, opened 2017-04-26
when the number data is a float, default to the decimal amount

### LINKED ISSUE #740: Interest in possible new Field::Float component
I have a project with coordinates for tons of locations and their precision matters but it also varies (e.g. some are 4 decimals some are 6). Administrate works fine detecting that I have a float field (Postgres) and uses `Field::Number`. But the problem is it sets `decimals: 2` and if you remove it, it uses a default of 0 so you lose all decimals. I figured this is fine for most numbers but not all the time yet I have no way around it. Since I have data that varies in precision its very important to see its exact value and not round it up or down. I did try using `Field::String` but then that causes search to fail.

I'd be happy to contribute a PR if there is interest in accepting a new `Field::Float` or something like that, where it has inherits or uses the same functionality of `Field::Number` but without the default of 0 for the decimals option. Would that be of interest?

BenMorganIO:
fixes #740 

Calling a class `Number` should probably incorporate integers, floats, and friends. When we want to render a decimal, we want to render a _decimal_. It is a number, but it should be treated fairly. We should always default to render the next half of a number.

### DIFF
--- a/spec/lib/fields/number_spec.rb
+++ b/spec/lib/fields/number_spec.rb
@@ -21,10 +21,7 @@ describe Administrate::Field::Number do
   describe "#to_s" do
     it "defaults to displaying no decimal points" do
       int = Administrate::Field::Number.new(:quantity, 3, :show)
-      float = Administrate::Field::Number.new(:quantity, 3.1415926, :show)
-
       expect(int.to_s).to eq("3")
-      expect(float.to_s).to eq("3")
     end
 
     context "with `prefix` option" do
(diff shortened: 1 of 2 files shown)


spec/lib/fields/number_spec.rb
```diff
@@ -21,10 +21,7 @@
   describe "#to_s" do
     it "defaults to displaying no decimal points" do
```
BenMorganIO: I should update this description...

PR #845, opened 2017-04-29
Remove delayed_job_active_record dependency

ianagne:
The delayed_job_active_record gem was added by suspenders automatically
during the generation of this gem.  The delayed_job_active_record gem is
not currently used.

Of note, delayed_job_active_record does not currently support Rails 5.1.  Removing the dependency has the side bonus of moving us closer to supporting Rails 5.1.

* Remove delayed_job_active_record dependency
* Re-generate appraisal gemfiles to exclude delayed_job_active_record
* Remove delayed_job binstubs
* Remove BackgroundJobs spec support module


pustomytnyk:
"Some checks were not successful"
I've created issue #846 about that test being failed sometimes.

BenMorganIO:
Yessssss one less gem!!!

BenMorganIO:
Alright, I really love this gem, but it looks like that random test failure happened twice in the CricleCI. This leads me to believe that remove DJ may actually be running code async somewhere. I don't like that and it's

nickcharlton:
Thanks!

The tests are failing because of a known flakey test, so I'm going to go ahead and merge anyway.

PR #849, opened 2017-05-01
Use Ruby Methods When Views are KO

BenMorganIO:


pustomytnyk:
I have doubts about fields knowing about controller actions. What should I do if I have added new action to admin controller, say `:refund`, and I want to show same row with fields for a resource as on `:index` page?

BenMorganIO:
@pustomytnyk yeah, I was thinking something similar. But I do think that we should have a way to call certain methods and get information from them. I've been tinkering with CSV exportation and this is definitely a win. A better API to more clear ruby methods would be nice.

Do you have an alternative that you'd prefer?

pustomytnyk:
@BenMorganIO after thinking on this, I suggest to keep those methods in Fields but give them next names: `#to_html_short`, `#to_html_verbose`, `#to_text_short`, `#to_text_verbose`. So CSV export would use `to_text` methods, and `to_html` would either render partial or `to_text` for primitive attributes.

pustomytnyk:
Played with removing partials. Don't like the result. It is not worth updating tests and generators. I would only add `#to_text_short` (for collection csv/json) and `#to_text_verbose` (for item csv/json) to fields.

BenMorganIO:
@pustomytnyk For performance gains, I totally think it's worth it. We probably shouldn't have been testing to the point where we made sure we loaded a file; internal testing. We can keep the generators the same right now and update them later on once we've solidified on an API.

pustomytnyk:
btw, I've experimented with this (here was the idea https://github.com/pustomytnyk/administrate/commit/400f305d0c814ac33a8da350b5c597aa3c254e84), and haven't noticed significant performance gain. Tried to benchmark it, but got very inaccurate values every time. I'll return to this once.

PR #791, opened 2017-03-23
Remove ActionCable and ActionMailer Dependency

BenMorganIO:
We never use this guy and it just adds to a developers' dependencies. Right now, The whole Rails gem still gets loaded, but at least that's changed from 3 gems with Rails as a dependency down to 2. The other two are:

1. ~administrate-field-image (https://github.com/thoughtbot/administrate-field-image/pull/11)~
2. ~markdown-rails~

PR #851, opened 2017-05-02
Remove dependency from ActionMailer

pustomytnyk:
Fix for #791


BenMorganIO:
OMG it's _all_ deleted code!

PR #854, opened 2017-05-03
Remove worker from Procfile

ianagne:
The delayed_job_active_record gem was removed in #845 and the jobs:work
rake task is no longer available.  Remove the worker setup from the
Procfile so `rake jobs:work` will not be called.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,1 @@
 web: bundle exec unicorn -p $PORT -c ./config/unicorn.rb
-worker: bundle exec rake jobs:work


ianagne:
Of note, there's also a migration for the delayed_jobs table that should be removed.  Would it be preferable to delete the migration outright, or should I add a drop_tables migration for posterity?

nickcharlton:
Hi @ianagne!

Thanks for this. Let's do a `drop_table` migration (in another PR).

PR #857, opened 2017-05-04
Abstract model classes should be skipped without warning.

nickcharlton:
This is a rebased version of #582.


nickcharlton:
The tests are failing because we removed DelayedJob in #845. We can probably make an `ApplicationRecord` like Rails 5 does by default instead.

PR #866, opened 2017-05-06
Remove bourbon gem dependency

jonatack:


jonatack:
Apologies, duplicate of #789.

PR #688, opened 2016-11-08
Add support for namespaced models, resolve #179

shvetsovdm:
The changes adds support for one level namespace of models. For example `Catalog::Product` or `Blog::Post` and `Blog::Comment`. Now generators, controllers, views and dashboards work with namespaced models.

I feel that attribute name require some abstraction because knowledge about that  attributes can have '/' spread through whole the codebase. But this is a big changes and I feel that need to show the approach and get feedback.

Thanks.

PR #871, opened 2017-05-07
Add support for namespaced models

shvetsovdm:
This PR is a replacement for the #688 PR and it is partially address the #179 issue.

This changes allow to use namespaced models like `Blog::Post` in Administrate gem.

This is one of the steps to implement a full support for namespaced models in Administrate gem.

Further it is required to implement:

* support associations with namespaced model
* support namespaced model for the dashboard generator
* support namespaced model for the routes generator

## related

spec/lib/administrate/resource_resolver_spec.rb was refactored to not interfere with existed models. The spec is removing constants and it is affects the other specs. This is simple solution for the problem, but it is not removing the root of the problem with this spec. Advices are welcome.


shvetsovdm:
@nickcharlton Hound is "stuck", can we do something with this?

tysongach:
@shvetsovdm If you amend your commit (no code change necessary) and force push, it should kick Hound into gear.

spec/example_app/app/models/blog.rb
```diff
@@ -0,0 +1,5 @@
+module Blog
+  def self.table_name_prefix
+    "blog_"
```
nickcharlton: I've not used namespaced models in quite a while. Is this necessary to override? Will be expecting all of our users to do the same?

spec/lib/administrate/resource_resolver_spec.rb
```diff
@@ -7,23 +7,23 @@
   describe "#dashboard_class" do
     it "handles global-namepsace models" do
```
nickcharlton: It's not in this diff, but that should be "namespace".

spec/lib/administrate/resource_resolver_spec.rb
```diff
@@ -7,23 +7,23 @@
...
-        class UserDashboard; end
-        resolver = Administrate::ResourceResolver.new("admin/users")
+        class AccountDashboard; end
+        resolver = Administrate::ResourceResolver.new("admin/accounts")
 
-        expect(resolver.dashboard_class).to eq(UserDashboard)
+        expect(resolver.dashboard_class).to eq(AccountDashboard)
       ensure
-        remove_constants :UserDashboard
+        remove_constants :AccountDashboard
```
nickcharlton: How come we're replacing `User` with `Account` here?

nickcharlton:
I've left a couple of comments. If you could also address the Hound comments, that'd be great!

spec/example_app/app/models/blog.rb
shvetsovdm: This is produced by the `rails generate model Blog::Post` command. Yes, I expect that most of the users will be do the same.

spec/lib/administrate/resource_resolver_spec.rb
shvetsovdm: `"handles namepsace models"`?
I am not sure. User model in a top/global namespace, I am miss something?

spec/lib/administrate/resource_resolver_spec.rb
shvetsovdm: It's been a while, but it seems like a premature improvement.

I will do this to the way as it was.

spec/example_app/app/dashboards/blog/post_dashboard.rb
```diff
@@ -0,0 +1,33 @@
+require "administrate/base_dashboard"
+
+module Blog
+  class PostDashboard < Administrate::BaseDashboard
+    ATTRIBUTE_TYPES = {
```
shvetsovdm: @nickcharlton  this is a generated dashboard by Administrate itself. It will be nice to fix it in the generator, not in generated files. I can do it in a another PR. What do you think?

spec/example_app/app/models/blog.rb
BenMorganIO: Yep, this is quite common.

martinstiago:
@shvetsovdm @nickcharlton Any updates on this?

shvetsovdm:
@nickcharlton @martinstiago waiting for feedback or merge ✌️

spec/example_app/app/dashboards/blog/post_dashboard.rb
nickcharlton: Ah, I see. Let's leave it like this for now.

nickcharlton:
Thanks for this! I'm going to go ahead and merge.

proveeder:
In case someone would need quickstart instructions:

```shell
rails generate administrate:dashboard BarModule::BazClass --namespace=<custom_namespace>
```

```ruby
  # config/routes.rb

  namespace :custom_namespace do
    resources :foo
    namespace :bar_module do
      resources :baz_class
    end

    root to: 'foo#index'
  end
  ```

PR #872, opened 2017-05-08
Fix build

### LINKED ISSUE #846: spec/generators/dashboard_generator_spec.rb failing randomly?
Check https://circleci.com/gh/ianagne/administrate/37 or https://circleci.com/gh/thoughtbot/administrate/1966

`spec/generators/dashboard_generator_spec.rb:162` seem to fail randomly. Both on CI and locally.

deivid-rodriguez:
* Fixes #846. We need to reset model column information in the `User` model after creating the new `users` table, so that model schema cache is properly refreshed.

* Unblocks #857. There was some leftovers of delayed_job from #845 that were being used to check that namespaced models are not yet supported. I removed them and replaced them with a custom namespaced model to fix the failures #857 is running into. The new model is the same as in #871, so this should make that PR simpler.

* Improves other tiny things in the test setup.


nickcharlton:
Oh, yes! This is wonderful. Merging away.

PR #873, opened 2017-05-08
Add support for cyrillic search

dminchev:
This update adds support for cyrillic and other language search and doesn't change the original search behaviour.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -30,7 +30,7 @@ module Administrate
     end
 
     def search_terms
-      ["%#{term.downcase}%"] * search_attributes.count
+      ["%#{term.mb_chars.downcase}%"] * search_attributes.count
     end
 
     def search_attributes
--- a/spec/lib/administrate/search_spec.rb
+++ b/spec/lib/administrate/search_spec.rb
@@ -59,5 +59,24 @@ describe Administrate::Search do
         remove_constants :User
       end
     end
+
+    it "converts search term lower case for latin and cyrillic strings" do
+      begin
+        class User < ActiveRecord::Base; end
+        resolver = double(resource_class: User, dashboard_class: MockDashboard)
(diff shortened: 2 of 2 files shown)


lib/administrate/search.rb
```diff
@@ -30,7 +30,7 @@ def query
     end
 
     def search_terms
-      ["%#{term.downcase}%"] * search_attributes.count
+      ["%#{term.mb_chars.downcase.to_s}%"] * search_attributes.count
```
BenMorganIO: Looking at this and yes, you shouldn't need `#to_s` since you're interpolating and the return value of `String#downcase` is a string.

nickcharlton:
Hi @dminchev! I think this is good to merge, if you'd be able to address Hound's comments?

dminchev:
Hi @nickcharlton, thanks for the feedback! I'll update it and let you know.

dminchev:
@nickcharlton I've updated the redundant usage and the tests are passing successfully local. However on circleci there is failure which I think is not related to the changes. I'm not able to rerun the circleci build. Can you please check it?

nickcharlton:
Oh, that's interesting. We'd thought that #872 had fixed this. What if you rebase your branch on top of current master? _Hopefully_ it should go away.

dminchev:
I've updated my master branch with the upstream master branch however there is one new violation found by hound.

nickcharlton:
Oh, it looks like you've picked up a bunch of previous commits.

This sometimes happens if the history is slightly older and git can't tell. I'd expect to see just your "Add support for cyrillic search", "Fix redundant use of Object#to_s in interpolation." commits. Would you mind rebasing again?

dminchev:
All should be ok now

nickcharlton:
Perfect! Thanks. Merging.

dminchev:
Great, thanks!

PR #877, opened 2017-05-09
Add appraisal and foreman commands to docs

shvetsovdm:
For contributors that not aware about appraisal and foreman it is hard to know how to run the development server and tests with appraisal.

As a suggestion, it will be good to move the `appraisal install` command to `bin/setup`, but not sure how CI will react on this change.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,22 +10,22 @@ issue trackers, chatrooms, and mailing lists.
 
 1. Fork the repo.
 
-2. Run `./bin/setup`.
+1. Run `./bin/setup`.
 
-3. Run `appraisal install`
+1. Run `appraisal install`
 
-4. Run the tests. We only take pull requests with passing tests, and it's great
+1. Run the tests. We only take pull requests with passing tests, and it's great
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -12,25 +12,31 @@ issue trackers, chatrooms, and mailing lists.
...
-4. Add a test for your change. Only refactoring and documentation changes
+4. Run the tests. We only take pull requests with passing tests, and it's great
+   to know that you have a clean slate: `bundle exec rake && bundle exec appraisal rake`
+
+5. Add a test for your change. Only refactoring and documentation changes
    require no new tests. If you are adding functionality or fixing a bug,
    we need a test!
 
-5. Make the test pass.
+6. Make the test pass.
```
tysongach: While we’re in here, thoughts on swapping all of these to `1`s and having Markdown build the ordered list for us?

shvetsovdm: Nice! I did not know about this feature

shvetsovdm:
@nickcharlton I see that CI do not use `bin/setup`. I think it's good to separate CI commands and those which prepare development environment.

Will we make changes to the `bin/setup` command?

PR #882, opened 2017-05-10
Update normalize.css and change how we include it

tysongach:
- Update normalize.css to v7.0.0.
- Include normalize.css as raw CSS, instead of through RubyGems because
  the normalize-rails gem has fallen much behing the release cycle of
  normalize.css.


BenMorganIO:
I deleted the @HoundCI comments since you marked that file as excluded (probably to avoid more Hound comments).

tysongach:
> Should we place the normalize.css in the vendor folder instead of the app folder?

Good thinking! We don’t have a `vendor` directory in this project, currently. Is that something we can simply add and it will work given that this is a gem/generator? Or will we have to explicitly include it?

BenMorganIO:
@tysongach Rails already loads up and checks for a vendor folder by default. So, you're all good.

nickcharlton:
@BenMorganIO Is that still the case with a Rails Engine, though?

BenMorganIO:
Errr... That's a good question. If we require it from administrate, it should still work... Administrate itself should be aware of a vendor folder. BUT an application should not be fully aware of Administrates vendor folder when it searches a tree. So, I'm going to say yes, but we should test that out. I could be completely wrong.

tysongach:
What’s a good way to test that?

BenMorganIO:
To test? Make sure normalize.css gets included 📦

tysongach:
d4f5fb354030c1b9f7e616aa62db02394ab2dc4b places `normalize.css` in the `vendor` directory.

nickcharlton:
I've tested this on a fresh Rails 5.1 app and it seems fine.

PR #883, opened 2017-05-10
Nested collection links should honour routes

### LINKED ISSUE #790: 500 error - #valid_action? called with the wrong arguments when rendering Field::HasMany
This issue applies to the current, unreleased `master` branch, not to 0.4.0. And first off-- thanks for the awesome gem! Administrate has been a huge boon to my org, and I really appreciate all the hard work.

If I have two Resources with routes like:

```
resources :users
resources :orders, except: [:edit, :update]
```

And `UserDashboard` has `orders: Field::HasMany,` under `SHOW_PAGE_ATTRIBUTES`, when someone navigates to `admin/users/:id` (the show action) Administrate will render, among other things, `administrate/app/views/fields/has_many/_show.html.erb`, which looks like:

```
<% if field.resources.any? %>
  <%= render(
    "collection",
    collection_presenter: field.associated_collection,
    resources: field.resources(params[field.name])
  ) %>
  ...
```

Inside the collection partial, we have:

```
        <% if valid_action? :edit %>
          <td><%= link_to(
            t("administrate.actions.edit"),
            [:edit, namespace, resource],
            class: "action-edit",
          ) %></td>
        <% end %>
```

When we call `valid_action?` here, it looks up `resource_name` on the rendering Controller's `resource_resolver`, which is for the User Resource, *not* the Order Resource.

Since the Order Resource doesn't have an edit route declared in `routes.rb`, a 500 error gets thrown because the path `link_to` goes to look up doesn't exist.

The same issue exists for the other routes, not just `:edit` and `:update`.

- - - 

Stop-gap solution (for anyone else running into this issue):

Put the following in your `routes.rb`, after your Administrate route declarations:

```
    get 'orders/:id/edit', to: 'orders#show', as: :edit_admin_order
```

This deals with the 500s, but the "Edit" links still get generated in collection partials.

- - - 

Actual solution:

I spent about an hour unsuccessfully trying to come up with a fix for this in a PR. It's easy to write a heuristic to pass `resource_name` into `valid_action?` for the Resource being rendered, but I'm very unsure what would be required to make that correct in all cases, or if there's a larger refactor that would be more appropriate. `Administrate::ResourceResolver` defines a clear mapping from the URL path to the Resource, but I'm not sure about going the other way without breaking things in some cases.

karlentwistle:
Prior to this commit if any routes where omitted for a resource like:

```ruby
resources :users
resources :orders, except: [:edit, :update]
```

Navigating directly to a orders show page would correctly omit the edit link but if another associated dashboard utilised a field that rendered using the partial `_collection.html.erb` like:

```ruby
class UserDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    orders: Field::HasMany,
  }

end
```

when navigating to the users show page an error 500 would be thrown because the rendering of orders would incorrectly try to link to the non existent edit route

Fixes #790

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -47,7 +47,8 @@ to display a collection of resources in an HTML table.
           <% end %>
         </th>
       <% end %>
-      <% [valid_action?(:edit), valid_action?(:destroy)].count(true).times do %>
+      <% [valid_action?(:edit, collection_presenter.resource_name),
+          valid_action?(:destroy, collection_presenter.resource_name)].count(true).times do %>
         <th scope="col"></th>
       <% end %>
     </tr>
@@ -57,7 +58,9 @@ to display a collection of resources in an HTML table.
     <% resources.each do |resource| %>
       <tr class="js-table-row"
           tabindex="0"
-          <%= %(role=link data-url=#{polymorphic_path([namespace, resource])}) if valid_action? :show -%>
+          <% if valid_action? :show, collection_presenter.resource_name %>
+            <%= %(role=link data-url=#{polymorphic_path([namespace, resource])}) %>
+          <% end %>
           >
         <% collection_presenter.attributes_for(resource).each do |attribute| %>
           <td class="cell-data cell-data--<%= attribute.html_class %>">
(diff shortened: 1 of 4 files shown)


BenMorganIO:
@karlentwistle very approved! Thank you ❤️

nickcharlton:
Thanks! Merging.

PR #884, opened 2017-05-11
Add `aria-sort` attribute to table headers

tysongach:
Adding the `aria-sort` attribute indicates to assistive technology if
items in a table or grid are sorted in ascending or descending order. It
accepts values of `ascending`, `descending` and `none`.

I’m no Ruby developer, so I have a hunch that this could be 
implemented in a much better way.

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -20,12 +20,10 @@ module Administrate
     end
 
     def sort_order(order)
-      if order == "asc"
-        "ascending"
-      elsif order == "desc"
-        "descending"
-      else
-        "none"
+      case order
+      when "asc" then "ascending"
+      when "desc" then "descending"
+      else "none"
       end
     end
 
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -19,6 +19,16 @@ module Administrate
         )
     end
 
+    def sort_order(order)
+      if order == "asc"
+        "ascending"
+      elsif order == "desc"
+        "descending"
+      else
+        "none"
+      end
+    end
+
     def sanitized_order_params
       params.permit(:search, :id, :order, :page, :per_page, :direction, :orders)
     end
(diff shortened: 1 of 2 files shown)


app/helpers/administrate/application_helper.rb
```diff
@@ -19,6 +19,16 @@ def display_resource_name(resource_name)
...
     end
 
+    def sort_order(order)
+      if order == "asc"
+        "ascending"
+      elsif order == "desc"
+        "descending"
+      else
+        "none"
+      end
```
BenMorganIO: I feel like a `case` statement would be better.

```ruby
def sort_order(order)
  case order
  when "asc" then "ascending"
  when "desc" then "descending"
  else "none"
  end
end
```

🚲 🏠

BenMorganIO:
> I’m no Ruby developer, so I have a hunch that this could be
> implemented in a much better way.

Nope, you got it. But I have a tendency to bikeshed, so left a comment to remove 2 lines of code...

app/helpers/administrate/application_helper.rb
tysongach: 👍  I know there might be a better way!

tysongach: Is it common to indent within a `case`?

```ruby
def sort_order(order)
  case order
    when "asc" then "ascending"
    when "desc" then "descending"
    else "none"
  end
end
```

BenMorganIO: I tend not to because that `end` used for it all. Whitespacing example:

```ruby
case foo
when "bar"
  "baz"
end

case foo
  when "bar"
    "baz"
end
```

Some people don't like the four character whitespace from `"baz"` to the `end`. I think Hound won't like it either. I'm just using an extra ruby keyword to inline them; `then`.

nickcharlton: We usually don't indent, no.

tysongach: Thanks!

tysongach: I updated the helper to use a case statement in 33a0da4233b275f2fda1d48ef9422d65e14fb758. It can be squashed before merging.

Issue #895, opened 2017-05-18
Entire search bar is not clickable


ianagne:
![customers - administrate prototype 2017-05-18 10-33-28](https://cloud.githubusercontent.com/assets/9160106/26210759/47a129fe-3bb6-11e7-8514-01d79f4d740b.png)

This was picked up during QA while testing v0.7.0.  The `.search__hint` blocks a portion of the search bar from being clicked.  Users attempting to activate the search input by clicking on the word "Search" are unable to.  The rest of the bar is clickable.

nickcharlton:
Closing as we've merged #896.

PR #881, opened 2017-05-10
Redesign search

tysongach:
- Moved the search into the main content header to make better use of
  vertical space in the layout.
- Search is now visually (and in the DOM) _after_ the page title, which 
  makes for a more logical and expected flow because it is a page-based
  search.
- Improved accessibility:
  - Navigating the search via keyboard now works in Safari.
  - Meaningful accessible names and labels were added.
  - Include the name of the resource being searched in labels.
- Reduced the CSS needed for search by 44%.
- Remove unnecessary `content_for` helper that wrapped the rendering of
  the search partial.

**Note:** This PR depends on #880 being merged.

## Before

screen shot 2017-05-10 at 16 12 30

## After

screen shot 2017-05-10 at 16 12 18

PR #896, opened 2017-05-18
Make entire search bar clickable

ianagne:
The .search__hint div blocks a portion of the search input.  If the
user clicks around the word "Search" they are unable to activate the
search input.

* Remove z-index from .search__hint before it is active
* Add z-index to .search__hint--active so it only blocks a portion of
the search bar after the user has already clicked the search bar
* Add z-index to .search__input so it is clickable above other elements

### DIFF
--- a/app/assets/stylesheets/administrate/components/_search.scss
+++ b/app/assets/stylesheets/administrate/components/_search.scss
@@ -50,6 +50,7 @@
   flex-grow: 1;
   margin: 0;
   padding: ($small-spacing * 1.5) $small-spacing;
+  z-index: 1;
 
   &:focus {
     box-shadow: none;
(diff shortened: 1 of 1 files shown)


ianagne:
closes #895

nickcharlton:
Nice one, thanks!

ianagne:
That's really cool.  I didn't know about `pointer-events`.  I enjoy that it perfectly describes the intention (i.e. to block `.search__hint` from accepting clicks).

#881 makes this PR obsolete, but I'm glad you shared 👍

lokenx:
It's a rarely used but helpful property, and well supported last I checked.

Yep just saw that PR, like the direction! As long as the full bar is clickable 😉

PR #902, opened 2017-05-23
Fixes to Dutch translation

robinvdijk:
As per title.


tysongach:
@robinvdijk Would you mind taking a look at https://github.com/thoughtbot/administrate/pull/881, as well, and see if that needs any fixing?

PR #910, opened 2017-06-13
Allow custom scope for dashboard resource

bobf:
Based on @nickcharlton 's original branch here:
https://github.com/thoughtbot/administrate/tree/nc-allow-overriding-resource-class

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -105,7 +105,7 @@ module Administrate
     end
 
     def find_resource(param)
-      resource_class.find(param)
+      resource_scope.find(param)
     end
 
     def resource_includes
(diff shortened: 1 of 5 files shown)


nickcharlton:
Thanks for this! I didn't have any specific comments, so I've merged it.

bobf:
@nickcharlton Ah, bad timing ! A colleague of mine has been working on what I consider to be a better implementation. He has moved the scoping into the controller, so you can override a method in each controller to provide the relevant scope. This allows using things like properties of `current_user` to determine the scope. Using the solution I implemented, I had to do some nasty hacks to make `current_user` available in the Dashboard.

I was actually going to close this PR as I think my colleague's implementation is superior. He should have a PR Soon. What do you think is the best way to proceed ?

bobf:
@nickcharlton Just discussed with my colleague - he's going to submit a new PR that reverts my changes and adds his changes instead. I'm optimistic you'll agree that it's a better implementation as it allows the controller to decide the scope, optionally basing it on the request, which I think is much cleaner.

nickcharlton:
Oh dammit!

I look forward to the new PR!

bobf:
@nickcharlton Here ya go: https://github.com/thoughtbot/administrate/pull/914

We're very keen to see this merged so if you have any feedback just let us know and we will act on it ASAP.

Thanks for getting back to us on this !

PR #913, opened 2017-06-19
More flexible searching

jason-uh:
This PR adds a pluggable search feature to Administrate. Instead of a simple boolean, `searchable` can now also take a class that implements `#query/2`, `#search_term`, and `#with_context/1`. There is a [`DefaultSearch`](https://github.com/upper-hand/administrate/blob/make-search-pluggable/lib/administrate/default_search.rb) implementation that other searches can extend.

### Usage
Searches can be labeled with their attribute or, without an attribute, they will be grouped in the "special" `:all` label that will be run against any searchable attribute. Labeled searches will always be preferred to the `:all` label. In addition, labels can be specified more than once to enable searching for multiple values within the attribute. Finally, "op: and" can be used to turn the default "OR" operator to an "AND" operator when creating the outer query conditions (multiple occurrences of a labeled search will always use an "OR" search unless overridden by the search definition).

### Examples:
  On the Line Items page of the example application this search will find
  all line items which are either for The Game of Life product, have a
  unit price greater than or equal to $63.00, or whose order comes to
  a total price greater than or equal to $190.00.

    `product: Game of Life, unit_price: 63, total_price: 190`

  This search is similar to the above but uses "AND" for the outer
  query, and thus requires all of the search terms to be true.

    `op: and, product: Game of Life, unit_price: 63, total_price: 190`


jason-uh:
4 specs are failing in CircleCI, which I cannot reproduce locally. If anyone has suggestions, I'm open to hear them.

```
rspec ./spec/administrate/views/fields/has_one/_index_spec.rb:18 # fields/has_one/_index with an associated record renders a link to the record
rspec ./spec/administrate/views/fields/polymorphic/_index_spec.rb:18 # fields/polymorphic/_index with an associated record renders a link to the record
rspec ./spec/administrate/views/fields/polymorphic/_show_spec.rb:23 # fields/polymorphic/_show with an associated record renders a link to the record
rspec ./spec/administrate/views/fields/has_one/_show_spec.rb:23 # fields/has_one/_show with an associated record renders a link to the record
```

jason-uh:
I just merged from `thoughbot/master` and I'm still getting CI failures for completely unrelated specs. Any suggestions?

nickcharlton:
@jason-uh You can replicate those failing specs by doing:

```sh
appraisal rails42 rspec spec/administrate/views/fields
```

It's a regression from elsewhere, so I'd not worry too much for your specific PR.

jason-uh:
@nickcharlton thanks. Ignoring for now.

lib/administrate/search.rb
```diff
@@ -1,15 +1,112 @@
...
+                           search_attributes,
+                           @term,
+                           resource_class: resource_class,
+                           )
+                         when NilClass
+                           search_strategy_cls = NilSearchStrategy.new
+                         else
+                           raise ArgumentError, "cannot determine search"\
+                             " strategy for a #{term.class}"
+                         end
```
pablobm: I'm surprised that Hound hasn't flagged this section. Particularly lines 96-98.

spec/controllers/admin/line_item_controller_spec.rb
```diff
@@ -2,12 +2,11 @@
 
 describe Admin::LineItemsController, type: :controller do
   describe "GET index" do
-    it "hides the search bar" do
+    it "does not hide the search bar" do
       line_item = create(:line_item)
 
       locals = capture_view_locals { get :index }
-      expect(locals[:show_search_bar]).to be_falsey
+      expect(locals[:show_search_bar]).to be_truthy
```
pablobm: This change means that we don't have any place left to check that the search bar is hiding when expected. Could this be addressed?

lib/administrate/search_term_parser.rb
```diff
@@ -0,0 +1,24 @@
+require "csv"
+
+module Administrate
+  class SearchTermParser
+    def self.parse(term)
+      all_terms = CSV.parse(term).first
```
pablobm: This usage of CSV is not explained. I guess it's done to ensure quotes and commas are parsed correctly. Perhaps add a spec to exemplify this? This way, if a surprised developer removes this in the future, they'll get a failure as an indication of what this is for.

pablobm:

PR #914, opened 2017-06-20
Scoping index action

damau:
Hi

@bobf discussed this and feel this is more of a railsy implementation for scoping the index action.

The purpose is to allow people to override the controller action

```ruby
def scoped_resource
     # Do something special with what's displayed in the index action
end
```

In order to manage what's displayed on the index page.

Which will certainly help Bob and I use roles within administrate.

Ta
Ad

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -105,7 +105,7 @@ module Administrate
     end
 
     def find_resource(param)
-      resource_scope.find(param)
+      resource_class.find(param)
     end
 
     def resource_includes
@@ -117,8 +117,7 @@ module Administrate
         permit(dashboard.permitted_attributes)
     end
 
-    delegate :resource_class, :resource_name, :namespace, :resource_scope,
-             to: :resource_resolver
+    delegate :resource_class, :resource_name, :namespace, to: :resource_resolver
     helper_method :namespace
     helper_method :resource_name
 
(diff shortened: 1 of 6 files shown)


damau:
Rails 4 is failing on master as well due.

bobf:
@nickcharlton FYI I am using this in a real application and it works great. Having scoping applied at the controller level is a huge improvement. Previously I was storing `current_user` in `Thread.current` so that the dashboards could access it, which was horrible at best.

damau:
@nickcharlton any chance of a sneaky merge next week?

nickcharlton:
Hi @damau! This is looking good. I'm going to go ahead and merge!

bobf:
@nickcharlton Awesome ! You just made our lives MUCH easier ! Thanks a lot for this. Glad to be back on `master` :-)

app/controllers/administrate/application_controller.rb
```diff
@@ -117,11 +123,14 @@ def resource_params
         permit(dashboard.permitted_attributes)
     end
 
-    delegate :resource_class, :resource_name, :namespace, :resource_scope,
-             to: :resource_resolver
+    delegate :resource_class, :resource_name, :namespace, to: :resource_resolver
     helper_method :namespace
     helper_method :resource_name
 
+    def dashboard_class
```
pusewicz: This could also be delegated in the lines above, right?

nickcharlton: Looks like it! I'd merge such a PR if you wanted to open one.

damau: Yes I think so.  If I'm honest I'm drawing a blank as to why I even changed the "dashboard" function at all.

damau:
@nickcharlton thanks so much for the merge and all the work you put into the engine!!!
@pusewicz good spot.

PR #916, opened 2017-06-23
Adds _stylesheet to the resources to be generated.

### LINKED ISSUE #888: administrate:views:layout task doesn't generate stylesheet partial
I'm using 0.7.0. The `administrate:views:layout` task doesn't generate `app/views/admin/application/_stylesheet.html.erb` file.

nickcharlton:
This produces a file like the following:

    <%#
    # Stylesheet Partial

    This partial imports the necessary stylesheets on each page.
    By default, it includes the application CSS,
    but each page can define additional CSS sources
    by providing a `content_for(:stylesheet)` block.
    %>

    <% Administrate::Engine.stylesheets.each do |css_path| %>
      <%= stylesheet_link_tag css_path %>
    <% end %>

    <%= yield :stylesheet %>

In `app/views/admin/application/_stylesheet.html.erb`. Fixes #888.

### DIFF
--- a/lib/generators/administrate/views/layout_generator.rb
+++ b/lib/generators/administrate/views/layout_generator.rb
@@ -13,6 +13,7 @@ module Administrate
           )
 
           call_generator("administrate:views:navigation")
+          copy_resource_template("_stylesheet")
           copy_resource_template("_javascript")
           copy_resource_template("_flashes")
         end


professorNim:
@nickcharlton not sure if this is intentional, but are you rendering the stylesheet outside of turbolinks on purpose?

nickcharlton:
@professorNim There's no direct dependency on turbolinks, as we've tried to keep the dependencies small.

MyklClason:
Not related to this, but I wanted to override the layout and found that using `layout :application` and adding:

```
<% Administrate::Engine.stylesheets.each do |css_path| %>
  <%= stylesheet_link_tag css_path %>
<% end %>
```

to the `layouts/application.html.erb` (before the application layouts are loaded so they can be overrode) to work like a charm. As changing the layout messes up the style sheets. So kudos for this.

PR #785, opened 2017-03-17
Add translations for additional strings

buren:
This replaces #478. I had to do this since I removed my fork of administrate while that PR has been open and forking it again didn't "reattach" it to my fork.

PR #918, opened 2017-06-26
Use `display_resource_name` helper in page header

bronislav:
This is one of the steps to reuse i18n everywhere.


bronislav:
This is partial duplicate of #785

PR #920, opened 2017-07-02
Remove NameError rescue in page base

### LINKED ISSUE #480: NoMethodError is silently ignored
Following a previous issue, I added a custom field to my model:

``` ruby
class CurrencyCard < ActiveRecord::Base
  def value
    Money.from_amout(available, currency)
  end
end
```

The above code has a typo. the method's name should be `from_amount` and not `from_amout`.
But the gem is swallowing the error in this rescue: https://github.com/thoughtbot/administrate/commit/351f80ba9668fa59d2d69208a8071b89f428d63f#diff-1524c94a187c1ab4bf8c2aeceebbbc35R25

I get a `NoMethodError` exception, which is a descendant of `NameError.`
So instead of the actual error, I get a `private method 'format' called for nil:NilClass`, because an unexpected `nil` ended up in the view

The commit I referenced adds this to deal with polymorphic associations.
I couldn't find any tests dealing with that particular `rescue`, so I'm not having a lot of luck in fixing this myself

ndrluis:
Fixes #480


ndrluis:
The rescue statement generates a false positive on payments feature specs, the timestamps migrations are created to fix this.

nickcharlton:
This looks good! Could I get you to fix the Hound comment above?

pablobm:
Stale :-( Happy to see this reopen if someone wants to take over.

pablobm:
Actually, I thought better of this one. I'm going to try see it through.

wkirby:
This PR appears to break any support for virtual fields by enforcing the the resource _must_ define a method to respond to all keys in the `Dashboard` class. Maybe this is the desired behavior, but the following use case was possible in `0.12.0`, and was very useful:

```ruby
class ReportDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    id: Field::Number,
    filename: Field::String,
    mimetype: Field::String,
    s3_key: Field::String,
    download: S3DownloadField.with_options({ download_path: :download_lab_report_path }),
    page_count: Field::Number,
    created_at: Field::DateTime,
    updated_at: Field::DateTime
  }.freeze

  COLLECTION_ATTRIBUTES = [
    :id,
    :filename,
    :download,
    :mimetype,
    :created_at
  ].freeze

  SHOW_PAGE_ATTRIBUTES = [
    :id,
    :download,
    :filename,
    :mimetype,
    :s3_key,
    :page_count,
    :created_at,
    :updated_at
  ].freeze

  FORM_ATTRIBUTES = [
    :lab,
    :filename,
    :mimetype,
    :s3_key,
    :page_count
  ].freeze
end
```

```ruby
class S3DownloadField < Administrate::Field::Base
  def to_s
    data
  end

  def downloadable?
    options.key?(:download_path)
  end

  def destroyable?
    options.key?(:destroy_path)
  end

  def destroy_path(field, attachment)
    path_helper(:destroy_path, field, attachment)
  end

  def download_path(field, attachment)
    path_helper(:download_path, field, attachment)
  end

  private

  def path_helper(key, field, _attachment)
    path_helper = options.fetch(key)
    Rails.application.routes.url_helpers.send(path_helper, field.resource)
  end
end
```

This allowed us to inject a download link for our S3 resource _without_ editing the underlying views, or modifying the display for any of the actual attributes.

Is there any advice on how to fix this _without_ defining useless methods on our models?

sedubois:
@wkirby I also filed regressions from this PR here:
https://github.com/thoughtbot/administrate/issues/1570

pablobm:
@wkirby - Thank you for your report. I have filed a separate issue at https://github.com/thoughtbot/administrate/issues/1586

PR #921, opened 2017-07-04
Allow sorting by specified 'order' option

jondkinney:
I'm sure this code could be cleaned up, and there are likely edge cases, but I wanted to get this out here to start a discussion as to whether or not people would find this useful? Here are my two use cases: 

1. I want to present one of the attributes differently so I create a new method on the model called, for example, `display_phone`, where I do some formatting of the sanitized stored phone number data. The problem is that once I do that, I can't sort by that filed because the db only knows about `phone` not `display_phone`. With this change I can specify an `order: 'phone'` option to the `display_phone` field on the dashboard and allow that pretty displayed data to be sorted by its underlying db column. 

        phone: Field::String,
        display_phone: Field::String.with_options(
          searchable: false,
          order: 'phone',
        ),

1. I am displaying a list of Clinics that belong to an Organization and I would like to be able to sort the clinic list by this associated model's `name` field. By allowing `order: 'organization.name'` to be specified, we can detect that we're sorting across a join table with the presence of a `.` and join to that table and order by the specified attribute. One trick/bummer here is that with this implementation, we need to leave the plurality of the join table name to be specified by the user. So in this case with a Clinic#index view where a Clinic belongs to an Organization it'd be a singular `organization.name` specified.

        organization: Field::BelongsTo.with_options(
          order: 'organization.name',
        ),

### DIFF
--- a/lib/generators/administrate/views/layout_generator.rb
+++ b/lib/generators/administrate/views/layout_generator.rb
@@ -13,6 +13,7 @@ module Administrate
           )
 
           call_generator("administrate:views:navigation")
+          copy_resource_template("_stylesheet")
           copy_resource_template("_javascript")
           copy_resource_template("_flashes")
         end


jondkinney:
The specs that are failing on circle do not fail for me locally. Not sure what's up with that.

PR #923, opened 2017-07-04
Fix 'found unpermitted param orders' dynamically

jondkinney:
Hard coding `:orders` here gets the tests to pass, but is naive. Unless the intention is to override the `sanitized_order_params` for each app in the `admin/application_controller.rb` file and manually specify every new resource that is added?

Instead, what is working for me is a dynamic list of any possible `has_many` resource names in the system as shown in this PR. Lack of specifying these permitted params will result in the following error messages.

`ActionController::UnpermittedParameters in Admin::Customers#show`

`found unpermitted parameter: :orders`


nickcharlton:
This looks good! Do you think there's a way we can write a spec for this?

jondkinney:
There are existing tests that check whether or not the `:orders` param is allowed. If you remove this code some specs will fail. Are you looking for something that more directly tests this implementation?

nickcharlton:
Yeah, I'd like to see something which fails without us providing `resources` like is done here.

jondkinney:
@nickcharlton a tests already exists that will fail. Without `:orders` hard coded, and without the splatting of the resources the following failure occurs

`rspec ./spec/features/show_page_spec.rb:26`

```
     ActionView::Template::Error:
       found unpermitted parameter: :orders
```

With the dynamically provided splatted resource names, it's all green. So I think we're good? Thanks!

nickcharlton:
I think that we should test the implementation in some form of unit test and not rely on the feature spec to catch it.

There's already an existing helper spec in: `spec/helpers/administrate/application_helper_spec.rb`, which would be a great place to start in.

jondkinney:
@nickcharlton I'll be revisiting this soon. I have a few PRs that I'll get tidy'd up either this weekend or next week.

nickcharlton:
@jondkinney, is this something you'd be able to look at again?

pablobm:

PR #924, opened 2017-07-04
Replace dashboard_class method with a delegate

pusewicz:
@nickcharlton as per #914–`dashboard_class` is a method on the `resource_resolver`, so we can delegate to that object instead of defining another method.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -123,7 +123,8 @@ module Administrate
         permit(dashboard.permitted_attributes)
     end
 
-    delegate :dashboard_class, :resource_class, :resource_name, :namespace, to: :resource_resolver
+    delegate :dashboard_class, :resource_class, :resource_name, :namespace,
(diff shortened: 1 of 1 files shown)


app/controllers/administrate/application_controller.rb
```diff
@@ -123,14 +123,10 @@ def resource_params
         permit(dashboard.permitted_attributes)
     end
 
-    delegate :resource_class, :resource_name, :namespace, to: :resource_resolver
+    delegate :dashboard_class, :resource_class, :resource_name, :namespace, to: :resource_resolver
```
pusewicz: Should I split it into multiple lines?

nickcharlton: Yes.

nickcharlton:
This looks great! Once the style comment is resolved, I'll go ahead and merge it.

pusewicz:
@nickcharlton Updated!

nickcharlton:
Great thanks! Merging!

PR #926, opened 2017-07-06
Accepts string instead of symbol on order

### LINKED ISSUE #341: Suggestion: Allowing ordering to accept more options
I have to order the index page by default by an attribute `quality`, which can be `NULL` or any value from 0 to 10.

``` ruby
module Admin
  class OrganizationsController < Admin::ApplicationController

    private

    def order
      @_order ||= if params[:order]
        Administrate::Order.new(params[:order], params[:direction])
      else
        Administrate::Order.new('quality', 'desc NULLS last')
      end
    end
  end
end
```

If I order with `desc` (highest quality first), `NULL`'s will come first. Postgres has a nice syntax to control whether `NULL`'s should come first or not: `Organization.all.order('quality DESC NULLS last')`.

However, I can't use the code posted above because AR is not aware of this Postgres syntax when attribute and order is passed in as a key / value like so: https://github.com/thoughtbot/administrate/blob/master/lib/administrate/order.rb#L10. I get the following error: `Direction "desc NULLS last" is invalid. Valid directions are: [:asc, :desc, :ASC, :DESC, "asc", "desc", "ASC", "DESC"]`

However, if we changed this line to accept any SQL ordering (without AR validating it), it would work: `relation.order("#{attribute} #{direction}")`

ndrluis:
This makes custom ordering possible

E.g.

 Administrate::Order.new('quality', 'desc NULLS last')

Fixes #341

### DIFF
--- a/spec/lib/administrate/order_spec.rb
+++ b/spec/lib/administrate/order_spec.rb
@@ -36,7 +36,7 @@ describe Administrate::Order do
 
         ordered = order.apply(relation)
 
-        expect(relation).to have_received(:order).with(name: :asc)
+        expect(relation).to have_received(:order).with("name asc")
         expect(ordered).to eq(relation)
       end
 
@@ -47,7 +47,7 @@ describe Administrate::Order do
 
         ordered = order.apply(relation)
 
-        expect(relation).to have_received(:order).with(name: :desc)
+        expect(relation).to have_received(:order).with("name desc")
         expect(ordered).to eq(relation)
       end
     end
--- a/lib/administrate/order.rb
+++ b/lib/administrate/order.rb
@@ -7,7 +7,7 @@ module Administrate
 
     def apply(relation)
       if relation.columns_hash.keys.include?(attribute.to_s)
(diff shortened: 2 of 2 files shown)


spec/lib/administrate/order_spec.rb
```diff
@@ -36,7 +36,7 @@
 
         ordered = order.apply(relation)
 
-        expect(relation).to have_received(:order).with(name: :asc)
+        expect(relation).to have_received(:order).with("name asc")
```
nickcharlton: Do you think there would be any value in testing for both sorts? (i.e.: string _and_ symbols)

ndrluis: I think that would be redundant

nickcharlton: Yeah, I think I'm with you on that. I was curious to see if we were testing the old option as well as the new, but I think we're good.

nickcharlton:
Thanks @ndrluis! Merging away.

PR #927, opened 2017-07-07
Fix specs with Rails 4.2

odlp:
Via this [`rspec-rails` issue](https://github.com/rspec/rspec-rails/issues/1644), I discovered [this issue](https://github.com/rails/rails-controller-testing/issues/24) where `rails-controller-testing` is breaking URL helpers in view tests.

This was [fixed](https://github.com/rails/rails-controller-testing/commit/e56cee404a67a812514e95f92282632ada1912e9) in `rails-controller-testing` v1.0.2 but unfortunately we can't use this in the `rails42.gemfile` because the v1.0.2 version depends on ActionView / ActionPack at v5.x.

Removing the `rails-controller-testing` gem and rewriting the two affected specs which used `assert_render_template` fixed the failures on my machine.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,7 +28,6 @@ group :test do
   gem "formulaic"
   gem "launchy"
   gem "poltergeist"
-  gem "rails-controller-testing"
   gem "shoulda-matchers", "~> 2.8.0", require: false
   gem "timecop"
   gem "webmock"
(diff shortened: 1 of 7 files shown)


nickcharlton:
Thanks! This has been bugging me for ages!

PR #928, opened 2017-07-09
Fix partial name in view customization docs

carlosramireziii:
The docs for customizing page views was referencing the wrong partial name for the index view's collection.


nickcharlton:
Thanks @carlosramireziii!

PR #931, opened 2017-07-14
Date and Time fields

bronislav:
This adds date and time fields. #741

Continuation of #757


spec/example_app/db/migrate/20170224010823_add_birthdate_to_customer.rb
```diff
@@ -0,0 +1,5 @@
+class AddBirthdateToCustomer < ActiveRecord::Migration[4.2]
```
nickcharlton: Could you set the migration to `5.1`? We've been doing that for recent migrations.

spec/example_app/db/migrate/20170224024549_add_example_time_to_customer.rb
```diff
@@ -0,0 +1,5 @@
+class AddExampleTimeToCustomer < ActiveRecord::Migration[4.2]
+  def change
+    add_column :customers, :example_time, :time
```
nickcharlton: What do you think about doing `birth_date` and `birth_time` to be consistent?

(It's the closest I can think which might make sense and avoid us having an `example_time` field.)

spec/lib/fields/date_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
+require "administrate/field/date"
+# require "support/field_matchers"
```
nickcharlton: Could we lose this comment?

nickcharlton:
This is looking good! Do you think you'd also be able to contribute some docs, too? We've got a list of included fields which this should be part of in `docs/customizing_dashboards.md`.

bronislav:
@nickcharlton Sure. I will update docs

There is still one failing spec with which I'm fighting. It's related to the timezone handling. I plan to fix it this weekend.

bronislav:
@nickcharlton I've added fields to the docs, but there is still one issue with date picker. Datepicker tries to show time in the local timezone when original value is in UTC.

nickcharlton:
Is that because the time being stored doesn't have a time zone, or is it losing the timezone? Or, is it just doing the Rails thing of assuming you'd want it in a local time?

I think as a user I'd expect it to display in the same way I'd stored it, but that makes it difficult if we're not storing that data.

bronislav:
@nickcharlton We are storing it in the UTC and display on the form in the UTC. But, datetime_picker shows it in the local timezone, which brings inconsistency.

From my perspective, I'd say that we don't need to save timezone and show exactly save value as saved. I wasn't managed to configure datetime_picker to not add timezone to the value. Still trying.

bronislav:
@nickcharlton I won't work on this in the nearest feature. I don't use Rails for a while now. If anyone would like to take ownership on this or feel free to just close it.

Issue #932, opened 2017-07-14
Normalize asset file not included in gem


gchaturvedi:
Since the change to put the normalize css file in the assets subdirectory, the asset isn't included in the gem.  It has to be manually added into your own Rails app under vendor.

tysongach:
Yep, I’m seeing the same thing. This PR is relevant: https://github.com/thoughtbot/administrate/pull/882

![localhost-5000- laptop with hidpi screen](https://user-images.githubusercontent.com/903327/28217939-f9f97e94-6884-11e7-9f8b-14593a22261b.png)

dsolomadin:
@nickcharlton would be really nice if you could follow up with new release

tysongach:
@dsolomadin We will; only been 10 minutes ;)

dsolomadin:
cool, no pressure guys 😄

nickcharlton:
I released 0.8.1 which fixes this: https://github.com/thoughtbot/administrate/releases/tag/v0.8.1

PR #935, opened 2017-07-14
Move normalize.css out of vendor

tysongach:
Having this in vendor resulted in an importing error.

Fixes https://github.com/thoughtbot/administrate/issues/932

I tested this locally with [Hound](https://houndci.com/) by pointing it to this branch and it appears to have fixed the problem.


nickcharlton:
Thanks @tysongach!

gchaturvedi:
Wow, thank you for addressing this so quickly.  Much appreciated.

PR #936, opened 2017-07-15
Refactor: Extract Search::Query object

koppen:
This is a step towards making it easier to integrate the solution built by @nando in #276.

Adding collection scopes/filters means we need to add more involved search query parsing; this pull request gives us a place for that, instead of cluttering the `Administrate::Search` class with string-parsing capabilities.

This pull request changes no public behaviours.


koppen:
The changes here are included in #947, which is the pull request they were made for. Closing this one.

PR #939, opened 2017-07-18
add order option for belongs_to field

MittchoBaroco:
add the posibility to order by one or more columns

- edit candidate_resources method
- add tests
- edit the doc

### DIFF
--- a/spec/lib/fields/belongs_to_spec.rb
+++ b/spec/lib/fields/belongs_to_spec.rb
@@ -66,4 +66,19 @@ describe Administrate::Field::BelongsTo do
       end
     end
   end
+
+  describe "#resources" do
+    context "with `order` option" do
+      it "returns the resources in correct order" do
+        5.times{ FactoryGirl.create(:customer) }
+        options = { order: "name" }
+        association = Administrate::Field::BelongsTo.with_options(options)
+        field = association.new(:customers, [], :view)
+
+        correct_order = Customer.all.order("name").map(&:id)
+
+        expect(field.associated_resource_options.drop(1).map {|row| row[1]}).to eq correct_order
+      end
+    end
+  end
 end
--- a/spec/lib/fields/belongs_to_spec.rb
+++ b/spec/lib/fields/belongs_to_spec.rb
@@ -70,14 +70,14 @@ describe Administrate::Field::BelongsTo do
   describe "#resources" do
     context "with `order` option" do
       it "returns the resources in correct order" do
-        5.times{ FactoryGirl.create(:customer) }
+        5.times { FactoryGirl.create(:customer) }
         options = { order: "name" }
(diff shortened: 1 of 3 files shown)


MittchoBaroco:
updated version of #386 with tests and docs

MittchoBaroco:
updated version of #386 with tests and docs

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -66,4 +66,19 @@
       end
     end
   end
+
+  describe "#resources" do
+    context "with `order` option" do
+      it "returns the resources in correct order" do
+        5.times { FactoryGirl.create(:customer) }
```
nickcharlton: We could use `create_list` here, like: `create_list(:customer, 5)`, which will create 5 customers.

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -66,4 +66,19 @@
...
+        options = { order: "name" }
+        association = Administrate::Field::BelongsTo.with_options(options)
+        field = association.new(:customers, [], :view)
+
+        correct_order = Customer.all.order("name").map(&:id)
```
nickcharlton: Could we use `pluck` here, instead of calling `map`?

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -66,4 +66,19 @@
...
+        resources = field.associated_resource_options.drop(1).map { |r| r[1] }
```
nickcharlton: I wonder if this line is a little complex. Do you think we'd be able to simplify it a little?

MittchoBaroco: I have updated it for `resources = field.associated_resource_options.compact.to_h.values` it's the simplest that I can think of.

the only way I found to test the data was to use the associated_resource_options method which returns an array like that: [nil, ["John doo", 1], ["master Yoda", 2}]

MittchoBaroco:
updated 😄

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -66,4 +66,20 @@
...
+
+  describe "#resources" do
+    context "with `order` option" do
+      it "returns the resources in correct order" do
+        FactoryGirl.create_list(:customer, 5)
+        options = { order: "name" }
+        association = Administrate::Field::BelongsTo.with_options(options)
+        field = association.new(:customers, [], :view)
+
+        correct_order = Customer.all.order("name").pluck(:id)
```
nickcharlton: We won't need to do `.all` here, as `.order` will be sufficient.

MittchoBaroco:
@nickcharlton done 😄

nickcharlton:
Great, thanks! I’m going to merge.

PR #945, opened 2017-07-21
Sort has_many fields

pauldub:
Hi!

I'm opening this pull request to disscuss how sorting `has_many` dashboard fields could be implemented. This is the implementation we are using but it could be done better.

I'm copying the text from the issue I opened with comments regarding improvements:

It is currently not possible to sort an HasMany field on the model's show view such as http://administrate-prototype.herokuapp.com/admin/customers/9085?direction=asc&order=address_state. There was a similar issue with its pagination which was fixed with release 0.5.0.

I have implemented it on a fork but I think it would require a bit more work. My implementation is based on the work done in PR #736

I suggest that we use a convention of #{attr_name}[param], for example orders[page] for the pagination and orders[direction] & orders[order] for sorting. But the issue is that there is currently no proper way to do this except by adding more arguments to the Page::Collection methods. Maybe a better solution would be to be able to propagate request parameters through Page#attribute_field overriding options for the field. It would be consistent with the HasMany field implementation and its pagination could be implemented that way too (having page as an option).

What are your thoughts about this? I am willing to push this (small) feature as it is a great quality of life improvement and follows the extensibility goals set by the project.

### DIFF
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -20,17 +20,17 @@ from the associated resource class's dashboard.
 
 <% if field.resources.any? %>
   <% order = field.order_from_params(params.fetch(field.name, {})) %>
-  <% page = params.fetch(field.name, {}).fetch(:page, nil) %>
+  <% page_number = params.fetch(field.name, {}).fetch(:page, nil) %>
   <%= render(
     "collection",
     collection_presenter: field.associated_collection(order),
     collection_field_name: field.name,
-    resources: field.resources(page, order),
+    page: page,
+    resources: field.resources(page_number, order),
     table_title: field.name,
-	association_includes: association_includes,
   ) %>
   <% if field.more_than_limit? %>
-    <%= paginate field.resources(page), param_name: "#{field.name}[page]" %>
+    <%= paginate field.resources(page_number), param_name: "#{field.name}[page]" %>
   <% end %>
 
 <% else %>
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -26,7 +26,8 @@ from the associated resource class's dashboard.
     collection_presenter: field.associated_collection(order),
     collection_field_name: field.name,
     resources: field.resources(page, order),
-    table_title: field.name
+    table_title: field.name,
+	association_includes: association_includes,
   ) %>
   <% if field.more_than_limit? %>
     <%= paginate field.resources(page), param_name: "#{field.name}[page]" %>
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -22,12 +22,12 @@ from the associated resource class's dashboard.
   <%= render(
     "collection",
     collection_presenter: field.associated_collection,
-    resources: field.resources(params[field.name]),
+    resources: field.resources(params.fetch(field.name, {}).fetch(:page, nil))
     table_title: field.name
   ) %>
 
   <% if field.more_than_limit? %>
-    <%= paginate field.resources(params[field.name]), param_name: "#{field.name}" %>
+    <%= paginate field.resources(params.fetch(field.name, {}).fetch(:page, nil)), param_name: "#{field.name}[page]" %>
   <% end %>
 
 <% else %>
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -25,7 +25,7 @@ from the associated resource class's dashboard.
     "collection",
     collection_presenter: field.associated_collection(order),
     collection_field_name: field.name,
-    resources: field.resources(page, order)
+    resources: field.resources(page, order),
     table_title: field.name
   ) %>
   <% if field.more_than_limit? %>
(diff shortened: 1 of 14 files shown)


pauldub:
Will rebase on master

pauldub:
Hi, any thoughts on this? I'd like to know whether it will ever get merged or not as maintaining the fork is quite cumbersome.

app/views/fields/has_many/_show.html.erb
```diff
@@ -19,15 +19,17 @@ from the associated resource class's dashboard.
 %>
 
 <% if field.resources.any? %>
+  <% order = field.order_from_params(params.fetch(field.name, {})) %>
+  <% page = params.fetch(field.name, {}).fetch(:page, nil) %>
```
pablobm: Hm, ideally we shouldn't be using `params` here, but instead be something that we pass down from the controller in some way. Here we would access them, eg. as properties of `field`.

But then again, it's not really your change that is doing that, as we were already doing that before :-/

nickcharlton:
Hi @pauldub! Do you think you'd be able to contribute the test @pablobm mentions above?

pauldub:
Hi @nickcharlton I will probably be able to do it within a few days

dukex:
Hi @pauldub Can I help you with something about this PR?

pauldub:
@pablobm regarding your comment 

> At first it's not clear why this uses query params in the format ${association}[order]=${attribute}, so this spec would help clarifying that should anyone try to simplify.

it uses this format in order to inform the controller of which association is being sorted. Unfortunately it does not allow sorting multiple associations on the same page which would require creating sort links by merging existing params.

I've tried to implement this feature with the smallest surface possible in order to avoid adding too much complexity to a system that will probably be refactored at some point to support such use-cases.

I've just rebased on administrate/master and will fix the failing test as well as add new tests for the show page (I think the existing search spec somewhat cover the other case). If you have any suggestion on how to proceed with this, you're welcome!

pablobm:
@pauldub - Sorry, I wasn't very clear on my comment.

When I reviewed the code, it took me a bit to understand the format of the query params. It was the sort of thing that looks overkill on a first pass, but then makes complete sense after looking more closely. I'm 100% happy with your implementation.

Since I got that initial impression, my concern was that others may have it too, and may try to "simplify" it in the future, this time without the context provided by this PR. Therefore, I think there should be a test that will break should a misguided future developer attempt this.

This test probably should be an acceptance test on a page that has two of these lists. Other things may work too; I'm open to alternatives.

pauldub:
That's great, this is the kind of tests I've started to write, I'll update the PR today.

pauldub:
There is one thing that is not supported with the current PR: sorting multiple associations on the same page concurrently. 

It could be done, but we would have to give `#sanitized_order_params` the list of associations that could be sorted, which might be tricky. 

IMO this is a nitpick because current sort link generation does not merge previous parameters and end up sorting only one association at a time, but it could be addressed in a further PR

pablobm:
@pauldub I just had a look and ran your code. I can sort log_entries and orders independently, and the feature spec works for me. Did you fix this since the time you said this wasn't working?

pauldub:
Yes I have pushed a fix, because it would also return a strong parameter exception. I'm not sure the fix is good tho, it essentially pass the dashboard's association_includes to the page in order to expose them in `#sanitized_order_params`

pablobm:
@pauldub - I think your solution is good. If anything, I'm thinking we should just pass the whole `page` object down, rather than `association_includes`. In my opinion, when reading the code it's easier to see "this partial needs more information about the page. OK, that's fair enough" than to see "this partial needs 'association_includes', what is that?".

I have experimented with my approach in my own repo. Here are the differences from your branch: https://github.com/tymate/administrate/compare/6baab4f...pablobm:tymate-sort_has_many

If you do that (no need to merge from mine, just do what's easier for you), and rebase on top of master, this may be good enough to merge at long last.

@nickcharlton thoughts?

pauldub:
@pablobm Yes, passing the whole `page` object makes more sense, I'll update the PR with that change. Hopefully we can see this merged!

pablobm:
@pauldub Thank you for that. I think it's good. Don't worry about the CI issues, it's a known problem.

If you can just fix those two Hound warnings, I think @nickcharlton will agree that this can be finally merged.

pablobm:
I got @nickcharlton's blessing. Let this be merged.

jondkinney:
@pablobm @pauldub @nickcharlton what are the implications of this change? I get errors about unpermitted parameters for order and direction now when using this trick in a controller:

```ruby
    def default_params
      params[:order] ||= 'name'
      params[:direction] ||= 'asc'
    end
```

It'd be great if PRs came with some documentation for what's changing and how to take advantage of the new feature(s). A ton of time is spent when working with Administrate to dig into the source, PR discussion, issue discussion, etc. to figure out how things work.

Also, does this PR allow for ordering by a belongs_to association at all? If I'm on a (soccer) team index dashboard and want to list the club name that each team belongs to, and want to order the list asc or desc by the belongs_to (club name), is that possible with this code? Doesn't look like it, but wanted to double check.

pablobm:
We are tracking down that issue with unpermitted parameters.

Re: ordering by a belongs_to attribute, that unfortunately doesn't work at the moment. Only scalar fields. Eventually we'll have to either provide a way to allow that or to remove the interface elements that suggest it's possible.

Issue #938, opened 2017-07-17
How to rename dashboard


rozhok:
By default it generates name from the model name, but I want to pass i18n'ed name here or whatever thing I like. Where could I do this?

Thanks.

pustomytnyk:
@rozhok do you want to pass argument to the generator or to show different name in sidebar?

rozhok:
I dunno what is generator.

I have name in the sidebar, name in the header and name in the searchbar. I want to localize them.

pustomytnyk:
if you check gem's source, you'll see that it uses http://guides.rubyonrails.org/i18n.html#translations-for-active-record-models

rozhok:
Cool, thanks.
The only lasting question is localizing of search string (it won't work with separate words in Russian for instance, I need to localize whole sentence)

nickcharlton:
Could you provide an example of what that might look like?

rozhok:
For instance, I have model named Cat, search string placeholder will be "Search cats"
If I localize it, it would look like "Кіт" (model name), "Шукати Коти" (search string placeholder), but it should be "Шукати Котів" because regular plural form can't be used here.

nickcharlton:
I didn't really understand, so I asked one of the developers at the client I'm working with at the moment to try and explain (he speaks Russian and not Ukrainian, but that was close enough for us to figure it out!).

From what I now understand:

* Depending on context there are different kinds of pluralisation.
* This means that you can't do "cats" and "search for cats", as pluralising for the second one should be a different word.

This lead us to this issue: https://stackoverflow.com/questions/12488705/incorrect-pluralizing-of-model-in-rails  
And this commit: https://github.com/rails/rails/commit/c300dca9963bda78b8f358dbcb59cabcdc5e1dc9

Anyway. The recommendation I got was to hard code the translation (assuming that this is practical). If there's not an easy way to do this, we should look at changing the way we localise strings to make it possible. There may or may not already be some open issues and/or PRs for this. I'm not really sure as I didn't until today really understand it!

rozhok:
I can't hardcode because it will change dashboard name.
So, what I'd like to do something like is:
Model – "Cat" (`activerecord.model.cat.one: Cat`)
Dashboard name – "Cats" (`activerecord.model.cat.other: Cats`)
Search string – "Search for kitties" <--- how do I hardcode this?

rozhok:
[...]

PR #946, opened 2017-07-22
Document renaming dashboard

rozhok:
#938

Add docs of how to i18n dashboard name.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -158,6 +158,17 @@ end
 
 [define your own]: /adding_custom_field_types
 
+To change the dashboard name in sidebar menu, sub-header and search string use default ActiveRecord i18n translations for models:
+
+```yaml
+en:
+  activerecord:
+    models:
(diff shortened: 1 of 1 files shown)


docs/customizing_dashboards.md
```diff
@@ -158,6 +158,17 @@ end
 
 [define your own]: /adding_custom_field_types
 
+To change dashboard name in sidebar menu, sub-header and search string simply use default ActiveRecord i18n translations for models:
```
nickcharlton: I think we need some "the"s in the start here:

> To change _the_ Dashboard name in _the_ sidebar

And would you mind removing "simply"? Whilst in this case it's true, I like to avoid words in our docs like "simply", "easily", etc as this is often not true for everyone.

rozhok:
Sorry for broken English. Fixed.

nickcharlton:
No worries, it always takes a bit to get the wording right. I'm going to merge this now, so thanks!

PR #956, opened 2017-08-01
add namespace option to generators

### LINKED ISSUE #754: Add `namespace` option to generators
https://github.com/thoughtbot/administrate/pull/339 lets Administrate work with namespaces other than `admin`, but the generators are still hard-coded to use that namespace.

The generators should either accept an optional namespace argument, or read the namespace from a configuration file or initializer.

Rodrigora:
fixes #754 

With this change, we'll be able to run the generators with the namespace option:

```
rails generate administrate:install --namespace manager
rails generate administrate:dashboard --namespace manager
```

### DIFF
--- a/spec/generators/install_generator_spec.rb
+++ b/spec/generators/install_generator_spec.rb
@@ -4,7 +4,9 @@ require "support/generator_spec_helpers"
 require "support/constant_helpers"
 
 describe Administrate::Generators::InstallGenerator, :generator do
-  describe "admin/application_controller" do
+  after { reset_routes }
+
+  describe "application_controller" do
     it "is copied to the application" do
       stub_generator_dependencies
       controller = file("app/controllers/admin/application_controller.rb")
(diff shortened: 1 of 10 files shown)


spec/generators/install_generator_spec.rb
```diff
@@ -18,35 +20,59 @@ module Admin
...
-        expect(Rails::Generators).not_to invoke_generator("administrate:routes")
-      ensure
-        reset_routes
+      expect(Rails::Generators).not_to invoke_generator(
+        "administrate:routes", ["--namespace", "admin"]
+      )
+    end
+
+    it "does not invoke routes generator if namespace routes already exist \
+    when using custom namespace" do
```
nickcharlton: Do you think we'd be able to use a `context` here, instead of having a super long `it`?

spec/generators/install_generator_spec.rb
```diff
@@ -18,35 +20,59 @@ module Admin
...
       RB
     end
+
+    it "uses the namespace option" do
+      stub_generator_dependencies
+      controller = file("app/controllers/manager/application_controller.rb")
+
+      run_generator ["--namespace", "manager"]
+
+      expect(controller).to exist
```
nickcharlton: Do we need to check it exists if we're also ensuring it looks right?

Rodrigora: I like this verification as it will raise a proper error message when the file doesn't exist (`expected "/home/rodrigo/Projects/administrate/tmp/app/controllers/manager/application_controller.rb" to exist`)

If we use only the `have_correct_syntax` matcher and the generator fails to create the file, the error message would be `No such file or directory @ rb_sysopen - /home/rodrigo/Projects/administrate/tmp/app/controllers/manager/application_controller.rb`

What do you think?

spec/generators/install_generator_spec.rb
Rodrigora: Sure

spec/generators/install_generator_spec.rb
nickcharlton: That's a good reason! Let's leave it in there.

Rodrigora:
rebased

Rodrigora:
Seems that it was fixed by https://github.com/thoughtbot/administrate/pull/339

nickcharlton:
Hi @Rodrigora! I noticed you closed this, but I don't quite think we've solved it.

Looking at the current `master`, we don't have an option for the generator (but everything else seems to be ready). This issue is relevant,too: #754

Issue #960, opened 2017-08-09
Dashboard file still created if errors occur


lun4i:
Tiny issue for you guys =)
Situation:
after
`rails g administrate:dashboard SomeModel`
I had an _'uninitialized constant SomeModel'_ error
However, it was my fault, becouse i named my model like this: 'Somemodel'

The dashboard files were created no mater what errors occurred in generator, IMO this should be fixed

nickcharlton:
That sounds like a good idea. Do you think you'd be able to contribute a PR for this?

lun4i:
@nickcharlton Sounds great! I think I'll handle this

nickcharlton:
Wonderful! I look forward to seeing your PR!

das3in:
Took a stab at this as my first non-documentation open source contribution - wrote about it in the commit message but was having some trouble with the test. Unsure about how to properly test for raising a `NameError`

nickcharlton:
Hi @das3in! That's off to a good start!

You should be able to do something like:

```ruby
expect { ... }.to raise_error(NameError)
```

Or, if there's a message with it:

```ruby
expect { ... }.to raise_error(NameError, "The name was invalid…")
```

das3in:
@nickcharlton - Yeah I thought that'd work too and it's still not passing that test. The spec is in the PR I linked if you could see if maybe I was doing something else wrong.

nickcharlton:
Ah, I see what you need to do. You need to wrap your `run_generator` call around the `expect` block so that it can catch the exception, a bit like this:

```ruby
expect do
  run_generator ["invalid_model"]
end.to raise_error(NameError)
```

purinkle:
@lun4i, is this still an issue for you?

PR #966, opened 2017-08-26
No dashboard generation when NameError #960

das3in:
When using the `rails g administrate:dashboard` generator with an
uninitialized model name, the method will raise a NameError but the
`app/dashboards/#{foo}_dashboard.rb` file will be created regardless.
This happens because the generator uses the built-in `file_name`
parameter to create the file name regardless of whether or not the rest
of the generator runs.

To solve this, we simply check whether or not `klass` is truthy before
actually generating the file because this is the same guard that we use for the
rest of the generator.

I did have a problem writing the test case for this. I expect the `run_generator` method to raise and expect the file to not exist. It's interesting because... it definitely does raise a `NameError` but it raises before the test expectations happen. Could use some direction there

### DIFF
--- a/spec/generators/dashboard_generator_spec.rb
+++ b/spec/generators/dashboard_generator_spec.rb
@@ -18,6 +18,16 @@ describe Administrate::Generators::DashboardGenerator, :generator do
       expect(dashboard).to have_correct_syntax
     end
 
+    it "should not create file with uninitialized model" do
+      dashboard = file("app/dashboards/customer_dashboard.rb")
+
+      expect do
+        run_generator ["invalid_model"]
+      end.to raise_error(NameError)
+
+      expect(dashboard).not_to exist
+    end
+
     describe "#attribute_types" do
       it "includes standard model attributes" do
         begin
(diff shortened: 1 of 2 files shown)


.ruby-version
```diff
@@ -1 +1 @@
-ruby-2.4.1
+2.4.0
```
nickcharlton: Could we remove this from this PR?

spec/example_app/db/schema.rb
```diff
@@ -23,38 +23,22 @@
     t.datetime "updated_at", null: false
   end
 
-  create_table "countries", force: :cascade do |t|
```
nickcharlton: As we didn't make any changes to the schema, could we drop this out, too?

In a future PR we can rerun migrations and ensure this is tidied up.

spec/generators/dashboard_generator_spec.rb
```diff
@@ -18,6 +18,15 @@
       expect(dashboard).to have_correct_syntax
     end
 
+    it "should not create file with uninitialized model" do
+      dashboard = file("app/dashboards/customer_dashboard.rb")
+
+      run_generator ["invalid_model"]
+
+      expect(run_generator).to raise_error(NameError)
```
nickcharlton: The right way to get this to work is to wrap `run_generator` in an `expect { … }`, as the execution of line 24 is what'll throw the exception. Something like this:

```ruby
expect do
  run_generator ["invalid_model"]
end.to raise_error(NameError)
```

pablobm:

PR #971, opened 2017-09-13
Add authorization features and a Pundit mixin

pedantic-git:
This PR adds basic *authorization* features to Administrate, allowing different admins access to different data/actions.

Users are free to roll their own authorization, or use any authz gems they like, but since I personally think [Pundit](https://github.com/elabs/pundit) is the best of the bunch, I've implemented a controller concern that allows users to get Pundit support in one line.

Pundit isn't added as a dependency, except for the tests.

I've added an authorization.md file to the `docs/` with full instructions... let me know what you think and if you'd like any changes!

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -106,7 +106,9 @@ module Administrate
     end
 
     def requested_resource
-      @_requested_resource ||= authorize_resource(find_resource(params[:id]))
+      @_requested_resource ||= find_resource(params[:id]).tap do |resource|
+        authorize_resource(resource)
+      end
     end
 
     def find_resource(param)
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -27,8 +27,10 @@ module Administrate
     end
 
     def new
+      resource = resource_class.new
+      authorize_resource(resource)
       render locals: {
-        page: Administrate::Page::Form.new(dashboard, resource_class.new),
+        page: Administrate::Page::Form.new(dashboard, resource),
       }
     end
 
@@ -40,6 +42,7 @@ module Administrate
 
     def create
       resource = resource_class.new(resource_params)
+      authorize_resource(resource)
 
       if resource.save
         redirect_to(
@@ -71,7 +74,7 @@ module Administrate
       flash[:notice] = translate_with_resource("destroy.success")
       redirect_to action: :index
     end
-
+ 
     private
 
     helper_method :nav_link_state
@@ -104,6 +107,8 @@ module Administrate
 
     def requested_resource
       @_requested_resource ||= find_resource(params[:id])
+      authorize_resource(@_requested_resource)
+      @_requested_resource
     end
 
     def find_resource(param)
@@ -145,5 +150,22 @@ module Administrate
         dashboard.collection_attributes
       ).any? { |_name, attribute| attribute.searchable? }
     end
+    
+    # Override to conditionally remove action links
+    def show_action?(action, resource)
+      true
+    end
+    helper_method :show_action?
+    
+    # How to instantiate a blank instance of this controller's resource
+    def new_resource
+      resource_class.new
+    end
+    helper_method :new_resource
+    
+    # Raise an exception if the resource isn't authorized
+    def authorize_resource(resource)
+    end
+    
   end
 end
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -74,7 +74,7 @@ module Administrate
       flash[:notice] = translate_with_resource("destroy.success")
       redirect_to action: :index
     end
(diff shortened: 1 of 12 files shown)


app/controllers/administrate/application_controller.rb
```diff
@@ -104,6 +107,8 @@ def dashboard
 
     def requested_resource
       @_requested_resource ||= find_resource(params[:id])
+      authorize_resource(@_requested_resource)
+      @_requested_resource
     end
```
pablobm: This breaks the existing memoization. Here's an alternative, saving us from (potentially) checking the authorization multiple times:

```ruby
def requested_resource
  @_requested_resource ||= find_resource(params[:id]).tap do |r|
    authorize_resource(r)
  end
end
```

But then I think... we could go further. If the spec for `authorize_resource` required it to return the resource on success, we could do this:

```ruby
def requested_resource
  @_requested_resource ||= authorize_resource(find_resource(params[:id]))
end
```

In fact that's what Pundit's own `authorize` does, which in turn means that the implementation provided by `Punditize` does this, although not the default, blank implementation.

But I'm probably complicating this too much for too little gain, so I'll be happy with something along the lines of my first alternative.

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+      create_list :order, 2, customer: user
+    end
+    
+    # Policies are defined in order_policy.rb
+    describe 'GET index' do
+      it 'shows only the records in the admin scope' do
+        locals = capture_view_locals { get :index }
+        expect(locals[:resources].count).to eq(9) # only my orders
+      end
+    end
```
pablobm: Please separate these with blank lines. I'm surprised that Hound is not complaining about this and the single quotes?

nickcharlton:
Thanks @pedantic-git! I know @pablobm left some comments, but I'm thinking it'd be good to get this into `0.9`, which should be in a few weeks time (i.e.: the next release).

pedantic-git:
Sure! I'll try to get onto these tomorrow. Just working on *another* Administrate PR right now (trying to get editing working for Field::Polymorphic... nearly there too!)

spec/example_app/app/policies/application_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+    @user = user
+    @record = record
+  end
+
+  def index?
+    false
+  end
+
+  def show?
+    scope.where(:id => record.id).exists?
```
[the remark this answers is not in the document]
pedantic-git: I'd rather not address this comment because application_policy.rb is included verbatim as generated by the "pundit:install" generator.

app/controllers/administrate/application_controller.rb
```diff
@@ -145,5 +148,20 @@ def show_search_bar?
         dashboard.collection_attributes
       ).any? { |_name, attribute| attribute.searchable? }
     end
+
+    def show_action?(action, resource)
```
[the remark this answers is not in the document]
pedantic-git: Not addressing this - it's a method signature that is meant to be overridden in subclasses so the arguments and their names are important.

pedantic-git:
OK! I've addressed all but two of the Hound violations (see above for my reasoning), I've changed the way `authorize_resource` works (thanks for spotting this, @pablobm!) and I've removed the comments.

[Removing comments is still something I struggle with - I can understand other elements of house style that I disagree with but this one always seems very counterintuitive to me.]

Let me know if I can do anything else!

pedantic-git:
Just realized that Pundit 1.1.0 (the latest gem) doesn't return the object from `authorize` - it returns `true`. Shocking there hasn't been a gem release in over a year, but in the meantime I'll change the `requested_resource` to use tap instead.

app/views/administrate/application/edit.html.erb
```diff
@@ -27,7 +27,7 @@ It displays a header, and renders the `_form` partial to do the heavy lifting.
       "#{t("administrate.actions.show")} #{page.page_title}",
       [namespace, page.resource],
       class: "button",
-    ) if valid_action? :show %>
+    ) if valid_action?(:show) && show_action?(:show, page.resource) %>
```
pustomytnyk: Is it possible to use only `show_action?` in all cases?
Like
```ruby
  valid_action?(action, resource) && authorized?(action, resource)
end
```

pedantic-git: Hi @pustomytnyk - `#show_action?` is a method that users of administrate can (and should) override in their controllers. If we implement it in terms of `#valid_action?` in the base controller and again in the Pundit mixin, there's still opportunity for administrate users to accidentally break the expected behaviour by overriding it to something that's not.

nickcharlton:
Thanks, @pedantic-git! I'm (finally!) going to merge this now.

pedantic-git:
Yay! Thanks @nickcharlton !!

PR #972, opened 2017-09-17
Fix interpolation highlighting at documentation examples

harrykiselev:
Hi, thoughtbot's!

When I read [the documentation](http://administrate-prototype.herokuapp.com/adding_custom_field_types) of this gem I had face a bug of interpolation styles at code examples.

Returned string here use interpolation for digest hash.

before

I have researched a process of parsing by highlight.js library and structure of default highlight.js styles to know that text color within interpolation class and main text of code block must be equal.

This PR fix this issue. It would be useful for another documentation blocks of code in a future that will use interpolation.

after

### DIFF
--- a/app/assets/stylesheets/docs.scss
+++ b/app/assets/stylesheets/docs.scss
@@ -80,6 +80,7 @@ code {
     padding-left: 1rem;
 
     .hljs-string { color: $code-green; }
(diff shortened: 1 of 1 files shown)


nickcharlton:
Nice catch, thank you! Merging away.

PR #976, opened 2017-09-21
Explain that installer requires models

pablobm:
I ran the installer on a pristine Rails setup and was confused when /admin yielded a 404.


nickcharlton:
Thanks!

PR #981, opened 2017-09-26
Add localization for Bosnian language

nedimdz:


nickcharlton:
Hi @nedimdz! Thank you for your language contribution!

Something I'm trying to do is have someone keep maintaining the languages they contribute, because if one us doesn't speak it then they end up getting either left behind (and breaking the build), or filled in with not so great Google Translate answers. Do you think you'd be able to do this?

nedimdz:
Yes, I will be glad if I can help the popularization of the Bosnian language

nickcharlton:
Fantastic! I'm going to merge this now, too.

PR #982, opened 2017-09-27
Update name of specification example.

harrykiselev:
Hi, thoughtbot's!

Name of specifications example don't corresponds code inside it.

Probably is a "copy/paste" mistake because example with the same name is follow below but it in a right context ("when provided a namespaced dashboard").

Thanks,
Harry :)


nickcharlton:
Perfect! Nice catch.

PR #984, opened 2017-09-28
Add support for editing polymorphic fields

pedantic-git:
This PR allows users to edit `Field::Polymorphic` fields (previously they were just told this was not possible).

The `<select>` is a grouped select, populated using the contents of zero-or-more model classes specified in the `:classes` option.

I appreciate I haven't really added much in the way of testing for this, but I think it's comparable with the amount of testing the other complex built-in types already have (e.g. `Field::HasMany`).

The one place in the core that's changed here and might be controversial, is that `#resource_params` now auto-converts anything that looks like a GlobalID. This is necessary for a polymorphic field without JavaScript as GlobalIDs are the only IDs that work across multiple models. Let me know if you need this implementing a different way.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -120,7 +120,14 @@ module Administrate
 
     def resource_params
       params.require(resource_class.model_name.param_key).
-        permit(dashboard.permitted_attributes).transform_values {|v|  v.respond_to?(:starts_with?) && v.starts_with?("gid://") ? GlobalID::Locator.locate(v) : v }
+        permit(dashboard.permitted_attributes).
+        transform_values do |v|
+          if v.respond_to?(:starts_with?) && v.starts_with?("gid://")
+            GlobalID::Locator.locate(v)
+          else
+            v
+          end
+        end
     end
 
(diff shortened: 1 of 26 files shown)


ewilliam:
any plans to implement this? or any polymorphic field solution?

pablobm:
@pedantic-git - Thank you for your PR. I've been having a look at this and I like the general direction. You are right that the implementation of `#resource_params` is a pickle, so I gave it some thought. I think I have with a good solution now, and I have implemented it on a separate PR https://github.com/thoughtbot/administrate/pull/1020, on top of your work.

The main change is in how to deal with this global id. Instead of relying on the global id only, I split the field in two:

  * `my_field[value]`: the global id
  * `my_field[type]`: literally `"polymorphic"`

Then `#resource_params` can look at the param values and treat them appropriately. If it's a simple value such a string, it does nothing special. However, if it comes in this value/type form, we treat it as polymorphic, reading the global id.

Something to note is that the `globalid` gem is not a default in old versions of Rails. Therefore I had to add it as a dependency. I could have done without it, but I liked it when I saw it on your code.

Additionally, I added some tests. I may add a couple more, but this does pretty well for now.

Closing this as it's superseded by the new PR.

PR #987, opened 2017-09-28
Fix cell sort indicator positioning

### LINKED ISSUE #864: Fix overlapping icon
This bug appears to only happen when clicking (and therefore sorting by) right-aligned labels/cells.

screen shot 2017-05-05 at 21 48 31

lookitsatravis:
Happy Hacktoberfest!

To fix #864, we float the element instead of using absolute positioning with a slight margin for fields with long labels.

Please let me know if you need anything else. Cheers!

### DIFF
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -26,10 +26,9 @@
 }
 
 .cell-label__sort-indicator {
-  display: inline-block;
   overflow: hidden;
-  position: absolute;
-  right: 0;
+  float: right;
+  margin-left: 5px;
 
   svg {
     fill: $hint-grey;
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -1,6 +1,4 @@
 .cell-label {
-  position: relative;
-
   &:hover {
     a {
       color: $action-color;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/administrate/components/_cells.scss
```diff
@@ -26,10 +26,9 @@
 }
 
 .cell-label__sort-indicator {
-  display: inline-block;
   overflow: hidden;
```
tysongach: Is this `overflow` needed anymore?

app/assets/stylesheets/administrate/components/_cells.scss
```diff
@@ -26,10 +26,9 @@
...
-  position: absolute;
```
tysongach: Now that we are no longer using absolute positioning, can the `position: relative;` go away on Line 2?

app/assets/stylesheets/administrate/components/_cells.scss
lookitsatravis: I don't believe so, I'll remove.

app/assets/stylesheets/administrate/components/_cells.scss
lookitsatravis: I think you're correct. I'll update it.

tysongach:
Thanks, @lookitsatravis!

Issue #720, opened 2016-12-27
DashboardGenerator does not update routes


mcmire:
Apologies if someone else has already submitted this.

If you use `rails g administrate:dashboard` to create a new dashboard file, the new dashboard is not added to `config/routes.rb`, and thus it does not appear automatically in the sidebar. I realize this is probably on purpose; after all, how is the generator going to know where in the routes file it needs to add a new line, especially if you've rearranged things? If that's the case, then I feel like the generator needs to display some kind of instructions when you run it to inform the user that they need to do this manually. For instance, if you ran `rails g administrate:dashboard Foo`, you would see something like:

> As you can see above, Administrate created files for the new dashboard. However, right now you won't see a new item in the sidebar. To make that happen, you'll need to add a new line to your existing admin-related routes. For instance:
>
>     resources :foos

rnice01:
Is this still open for PRs? I'm new to the Rails community is it common for the routes.rb file to be in a different folder? Seems like a lot of stuff in the install generator process relies on routes.rb to be in the config folder and not just the routes generator.

nickcharlton:
Hi @rnice01!

Sure is!

It might be in a different place if you're using a lot of Rails Engines and so administrate (an Engine itself) might not be able to determine where to put things.

rnicedep:
@nickcharlton Thanks for getting back to me, I'd love to work on something for this. After looking at the install and routes generator, it looks like both of them are looking directly at the /config directory from the Rails root for the routes.rb file. 

My idea for this PR is to create a generator helper method to first check where the routes.rb would be, like it is now, and if it can't find it then do a quick search through the project's directories from the Rails root for the routes.rb file.  This could then be used in any generator needing to find the routes.rb file like the install and routes generators. What are your thoughts for this?

nickcharlton:
That sounds like a good idea! I look forward to seeing it.

pablobm:
I'm going to close this one as there hasn't been any activity in a long time.

PR #991, opened 2017-09-30
Find routes config

rnice01:
PR for issue #720 : Both the Install and Routes generators were looking for the project's routes.rb file in /config from the Rail's root path. Which the helper method I added will look for first, if it cannot find the routes.rb file there, then it will search through the project until it finds the path for that file.

### DIFF
--- a/lib/administrate/generator_helpers.rb
+++ b/lib/administrate/generator_helpers.rb
@@ -4,10 +4,31 @@ module Administrate
       Rails::Generators.invoke(generator, args, generator_options)
     end
 
+    def find_routes_file
+      routes_file = Rails.root.join("config/routes.rb")
+      if File.exist?(routes_file)
+        routes_file = find_file("routes.rb")
+        raise "Unable to locate your routes.rb file" if routes_file.nil?
+      end
+      routes_file
+    end
+
     private
 
     def generator_options
       { behavior: behavior }
+    end
+
+    def find_file(file_name)
+      file = nil
+      Find.find(Rails.root) do |path|
+        Find.prune if path.include? ".git"
+
+        next unless path.include? file_name
+
+        file = path
+      end
+      file
     end
   end
 end
--- a/spec/lib/administrate/generator_helpers_spec.rb
+++ b/spec/lib/administrate/generator_helpers_spec.rb
@@ -1,0 +1,15 @@
+require "rails_helper"
+
+RSpec.describe Administrate::GeneratorHelpers do
+  include Administrate::GeneratorHelpers
+
+  describe "#find_routes_file" do
+    it "returns the typical path for a Rails routes file if found" do
+      typical_routes_path = "config/routes.rb"
+
+      returned_path = find_routes_file
+
+      expect(returned_path.to_s).to include(typical_routes_path)
+    end
+  end
+end
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -13,7 +13,7 @@ module Administrate
       def run_routes_generator
         if dashboard_resources.none?
           call_generator("administrate:routes", "--namespace", namespace)
-          load Rails.root.join("config/routes.rb")
+          load find_routes_file
         end
       end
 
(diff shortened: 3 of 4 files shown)


spec/lib/administrate/generator_helpers_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+include Administrate::GeneratorHelpers
+
+RSpec.describe Administrate::GeneratorHelpers do
+  describe "#find_routes_file" do
+  	before (:all) do
```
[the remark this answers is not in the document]
nickcharlton: Could you remove the space between the `before` and `(:all)`, too?

lib/administrate/generator_helpers.rb
```diff
@@ -4,10 +4,29 @@ def call_generator(generator, *args)
       Rails::Generators.invoke(generator, args, generator_options)
     end
 
+    def find_routes_file
+      routes_file = Rails.root.join("config/routes.rb")
+      if !File.file?(routes_file)
+        routes_file = find_file("routes.rb")
+        raise "Unable to locate your routes.rb file" unless !routes_file.nil?
```
nickcharlton: How do you think about doing this like:

`raise "…" if routes_file.nil?`

It saves us from checking for a negative negative (`unless not`).

lib/administrate/generator_helpers.rb
```diff
@@ -4,10 +4,29 @@ def call_generator(generator, *args)
       Rails::Generators.invoke(generator, args, generator_options)
     end
 
+    def find_routes_file
+      routes_file = Rails.root.join("config/routes.rb")
+      if !File.file?(routes_file)
```
nickcharlton: There's a `exist?` method on `File` which might avoid us flipping with `!`. What do you think?

lib/administrate/generator_helpers.rb
```diff
@@ -4,10 +4,29 @@ def call_generator(generator, *args)
...
     private
 
     def generator_options
       { behavior: behavior }
     end
+
+    def find_file(file_name)
+      file = nil
+      Find.find(Rails.root) do |path|
+        Find.prune if path.include? ".git"
```
nickcharlton: Do you think there could ever be other similar directories we should be ignoring here? I wonder if we could be more selective with what we're looking for to help avoid this check. What do you think?

lib/administrate/generator_helpers.rb
rnice01: Sounds great, I'll update that.

lib/administrate/generator_helpers.rb
rnice01: I think your suggested changes reads a lot nicer, I'll update that.

lib/administrate/generator_helpers.rb
rnice01: Hmm now that I think about it, we should probably avoid vendor, test, spec, and public, maybe db too?

I can't really think of another way to avoid the check unless maybe we only look through the entire config and app tree and then "give up" if routes.rb is not there. What are your thoughts on that?

nickcharlton: Yeah, I'd agree on both of those. I'd probably create an array of possible excludes, add those to a constant and then check that for presence.

nickcharlton:
Hi @rnice01!

Sorry, it's been a while. I just answered your question left above. Do you think you'd be able to rebase against current master and fix up the remaining comments?

purinkle:
@rnice01, is this a change you still want to make?

rnice01:
@purinkle, yeah I can get it cleaned up.

lib/administrate/generator_helpers.rb
```diff
@@ -4,10 +4,29 @@ def call_generator(generator, *args)
...
+
     private
 
     def generator_options
       { behavior: behavior }
     end
+
+    def find_file(file_name)
+      file = nil
+      Find.find(Rails.root) do |path|
```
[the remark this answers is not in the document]
nickcharlton: This is not what you think it is, Hound. There isn't a `detect` on this module.

Issue #486, opened 2016-02-27
List parameters on "Customizing attribute partials" docs page


djsegal:
[The docs](http://administrate-prototype.herokuapp.com/adding_custom_field_types) refer to custom field type partials, but they don't list them out.

Also, the command `rails generate administrate:views:field` without any parameters errors out. Maybe it should generate all the defined custom fields, which are:

---
- belongs_to
- boolean
- date_time
- email
- has_many
- has_one
- image
- number
- polymorphic
- string
- text

graysonwright:
@djsegal, thanks! It may take me a while to get to this - feel free to open a pull request if you'd like to speed it up. All of the documentation lives in the `/docs` folder.

Otherwise, I'll take a look when my schedule frees up, after March 7.

djsegal:
@graysonwright, just following up. are you still able to do this?

thecalliopecrashed:
Seems related to #1108?

nickcharlton:
Closing as we merged #992.

PR #992, opened 2017-10-01
Field generators

ryanhertz:
This addresses https://github.com/thoughtbot/administrate/issues/486. Instead of copying the partials for all field types when no argument is given as suggested in the issue, I added support for passing "all" in place of the field name.


nickcharlton:
This is looking great! I'm going to merge this in now!

PR #993, opened 2017-10-02
Fix overlapping icon

radvc:
Fix for this issue https://github.com/thoughtbot/administrate/issues/864. I remove the right property and add left, top property in `.cell-label__sort-indicator` class


tysongach:
This is a duplicate of https://github.com/thoughtbot/administrate/pull/987. I think using floating is a more robust approach, as well.

radvc:
@tysongach sorry I didn't see.

Issue #860, opened 2017-05-05
Improve error state visual design


tysongach:
The flash messages could be improved to better fit within the layout, as well as group together multiple errors and their label/description.

![screen shot 2017-05-05 at 14 14 21](https://cloud.githubusercontent.com/assets/903327/25758592/2c951af8-319d-11e7-8303-35cb561c9dd4.png)

PapePathe:
+1

admir23:
Hey @tysongach ,
I've would like to tackle this please :)

Thanks
Admir

PR #994, opened 2017-10-02
Group multiple errors

radvc:
Reference to this issue https://github.com/thoughtbot/administrate/issues/860, just let me know what are the things that I need to add. Thanks


purinkle:
@radvc, do you still want to make this change?

PR #1002, opened 2017-10-11
Improve german translation.

hanneskaeufler:
German translation of "Clear search", "Search Resource" and "New resource" did not make any sense.

In order to be able to improve the german translation of "New post",
the string contatenation of "new" + resource had to be moved into
the i18n string instead of being performed inside the template. This
will help other locales as well I assume. To keep the current
behaviour for all other locales, I appended "%{resource}".

### DIFF
--- a/config/locales/administrate.ar.yml
+++ b/config/locales/administrate.ar.yml
@@ -5,7 +5,8 @@ ar:
       confirm: "هل أنت متأكد ؟"
       destroy: "حذف"
       edit: "تعديل"
-      show: "إظهار"
+      edit_item: "تعديل %{item}"
+      show: "إظهار %{item}"
       new: "جديد  %{resource}"
       back: "الى الخلف"
     controller:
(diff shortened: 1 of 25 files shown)


config/locales/administrate.ar.yml
```diff
@@ -5,8 +5,9 @@ ar:
       confirm: "هل أنت متأكد ؟"
       destroy: "حذف"
       edit: "تعديل"
-      show: "إظهار"
-      new: "جديد"
+      edit_item: "تعديل %{item}"
+      show: "إظهار %{item}"
+      new: "جديد  %{resource}"
```
buren: hm.. 🤔 This actually might be the other way around, name/item before the string... (but I don't know, since I personally don't speak arabic could check with a friend who does though).

nickcharlton: I had a chat with a colleague about this and apparently it is correct. You'll read it right-to-left, but because it should be rendered that way it'll make sense.

I'd be happy to be shown we're wrong though!

hanneskaeufler:
Closing this in favor of a new one as suggested in #1004

PR #1004, opened 2017-10-16
Add and improve I18n for application templates

buren:
I finally got around replacing/updating #785, which adds better I18n support for administrate.

---

- [x] Removes "hardcoded" placement of resource name for edit/new/show links (problematic on Arabic and other right-to-left languages since the positioning of the name will be flipped too).
- [x] Update `config/locale/administrate.en.yml` 
- [x] Includes the latest commit from `master`
- [x] Use `display_resource_name` consistently when calling `page.resource_name` in view templates
- [x] Update all other locale files
    * Any comments/suggestions on how I best should go about this?
    * If I use `i18n-tasks add-missing` other updates (formatting etc) are made in the locale files
    * Should we try to re-use the existing strings or do we deem that dangerous and think its better that translations are re-done by someone who actually understands the language? 😅 

⚠️ There is one expected test failure for `rspec ./spec/i18n_spec.rb:9 # I18n does not have missing keys` (see TODO)

### DIFF
--- a/app/views/administrate/application/_form.html.erb
+++ b/app/views/administrate/application/_form.html.erb
@@ -18,8 +18,11 @@ and renders all form fields for a resource's editable attributes.
   <% if page.resource.errors.any? %>
     <div id="error_explanation">
       <h2>
-        <%= pluralize(page.resource.errors.count, "error") %>
-        prohibited this <%= page.resource_name %> from being saved:
+        <%= t(
+          "administrate.form.errors",
+          pluralized_errors: pluralize(page.resource.errors.count, t("administrate.form.error")),
+          resource_name: display_resource_name(page.resource_name)
+        ) %>
       </h2>
 
       <ul>
--- a/app/views/administrate/application/edit.html.erb
+++ b/app/views/administrate/application/edit.html.erb
@@ -15,7 +15,7 @@ It displays a header, and renders the `_form` partial to do the heavy lifting.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Page/Form
 %>
 
-<% content_for(:title) { "#{t("administrate.actions.edit")} #{page.page_title}" } %>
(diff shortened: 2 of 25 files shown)


hanneskaeufler:
I improved the german translation in a similar pr. See #1002. To preserve the existing behaviour for all other locales I reused the existing translation and tackes on the resource token. But surely this is not optimal for many more locales other then german as well.

buren:
*tl;dr* Updated locales, tests are passing 🎉 


I've updated non-default locales now I ended up including most of the locale changes @hanneskaeufler did in #1002. Thanks for that! Saved me a lot of time my editor didn't play nice with the arabic strings :)

@hanneskaeufler I didn't include your German translation update (keys: `search.label`, `search.clear`). Thought that would be better to put in a separate PR. 

@nickcharlton let me know if there are any changes you want and if I should squash all the commits together as one :)

nickcharlton:
Hi @buren!

This is looking good to me. I like the way the commits are split up, so I'm not going to squash them here.

What do you think about me merging #1002, then you rebasing on top of that? It'll solve the separate PR issue you mention (I think?).

buren:
@nickcharlton perhaps merging this and then just opening a new PR with just the 2-line German fix? If @hanneskaeufler is unavailable to do it, I can :) Then the I18n fix for German would be as clean as possibly can be.

hanneskaeufler:
Sure go ahead and merge this and I'll update/reopen something similar to #1002.

buren:
@nickcharlton something blocking this? :)

Can I do anything to help?

nickcharlton:
@buren Just my time to hit the magic button!

I'm going to merge this now. Thanks for all of your time on it!

@hanneskaeufler if you could go ahead and do the German language fix in a new PR, that'd be wonderful.

buren:
@nickcharlton  🎉 Sweat!

Thank you!

PR #1007, opened 2017-10-18
Make overriding index actions possible

### LINKED ISSUE #320: Overriding the index method on a controller
I'm trying to display all the users that aren't administrators in my users dashboard.

``` ruby
def index
  super
  @resources = User.where(admin: false).page(params[:page])
end
```

This is essentially from the boilerplate provided. The collections partial is rendering first and then the index method is being hit - it doesn't pass my `@resources` into the partial. Anything I am missing?

anhari:
Fixes #320.

- Replaces the use of locals with instance variables.
- Removes `ControllerHelpers` module.
- Adds the rails-controller-testing gem to test assigning instance
variables, and updates the specs accordingly.
- Updates views to use newly implemented instance variables.


Gemfile
```diff
@@ -24,6 +24,7 @@ end
 
 group :test do
   gem "ammeter"
+  gem "rails-controller-testing"
```
nickcharlton: We've previously had some problems with this gem and compatibility across different Rails versions. Do you think there's a way that we could pull off testing for this without bringing it back in?

app/controllers/administrate/application_controller.rb
```diff
@@ -3,53 +3,42 @@ class ApplicationController < ActionController::Base
     protect_from_forgery with: :exception
 
     def index
-      search_term = params[:search].to_s.strip
-      resources = Administrate::Search.new(scoped_resource,
+      @search_term = params[:search].to_s.strip
```
nickcharlton: I'm concerned here that we're making a lot of variables available with these changes (and there's quite a few of them). Are we able to pull this off without making these changes?

nickcharlton:
Hi @anhari! I just got to seeing a notification about #320. Do you think you'd be able to look at this again?

anhari:
@nickcharlton accidentally deleted my fork of administrate. I have another PR I'm about to submit that tries to address your comments. :)

PR #1009, opened 2017-10-20
Option for Date(Time) objects be **in a timezone**

jayroh:
Often times a layperson using a dashboard can be misled when they look
at a record's creation/update time and not realize it's rooted in the
UTC timezone. This commit adds the option to display `Field::DateTime`
fields as being within a provided timezone.

Also - update the documentation to contain the two options for that
field: `:format` and `:timezone`.

Note: By default the timezone in that field is not explicitly set as
UTC.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -124,6 +124,14 @@ Or, to display a distance in kilometers:
 `:classes` - Specify a list of classes whose objects will be used to populate select boxes for editing this polymorphic field.
 Default is `[]`.
 
+**Field::DateTime**
+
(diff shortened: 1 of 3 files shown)


jayroh:
Note: still needs a test for the Date. Currently only has one for DateTime.

nickcharlton:
Looks good! Let's do the spec for `Date` and then we'll be good to go.

jayroh:
Good to go, @nickcharlton!  Thanks for taking a look, mate 😄

jayroh:
🎉🎉🎉🎉

Issue #430, opened 2016-01-22
Feature: deal with huge result sets


geoffharcourt:
We have an app with tens of thousands of people in it. There are several records that have a person as an association. On the form page for a model that belongs to a person, trying to populate a select list of people results in the app timing out on Heroku or choking when the JS-enabled select field is loaded with all the possible values.

A widget with a search functionality if the potential result set is larger than 10 or so items would be awesome. Administrate already has a search functionality, so this could piggyback on that API.

graysonwright:
Agreed, I started to look into this a little bit earlier, but punted until later. I'm going to add this to the "flexibility" milestone, 'cuz I'd like to see this on the roadmap, but I think extensibility takes priority.

If anyone wants to take a crack at this, have at it!

igorbozato:
Is anyone working on this one? I would like to work on it.

nickcharlton:
Hi @igorbozato! Please do! I think it'd be interesting to see a solution that only involved JavaScript on really huge result sets (so we don't load things which aren't needed) and so forth.

PR #1014, opened 2017-10-25
Load huge result sets for belongs_to remotely

igorbozato:
Started working on #430 and need some feedback, I hope the implementation makes sense.

The `Administrate::Search` doesn't handle cases where the select options name is the `id` (e.g., `"Order #22"`), should `id` be included in the query for all models or only for cases that is needed?

### DIFF
--- a/app/assets/javascripts/administrate/components/belongs_to_form.js
+++ b/app/assets/javascripts/administrate/components/belongs_to_form.js
@@ -1,0 +1,23 @@
+$(function() {
+  $(".field-unit--belongs-to select[data-remote-url]").each(function(index, field) {
+    $(field).selectize({
+      valueField: 'id',
+      labelField: 'name',
+      searchField: 'name',
+      preload: 'focus',
+      load: function(query, callback) {
+          if (!query.length) return callback();
+          $.ajax({
+              url: $(field).data('remote-url') + '?search=' + encodeURIComponent(query),
+              type: 'GET',
+              error: function(res) {
+                  callback();
+              },
+              success: function(res) {
+                  callback(res);
+              }
+          });
+      }
+    });
+  });
+});
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -23,7 +23,7 @@ module Administrate
         end
 
         format.json {
-          render json: resources.map { |resource| { name: dashboard.display_resource(resource), id: resource.send(resource_class.primary_key) } }
+          render json: resources_json(resources, dashboard, resource_class)
         }
       end
     end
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/administrate/components/belongs_to_form.js
```diff
@@ -0,0 +1,23 @@
+$(function() {
+  $(".field-unit--belongs-to select[data-remote-url]").each(function(index, field) {
+    $(field).selectize({
+      valueField: 'id',
+      labelField: 'name',
+      searchField: 'name',
+      preload: 'focus',
+      load: function(query, callback) {
+          if (!query.length) return callback();
```
dsolomadin: So the problem with that is that dropdown is blank unless user starts typing which is not what we want: in case user does not have a particular record in mind he won't be able to select anything.

I'm not sure what would be good UX here as a start I'd recommend:
1. Pressing `keydown` should populate select with all records
2. Pressing an arrow on the dropdown should populate select with all records

app/assets/javascripts/administrate/components/belongs_to_form.js
```diff
@@ -0,0 +1,23 @@
+$(function() {
+  $(".field-unit--belongs-to select[data-remote-url]").each(function(index, field) {
+    $(field).selectize({
+      valueField: 'id',
```
dsolomadin: Why do we need `valueField`, `labelField` etc, and why those are set `id/name`?

app/controllers/administrate/application_controller.rb
```diff
@@ -145,5 +153,9 @@ def show_search_bar?
         dashboard.collection_attributes
       ).any? { |_name, attribute| attribute.searchable? }
     end
+
+    def resources_json(resources, dashboard, resource_class)
+      resources.map { |resource| { name: dashboard.display_resource(resource), id: resource.send(resource_class.primary_key) } }
```
dsolomadin: So the Hound and `rubocop` is giving weird error messages for your branch.
Try rebasing from latest master and see what happens, there are few style errors which Hound should've pointed out, i.e this line is too long

dsolomadin:
Pinging @tysongach @nickcharlton for reviews, this could be great addition to the project.

nickcharlton:
I haven't looked in depth yet, but if you could rebase against current master some of the highlighted errors should disappear.

/cc @igorbozato

app/assets/javascripts/administrate/components/belongs_to_form.js
cheeseweasel: I'm not the author (obviously), but the PR is to resolve huge result sets bringing the browser down because the whole result set is loaded into a dropdown.

If you have 100,000 users, you're going to run into the same problem any way you fill the dropdown with all records.

Maybe it would be worth making this configurable somehow - so being able to turn on the remote search / query?

app/assets/javascripts/administrate/components/belongs_to_form.js
cheeseweasel: This is just telling the selectize plugin which fields from the json returned by the remote query to use for the value (valueField) and text (labelField) on the select.

purinkle:
@igorbozato, is this still an issue for you?

PR #1016, opened 2017-10-27
FactoryGirl is now FactoryBot

nickcharlton:
https://robots.thoughtbot.com/factory_bot


nickcharlton:
@sidraval Thanks for the review!

PR #1020, opened 2017-11-01
Add support for editing polymorphic fields - Revisited

pablobm:
This adds support for editing polymorphic fields. Props go to @pedantic-git, on whose work this is based. This PR supersedes https://github.com/thoughtbot/administrate/pull/984.

An important aspect of this PR is how it handles the polymorphic value. At heart, polymorphic values are split in two: an id and a type. Using [the `globalid` gem](https://github.com/rails/globalid) we can obtain a string that represents both. However, we still need to tell that this string is indeed a globalid and not a literal value when we get to `Administrate::ApplicationController#resource_params`. I handle this by splitting the field in two:

  * `my_field[value]`: the global id, which includes both type and instance id.
  * `my_field[type]`: literally `"Administrate::Field::Polymorphic"`

Then `#resource_params` can look at the param values and treat them appropriately. If it's a simple value such a string, it does nothing special. However, if it comes in this value/type form, we treat it as polymorphic, reading the global id.


pedantic-git:
Wow! This is amazing! Thanks for your additional work to bring this up to the right standard!

pablobm:
I'm going to leave this to rest for a week, just in case there is any more feedback. I'll merge it Friday next week if there are no issues.

ewilliam:
will this be merged soon? thanks @pablobm for your work!

pablobm:
I shall...

make it so

Issue #973, opened 2017-09-18
Inconsistent paths between view/partial generators


nealshepherd:
Is there a way to neatly override where field partials are generated or expected to be? 
For page templates, they are placed in `app/views/admin/`, which makes sense. Field partials are instead placed in `app/views/fields` while it seems more logical that they would be generated into `app/views/admin/fields/`

Furthermore, is there a reason for this inconsistency? The other generators to my knowledge all nest contents under an `/admin` directory, which keeps things nicely separated.

Thanks!

nickcharlton:
This is something I've wondered myself. I don't know if there is a big motivation for it and I've wondered it'd be better under `app/views/admin`, too.

Would you be interested in exploring what a change might look like? Something to consider is breaking compatibility for plugin authors, but we could do this now (pre-`1.0.0`) before it'd be much more disruptive.

nealshepherd:
I think it's an important thing to fix if nothing else but to keep a consistent design and organization paradigm. More importantly, it currently leaves open the possibility for conflicts in an application already using `app/views/fields`.

I've never contributed to an open source project before, but I'd be willing to give it a shot. My assumption is it would be relatively straightforward to modify; plugin issues notwithstanding. Is there anything else I should be aware of that this change would affect?

nickcharlton:
Nothing instantly springs to mind. How about we get a PR going and then we'll (hopefully) start to see issues that drop out?

nealshepherd:
Sounds good, will do.

igorbozato:
Hey @nealshepherd do you still intend to write an PR for this? If not I'm interest in writing it.

nealshepherd:
Hi @igorbozato, that would be great! I'm stretched at bit thin currently, so it would be a while until I had the chance to have a go at it.

pablobm:
Closing due to lack of activity 😞 Happy to look at a PR if anyone has one!

PR #1022, opened 2017-11-08
Move app/views/fields/ to app/views/administrate/fields/

igorbozato:
Opening this PR to continue the discussion #973.
I moved `app/views/fields/` to `app/views/administrate/fields/` and renamed all references in the code and docs.

The only specs that broke was when `administrate-field-image` gem was used, which raises the question on how to handle backwards compatibility if we decide to move forwards with this change.


nickcharlton:
This is looking good so far. Would you mind looking a the Hound comments and rebasing on current `master`?

I do agree we need to think about the backwards compatibility. Our releases are quite spread out at the moment, and we've been pre-`1.0.0` for a quite a while. So this makes me think that we should support the older style until we cross `1.0.0`.

The best way to do this might be to have a separate commit which gives us backwards compatibility (so a renaming one, then a backwards compatibility one), as it will help us remove it in the future. What do you think?

purinkle:
@igorbozato, is this a change you would still like to make?

pablobm:
Closing due to lack of activity 😞 Let us know if anyone wants to take over!

madhums:
@pablobm would you be up for considering this PR if I reworked and supported backward compatibility?

pablobm:
@madhums - I'm up for it. Would you be able to explain why you want it fixed, please?

Issue #951, opened 2017-07-28
has_one with options class_name burns in agony


dan-ding:
```ruby
class PhoneDashboard < Administrate::BaseDashboard
ATTRIBUTE_TYPES = {
  owner: Field::HasOne,
  ...
} ...

class OwnerDashboard < Administrate::BaseDashboard
ATTRIBUTE_TYPES = {
  account: Field::HasOne,
  cat: Field::HasOne.with_options(class_name: 'Pet')
  ...
}...
```

trying to edit Phone dies with a message such as:

> uninitialized constant CatDashboard
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/resource_resolver.rb:8:in 'const_get'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/resource_resolver.rb:8:in 'dashboard_class'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/field/has_one.rb:9:in 'initialize'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/field/deferred.rb:15:in 'new'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/field/deferred.rb:15:in 'new'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/page/base.rb:23:in 'attribute_field'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/page/form.rb:15:in 'block in attributes'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/page/form.rb:14:in 'map'
/stark_tower/gems/administrate-8b51445b9251/lib/administrate/page/form.rb:14:in 'attributes'
/stark_tower/gems/administrate-8b51445b9251/app/views/fields/has_one/_form.html.erb:17:in 'block in app_views_fields_has_one__form_html_erb'
...
>

nickcharlton:
* Which version of administrate do you see this with?
* What do your models look like?
* Could you provide an example project where this happens?

dan-ding:
version - 8.0/8.1
models - have seen no common pattern -- has_one straight, has_one through
example - no, but should be reproducible with the above snippets.

dan-ding:
The issue seems to be in the initialize method of administrate/field/has_one.rb
```ruby
      def initialize(attribute, data, page, options = {})
        resolver = Administrate::ResourceResolver.new("admin/#{attribute}")
        @nested_form = Administrate::Page::Form.new(
[...]

PR #1028, opened 2017-11-20
Fix has one class name

dmitry-solomadin:
Related issue: https://github.com/thoughtbot/administrate/issues/951

Currently, when passing `class_name` option to `has_one` field it does not infer correct class name for the dashboard which results in exception. 
This PR modifies `has_one` class to use `associative` to get the correct dashboard name.
Additionally, it fixes `permitted_attributes` method, which required me to modify the signature and pass `options` from `Deferred` attribute.

In scope of this I've also removed old locale keys for `has_one.not_supported` string.

/r @nickcharlton @tysongach

### DIFF
--- a/spec/lib/fields/deferred_spec.rb
+++ b/spec/lib/fields/deferred_spec.rb
@@ -20,7 +20,7 @@ describe Administrate::Field::Deferred do
         deferred.permitted_attribute(:foo)
 
         expect(Administrate::Field::String).
-          to have_received(:permitted_attribute).with(:foo)
+          to have_received(:permitted_attribute).with(:foo, {})
       end
     end
   end
--- a/config/locales/administrate.ar.yml
+++ b/config/locales/administrate.ar.yml
@@ -5,8 +5,9 @@ ar:
       confirm: "هل أنت متأكد ؟"
       destroy: "حذف"
       edit: "تعديل"
-      show: "إظهار"
-      new: "جديد"
+      edit_resource: "تعديل %{name}"
+      show_resource: "إظهار %{name}"
+      new_resource: "جديد  %{resource}"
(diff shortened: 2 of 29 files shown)


dmitry-solomadin:
Added some tests. PR is ready for review
/cc @nickcharlton @tysongach

nickcharlton:
Hi @dmitry-solomadin! Thank you for opening this!

As I've taken a while to look at it, it's fallen behind a bit. Would you mind rebasing it (it's mostly translations, it seems…)?

spec/lib/fields/deferred_spec.rb
```diff
@@ -20,7 +20,7 @@
         deferred.permitted_attribute(:foo)
 
         expect(Administrate::Field::String).
-          to have_received(:permitted_attribute).with(:foo)
+          to have_received(:permitted_attribute).with(:foo, {})
```
nickcharlton: If the method is providing a default (`nil`), is it valuable for us to be providing an empty hash here?

dmitry-solomadin: Yes, however in that particular case we are testing this method on String field in scope of Deferred attribute, in `deferred.rb` the options that we pass are actually initialized with `{}` by default

dmitry-solomadin:
@nickcharlton Hey Nick, rebased the branch off latest master, should be good now.
Let me know if you have any other concerns.

dmitry-solomadin:
Looks like some tests on master are failing the same ones are failing on my branch

nickcharlton:
Great, thanks! The failures are unrelated to your work, so I'm ignoring them until I get some time on Friday to look at them. I'll merge this now!

PR #1031, opened 2017-11-23
Switch to Circle 2.0.

nickcharlton:
* Use a Docker image for Ruby 2.4.1.
* Use an image with node for JavaScript dependencies of Rails.
* Use an image with a browser for integration tests.
* Configure Postgres with an adjacent image.
* Avoid race conditions by using `dockerize`.
* Cache bundler dependencies between runs.


.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+      # Bundle install dependencies
+      - run: bundle install --path vendor/bundle
+
+      # Install Appraisal gemfiles
+      - run: bundle exec appraisal install
+
+      # Cache Dependencies
+      - type: cache-save
+        name: Store bundle cache
+        key: houndci-{{ checksum "Gemfile.lock" }}
```
mygulamali: Should the key for this cache be the same as the one on line 22?

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
...
+
+      # Wait for DB
+      - run: dockerize -wait tcp://localhost:5432 -timeout 1m
+
+      # Setup the environment
+      - run: cp .sample.env .env
+
+      # Setup the database
+      - run: bundle exec rake db:create
+      - run: bundle exec rake db:schema:load
```
mygulamali: Could these two lines be replaced with one?

`bundle exec rake db:create db:schema:load`

mygulamali: Or alternatively: `bundle exec rake db:setup` ?

.circleci/config.yml
```diff
@@ -0,0 +1,49 @@
+version: 2
+jobs:
+  build:
+    working_directory: ~/administrate
+    docker:
+      - image: circleci/ruby:2.4.1-node-browsers
+        environment:
+          PGHOST: 127.0.0.1
```
mygulamali: Is it possible to use `localhost` instead of the actual IP address here? Just to keep it more consistent with line 38.

.circleci/config.yml
sardaukar: Yep, that's pretty much needed @nickcharlton should be `administrate-{{ checksum "Gemfile.lock" }}`

nickcharlton: Ah, yeah. I missed that.

PR #1039, opened 2017-12-04
Merge both Field::BelongsTo options_with sections.

nickcharlton:
These had split, probably with different PRs introduced more options.
This uses the first one (to try and keep it in alphabetical order) and
reformats a little.


nickcharlton:
@croaky Thanks for reviewing this!

Issue #1025, opened 2017-11-10
Adding a newly-migrated attribute to the dashboard


vcavallo:
Apologies if this is explained somewhere. I googled a bunch, read through the docs and searched open & closed issues and didn't see anything...

I've already installed Administrate and everything is working great. I've now added an attribute to one of my models and am looking for the best way to include this field on administrate's dashboards. I know I can add it in manually in the Dashboard file, but I opted to re-run `rails g administrate:install` - even though it was a bit cumbersome to skip/diff/Y/N various files.

What is the recommended way to do this?
Thanks!

nickcharlton:
Hi @vcavallo!

I think you could probably do either of what you'd suggested. Personally, I'd probably lean towards adding it manually to the Dashboard (such as in a PR which introduced this new attribute).

It'd be great to be able to pick them up automatically with a generator, but this is pretty difficult to do.

To help others in future, what do you think about adding a mention of this to the docs?

vcavallo:
@nickcharlton ! 

How does this look? https://github.com/thoughtbot/administrate/pull/1040

nickcharlton:
Closing as we solved this!

PR #1040, opened 2017-12-04
Add note re: updating dashboards with new model attributes

vcavallo:
This change updates the docs with a reminder and example code to update already-existant dashboards after models change. This change was recommended in https://github.com/thoughtbot/administrate/issues/1025#issuecomment-349026731 by @nickcharlton

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -110,7 +110,7 @@ Example:
 ```
 
 It's recommended that you make this change at the same time as you add the 
-attribute to the model
+attribute to the model.
 
 The alternative way to handle this is to re-run `rails g administrate:install` and 
 carefully pick through the diffs. This latter method is probably more cumbersome.
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -109,8 +109,8 @@ Example:
   ].freeze
 ```
 
-It is recommended you make this change in the PR that also adds the attribute to 
-your model.
+It's recommended that you make this change at the same time as you add the 
+attribute to the model
 
 The alternative way to handle this is to re-run `rails g administrate:install` and 
 carefully pick through the diffs. This latter method is probably more cumbersome.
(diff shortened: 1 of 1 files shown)


docs/getting_started.md
```diff
@@ -73,6 +73,48 @@ namespace :admin do
...
+  
+  COLLECTION_ATTRIBUTES = [
+    # ...
+    :the_new_attribute, # if you want it on the index, also.
+    # ...
+  ].freeze
+```
+
+It is recommended you make this change in the PR that also adds the attribute to 
+your model.
```
nickcharlton: I wonder if we'd be better off not being prescriptive in the process here? (Even though I do _exactly_ this.) What do you think about changing this to be:

> It's recommended that you make this change at the same time as you add the attribute to the model

docs/getting_started.md
```diff
@@ -73,6 +73,48 @@ namespace :admin do
...
+
+The alternative way to handle this is to re-run `rails g administrate:install` and 
+carefully pick through the diffs. This latter method is probably more cumbersome.
```
nickcharlton: Hah! I like this. It often _really_ is, too!

docs/getting_started.md
vcavallo: Yup, good point. will change this.

docs/getting_started.md
```diff
@@ -73,6 +73,48 @@ namespace :admin do
...
+  
+  COLLECTION_ATTRIBUTES = [
+    # ...
+    :the_new_attribute, # if you want it on the index, also.
+    # ...
+  ].freeze
+```
+
+It's recommended that you make this change at the same time as you add the 
+attribute to the model
```
nickcharlton: Sorry to pick on a tiny thing …but could you add a full stop at the end of this sentence?

PR #1041, opened 2017-12-04
Hk minor german translation improvements

hanneskaeufler:
As suggested in #1004 this follows up on #1002 with the minor german improvements. Please keep in mind that the translation of the `form.errors` is mediocre at best, since the sentence would need many variations based on the number of errors and the gender of the resource and so on. But ah well.


nickcharlton:
Hi @hanneskaeufler! Thanks for opening this alternative. I'm going to go ahead and merge this.

PR #1051, opened 2017-12-17
humanize has_many form label

ryanhertz:
The has_many field form label displays in underscore case. The second argument to the `label_tag` helper should be the display text, but here `field.attribute` is a symbol. Other fields, like the string field, aren't affected because the label tag helper in the string form partial only receives one argument. When the `label_tag` helper only receives one argument, it calls `.to_s.humanize` on it to create the display text, so that's what I've added to the has_many form partial.

### DIFF
--- a/app/views/fields/has_many/_form.html.erb
+++ b/app/views/fields/has_many/_form.html.erb
@@ -20,7 +20,7 @@ and is augmented with [Selectize].
 %>
 
 <div class="field-unit__label">
-  <%= f.label field.attribute_key, field.attribute %>
+  <%= f.label field.attribute_key, field.attribute.to_s.humanize %>
 </div>
 <div class="field-unit__field">
   <%= f.select(field.attribute_key, nil, {}, multiple: true) do %>
--- a/app/views/fields/has_many/_form.html.erb
+++ b/app/views/fields/has_many/_form.html.erb
@@ -20,7 +20,10 @@ and is augmented with [Selectize].
 %>
 
 <div class="field-unit__label">
-  <%= f.label field.attribute_key, field.attribute.to_s.humanize %>
+  <%= f.label field.attribute_key, 
+      t("helpers.label.#{resource_name}.#{field.attribute_key}", 
+          default: field.attribute.to_s.humanize) %>
+
 </div>
 <div class="field-unit__field">
   <%= f.select(field.attribute_key, nil, {}, multiple: true) do %>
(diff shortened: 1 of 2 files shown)


nickcharlton:
Hi @ryanhertz!

Do you think you'd be able to add some tests to this?

ryanhertz:
@nickcharlton I was surprised to find there was already a test for this, and it was passing without my fix. The problem was with [this line](https://github.com/thoughtbot/administrate/pull/1051/files#diff-9cc5b1bfb56d2996c51478516b9b46d0L23) in `fake_form_builder`. That was causing the test to pass, even though it didn't actually work with a real form builder.

evrimfeyyaz:
@nickcharlton Will this be merged soon?

evrimfeyyaz:
By the way, although this is a good step in the right direction, this ignores i18n translations.

I'd recommend either using `<%= f.label field.attribute_key %>`, or somehow checking i18n translations manually for `field.attribute`.

nickcharlton:
@ryanhertz What do you think about incorporating @evrimfeyyaz's suggestions?

ryanhertz:
It looks like `<%= f.label field.attribute_key %>` would output something like "Associated object ids", so I think checking the translations might be the better option. I'll work on it, but it might be a few days before I can get to it.

ryanhertz:
@nickcharlton @evrimfeyyaz  I found [a relevant PR](https://github.com/thoughtbot/administrate/pull/492) and [some documentation on how this should work](https://github.com/thoughtbot/administrate/blob/master/docs/customizing_dashboards.md#defining-labels) so I tried to follow those examples.

nickcharlton:
Excellent!

What's a good way for me to try this out? Do we have anything in the prototype where this is obvious?

ryanhertz:
@nickcharlton You can see the problem here in both "line_items" and "payments": http://administrate-prototype.herokuapp.com/admin/orders/83860/edit

nickcharlton:
I'm just getting time to go back and look at this. In the prototype link you mentioned, I'm just seeing "LINE ITEMS". From what I understand we should see "LINE_ITEMS" here, right?

nickcharlton:
I spent some time looking at this today and after rebasing onto current master, I can't replicate it: "Line items" is shown correctly if I spin up Administrate's demo app locally with or without this PR. Looking at my previous comments, I wonder if I had the same confusion previously.

We've changed a bunch since this PR was last actively worked on, so I'm going to close this for now. However, if someone is still finding this to be a problem we can take a look at it in another PR.

Issue #1042, opened 2017-12-04
HasMany index partial passes an already-pluralized field name to pluralize helper


tbroadley:
[The HasMany index partial](https://github.com/thoughtbot/administrate/blob/6661889711edd688c8768e48602fa4a25b82d7bf/app/views/fields/has_many/_index.html.erb) includes one line of code:

```erb
<%= pluralize(field.data.size, field.attribute.to_s.humanize.downcase) %>
```

In this partial, `field` is an instance of `Administrate::Field::HasMany`, so `field.attribute` returns the pluralized name of a model, e.g. "teams". When `field.data.size == 1`, the partial renders `pluralize(1, 'teams')` which equals `'1 teams'`, instead of `'1 team'` which is what I would expect.

To fix this, I would suggest using the following as the body of the partial instead:

```erb
<%= pluralize(field.data.size, field.attribute.to_s.humanize.downcase.singularize) %>
```

nickcharlton:
Hi @tbroadley!

Thanks for opening this issue. Would you be able to open a PR making the change you suggested?

tbroadley:
@nickcharlton Sure, I can do that!

PR #1055, opened 2017-12-21
Singularize field name in HasMany index partial

tbroadley:
Fixes https://github.com/thoughtbot/administrate/issues/1042.

`field.attribute` is already plural. Because of this, if `field.data.size == 1`, `pluralize` returns a string like "1 teams". To fix this, we call `singularize` on `field.attribute` before passing it to `pluralize`.


nickcharlton:
Hi @tbroadley!

Could I get you to rebase against `master`? This should fix the failing tests by this point.

tbroadley:
@nickcharlton Done! Thanks for letting me know I should do that.

nickcharlton:
Wonderful! This is looking good, so I'm going to merge this!

tbroadley:
Thank you!

PR #1046, opened 2017-12-06
Fix wrong variable name for new_resource I18n keys in locales

shvetsovdm:
This is how new_resource translation is used:

```erb
  <%# app/views/administrate/application/index.html.erb %>
  <div>
    <%= link_to(
      t(
        "administrate.actions.new_resource",
        name: display_resource_name(page.resource_name).titleize.downcase
      ),
      [:new, namespace, page.resource_path],
      class: "button",
    ) if valid_action?(:new) && show_action?(:new, new_resource) %>
  </div>
```
we are passing `name` variable instead of `resource`.

This PR fix the problem.

PR #1056, opened 2017-12-21
Don't pluralise "New [resource]" buttons.

nickcharlton:
In #1004 the handling of translation strings improved, but bought in a
change to the buttons for new resources. This started to pluralise them
when they shouldn't be.

So you'd have "New Line items", not "New Line item". This reverts that
change.


nickcharlton:
I'm going to merge this in favour of #1046, as the tests pass here but not there and it's been quiet for a while.

Issue #1021, opened 2017-11-05
Use number_field for Field::Number


rkbhochalya:
It would be nice to use `number_field` helper to render input field in the form partial view for `Field::Number` fields. You are using `text_field` at the moment.

```erb
<%#
# Number Form Partial

This partial renders an input element for number attributes.
By default, the input is a text field.

## Local variables:

- `f`:
  A Rails form generator, used to help create the appropriate input fields.
- `field`:
  An instance of [Administrate::Field::Number][1].
  A wrapper around the number pulled from the database.

[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Number
%>

<div class="field-unit__label">
  <%= f.label field.attribute %>
</div>
<div class="field-unit__field">
  <%= f.text_field field.attribute %>
</div>

```

nickcharlton:
Hi @rkbhochalya!

This sounds like a good idea. Do you think you'd be able to contribute a PR for this?

PR #1063, opened 2018-01-08
number_field for Field::Number

codergeek121:
I changed the partial for the Field::Number to use the  number_field view helper instead of the text_field helper. The tests failed for my cloned  master branch, but my first commit fixes it.
Is a new test for the input_field change necessary, and if so what would be a good test to write?

This PR is for Issue  #1021.

Happy to help.

### DIFF
--- a/app/views/fields/number/_form.html.erb
+++ b/app/views/fields/number/_form.html.erb
@@ -19,5 +19,5 @@ By default, the input is a text field.
   <%= f.label field.attribute %>
 </div>
 <div class="field-unit__field">
-  <%= f.text_field field.attribute %>
+  <%= f.number_field field.attribute, step: "any" %>
 </div>


nickcharlton:
Hi @codergeek121!

I'm not sure a test would be 100% valuable in this scenario. Is there some validation/presentation that `number_field` does that we can use to our advantage for one?

Could you also rebase against master? I wonder if that'll now help with the test failures you're seeing.

jdrosales17:
Hi @nickcharlton!

I was just working with the gem and wondering why `Field::Number` was using `text_field`, which led me to this PR. Do you know when it will be merge ?

pablobm:
@jdrosales17 - If you rebase this (skipping the first commit, now unnecessary) and create a new PR, I'll be happy to review it.

nickcharlton:
I think we're good here. Maybe there will be a regression and we can pick that up but the combination of using the right field type and "any" is good for now.

PR #1064, opened 2018-01-08
human_attribute_name instead of raw attr_name

masciugo:
this just a rebase of [this old PR](https://github.com/thoughtbot/administrate/pull/667)


nickcharlton:
Ah, nice, thanks!

Do you think you'd be able to rebase again (I recently fixed some tests that should go green), and potentially add some tests?

PR #1065, opened 2018-01-12
Fix new_resource interpolation variable name in translations

Knack:
The translations of the I18n key `new_resource` have an incorrect variable name, causing a "missing interpolation argument" error


Knack:
Sorry, didn't notice it was duplicated: #1046

PR #1068, opened 2018-01-14
Only define Punditize if Pundit has been loaded.

### LINKED ISSUE #1048: Eager load: uninitialized constant Administrate::Punditize::Pundit (NameError)
Hello!

In production (which eager load classes) I see this error:
```
uninitialized constant Administrate::Punditize::Pundit (NameError)
```

How to fix it?

Thanks!

nickcharlton:
When the `RAILS_ENV` is set to `production`, the concern will be eagerly
loaded. As we `include Pundit` here, this means that it'll fail if
`pundit` isn't included in the bundle.

Fixes #1048.

### DIFF
--- a/app/controllers/concerns/administrate/punditize.rb
+++ b/app/controllers/concerns/administrate/punditize.rb
@@ -1,33 +1,35 @@
-module Administrate
-  module Punditize
-    extend ActiveSupport::Concern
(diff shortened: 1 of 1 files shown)


pedantic-git:
Oops! Good spot! The fix looks good to me!

Issue #1069, opened 2018-01-15
Default timezone is better to be Rails.application.config.time_zone


nao58:
Hi,

I've just realized this change
https://github.com/thoughtbot/administrate/pull/1009
and in trouble now.
My datetime field's timezone is in local on input form, utc on view after the gem has updated. To solve this, I should add `with_options` to ALL current datetime fields.

I think the default timezone is better not to be fixed UTC but rails default configuration, something like this:
https://github.com/thoughtbot/administrate/blob/29166a33cdd4f09df3fe8d7aa330503e2d90343d/lib/administrate/field/date_time.rb#L27

```ruby
def timezone
  options.fetch(:timezone, Rails.application.config.time_zone || "UTC")
end
```

How do you find it?

nickcharlton:
Ah, interesting!

Is there something we can rely on in Rails to always provide an answer? What's the default timezone if one isn't set (I'd like to assume `UTC`).

nao58:
Yea, it returns `UTC` when you don't set default `time_zone` . And just in case the value is nil, still we can provide `UTC` by the line `Rails.application.config.time_zone || "UTC"` .
Or maybe it's better to use the value `Time.zone.name` instead of `Rails.application.config.time_zone` . I believe it returns same value.
http://api.rubyonrails.org/classes/ActiveSupport/TimeZone.html

nickcharlton:
Okay, let's give it a shot. Do you think you'd be able to contribute a PR for this?

nao58:
Okay, I'll work on it.

nao58:
Hi, current master branch seems to be failed test on CI, so my forked branch as well. Do you have any plan to fix it?

nao58:
@nickcharlton merged :)

PR #1070, opened 2018-01-17
Use default application timezone for datetime fields (#1069)

nao58:
Use application default timezone to avoid specifying options on all datetime fields.

This PR is for Issue #1069


nao58:
The test seems to be failed on the master branch. I will rebase the branch after the problems are fixed 😃

nao58:
has been already merged master and passed the test.

PR #1076, opened 2018-01-26
Nothing says they have to be in a specific order

pablobm:
This should fix one of the flaky tests that fails randomly on CI.

I looked around for a bit and saw nothing that specified that line items had to be in a specific order. So either:

1. We accept this and spec for it.
2. We change the models to make line items be ordered by id.

I went for option 1.


nickcharlton:
🎉 Great, thanks! I'll merge this.

PR #1077, opened 2018-01-26
Add Ruby 2.2.9 to Circle and clarify support.

### LINKED ISSUE #1010: Ruby version 2.1.6 throws SyntaxError from lib/administrate/field/has_one.rb:21
I was just installing administrate on a brand new app and my first launch of the dashboard threw this:

```
SyntaxError in Admin::UsersController#index
/Users/patrickcushing/.rvm/gems/ruby-2.1.6/gems/administrate-0.8.1/lib/administrate/field/has_one.rb:21: syntax error, unexpected ':', expecting => { "#{attr}_attributes": related_dashboard_attributes } ^ /Users/patrickcushing/.rvm/gems/ruby-2.1.6/gems/administrate-0.8.1/lib/administrate/field/has_one.rb:27: syntax error, unexpected end-of-input, expecting keyword_end
```

Here's the line in question: https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/has_one.rb#L21

Swiching over to ruby 2.2.2 seemed to resolve the issue. I don't see anything specific to ruby versions in the README so I figured I'd report. A simple mention there would be helpful. I'm not in the position to propose a more robust fix this moment, but can take a look next week if desired.

nickcharlton:
The version support is informed by:

* https://stackoverflow.com/a/23532476/83386
* http://guides.rubyonrails.org/5_0_release_notes.html

Closes #1010.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,20 @@ default_job: &default_job
     - run: bundle exec appraisal rake
 
 jobs:
+  ruby-2.2:
+    <<: *default_job
+    docker:
+      - image: circleci/ruby:2.2.9-node-browsers
+        environment:
+          PGHOST: localhost
+          PGUSER: administrate
+          RAILS_ENV: test
+      - image: postgres:9.5-alpine
+        environment:
+          POSTGRES_USER: administrate
+          POSTGRES_DB: ruby22
+          POSTGRES_PASSWORD: ""
+
   ruby-2.3:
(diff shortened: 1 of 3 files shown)


nickcharlton:
That's my thinking.

The problem I foresee is that the combination of Ruby EOL, Rails versions and people's actual deployments never really line up perfectly …but also neither will we in keeping track here.

PR #1078, opened 2018-01-26
Fix another issue with non-guaranteed order

pablobm:
Similar to https://github.com/thoughtbot/administrate/pull/1076, these two test are flaky because they expect records to be in a specific order.

In this case the fix is a bit more involved. I had to merge two examples in one to preserve the original meaning of the tests. It may raise some eyebrows.

Thoughts?


nickcharlton:
This looks like a good solution to me, thanks! I'll merge this now so we can get rid of the randomly failing tests.

PR #1079, opened 2018-01-27
Cast all search queries to text.

### LINKED ISSUE #620: cast type in search (function lower text does not exist error)
Current search try to use lower() function in PG, thsi funciton only accepts an string since PG 8.4, so it will be a good idea to add coercion to the query to avoid problems with nil columns.

rewrite the quey to:

```
"lower(text(#{quote_attr(attr)})) LIKE ?"
```

is enough

nickcharlton:
All Postgres versions since 8.4 require this to be a string. This casts
the table and column to be strings as it's possible that the column name
is nil.

Fixes #620.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -24,7 +24,7 @@ module Administrate
         table_name = ActiveRecord::Base.connection.
           quote_table_name(@scoped_resource.table_name)
         attr_name = ActiveRecord::Base.connection.quote_column_name(attr)
-        "lower(#{table_name}.#{attr_name}) LIKE ?"
+        "LOWER(TEXT(#{table_name}.#{attr_name})) LIKE ?"
       end.join(" OR ")
     end
 
--- a/spec/lib/administrate/search_spec.rb
+++ b/spec/lib/administrate/search_spec.rb
@@ -1,5 +1,4 @@
-require "spec_helper"
-require "support/constant_helpers"
+require "rails_helper"
 require "administrate/field/string"
 require "administrate/field/email"
 require "administrate/field/number"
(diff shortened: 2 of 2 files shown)


lib/administrate/search.rb
```diff
@@ -24,7 +24,7 @@ def query
         table_name = ActiveRecord::Base.connection.
           quote_table_name(@scoped_resource.table_name)
         attr_name = ActiveRecord::Base.connection.quote_column_name(attr)
-        "lower(#{table_name}.#{attr_name}) LIKE ?"
+        "lower(text(#{table_name}.#{attr_name})) LIKE ?"
```
geoffharcourt: We are writing the SQL keyword `LIKE` in all caps, how would you feel about doing `LOWER` and `TEXT` in all caps as well? I do this in my SQL style so that keywords and column names are clearly separated.

spec/lib/administrate/search_spec.rb
```diff
@@ -74,8 +73,8 @@ class User < ActiveRecord::Base; end
                                           MockDashboard,
                                           "Тест Test")
         expected_query = [
-          "lower(\"users\".\"name\") LIKE ?"\
-          " OR lower(\"users\".\"email\") LIKE ?",
+          "lower(text(\"users\".\"name\")) LIKE ?"\
+          " OR lower(text(\"users\".\"email\")) LIKE ?",
```
geoffharcourt: Same here, we write `OR` and `LIKE` in caps, but not `LOWER` or `TEXT`.

lib/administrate/search.rb
nickcharlton: Ah yeah, nice. I'll do that, it's much nicer.

PR #1081, opened 2018-01-28
Update Field::Polymorphic doc to include order option

Calyhre:
I just noticed there is a nice option that allow us to sort the polymorphic association select options, but it lacks documentation.


nickcharlton:
Wonderful, thanks!

PR #1096, opened 2018-02-14
Implement overridable index actions

### LINKED ISSUE #320: Overriding the index method on a controller

anhari:
Fixes #320.

- Replaces the use of locals with instance variables.
- Replaces `.capture_view_locals` in the `ControllerHelpers` module with a `.value_assigned_to` method which grabs the value assigned to an instance variable without requiring the [rails-controller-testing gem](https://github.com/rails/rails-controller-testing).
- Updates views to use newly implemented instance variables, and replaces comments referencing local variables with ones referencing instance variables.
- Moves the logic behind `show_search_bar` and `search_term` into the `Page::Collection` object.
- Alphabetizes the `Gemfile`.


hanchang:
Great progress! Any ideas on timeline for this PR?

andrepcg:
This would be amazing!

Yegorov:
Awesome!

anhari:
@nickcharlton can I close this? I haven't kept up with this issue and haven't used administrate in quite awhile. It also looks like there is another approach being discussed in #320.

Issue #220, opened 2015-11-11
Request: Custom-building a new record in controller


mcmire:
I have a form that contains nested attributes for an association. In order to render the `new` page, I need a way to pre-populate the association on the record with an empty object. Right now, I can override the `new` method like so:

``` ruby
def new
  new_resource = resource_class.new
  new_resource.my_has_many_association.build
  @page = Administrate::Page::Form.new(dashboard, new_resource)
end
```

However, I'm wondering if it would be better if there were a `new_resource` method I could simply override:

``` ruby
def new_resource
  resource_class.new.tap do |resource|
    resource.my_has_many_association.build
  end
end
```

Thoughts?

graysonwright:
@mcmire, that's a good idea. Also note that your solution won't work on the current version of Administrate, because we've moved away from instance variables in the controller.

mcmire:
Whoops! You're right. Well, you get the picture.

GustavoCaso:
@graysonwright is anyone working on this one ? if not I will like to tackle this one

carlosramireziii:
@GustavoCaso Not sure if you're still interested in tackling this one, but if you are let me know.

GustavoCaso:
@tysongach sorry for the really late response 😄  I will check for a solution on this issue

nickcharlton:
@GustavoCaso Were you able to look into this?

GustavoCaso:
I have. But I haven’t finish with it. Currently I’m not at home, since I’m on holidays. If it is urgent I will be able to push something in two weeks :)

nickcharlton:
Hi @GustavoCaso! No worries. Take your time, I look forward to seeing it!

nickcharlton:
I'm going to close this as we merged #1097 the other month and so this should be done now!

PR #1097, opened 2018-02-15
Use new_resource in new action

EugeneGrigorenko:
- Simplify making new form with prefilled associations
- Is supposed to close this issue https://github.com/thoughtbot/administrate/issues/220
- Now to have  simply override `new_resource` method instead of new action to init resource associations

P.S. Though, it works as expected on has_one association, when trying to build has_many association it doesn't work. The reason is that select is used on has_many form and only ids get send with the request. And because the record doesn't have an id it gets lost. I'm not sure if it's a bug and should be addressed or not, please let me know what you think.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -27,7 +27,7 @@ module Administrate
     end
 
     def new
-      resource = resource_class.new
+      resource = new_resource
       authorize_resource(resource)
       render locals: {
         page: Administrate::Page::Form.new(dashboard, resource),


nickcharlton:
Hi @EugeneGrigorenko!

Thanks for this! It sounds like we should make sure it still works with a `has_many` in this PR.

Do you also think you'd be able to add some tests for the feature this gives us, too?

EugeneGrigorenko:
hi @nickcharlton, sorry for delay
i've investigated a bit and I can say, that has_many works with ids. Like e.g. overriding new_resource like this would initialize customer form with first and last orders selected. 
```
      resource_class.new.tap do |resource|
        resource.order_ids = [::Order.first.id, ::Order.last.id]
      end
```
To supported creation of has_many via nested forms, I believe, we need to change https://github.com/nickcharlton/administrate-field-nested_has_many

As for tests, I'm not sure what I can test exactly, as, basically that's a one line refactoring so far. It's supposed to being overriden and then it gives new functionality

nickcharlton:
Okay, cool!

Could you try and confirm if we do indeed to change `administrate-field-nested_has_many`? If we do, I'll want to coordinate the releases.

pablobm:
On one hand, this PR looks like a slam dunk to me (provided there's a rebase first). I can't see how it would interfere with anything, and I can see how it's beneficial to have this API.

On the other hand, I'm looking into `administrate-field-nested_has_many` and I'm having issues making it work at all (unrelated to this PR). Specifically, I'm getting an error when trying to create a School resource in the dummy app... I'm not feeling quite like looking into that one now :sweat: 

@nickcharlton, do you have any thoughts?

nickcharlton:
I rebased and then merged. We'll worry about nested has many when it's a problem (assuming it is!).

sedubois:
Is this `new_resource` method documented somewhere?

nickcharlton:
I don't believe so, but I might be wrong. Would you be able to do that?

If it doesn't fit anywhere as is, we might want to create an "Administrate Internals" documentation page.

sedubois:
Sorry @nickcharlton I'm not familiar with this method, maybe someone else knows a bit more?

nickcharlton:
I don't either! I'll open a new issue to keep track of it.

PR #1103, opened 2018-02-24
Use cross-DB way to cast search queries to strings

ineu:
Tested on:
- Postgres 10.2
- MariaDB 10.1
- Sqlite 3.22


martinstreicher:
Would be great to merge this. Bit by it today.

nickcharlton:
Thanks for this! I'm going to merge this as is.

Something I'm going to note here for the future: We should probably be testing against more database engines at some point to avoid us having these sorts of problems in the future.

PR #1104, opened 2018-02-25
Add `scope` option to Fields::BelongsTo

ineu:
When working with large datasets, simple `.all` can cause N+1 if related objects are used in `display_resource`. This PR allows to pass the scope explicitly.

The `scope` is passed in a callable to avoid arguments being resolved while preloading dashboards. I.e.
```ruby
  ATTRIBUTE_TYPES = {
    collection: Field::BelongsTo.with_options(
      scope: -> { Collection.where(created_at < Date.today).includes(:something) }),
```


nickcharlton:
This looks great! Thanks. Merging now.

Issue #626, opened 2016-07-06
question: Can I hide dashboards from the left menu but keep them in as relationships?


l0c0luke:
I really only want to see `projects` and `users` on the left. But each model has a bunch of relationships. I find if I remove the dashboard file, I can remove the option in the menu, but then I cannot display that relationship when viewing a project or user.

michaelkl:
I'm using the following solution.

Create file `app/views/administrate/application/_sidebar.html.erb` which contains the code:
```
<ul class="sidebar__list">
  <% Administrate::Namespace.new(namespace).resources.each do |resource| %>
    <% unless [:photos].include? resource %>
      <li>
        <%= link_to(
          display_resource_name(resource),
          [namespace, resource],
          class: "sidebar__link sidebar__link--#{nav_link_state(resource)}"
        ) %>
      </li>
    <% end %>
  <% end %>
</ul>
```
Here you can see hardcoded array `[:photos]`. It is array of resource names to exclude from sidebar.

ID25:
I do a similar solution
```ruby
class ServiceObject
  #  set pages which you wanna remove from sidebar
  HIDDEN_PAGES = [:hide_me_from_sidebar].freeze

  def self.routes(admin)
    admin.resources.reject { |i| i.in?(HIDDEN_PAGES) }
  end
end
```
```erb
<ul class="sidebar__list">
  <% resources =  ServiceObject.routes(Administrate::Namespace.new(namespace)) %>
  <% resources.each do |resource| %>
    <li>
      <%= link_to(
        display_resource_name(resource),
        [namespace, resource],
        class: "sidebar__link sidebar__link--#{nav_link_state(resource)}"
      ) %>
    </li>
  <% end %>
</ul>
```

tehfailsafe:
Is there a good way to re-order them?

tatecarr:
@tehfailsafe I believe the order of the resources in the routes.rb will accomplish that pretty easily.  There are likely other options, but I used that approach successfully.

tehfailsafe:
Hah! Perfect, didn't even think to re-order them...

carlosramireziii:
Sounds like this one is resolved. 

FYI - there's now a generator you can use to customize the sidebar
`rails g administrate:views:sidebar`

Hope that helps!

FanaHOVA:
For future Googlers, the generator is now `rails g administrate:views:navigation`.

tonyvince:
Also now, @ID25's service object should be

[...]

PR #1106, opened 2018-03-05
Update Customizing Views docs with sidebar generator

FanaHOVA:
Per @nickcharlton's suggestion on #626. I've also reworded the other paragraph since it had a bit of a weird syntax.


nickcharlton:
Wonderful, thanks!

PR #1107, opened 2018-03-05
Exclude routes that start with namespace that are not in the namespace

GerardMopp:
Addresses #758 
Using the trailing slash when looking at for routes in the namespace ensures that only routes from the namespace will be used and not routes that start with the namespace. This stops the navigation looking for namespaced routes that do not exist.

### DIFF
--- a/lib/administrate/namespace.rb
+++ b/lib/administrate/namespace.rb
@@ -12,7 +12,7 @@ module Administrate
 
     def routes
       @routes ||= all_routes.select do |controller, _action|
-        controller.starts_with?("#{namespace.to_s}/")
+        controller.starts_with?("#{namespace}/")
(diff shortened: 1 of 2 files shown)


omake43:
👍

nickcharlton:
Great, thanks!

Do you think you'd be able to add a test which ensures we don't accidentally undo this in future?

GerardMopp:
@nickcharlton you are welcome,

The change I made to spec/administrate/namespace_spec.rb should ensure it does not get overridden unless I am mistaken. I if I am wrong could you please point me in the right direction, sorry this it mine first open source pull request.

nickcharlton:
Thanks! I tested it out, and that seems a reasonable way to catch it. Merged now.

PR #1115, opened 2018-03-14
Translate error message to Japanese

okuramasafumi:
I18n file for Japanese didn't contain properly translated error messages so I added them.


nickcharlton:
Wonderful, thank you!

PR #1120, opened 2018-03-17
Ability to export an index page in CSV

pustomytnyk:
Based on @BenMorganIO implementation from #782.

- [x] Show export button on index page if the action is allowed

- [x] Render into CSV same columns as on page. Render ALL records, not only those of current page.

- [x] Tests for CSV export

- [x] Tests for all field types rendered properly in CSV

- [x] Translations

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -12,12 +12,22 @@ module Administrate
       resources = resources.page(params[:page]).per(records_per_page)
       page = Administrate::Page::Collection.new(dashboard, order: order)
 
-      render locals: {
-        resources: resources,
-        search_term: search_term,
-        page: page,
-        show_search_bar: show_search_bar?,
-      }
+      respond_to do |format|
+        format.html do
+          render locals: {
+            resources: resources,
+            search_term: search_term,
+            page: page,
+            show_search_bar: show_search_bar?,
+          }
+        end
+
+        format.csv do
+          name = params[:controller].sub(/^admin\//, '')
+          csv = Administrate::CSV.new(resources, page).generate
+          send_data csv, filename: "#{name}-#{Date.today}.csv"
+        end
+      end
     end
 
     def show
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -22,10 +22,12 @@ module Administrate
           }
         end
 
-        format.csv do
-          name = params[:controller].sub(/^admin\//, '')
-          csv = Administrate::CSV.new(resources, page, view_context).generate
-          send_data csv, filename: "#{name}-#{Date.today}.csv"
+        if show_action?(:export_csv, resources)
+          format.csv do
+            name = params[:controller].sub(/^admin\//, "")
+            csv = Administrate::CSV.new(resources, page, view_context).generate
+            send_data csv, filename: "#{name}-#{Date.today}.csv"
+          end
         end
       end
     end
@@ -178,7 +180,7 @@ module Administrate
 
     def show_search_bar?
       dashboard.attribute_types_for(
-        dashboard.all_attributes,
+        dashboard.all_attributes
       ).any? { |_name, attribute| attribute.searchable? }
     end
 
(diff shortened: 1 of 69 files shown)


pustomytnyk:
Currently, the "Download as CSV" will download all the records despite selected page.

app/controllers/administrate/application_controller.rb
```diff
@@ -163,11 +178,11 @@ def translate_with_resource(key)
 
     def show_search_bar?
       dashboard.attribute_types_for(
-        dashboard.collection_attributes
+        dashboard.collection_attributes,
```
BenMorganIO: Trailing comma.

lib/administrate/field/associative.rb
```diff
@@ -11,6 +11,10 @@ def associated_class
         associated_class_name.constantize
       end
 
+      def short_plain_text
+        display_associated_resource
+      end
```
BenMorganIO: You can probably just alias the methods.

lib/administrate/field/base.rb
```diff
@@ -40,6 +40,10 @@ def to_partial_path
         "/fields/#{self.class.field_type}/#{page}"
       end
 
+      def short_plain_text
+        to_s
+      end
```
BenMorganIO: This method could just be aliased to `to_s` as well.

lib/administrate/field/email.rb
```diff
@@ -6,6 +6,10 @@ class Email < Field::Base
       def self.searchable?
         true
       end
+
+      def short_plain_text
+        data
+      end
```
BenMorganIO: This could be aliased.

lib/administrate/field/password.rb
```diff
@@ -11,6 +11,10 @@ def truncate
         data.to_s.gsub(/./, character)[0...truncation_length]
       end
 
+      def short_plain_text
+        truncate
+      end
```
BenMorganIO: aliasable.

lib/administrate/field/select.rb
```diff
@@ -11,6 +11,10 @@ def selectable_options
         collection
       end
 
+      def short_plain_text
+        data
+      end
```
BenMorganIO: Eh, you get the idea... Aliasable...

lib/administrate/field/associative.rb
pustomytnyk: @BenMorganIO should I use `alias short_plain_text display_associated_resource` or `alias_method :short_plain_text, :display_associated_resource`?

BenMorganIO: > This is because alias is a keyword and it is lexically scoped. It means it treats self as the value of self at the time the source code was read . In contrast alias_method treats self as the value determined at the run time.

- https://blog.bigbinary.com/2012/01/08/alias-vs-alias-method.html

They're both the same in this instance. The code is read at compile time so the performance should be the same. The lexical scoping presents no variation when doing it in the class. `alias` is a keyword and cannot be changed; just as much as you cannot change `do` or `end` in ruby. `alias_method` allows a person to change what happens since it is a method call.

Since the community has a preference for creating change in the system when possible, I would recommend `alias_method`. It's all down to philosophy here, and Ruby would opt for `alias_method`. If this was Elixir, we would be doing `alias`.

pustomytnyk: @BenMorganIO 
As I understand now, If you alias some method, say, `AncestorField#short_plain_text` > `AncestorField#to_s` and override `to_s` in descendant (`DescendantField#to_s`) then `DescendantField#short_plain_text` will still point to `AncestorField#to_s` instead of `DescendantField#to_s` unless you redeclare the alias in `DescendantField`. I think it is not intuitive and I would stick with current implementation.

BenMorganIO: Alright, these points are all very moot. I just felt refactoring to an alias would be nice, but if people need to extend that method, then this shouldn't be a problem.

nickcharlton:
@pustomytnyk How are you getting on with this? Can I help at all?

pustomytnyk:
Updated.

hudadiaz:
Any update on this?

nickcharlton:
This isn't something we're going to do in the short-term and so I'm going to close this. There's lots of good work here though and it should provide inspiration for when we revisit this in future.

sedubois:
Would love to see this supported in core, meanwhile there is https://github.com/SourceLabsLLC/administrate_exportable

PR #1121, opened 2018-03-18
Add fetch default value for options hash

Amrani:
As stated in the [documentation](https://github.com/thoughtbot/administrate/commit/8329b214cea4eada8febbf1552dcbe7e3b9a9153#diff-6f1e8f63b36ca03a54fed082fc955a49R125), a `Field::Polymorphic` should default its classes to an empty array.

- [x] Default `classes` for a `Field::Polymorphic`
- [x] Add spec coverage

### DIFF
--- a/lib/administrate/field/polymorphic.rb
+++ b/lib/administrate/field/polymorphic.rb
@@ -30,7 +30,7 @@ module Administrate
       end
 
       def classes
-        options.fetch(:classes) || []
+        options.fetch(:classes, [])
       end
 
       private
--- a/spec/lib/fields/polymorphic_spec.rb
+++ b/spec/lib/fields/polymorphic_spec.rb
@@ -51,4 +51,25 @@ describe Administrate::Field::Polymorphic do
       expect(field.selected_global_id).to eq(nil)
     end
   end
+
+  describe "#classes" do
+    let(:field) { Administrate::Field::Polymorphic.new(:foo, "hello", :show) }
+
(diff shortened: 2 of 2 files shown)


lib/administrate/field/polymorphic.rb
```diff
@@ -30,7 +30,7 @@ def associated_dashboard(klass = data.class)
       end
 
       def classes
-        options.fetch(:classes) || []
+        options.fetch(:classes, [])
```
FanaHOVA: `.fetch` raises a KeyError when the key is missing, it doesn't return `false`, so the second part of the conditional would have never ran in the previous code:

```
2.4.0 :001 > hash = {}
 => {} 
2.4.0 :002 > res = hash.fetch(:key1) || 'Something'
KeyError: key not found: :key1
	from (irb):2:in `fetch'
	from (irb):2
	from /Users/fana/.rvm/rubies/ruby-2.4.0/bin/irb:11:in `<main>'
2.4.0 :003 > res
 => nil 
```

I'm wondering if users have been rescuing `KeyError` as a result and this will be a breaking change for them?

Amrani: I think if someone were rescuing then it would be to return a default. In that case, the default is likely also an empty Array as it is consistent with other keys from `options`. Where do you think a user may have added the rescue? If they were overriding this method then this change would not affect them. 🤔 What do you think?

Issue #710, opened 2016-12-14
DateFormat Option in Field::DateTime


fogshot:
I found a few possible options for fields in the [Docs](administrate/docs/customizing_dashboards.md), added in #647. 

Unfortunately, no option for date formatting seems to be supported yet. Is this not possible yet or just missing from the docs?

If it is no possible yet, I'd be happy to try my hand in adding this.

mcmire:
@fogshot It's possible, albeit in a limited way, to format the date through i18n: if you set `date.formats.default` or `time.formats.default` then this will control how dates and times are formatted, at least on index and show pages.

mark100net:
```
In the index page, I am getting "00:00:00" in the time part of the DateTime for every record when I do this (in config/locales/en.yml):
```
    date:
      formats:
      default: "%d/%m/%Y %H:%M:%S"
    time:
      formats:
        default: "%d/%m/%Y %H:%M:%S"

```
or this (in the dashboard):


```    created_at: Field::DateTime.with_options(format: "%d/%m/%Y %H:%M:%S"),

It works fine on the `show` page.

mark100net:
I can see that it is calling `date` on the field in the index view and `datetime` in the show view. Was this intentional? Kind of largely defeats the idea of setting the format.

nickcharlton:
Hi @mark100net! That sounds like a bug. Do you think you'd be able to open a PR for that?

mark100net:
@nickcharlton I will try to get to it this weekend.

PR #1129, opened 2018-04-01
include time for date_time field in index

mark100net:
#710 

Just modifies the index view template for the date_time field to display the time (as well as the date) as is being done in the show template.


nickcharlton:
Looks good, thanks!

pablobm:
@nickcharlton - Should this also have deleted the method `Administrate::Field::DateTime#date` and related specs? As I see it, it was left as dead code 🤔

PR #1134, opened 2018-04-10
Make the resource action "show" optional

shvetsovdm:
This will allow to omit show routes and actions for admin resources.

We should consider to make a conditionally change of update/create redirects to index action if no show action is not provided. It will be handy if administrate will be able to automatically handle the show action absence.


shvetsovdm:
premature PR submission

Issue #1139, opened 2018-04-11
`.byebug_history` and `gemfiles/.bundle/` not in .gitignore


shvetsovdm:
```
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	.byebug_history
	gemfiles/.bundle/
```

This files have not been added to .gitignore intentionally? I think `.byebug_history` should be included. Not sure about `gemfiles/.bundle/config` should be ignored or included in the repo.

nickcharlton:
Thanks for raising this! I probably have it in my global `.gitignore` so I've never seen them.

PR #1145, opened 2018-04-13
Add `.byebug_history` and `gemfiles/.bundle/` to .gitignore

bikramwp:
Ref https://github.com/thoughtbot/administrate/issues/1139


nickcharlton:
Thanks!

Issue #1114, opened 2018-03-13
Randomly failing test in show_page_spec.rb:5


pustomytnyk:
@pablobm, it seem the `rspec ./spec/features/show_page_spec.rb:5` test from #1078 has some random issue not related to the ordering https://circleci.com/gh/thoughtbot/administrate/2983. I was unable to reproduce it and don't know what can be a possible cause.

pablobm:
@pustomytnyk Thank you for pointing out. After some playing with it, I think what must be happening is that we don't specify a sorting order, and therefore the DB doesn't necessarily return orders always in the expected order. In some occasions, this would result in the DB returning the same orders for the first and the second page, which I think is what happened at https://circleci.com/gh/thoughtbot/administrate/2983.

I'm creating a PR to ensure a consistent sorting order. Let's see if we see the same problem again.

PR #1150, opened 2018-04-20
There's a spec that requires consistent ordering here

pablobm:
https://github.com/thoughtbot/administrate/issues/1114 points out that `spec/features/show_page_spec.rb:28` may fail randomly (rarely, but it can). This is probably because the spec navigates a paginated list, but we are not specifying an order for this list.

This should fix the problem... if that is indeed the issue here. We can revisit if we see this failing randomly again.

Additionally, I rephrased that spec's description, to something that sounded a bit better to me.


nickcharlton:
Thanks @pablobm!

Issue #1111, opened 2018-03-08
automatically remove models from sidebar that are inaccessible?


jmmastey:
I have an administrate install with two types of users (`admin` and `staff`), authorized via Pundit.

Staff are not able to edit the record `Neighborhoods`, admins are. Example part of the policy:

```
class NeighborhoodPolicy
  def index?
    @user.admin?
  end
end
```

Really, staff can't do _anything_ with neighborhoods, so I'd like to remove that link from the left navigation entirely. Would y'all be amenable to a PR that changes `Administrate::Namespace.resources` to check authorization of the index action and conditionally drop that from navigation?

nickcharlton:
Yes, that'd be great!

jmmastey:
@nickcharlton submitted!

PR #1151, opened 2018-04-21
Don't show navigation items if the user can't view the collection anyway

jmmastey:
Per discussion on #1111, removes items from the left navigation if you aren't allowed the `index?` action on them.

I'm not 100% happy with the test for this, but this didn't fit well with the feature specs that were already in the app and I didn't want to drag a huge change into a small feature. Let me know if they still need work (and a hint of what you'd prefer would be appreciated).

Thanks!

### DIFF
--- a/app/views/administrate/application/_navigation.html.erb
+++ b/app/views/administrate/application/_navigation.html.erb
@@ -9,10 +9,12 @@ as defined by the routes in the `admin/` namespace
 
 <nav class="navigation" role="navigation">
   <% Administrate::Namespace.new(namespace).resources.each do |resource| %>
-    <%= link_to(
-      display_resource_name(resource),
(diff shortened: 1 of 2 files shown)


dinhvantiep:
How can we access new action ?

jnsoo123:
Wouldn't it be better if we don't allow the route to be accessed? Like for example if you have a `User` model  and index is not authorized, it should raise an error accessing `/admin/users`? same with other actions. Not just hiding the link?

jmmastey:
@jnsoo123 that'd probably _also_ be a very good idea. I don't know off the top of my head what the current auth logic is, but for some reason I think your case may already be handled?

either way, submitted this PR back in April and no activity on it, so I think I'm going to leave it be. looking at the most recently merged code, this codebase is more or less unmaintained.

PR #1153, opened 2018-04-21
Add Indonesian locale

styd:
Hi! I want to add Indonesian to the list of locales. Thanks.

### DIFF
--- a/config/locales/administrate.id.yml
+++ b/config/locales/administrate.id.yml
@@ -1,0 +1,28 @@
+---
+id:
+  administrate:
+    actions:
+      confirm: Anda yakin?
+      destroy: Hapus
+      edit: Perbarui
+      edit_resource: Perbarui %{name}
+      show_resource: Tampilkan %{name}
+      new_resource: "%{name} baru"
+      back: Kembali
+    controller:
+      create:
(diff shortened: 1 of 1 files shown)


nickcharlton:
Hi @styd! Thank you for your language contribution!

Something I'm trying to do is have someone keep maintaining the languages they contribute, because if one us doesn't speak it then they end up getting either left behind (and breaking the build), or filled in with not so great Google Translate answers. Do you think you'd be able to do this?

styd:
Hi, @nickcharlton! I understand what you mean and I'm willing to maintain the Indonesian locale. I too want it to always be updated with the correct translations since I also use administrate in my project. So, if there are additions/changes in the locale please let me know.

[Here is an example of how I want Indonesian locale to be supported in another project](https://github.com/radar/humanize/issues/35). In short, I did some research to get the best possible translations and tried to ask a language expert about some unknown terms even though Indonesian is my native language.

nickcharlton:
That sounds perfect! My main concern is ensuring that changes we make don't cause significant language regressions, so if you can watch out for changes and open new PRs if needed that'd be wonderful.

I'll go ahead and merge this now.

styd:
Ok, will do. Thanks for merging.

PR #1156, opened 2018-04-25
Add id HTML attribute for rows on resources index page

iamdidev:
This change will allow referencing rows by id not by data-url which will be useful when a resource will not have show action and will have no data-url.


purinkle:
@iamdidev, is this a change you would still like to make?

iamdidev:
@purinkle hey! I no longer an "Administrate" user, so I can't say will it be valuable now or not. But if this still has value as 1 year ago then I will be happy to contribute again to the project.

PR #1158, opened 2018-05-01
Fix Option Examples in Customization Documentation

GeekOnCoffee:


composerinteralia:
Thanks! It looks like this got address in #1166 and #1199

PR #1161, opened 2018-05-04
Allow parameters that are more permissive

tabfugnic:
When using the Clearance gem's backdoor to access views in
administrate, tests prematurely break because of the `as` parameter
used to do quick sign in. While it's probably fair that we don't want
persist that parameter through the GET request, it shouldn't matter
what parameters we pass in a GET request.

* Use `permit!` to allow anything through on a GET request

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -33,11 +33,11 @@ module Administrate
     end
 
     def sanitized_order_params
-      params.permit(:search, :id, :order, :page, :per_page, :direction, :orders)
+      params.permit!
     end
 
     def clear_search_params
-      params.except(:search, :page).permit(:order, :direction, :per_page)
+      params.except(:search, :page).permit!
     end
   end
 end


nickcharlton:
@tabfugnic, do you think you'd be able to fix the conflicts on this?

app/helpers/administrate/application_helper.rb
```diff
@@ -32,18 +32,12 @@ def resource_index_route_key(resource_name)
...
-      params.permit(:search, :id, :page, :per_page, association_params)
+    def sanitized_order_params
+      params.permit!
     end
 
     def clear_search_params
-      params.except(:search, :page).permit(
-        :per_page, resource_name => %i[order direction]
-      )
+      params.except(:search, :page).permit!
```
pablobm, line 40: Similarly to my other comment:

```suggestion
      params.except(:search, :page).to_unsafe_h
```

pablobm (changes requested):
I'm not clear whether this may have unexpected side-effects. Thoughts on my review comments?

pablobm:

PR #1164, opened 2018-05-12
Update bundle to fix tests on CI

pustomytnyk:
It actually requires only Capybara to be updated, but why not update all then.

I don't know if there is any difference which `Capybara.server` to use. Webrick just don't need adding new dependency to Gemfile.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -62,13 +62,13 @@ GEM
       bundler (~> 1.2)
       thor (~> 0.18)
     byebug (10.0.0)
-    capybara (2.18.0)
+    capybara (3.1.0)
       addressable
       mini_mime (>= 0.1.3)
-      nokogiri (>= 1.3.3)
-      rack (>= 1.0.0)
-      rack-test (>= 0.5.4)
-      xpath (>= 2.0, < 4.0)
+      nokogiri (~> 1.8)
+      rack (>= 1.6.0)
+      rack-test (>= 0.6.3)
+      xpath (~> 3.0)
     cliver (0.3.2)
     coderay (1.1.2)
     concurrent-ruby (1.0.5)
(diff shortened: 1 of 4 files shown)


Gemfile.lock
```diff
@@ -52,41 +52,41 @@ GEM
...
       execjs
     awesome_print (1.8.0)
     builder (3.2.3)
     bundler-audit (0.6.0)
       bundler (~> 1.2)
       thor (~> 0.18)
-    byebug (10.0.0)
-    capybara (2.18.0)
+    byebug (10.0.2)
+    capybara (3.1.0)
```
nickcharlton: Can we just update Capybara here? I'd like there to be just one commit fixing Capybara if that's what we need to be doing to fix CI.

spec/features/search_spec.rb
```diff
@@ -85,6 +85,6 @@ def submit_search
   end
 
   def order_row_match(order)
-    %r{#{order.id} #{order.customer.name} }
+    /#{order.id}\s+#{order.customer.name}\s+/
```
nickcharlton: Was this broken with the Capybara change too?

spec/rails_helper.rb
```diff
@@ -31,8 +31,10 @@ module Features
 end
 
 ActiveRecord::Migration.maintain_test_schema!
+
 Capybara.javascript_driver = :poltergeist
 Capybara.register_driver :poltergeist do |app|
   options = { phantomjs_options: ["--load-images=no"] }
   Capybara::Poltergeist::Driver.new(app, options)
 end
+Capybara.server = :webrick
```
nickcharlton: Currently, CI fails because it can't find Puma. Should we instead be adding that, rather than switching to Webrick here? If I recall correctly, Webrick won't execute any Javascript, which we may have in some tests?

pustomytnyk: As I understand, `Capybara.server` is just the server, not js-driver.
Maybe there is some performance difference, so I agree this should be replaced with Puma.

spec/features/search_spec.rb
pustomytnyk: Yes, the new version had put tabulation chars instead of just spaces here.

nickcharlton:
Thanks for doing this! I had a with @odlp about this, and realised that just configuring to use Webrick would be a better outcome than adding an extra dependency on Puma. I'm going to close this and merge #1168 instead for that.

PR #1117, opened 2018-03-14
Add translation for the Albanian language

acelina:
This PR adds the locale for the Albanian language.

PR #1165, opened 2018-05-13
Change Albanian two-letter code to "sq" as per ISO 639-1

pustomytnyk:
It would be logical if language codes correspond ISO 639-1.

### DIFF
--- a/config/locales/administrate.sq.yml
+++ b/config/locales/administrate.sq.yml
@@ -1,5 +1,5 @@
 ---
-al:
+sq:
   administrate:
     actions:
       confirm: A jeni te sigurtë?


nickcharlton:
Do you know if there's a reason that it was originally `al`?

pablobm:
Looks like `al` is the code for the country, while `sq` is the code for the language. No idea of how this plays into locales.

Looking at https://www.oracle.com/java/technologies/javase/jdk8-jre8-suported-locales.html, I see that the language tag is `sq-AL`. Comparing this with something else like `es-AR`, then I would infer that `sq` is the correct one to go with. But I can't tell for certain.

The original author of this file is @acelina (see https://github.com/thoughtbot/administrate/pull/1117). I wonder if he could shed some light into this?

acelina:
Hello. Yes, correct one to go is sq. That's the ISO code for the language. Apologies for the mistake that slipped through.

pablobm:
Thank you. @pustomytnyk, would you be able to rebase and push this? The build failed, and I think it's just a glitch that has been solved since.

pablobm:
OK, I tried running CI of this on my repo and it passed, so I'm going to try and merge this :crossed_fingers:

pablobm:
That worked! :D

PR #1166, opened 2018-05-21
Add forgotten .with_options in documentation

kevcha:
The customizing dashboard documentation was missing a `.with_options` when specifying if a belongs to association is searchable or not and on which field.

By the way, do you plan to update online documentation https://administrate-prototype.herokuapp.com/ ?


nickcharlton:
Thanks for adding this!

I usually update the prototype site when we do a release …which it sounds like I missed when we did v0.10.0. I'll sort that now.

PR #1170, opened 2018-05-30
Reduces controller comments

kdiogenes:
Reduces the amount of information that will be deleted :yum:


nickcharlton:
This is deliberate to help you get going (as documentation). I would of course expect you to delete it all when you start. So I'm going to close this.

PR #1172, opened 2018-05-31
Remove ids, capitalize and pluralize HasMany fields on form

GeekOnCoffee:
Original:
Original

With Change
New


nickcharlton:
Sorry this one never got look at until now; I think the best way to solve this is to provide localisations and so I'm going to close this (mostly so we avoid introducing `gsub`!). Thanks for your work here.

PR #1192, opened 2018-07-20
Fix i18n inconsistencies in forms for associations

### LINKED ISSUE #1105: i18n inconsistencies in forms for associations
There are some inconsistencies in the way Administrate handles i18n in places. For example, this is how form labels are rendered for forms in the `HasMany`/`BelongsTo`/`HasOne` cases:

* `HasMany` (eg: `Order#line_items`)
  - Rendered as `f.label field.attribute_key, field.attribute`
  - Gives "line_items", which is better than "line_item_ids", but isn't great
  - No i18n.

* `BelongsTo` (eg:  `Customer#country`)
  - Rendered as `f.label field.permitted_attribute`
  - Defaults to "Country code" (similar to "Country id" in this case).
  - Allows i18n as `helpers.label.customer.country_code`.

* `HasOne` (eg:  `Product#product_meta_tag`)
  - Rendered as `field.nested_form.resource_name.titleize`
  - Gives "Product Meta Tag".
  - No i18n.

I see that https://github.com/thoughtbot/administrate/pull/707/files touches on `app/views/fields/has_many/_index.html.erb`, which was something I had noted was inconsistent in this way. Otherwise, this is a relatively different thing from that PR.

cpytel:
Fixes #1105. As described in #1105, there are some inconsistencies in
the way Administrate handles labels for `HasMany`/`BelongsTo`/`HasOne`
associations.

* `HasMany` (eg: `Order#line_items`)
  - Gives "line_items" as the label
  - No i18n.

* `BelongsTo` (eg:  `Customer#country`)
  - Defaults to "Country code" (similar to "Country id" in this case).
  - Allows i18n as `helpers.label.customer.country_code`.

* `HasOne` (eg:  `Product#product_meta_tag`)
  - Gives "Product Meta Tag".
  - No i18n.

The behavior of the `BelongsTo` field is most correct, and is preserved
in this commit.

The behavior of `HasMany` and `HasOne` has been improved as follows:

* `HasMany` (eg: `Order#line_items`) now no longer overrides the text
  given to the `label` helper, and passes the attribute name. This results
  in a default label of "Line items" and one that can be overridden with
  i18n with a key at the standard location (`helpers.label.order.line_items`)
* `HasOne` (eg:  `Product#product_meta_tag`) now can be overriden with
  the key `helpers.label.order.line_items`.

The `HasOne` is rendered into a nested form, inside of a fieldset and
with a legend. The text of the "label" in this nested form is actually
in a legend tag. I considered the fact that this was not using the
`label` helper, whether it should be switch to a label from a legend, or
whether we should have a non-standard i18n key. In the end, I decided
that the least surprising behavior would be to customize the "label" for
that attribute at the expected label's i18n key.

### DIFF
--- a/app/views/fields/has_many/_form.html.erb
+++ b/app/views/fields/has_many/_form.html.erb
@@ -20,7 +20,7 @@ and is augmented with [Selectize].
 %>
 
 <div class="field-unit__label">
-  <%= f.label field.attribute_key, field.attribute %>
+  <%= f.label field.attribute, for: "#{f.object_name}_#{field.attribute_key}" %>
 </div>
 <div class="field-unit__field">
   <%= f.select(field.attribute_key, nil, {}, multiple: true) do %>
--- a/app/views/fields/has_one/_form.html.erb
+++ b/app/views/fields/has_one/_form.html.erb
@@ -18,7 +18,7 @@ The form will be rendered as nested_from to parent relationship.
 
 <%= f.fields_for field.attribute, field.data || field.nested_form.resource.class.new do |has_one_f| %>
   <fieldset class="field-unit--nested">
-    <legend><%= field.nested_form.resource_name.titleize %></legend>
+    <legend><%= t "helpers.label.#{f.object_name}.#{field.nested_form.resource_name}", default: field.nested_form.resource_name.titleize %></legend>
     <% field.nested_form.attributes.each do |attribute| -%>
       <div class="field-unit field-unit--<%= attribute.html_class %>">
         <%= render_field attribute, f: has_one_f %>
--- a/spec/administrate/views/fields/has_many/_form_spec.rb
+++ b/spec/administrate/views/fields/has_many/_form_spec.rb
@@ -13,14 +13,14 @@ describe "fields/has_many/_form", type: :view do
         locals: { f: fake_form_builder, field: has_many },
       )
 
-      expect(rendered).to include("Associated Objects")
+      expect(rendered).to include("Associated objects")
     end
   end
 
   def fake_form_builder
     double("Form Builder").as_null_object.tap do |form_builder|
       allow(form_builder).to receive(:label) do |*args|
-        args.second.to_s.titleize
+        args.first.to_s.humanize
       end
     end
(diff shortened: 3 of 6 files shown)


app/views/fields/has_many/_form.html.erb
```diff
@@ -20,7 +20,7 @@ and is augmented with [Selectize].
 %>
 
 <div class="field-unit__label">
-  <%= f.label field.attribute_key, field.attribute %>
+  <%= f.label field.attribute, for: "#{f.object_name}_#{field.attribute_key}" %>
```
cpytel: If there is an existing Rails helper I can use to generate this label key, I think that would be nice, but I couldn't find one.

app/views/fields/has_one/_form.html.erb
```diff
@@ -18,7 +18,7 @@ The form will be rendered as nested_from to parent relationship.
 
 <%= f.fields_for field.attribute, field.data || field.nested_form.resource.class.new do |has_one_f| %>
   <fieldset class="field-unit--nested">
-    <legend><%= field.nested_form.resource_name.titleize %></legend>
+    <legend><%= t "helpers.label.#{f.object_name}.#{field.nested_form.resource_name}", default: field.nested_form.resource_name.titleize %></legend>
```
cpytel: If there is an existing Rails helper or class method I could use to generate this i18n key, that would be nice, but I looked and couldn't find one.

spec/example_app/config/database.yml
```diff
@@ -1,4 +1,5 @@
 development: &default
+  host: localhost
```
nickcharlton: Was this necessary to get up and running? Should we include this in this PR, or elsewhere?

cpytel:
Yes, it was necessary for me, but this is because of my computer config. I meant to remove it before opening the PR, but forgot. Thanks for catching, I’ve removed it in bc37606.

---
Chad Pytel, Developer and CEO
617-482-1300 x113
https://thoughtbot.com 
https://twitter.com/thoughtbot

PR #1200, opened 2018-08-09
Fix error in suggested translation key

jean-tymate:
Rails expects `other` key and not `others`.


nickcharlton:
Wonderful, thanks for spotting and fixing this!

PR #1203, opened 2018-08-23
Support search over multiple fields

ap4y:
Support search on multiple fields. Renamed `:searchable_field` to `:searchable_fields`

### DIFF
--- a/lib/administrate/field/deferred.rb
+++ b/lib/administrate/field/deferred.rb
@@ -30,7 +30,18 @@ module Administrate
       end
 
       def searchable_field
+        ActiveSupport::Deprecation.warn(
+          "searchable_field is deprecated, use searchable_fields instead",
+        )
         options.fetch(:searchable_field)
+      end
+
+      def searchable_fields
+        if options.key?(:searchable_field)
+          [searchable_field]
+        else
+          options.fetch(:searchable_fields)
+        end
       end
 
       def permitted_attribute(attr, _options = nil)
--- a/spec/lib/administrate/search_spec.rb
+++ b/spec/lib/administrate/search_spec.rb
@@ -29,14 +29,21 @@ class MockDashboardWithAssociation
       searchable: true,
       searchable_field: "name",
     ),
+    author: Administrate::Field::BelongsTo.with_options(
+      searchable: true,
+      searchable_fields: ["first_name", "last_name"],
+    ),
     address: Administrate::Field::HasOne.with_options(
       searchable: true,
-      searchable_field: "street",
+      searchable_fields: ["street"],
     ),
   }.freeze
 end
 
 describe Administrate::Search do
+  before { ActiveSupport::Deprecation.silenced = true }
+  after { ActiveSupport::Deprecation.silenced = false }
+
   describe "#run" do
     it "returns all records when no search term" do
       begin
@@ -132,7 +139,11 @@ describe Administrate::Search do
       let(:expected_query) do
         [
           'LOWER(CAST("roles"."name" AS CHAR(256))) LIKE ?'\
+          ' OR LOWER(CAST("authors"."first_name" AS CHAR(256))) LIKE ?'\
+          ' OR LOWER(CAST("authors"."last_name" AS CHAR(256))) LIKE ?'\
           ' OR LOWER(CAST("addresses"."street" AS CHAR(256))) LIKE ?',
+          "%тест test%",
+          "%тест test%",
           "%тест test%",
           "%тест test%",
         ]
@@ -141,14 +152,14 @@ describe Administrate::Search do
       it "joins with the correct association table to query" do
         allow(scoped_object).to receive(:where)
 
-        expect(scoped_object).to receive(:joins).with(%i(role address)).
+        expect(scoped_object).to receive(:joins).with(%i(role author address)).
           and_return(scoped_object)
 
         search.run
       end
 
       it "builds the 'where' clause using the joined tables" do
-        allow(scoped_object).to receive(:joins).with(%i(role address)).
+        allow(scoped_object).to receive(:joins).with(%i(role author address)).
           and_return(scoped_object)
 
         expect(scoped_object).to receive(:where).with(*expected_query)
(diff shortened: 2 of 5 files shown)


tomasmcm:
@ap4y what is stopping this PR from being merged (apart from the conflict)? It would be amazing to get this merged soon 🤞

ap4y:
I will rebase this one today, shouldn't be too hard. The rest is not under my control unfortunately.

jankeesvw:
Would love this feature. I was looking for a way to do this ✨ Can I do anything to get this merged?

sedubois:
@ap4y now that a new Administrate release has been published and that some fixes related to search are also on master (e.g #1569), would it be possible for you to rebase this PR so that hopefully it can get merged?

ap4y:
@sedubois I have rebased my changes to the current master.

lib/administrate/field/deferred.rb
```diff
@@ -29,8 +29,8 @@ def searchable?
         options.fetch(:searchable, deferred_class.searchable?)
       end
 
-      def searchable_field
-        options.fetch(:searchable_field)
+      def searchable_fields
+        options.fetch(:searchable_fields)
```
nickcharlton, line 33: This will end up being a breaking change. To reduce the impact of breakages, what do you think about still keeping `searchable_field` method, but in a way that reformats it for the new method?

We could then use [`ActiveSupport::Deprecation`](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation.html) to highlight it and remove it in the release afterwards.

ap4y: Updated

nickcharlton:
Great! Could we add a test too, so that we know we're not breaking the old functionality until we remove it?

We'll likely want to suppress the deprecation warning output during that test, too, so that we avoid making our test suite noisy.

ap4y:
Updated `search_spec` to keep one instance of the `searchable_field`. Updated `spec/rails_helper` to suppress deprecation warnings.

spec/rails_helper.rb
```diff
@@ -31,6 +31,7 @@ module Features
 end
 
 ActiveRecord::Migration.maintain_test_schema!
+ActiveSupport::Deprecation.silenced = true
```
nickcharlton, line 34: Can we do this around the specific test instead of globally? We don't want to suppress all deprecation warnings, some of which may come from `ActiveSupport`/Rails itself.

This would also help show which test is there to ensure we're doing this in a backwards compatible way, as looking at the changes here it's hard to tell what we'll need to remove when we change the feature completely.

ap4y: Moved deprecation silencing to the spec

lib/administrate/field/deferred.rb
```diff
@@ -30,9 +30,20 @@ def searchable?
...
       def searchable_field
+        ActiveSupport::Deprecation.warn(
+          "searchable_field is deprecated, use searchable_fields instead",
+        )
         options.fetch(:searchable_field)
       end
 
+      def searchable_fields
+        if options.key?(:searchable_field)
+          [searchable_field]
```
pablobm, line 41: Am I right in thinking that the warning should also be shown here?

The method `searchable_field` is indeed deprecated, and it will be called by outdated code, so it's good that it shows the warning. On the other hand, a field defined before this change will define `:searchable_field` when it should be `:searchable_fields`. This method `searchable_fields` will be called by up-to-date code, making the translation, and no warning will be shown to users who used the deprecated option. I think it should be shown. What do you think?

spec/lib/administrate/search_spec.rb
```diff
@@ -29,14 +29,21 @@ class MockDashboardWithAssociation
...
       searchable: true,
-      searchable_field: "street",
+      searchable_fields: ["street"],
     ),
   }.freeze
 end
 
 describe Administrate::Search do
+  before { ActiveSupport::Deprecation.silenced = true }
+  after { ActiveSupport::Deprecation.silenced = false }
```
pablobm, line 45: I think stubbing this with RSpec's facilities might be marginally better, as it automatically un-stubs after the example. What do you think?

```ruby
before do
  allow(ActiveSupport::Deprecation).to receive(:silenced).and_return(true)
end
```

lib/administrate/field/deferred.rb
ap4y: Logic here is a bit different. Existing user code that relies on older administrate versions will be using `:searchable_field` option for the `Filed::` associations. Method `searchable_field` is  unlikely to be used by users, this method is only called inside `Administrate::Search::Query`. With this patch `Administrate::Search::Query` will switch to using `searchable_fields`. So what we are trying to do with this code is provide support for old `:searchable_field` option inside updated `Administrate::Search::Query`. Hope this makes it a bit more clear.

spec/lib/administrate/search_spec.rb
ap4y: To be honest I don't like this approach, it feels semantically wrong to use stubs for this use case. I think current version shows intent quite well and version with stubs only provides marginal syntactical improvements. I can change it if you feel strongly about it though.

pablobm: Nah, it's ok. Thank you for your thoughts :+1:

lib/administrate/field/deferred.rb
pablobm: Yes, let me rephrase. Say that a user is doing the following:

```ruby
    parts: Field::HasMany.with_options(
      searchable: true,
      searchable_field: :name,
    ),
```

After merging this PR, they'll be using a deprecated option, `:searchable_field`, which should now be `:searchable_fields`:

```ruby
    parts: Field::HasMany.with_options(
      searchable: true,
      searchable_fields: [:name],
    ),
```

It's great that this MR provides this compatibility, but I think that eventually we should phase out this option as it won't be documented any more, and it will reduce maintenance costs going forward.

Therefore I think there should be a warning. Afterwards we can actually remove the option, probably after a couple of versions, so that the change is not too sudden on our users.

What do you think?

ap4y: I think this is already handled. If first example will be executed in the context of the new version the following will happen:

- Query related code will access `#searchable_fields`
- `#searchable_fields` will check that there is a `:searchable_field` key available in the options
- This will trigger `[searchable_field]` call
- `#searchable_field` will trigger deprecation warning

pablobm: Oh, of course. You are correct. Thank you for taking the time to explain this :+1:

ap4y:
@nickcharlton Could you please check this PR one more time? Thanks!

PR #1206, opened 2018-08-26
Upgrade Ruby to 2.5.1

okuramasafumi:
Using latest version of Ruby is always a good idea :)


nickcharlton:
Thanks for opening this!

PR #1211, opened 2018-09-01
Do not pluralize resource name on new page

### LINKED ISSUE #1205: New page title should not be pluralized
This:
https://github.com/thoughtbot/administrate/blob/2c54226aeaebc1ac5316ccd22e87d3080e3608ed/app/views/administrate/application/new.html.erb#L21

Should probably use the same logic as this:
https://github.com/thoughtbot/administrate/blob/2c54226aeaebc1ac5316ccd22e87d3080e3608ed/app/views/administrate/application/index.html.erb#L47

dvanderbeek:
fixes #1205

### DIFF
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -79,7 +79,6 @@ textarea {
 }
 
 textarea {
-  min-height: 550px;
   resize: vertical;
 }
 
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -79,6 +79,7 @@ textarea {
 }
 
 textarea {
+  min-height: 550px;
   resize: vertical;
 }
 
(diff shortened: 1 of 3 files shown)


app/views/administrate/application/new.html.erb
```diff
@@ -18,7 +18,7 @@ to do the heavy lifting.
 <% content_for(:title) do %>
   <%= t(
     "administrate.actions.new_resource",
-    name: display_resource_name(page.resource_name).titleize
+    name: page.resource_name.titleize
```
nickcharlton: Is there a corresponding test for this, too? If not, would you be able to contribute one?

app/assets/stylesheets/administrate/base/_forms.scss
```diff
@@ -79,6 +79,7 @@ textarea {
 }
 
 textarea {
+  min-height: 550px;
```
nickcharlton: Do you mean to change this in this PR too?

dvanderbeek: oops, no I only meant to do that in my fork which I have an app using directly from github. I'll revert this change as soon as I can.

app/views/administrate/application/new.html.erb
dvanderbeek: Will do

vifreefly: Looks like it didn't work with i18n translated model name

dvanderbeek:
@nickcharlton I finally had a chance to clean up this PR. Let me know how it looks!

PR #1212, opened 2018-09-02
Fix typo in customizing_dashboards.md

AlbinOS:
There is a typo in the Field::BelongsTo searchable example causing copy paste code to fail


AlbinOS:
Dupplicates #1171 and #1203

PR #1214, opened 2018-09-15
Adds missing colon to name resouces on admin namespace of Rails API e…

lucasduarte:
There is a missing semicolon in a controller name of Rails API settings that results in an error since it is a symbol


nickcharlton:
Thanks!

Issue #1188, opened 2018-07-06
Ordering BelongsTo's column alphabetically


thealiilman:
Hallo. 😄 I need some help, as I've spent the past day or two digging through the repo's issues / PRs figuring out how to order a record's belongs_to association alphabetically in its collection.

These are the gems that I think are relevant to this issue.
```
administrate (0.10.0)
      actionpack (>= 4.2, < 6.0)
      actionview (>= 4.2, < 6.0)
      activerecord (>= 4.2, < 6.0)
      autoprefixer-rails (>= 6.0)
      datetime_picker_rails (~> 0.0.7)
      jquery-rails (>= 4.0)
      kaminari (>= 1.0)
      momentjs-rails (~> 2.8)
      sass-rails (~> 5.0)
      selectize-rails (~> 0.6)
    administrate-field-image (1.1.0)
      administrate (>= 0.2.0.rc1)
rails (5.0.7)
      actioncable (= 5.0.7)
      actionmailer (= 5.0.7)
      actionpack (= 5.0.7)
      actionview (= 5.0.7)
      activejob (= 5.0.7)
      activemodel (= 5.0.7)
      activerecord (= 5.0.7)
      activesupport (= 5.0.7)
      bundler (>= 1.3.0)
      railties (= 5.0.7)
      sprockets-rails (>= 2.0.0)
```

In my situation, I have a Dish model and a Restaurant model. Dish belongs to Restaurant and Restaurant has many dishes.
I noticed that when one clicks the `<th>` for Restaurant, it orders the column based on the restaurant's `id` and not the restaurant's name. I want to change it so that, it orders the column based on the restaurant's name and alphabetically.

Right now, it appears like so.
<img width="336" alt="screen shot 2018-07-06 at 15 42 48" src="https://user-images.githubusercontent.com/26957580/42366490-e88f6128-8133-11e8-9e63-d783538c11aa.png">

I've tried using the `order` option, but according to the docs, it's only for dropdown menu, not for a table.
Is there a way to let it be ordered alphabetically based on its name?

splashinn:
@thealiilman What you'll want to do is generate the views for your resource: (`rails g administrate:views Restaurant`) 

In the `_collection.html.erb` file within the `app/views/admin/restaurants`, you'll see `resources.each do |resource|...`. 

Change that to `resources.order("name ASC").each do |resource|...` and that should list your restaurants alphabetically.

thealiilman:
[...]

PR #1215, opened 2018-09-16
Enable sort by column inside associated record (BelongsTo)

w0st:
Hello!

I've have a proposition for solving problems like this one: https://github.com/thoughtbot/administrate/issues/1188

If your dashboard has a field which is a subclass of (or it is a) `Administrate::Field::BelongsTo` then these changes allows you sort by column inside associated record based on `order` option.

I would appreciate your comments

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -101,10 +101,24 @@ module Administrate
     end
 
     def order
-      @order ||= Administrate::Order.new(
-        params.fetch(resource_name, {}).fetch(:order, nil),
-        params.fetch(resource_name, {}).fetch(:direction, nil),
-      )
+      attribute = params.fetch(resource_name, {}).fetch(:order, nil)
+      direction = params.fetch(resource_name, {}).fetch(:direction, nil)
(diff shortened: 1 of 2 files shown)


nickcharlton:
This is looking good! Do you think you'd be able to add some tests for this feature, too?

horacio:
@w0st Hello! I'm interested in helping completing this PR. Let me know if you'd like me to write the tests for this new functionality. Thanks!

w0st:
Thanks for reminding me about this PR. I'll check the existing test suite in Administrate and notify you about the status. I remember that I spot small problems with existing unit tests when I was creating this PR. I wanted to fix them before this.

pablobm:
An alternative would be to implement https://github.com/thoughtbot/administrate/pull/1506 (still in discussion; may not make it through).

nickcharlton:
@w0st, where you able to look at this one again?

I've not looked into #1506 properly myself, but what do you think about that when you compare it to this approach?

pablobm:
Superseeded by https://github.com/thoughtbot/administrate/pull/2208

PR #1218, opened 2018-09-19
"Advanced" search

patrick-gleeson:
Forgive a rather presumptuous PR. I know that #157 is open and there has not been a general agreement on exactly what smart searching would look like.

I just needed more advanced search functionality for a site I'm building, and so I built what I need, and I want to see if anyone else would find it useful.

How this works:

Suppose you have a `User` class with the following in its Administrate dashboard
```ruby
ATTRIBUTE_TYPES = {
  name: Field::String,
  email: Field::String,
  activated: Field::Boolean
}
```

Search term: `name:Steve`

* This will search for 'steve' in the `name` field

Search term: `name:Steve example`

* This will search for 'steve' in the `name` field and 'example' in everything searchable __except__ the `name` field

Search term: `name:'Steve Jones'`

* This will search for 'steve jones' in the `name` field

Search term: `activated:true`

* This will search for records with `activated` set to `true`

Search term: `email:example.com name:"Steve Jones" activated:false hoopla`

* This will search for records where `email` matches 'example.com', `name` matches 'Steve Jones', `activated` is set to `true`, and it will __ignore__ 'hoopla' because there are no fields left for it to search.

I look forward to any thoughts and feedback!

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -1,8 +1,11 @@
 require "active_support/core_ext/module/delegation"
 require "active_support/core_ext/object/blank"
+require "administrate/field/boolean"
 
 module Administrate
   class Search
+    ADVANCED_CLAUSE_REGEX = /(\w*)\:((?:["'].*?["'])|(?:\w+))/
+
     def initialize(scoped_resource, dashboard_class, term)
       @dashboard_class = dashboard_class
       @scoped_resource = scoped_resource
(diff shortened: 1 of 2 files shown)


patrick-gleeson:
@conradbeach In its current form it doesn't allow specifying associations to search. It might be relatively straightforward to implement that, though - I just haven't done much with associations yet so didn't think to try to support it.

mmrobins:
Any word on getting something like this merged?  This would be awesome to have

pablobm:
Search in Administrate could definitely get some improvement, but perhaps not as super-smart default option, but rather by providing an easy way to provide your own.

The problem with implementing anything is that different people expect search to work differently in different situations. Take this PR as an example. When you search for `name:Steve example`, it "will search for 'steve' in the name field and 'example' in everything searchable except the name field". That's not what I would expect: I would expect a search for `steve` in the name field and for `example` across all fields with no exception.

Therefore I hesitate to accept any specific changes.

In my mind, search should be something for which Administrate provides a bare minimum default (eg: all words across all searchable fields, perhaps with quoted phrases), and users would provide their own custom algorithm by overriding controllers.

Currently, search is implemented in `Administrate::ApplicationController#index`. This is what this action looks like at the moment:

https://github.com/thoughtbot/administrate/blob/c16b8d1ee3a5ef1d622f9470738e89d73dbb8f1b/app/controllers/administrate/application_controller.rb#L5-L22

Line 8 is the key. It will take the search query and return some results based on it. If you override this, you can implement whichever search you wish.

A simple first approach would be to move this line to a separate method in the same controller. This new method can read the query params (line 7) and perform the search (line 8), returning the results. Once a separate method, it should be more readily available for users to override.

Would anyone be up for providing a PR that does just that?

(Meanwhile, I'm afraid I'm going to close this PR for the reasons I explain above).

pldavid2:
Hello @pablobm , I'm rather new on administrate and was looking to have some kind of elasticsearch/searchkick integration on it. I understand your solution could help with that?
My first try was looking on the administrate plugin list but unfortunately it seems there is none for elasticsearch/searchkick, only for ransack.
Thanks

pablobm:
I don't know anything about searchkick, but I'm going to guess that yes, it should work.

To override Administrate as I explain above, you can edit `Admin::ApplicationController` and provide your own version of `index` (or whatever method you want to change). You probably want to copy the original method and then tweak it to suit your needs.

However bear in mind that you'll have to double check that things still work whenever you upgrade Administrate. We change the implementation of controllers, templates, etc every now and then, and this may cause an incompatibility with your overrides, which will need to be updated.

PR #1167, opened 2018-05-25
Demonstrate / test controller #scoped_resource override

odlp:
Relating to #320, this test demos how to provide a custom scope for the resources displayed in an Administrate controller's `#index` action.

PR #1219, opened 2018-09-20
Test/demo scope overrides and update comments

odlp:
Relating to #320, this test demos how to provide a custom scope for the resources displayed in an Administrate controller's `#index` action.

It also updates the comments in the controller template which are out of date; referencing instance variables (such as `@resources`) which aren't in use any more (pointed out by @vasco [here](https://github.com/thoughtbot/administrate/issues/320#issuecomment-421645553)).

This PR replaces #1167


nickcharlton:
Thanks Oli!

Issue #533, opened 2016-04-04
Enum fields


hauleth:
It would be really useful to have some way to entry values for `has_entry`.

stevenpetryk:
This took about 3 minutes. Didn't even have to change the generated class.

```
rails g administrate:field enum
```

_form.html.erb
```erb
<div class="field-unit__label">
  <%= f.label field.attribute %>
</div>
<div class="field-unit__field">
  <%= f.select field.attribute, Wireframe.orientations.keys.map { |v| [v.titleize, v] } %>
</div>
```

(_show/_index).html.erb
```erb
<%= field.to_s.titleize %>
```

carlosramireziii:
@hauleth Does the custom field that @stevenpetryk illustrated above work for your needs? If so, would you or he like to take a stab at creating it as a plugin? 

Here's an example: 
https://github.com/thoughtbot/administrate-field-image

brndn4:
Worked for me. You might want to change EnumField#to_s to
```
def to_s
  data.to_s
end
```
In case the enum data is nil

carsonbland:
@stevenpetryk what is the benefit over Field::Select.with_options(collection: Model::ENUM) ?

stevenpetryk:
@carsonbland I'm not sure – it may very well be the case that that was undocumented/impossible back when I posted that response. If that works, then go for it!

BenMorganIO:
Closing this, since a custom enum field with `bin/rails g administrate:field enum` solves the issue.

fercreek:
@stevenpetryk great solution! Thanks!!

joallard:
@BenMorganIO Sounds like this should be part of core

PR #1220, opened 2018-09-25
Make Field::Select more flexible

jumjamjohn:
- This PR adds more flexibility to `Field::Select` so we can set an array of arrays and a hash to `collection` option. This behavior would be more intuitive since it follows Rails form helper convention. 
- This also solves the absence of `Enum` field referred to #322 and #533 so we don't have to create a Enum field as a custom or built-in field anymore.
- Added select_spec.rb so we can test Select field

By default, it displays label data(e.g. "Pending") instead of raw persisted data(e.g. 1) on show/index page. Search feature would be an issue because users would want to search by label data but the system searches only by persisted data for the moment. However, that issue should be solved by another PR, not this one. Thank you for this awesome gem 😃 

Basic usage:
```ruby
status: Field::Select.with_options(collection: [ ['Pending',1], ['Approved',2], ['Rejected',3] ])
```
=>
```html
<select name="post[status]" id="post_status">
  <option selected="selected" value="1">Pending</option>
  <option value="2">Approved</option>
  <option value="3">Rejected</option>
</select>
```

For Enum:
```ruby
status: Field::Select.with_options(collection: Post.localized_statuses.invert)
```

```ruby
class Post < ApplicationRecord
  enum status: { submitted: 0, approved: 1, rejected: 2 }
  def self.localized_statuses
    I18n.t 'activerecord.attributes.post/status'
  end
  # or
  # def self.localized_statuses
  #   Post.statuses.keys.map {|k| [k, Post.human_attribute_name("post.status.#{k}")] }
  # end
end
```

```yaml
activerecord:
  attributes:
    post/status:
      submitted: "承認待ち"
      approved: "承認済み"
      rejected: "取り消し"
```


bjgaynor:
Any word on this?  It would be really useful to have.  I feel like Administrate has so much potential and then stuff like this doesn't get merged.  Can I do something to fix it and get it merged?  I'd be happy to update the syntax if that's the issue here.

pablobm:
Closed due to lack of activity. Additionally, I am not sure about overloading `collection`. In order to make Administrate work with minimum surprises, I think `collection` should behave exactly like `options_for_select`. I have done this at https://github.com/thoughtbot/administrate/pull/2348, which also implements enums.

However this means that `collection` would not be able to provide groups. I think this is fine and instead there should be a separate option for grouped options. Eg: `grouped_options`, which is the name it receives in the documentation of [`grouped_options_for_select`](https://api.rubyonrails.org/classes/ActionView/Helpers/FormOptionsHelper.html#method-i-grouped_options_for_select).

PR #1222, opened 2018-10-02
Include_blank in Belongs to form

brynmrk:
Replaces the functionality wherein nil is already included in the possible options of belongs_to. In the current functionality of administrate, if you add `include_blank: true` in f.select it would have double blank option in the selections.

Changes in spec/factories.rb where made by running:
```
rubocop \
  --require rubocop-rspec \
  --only FactoryBot/AttributeDefinedStatically \
  --auto-correct
```


nickcharlton:
Thanks for opening this!

PR #1226, opened 2018-10-15
Fix Time fields bug which occurs when Time value is nil

rnitta:
When there are null-allowed Time column on DB,
https://github.com/thoughtbot/administrate/blob/8d86f9f0833aed9394ef1bd1f0edef1de4e89b83/app/views/fields/time/_index.html.erb#L17

This code can raise error, because`undefined method 'strftime' for nil:NilClass` if `field.data` is nil

so I add code to check if time value is nill and some specs

please check it.

and,
CircleCi tests failed due to dependencies vulnerability.
please check this, too. https://github.com/thoughtbot/administrate/pull/1225

### DIFF
--- a/app/views/fields/time/_index.html.erb
+++ b/app/views/fields/time/_index.html.erb
@@ -14,4 +14,6 @@ By default, the attribute is rendered as a text tag.
 
 %>
 
-<%= field.data.strftime("%I:%M%p").to_s %>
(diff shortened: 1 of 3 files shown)


nickcharlton:
Thanks for your work here! I rebased and now I'll merge it.

PR #1230, opened 2018-10-24
Translate resource name for create link on index page.

Samemura:
I am using administrate in ja local and realize resource name is not translated.
Can we replace it with translated ?

Place I meant here. (It is en local but)
image

### DIFF
--- a/app/views/administrate/application/index.html.erb
+++ b/app/views/administrate/application/index.html.erb
@@ -44,7 +44,7 @@ It renders the `_table` partial to display details about the resources.
     <%= link_to(
       t(
         "administrate.actions.new_resource",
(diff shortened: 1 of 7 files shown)


Samemura:
@nickcharlton Hello! Can you review this PR, please ? If OK, I will fix Circle CI error(nokogiri version error).

nickcharlton:
I just rebased this to catch up with it; where does `class_from_resource` come from?

pablobm:
I believe this is the same as https://github.com/thoughtbot/administrate/pull/1777

nickcharlton:
Oh yes, you're right. I'm going to close this.

PR #1231, opened 2018-10-25
Associations button

KjellMorgenstern:
This pull request adds a button for creating new items underneath each "has_many" field.
Example:
class order has many line_items.
In the GET admin/orders/1 view, the line_items are listed. Now, after that list, a "New line_item" button is added. When clicked, it leads to GET admin/orders/1/line_items#new , and the line_item form has order#1 preselected in the order_id field.

I use this by default for each model. Maybe others find it useful, so here is a generic version of it.
I think the layout could use some polishing (using 'div' is probably not in line with adminstrate), but first I'd like to collect some feedback.


purinkle:
@KjellMorgenstern, is this still a change you want to make?

KjellMorgenstern:
I still think its an improvement, but I didn't follow the development of administrate (or use it) since then.

nickcharlton:
It's been a long time since this was last looked at and so I'm going to close it for now.

PR #1259, opened 2018-12-11
Use Show Page Attributes When Rendering a has_one Relationship On Another Record's Show Page

conradbeach:
When rendering a `has_one` relationship on the show page for a record, we're presently displaying the form attributes of the child record (as defined in the `FORM_ATTRIBUTES` constant from the child record's dashboard file).

I think it makes more sense to use the child record's show page attributes (as defined in the `SHOW_PAGE_ATTRIBUTES` constant from the child record's dashboard file). That just fits better when rendering a record on a show page.

Furthermore, using the form attributes can lead to displaying odd attributes such as "Password Confirmation".

This pull request makes the necessary changes to use the show page attributes of the child record instead of the form attributes.

It also makes use of the `render_field` helper to render each field a little more nicely than the raw data rendering we're doing right now.

**Current**
snip20181207_5_edited

**Changed by This Pull Request**
snip20181207_8

### DIFF
--- a/app/views/fields/has_one/_show.html.erb
+++ b/app/views/fields/has_one/_show.html.erb
@@ -4,7 +4,7 @@
 This partial renders a has_one relationship,
 to be displayed on a resource's show page.
 
-All fields of has_one relationship would be rendered
+All show page attributes of has_one relationship would be rendered
 
 ## Local variables:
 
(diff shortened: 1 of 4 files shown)


nickcharlton:
Thanks for your contribution and I'm sorry that it took so long before we got around to merging it!

conradbeach:
Thanks!

PR #1264, opened 2018-12-20
Fix association search when passing class_name option

sadhu89:
To get the right table name for complex cases like when using postgres schemas.


andrepcg:
How's this PR going?

lib/administrate/search.rb
```diff
@@ -44,7 +44,13 @@ def attribute_types
 
     def query_table_name(attr)
       if association_search?(attr)
-        ActiveRecord::Base.connection.quote_table_name(attr.to_s.pluralize)
+        table_name = if attribute_types[attr].options[:class_name].present?
+                       attribute_types[attr].options[:class_name].constantize.table_name
```
pablobm: Since the line `attribute_types[attr].options[:class_name]` is used twice, I think it's worth moving it into a variable.

sedubois:
@pablobm are the questions that you asked blocking? As there is no reply of @sadhu89 maybe the PR can be merged as-is and we can improve whatever is required in another PR? Currently the search is broken and this appears to improve things a lot already.

pablobm:
The issues are, unfortunately, blocking. It's extraneous code that really doesn't belong here. However, I have created an alternative PR: https://github.com/thoughtbot/administrate/pull/1569.

Closing this PR in favour of the new one.

PR #1284, opened 2019-01-28
Update _form.html.erb

mparramont:
😄


purinkle:
@mparramont, thank you for your first contribution!

PR #1305, opened 2019-03-27
 Change response by request format

d-mato:
I'm using with ajax but application_controller responds for html.
So I added json views and change response format by request format.


nickcharlton:
What's reason for adding a JSON version for this? What use case is this enabling?

d-mato:
@nickcharlton Thanks feedback.
My Rails app submits form through ajax with `remote: true` or `local: false` option. In that case I want not html but json.

nickcharlton:
Unfortunately this ended up sitting for quite a while. I don't think this is a good fit for Administrate now and so I'm going to close it. We might reconsider in future.

PR #1319, opened 2019-04-17
Update README to show support for Ruby 2.4 and above

reubenbrown:
Based on changes made in #1318 the README now reflects the correct Ruby versions this gem supports


nickcharlton:
Great! Thanks for updating this.

PR #1331, opened 2019-04-23
Drop active_job from the dependencies.

nickcharlton:
In 8bc80cfecc4a438a09e877397b0913184ba4aba7, we needed to add `active_job` to the dependencies to allow database seeding to continue to work.

This was raised in https://github.com/rails/rails/issues/35782 and fixed in https://github.com/rails/rails/commit/d3dc651a819d95f0467b0bdad6bc85080debcc43


thecalliopecrashed:
@nickcharlton, I'm poking around here.

This requires us to depend on Rails `6.0.0.rc1` or greater, yeah?

thecalliopecrashed:
I'm not sure how to read the Rails' team's notation on https://github.com/rails/rails/commit/d3dc651a819d95f0467b0bdad6bc85080debcc43:
Screen Shot 2019-05-20 at 10 37 29 AM

nickcharlton:
👋 @thecalliopecrashed 

I think you're right. That tag symbol should mean that the commit will be included in that release.

I do wonder if this has additional implications when Rails 6 is released. Will we need to conditionally add `active_job` for Rails versions lower?

PR #1334, opened 2019-04-26
Fix warning messages

jonatasrancan:
- Fix warning message related to the seed file
```
NOTE: Faker::LordOfTheRings.character is deprecated; use Faker::Movies::LordOfTheRings.character instead. It will be removed on or after 2019-01-01.
Faker::LordOfTheRings.character called from .../administrate/spec/example_app/db/seeds.rb:51
```
 

- Fix warning message relate to style

```
autoprefixer: administrate/app/assets/stylesheets/administrate/application.scss:1934:3: Replace text-decoration-skip: ink to text-decoration-skip-ink: auto, because spec had been changed
```

### DIFF
--- a/app/assets/stylesheets/administrate/base/_typography.scss
+++ b/app/assets/stylesheets/administrate/base/_typography.scss
@@ -23,7 +23,7 @@ p {
 
 a {
   color: $action-color;
-  text-decoration-skip: ink;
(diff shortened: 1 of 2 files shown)


composerinteralia:
I just ran into these warnings today. Thanks for fixing.

Issue #1332, opened 2019-04-25
Authorization with Pundit doesn't allow Policy Namespacing


bhtabor:
[Punditize](https://github.com/thoughtbot/administrate/blob/b85abcf965ecc843ee9b25e3a2ea65dfeb1b0b3c/app/controllers/concerns/administrate/punditize.rb) assumes that there is no Pundit [Policy Namespacing](https://github.com/varvet/pundit#policy-namespacing) being used. If we allow namespaces, we could even avoid the custom policy_scope_admin.

Quick work around is, instead of including Punditize, to use the following code:

```
# All Administrate controllers inherit from this `Admin::ApplicationController`,
# making it the ideal place to put authentication logic or other
# before_actions.
#
# If you want to add pagination or other controller-level concerns,
# you're free to overwrite the RESTful controller actions.
module Admin
  class ApplicationController < Administrate::ApplicationController
    include Pundit

    def policy_scope(scope)
      super([:admin, scope])
    end

    def authorize(record, query = nil)
      super([:admin, record], query)
    end

    def scoped_resource
      policy_scope super
    end

    def authorize_resource(resource)
      authorize resource
    end

    def show_action?(action, resource)
      Pundit.policy!(pundit_user, [:admin, resource]).send("#{action}?".to_sym)
    end

  end
end

```

robinboening:
I had the same thoughts for namespacing after stumbling into an error because I like using Pundit's `verify_policy_scoped` to verify the use of `policy_scope`.

My approach for changing administrate's punditize is a bit different, though. I believe deprecating the `policy_scope_admin` and removing it on the next major version is the right way to go, not to build on top of it. I believe software should needs to get rid of legacy code if we agree on a better approach.

bhtabor:
I agree @robinboening We should just depricate `resolve_admin`. The whole idea behind `policy_scope_admin` is stated in the comment below:
```
# Like the policy_scope method in stock Pundit, but allows the 'resolve'
# to be overridden by 'resolve_admin' for a different index scope in Admin
# controllers.
```
This is the problem. It is trying to do what Pundit already does with [Policy Namespacing](https://github.com/varvet/pundit#policy-namespacing). Admin scope is just like a Pundit Namespace.

We still need to allow Pundit Namespacing though, like I did in this pull request - https://github.com/thoughtbot/administrate/pull/1335. I kept the `resolve_admin` feature for backward compatibility only. But, it could also be used to even have an additional admin scope over a namespace.

[...]

PR #1335, opened 2019-04-27
Allow Authorization with Pundit Policy Namespacing

bhtabor:
Fix https://github.com/thoughtbot/administrate/issues/1332

To use specific [Policy Namespacing](https://github.com/varvet/pundit#policy-namespacing) for administrate, like admin namespace policies for example, just define the pundit_namespaces in an array format, like below. It is also backward compatible with resolve_admin.

```
module Admin
    include Administrate::Punditize
    def pundit_namespaces
      [:admin]
    end
  end
end
```

### DIFF
--- a/app/controllers/concerns/administrate/punditize.rb
+++ b/app/controllers/concerns/administrate/punditize.rb
@@ -5,16 +5,22 @@ if Object.const_defined?("Pundit")
       include Pundit
 
       included do
+        def pundit_namespaces
+          []
+        end
+
         def scoped_resource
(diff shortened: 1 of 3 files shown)


pablobm:
Closing due to lack of activity. However I generally agree with this, particularly deprecating `policy_scope_admin` in favour of policy namespacing. I think we will revisit this soon when addressing #1332.

Having said that, if someone feels like rebasing this and cleaning it up, I'll all for it!

PR #1336, opened 2019-04-28
Test against Rails 6.0.0rc1

composerinteralia:
To get ready for Rails 6.0.0 we can start testing against 6.0.0.rc1

Rails 6.0.0 has some breaking changes that don't allow it to work with
older versions of rspec-rails, so we will need to use rspec-rails 4.0.
rspec-rails 4.0 is not compatible with older versions of rails.

I also removed the pessimistic version constraints for the Rails
libraries in the gemspec. These constraints make it harder for people to
upgrade, and it is something Rails itself is moving away from.

The tests are passing on Ruby 2.6. ~~On the Ruby 2.5 build bundler doesn't seem to be able to resolve the dependencies. I will try `gem "railties", "~> 6.0.0.rc1"` in the Appraisal file instead of loading each library separately to see if that gets what we need.~~ Edit: the build is now passing for Ruby 2.5 as well.

Rails 6 is not compatible with Ruby 2.4 ~so we will need to either stop running against Ruby 2.4 altogether, or find something similar to Travis's `exclude` (see the [factory_bot_rails travis setup](https://github.com/thoughtbot/factory_bot_rails/blob/master/.travis.yml)) so we can skip running  Rails 6 with Ruby 2.4. I am still not all that comfortable with circleci config, so this might take me some time.~ Updating to circleci 2.1 and using `commands` seems to have done the trick.


administrate.gemspec
```diff
@@ -27,6 +27,8 @@ Gem::Specification.new do |s|
   s.add_dependency "sassc-rails", "~> 2.1"
   s.add_dependency "selectize-rails", "~> 0.6"
 
+  s.add_development_dependency "rspec-rails"
```
composerinteralia: Having this in the Gemfile wasn't playing well with the specific versions needed in the Appraisals. I was seeing an error about specifying the same gem twice.

.circleci/config.yml
```diff
@@ -1,42 +1,48 @@
...
+  working_directory: ~/administrate
+  steps:
+    - shared_steps
+    # Run the tests against multiple versions of Rails
+    - run: bundle exec appraisal install
     - run: bundle exec appraisal rake
 
 jobs:
-  ruby-2.4:
+  ruby-24:
```
nickcharlton: Does Circle CI 2.1 require the removal of the dot here?

.circleci/config.yml
```diff
@@ -49,8 +55,19 @@ jobs:
           POSTGRES_USER: administrate
           POSTGRES_DB: ruby24
           POSTGRES_PASSWORD: ""
+    steps:
+      - shared_steps
+      # Run the tests against multiple versions of Rails
+      - run: BUNDLE_GEMFILE=gemfiles/rails42.gemfile bundle install
+      - run: BUNDLE_GEMFILE=gemfiles/rails50.gemfile bundle install
+      - run: BUNDLE_GEMFILE=gemfiles/rails51.gemfile bundle install
+      - run: BUNDLE_GEMFILE=gemfiles/rails52.gemfile bundle install
```
nickcharlton: Are we not able to use `bundle exec appraisal install` here, like we are with the others?

.circleci/config.yml
composerinteralia: Yeah it does. I had mentioned that in the commit, but forgot to add it to the PR description.

.circleci/config.yml
composerinteralia: ‘bundle exec appraisal install’ fails when trying to install Rails 6 on Ruby 2.4

gemfiles/rails51.gemfile
```diff
@@ -13,6 +13,7 @@ gem "unicorn"
 gem "actionpack", "~> 5.1.0"
 gem "actionview", "~> 5.1.0"
 gem "activerecord", "~> 5.1.0"
+gem "rspec-rails", group: [:development, :test]
```
danbee: Why did this move here? Could it not have stayed in the `:development, :test` group below?

gemfiles/rails52.gemfile
```diff
@@ -13,6 +13,7 @@ gem "unicorn"
 gem "actionpack", "~> 5.2.0"
 gem "actionview", "~> 5.2.0"
 gem "activerecord", "~> 5.2.0"
+gem "rspec-rails", group: [:development, :test]
```
danbee: Same comment as above.

gemfiles/rails51.gemfile
composerinteralia: These gemfiles are generated by Appraisal. I needed to set up the appraisals to use different versions of rspec-rails, and the changes in this PR were the only way I could think to do it. I’ll double check whether there is another way to do this that leaves these gemfiles as they were.

danbee: Ah fair enough. I don't think it's a problem. I was mostly curious.

.circleci/config.yml
nickcharlton: Exciting.

.circleci/config.yml
nickcharlton: Hmm. I guess this is a bug in `appraisal`.

composerinteralia: Maybe we could add something like `appraisal install rails42 rails50 rails51 rails52` to install a subset of the appraisals. We have something like that for `appraisal update`.

gemfiles/rails51.gemfile
composerinteralia: Using 

```
  group :development, :test do
    gem "rspec-rails"
  end
```

in the Appraisal file seems to have done the correct thing.

PR #1350, opened 2019-05-20
Remove NameError rescue in page base

### LINKED ISSUE #480: NoMethodError is silently ignored

thecalliopecrashed:
Closes #920;
Fixes #480.

This pull request overrides #920 in an effort to keep the ball moving.

### DIFF
--- a/lib/administrate/page/base.rb
+++ b/lib/administrate/page/base.rb
@@ -33,8 +33,6 @@ module Administrate
 
       def get_attribute_value(resource, attribute_name)
         resource.public_send(attribute_name)
-      rescue NameError
-        nil
       end
 
       attr_reader :dashboard, :options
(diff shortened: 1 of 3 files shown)


GeoffAbtion:
* How are the timestamps related to the NameError rescue?

* Should there be tests?

pablobm:
Duplicate of https://github.com/thoughtbot/administrate/pull/920. That PR explains the (very relevant) questions about timestamps:

> The rescue statement generates a false positive on payments feature specs, the timestamps migrations are created to fix this.

I think that also sorts the need for tests, fixing that false positive that is raised in the existing ones.

PR #1357, opened 2019-06-09
Add word-break to attribute-data

mikegabriel:
Similar to #306 - but improves functionality.

Long values such as MD5 hashes are currently being outputted on a single line that expands the page for either `Field::String` or `Field::Text`.  This commit adds a `word-break` to `.attribute-data` which allows word-wrapping to occur while maintaining copy and paste-ability.

Before (page scrolls a lot and has 'h-align' padding to centre it:
Screen Shot 2019-06-09 at 10 52 40 AM

After:
Screen Shot 2019-06-09 at 10 52 54 AM


nickcharlton:
Thanks! This one must've slipped through, so I rebased and merged.

PR #1371, opened 2019-07-03
Remove the need for resource index path

0urobor0s:
This removes the need to have an index path if you have a resource that only has a show path requirement

### DIFF
--- a/app/views/administrate/application/_navigation.html.erb
+++ b/app/views/administrate/application/_navigation.html.erb
@@ -13,6 +13,8 @@ as defined by the routes in the `admin/` namespace
       display_resource_name(resource),
       [namespace, resource_index_route_key(resource)],
       class: "navigation__link navigation__link--#{nav_link_state(resource)}"
+    ) if Rails.application.routes.url_helpers.method_defined?(
+        "#{resource.namespace}_#{resource.path}_path".to_sym
     ) %>
   <% end %>
 </nav>


purinkle:
Would you be able to give a more concrete example of where you may need this?

Is it something that can be controlled by [customising the dashboard]?

[customising the dashboard]: https://administrate-prototype.herokuapp.com/customizing_dashboards

0urobor0s:
Let's say you have a resource A that has many resources B. When I enter the show page for A I want to see all Bs that belong to A and also see the show page for B without having the index page for B.
```
namespace :admin do
  resources :a
  resources :b, only: [:show]
end
```

robinboening:
👍 This is much simpler than my approach https://github.com/thoughtbot/administrate/pull/1418 and solves the same issue.

0urobor0s:
@purinkle ping

pablobm:
Hello @0urobor0s, thank you for your PR and apologies for the delay.

There are two helpers that I think can be used instead of `Rails.application.routes.url_helpers.method_defined?`.

One is `valid_action?`, which checks if the route exists (similar to your line), and then there's `show_action?`, which checks if the user is authorised to see it (if using Pundit as per the docs).

So I think the line could instead be `valid_action?(:index, resource) && show_action?(:index, resource)`.

Also, we'd need a spec for this. Perhaps one around payments only showing as a has_many of orders? (As per the current example app and feature specs).

pablobm:
Fixed at https://github.com/thoughtbot/administrate/pull/1524

Issue #271, opened 2015-11-24
Feature Request: Add Top Nav


BenMorganIO:
While I was cruising about, I noticed that I wanted to return to the app, but that link wasn't in the admin. Also, sometimes when you're running a job, you get a dashboard to look at jobs. There's no link anywhere to find them. You're also logged in (most of the time) and there's no conventional logout button.

What I'm asking:

Basically add a nav to the top right near the search to logout if `user_signed_in?` is `true` and `defined?`. And maybe a link to `main_app.root_url` as well.

An API to add new links (such as giving users the ability to link to a CMS admin and/or a jobs dashboard possible) would be great. I know Solidus/Spree has this.

graysonwright:
@BenMorganIO A top nav is something that will probably vary a lot between different implementations of Administrate. I'm not sure how to add a top nav that will be generally useful.

On `master`, we just added #269. This gives you the command `rails generate administrate:views:layout`. It'll go out in version 0.1.3, and until then you can get it by using:

```ruby
gem "administrate", github: "thoughtbot/administrate"
```

If you (and others) want to share your top navbar implementation here, that would be a good first step to see if we can add a generally-useful API for it.

Cheers!

zamith:
I added a top navbar on my app by generating the layout and adding it to that layout. Here's the code in Slim, but it should be easily translated to ERb.

```slim
doctype html
html lang='en'
head
  meta charset="utf-8"
  meta name="ROBOTS" content="NOODP"
  meta name="viewport" content="initial-scale=1"
  title
    = content_for(:title)
    | |
    = Rails.application.class.parent_name.titlecase
  = stylesheet_link_tag "//fonts.googleapis.com/css?family=Lato:300,400,900", media: "all"
  = stylesheet_link_tag "administrate/application", media: "all"
  = stylesheet_link_tag "admin", media: "all"
  = csrf_meta_tags

body.administrate
  = render "application/navbar"  # The top nav
  main.main
    .sidebar
      = render "sidebar"

    .container
      = content_for(:search)
      .content
        = render "flashes"
        = yield

  = render "javascript"
```

Hope it helps.

graysonwright:
Great, thanks! I'll leave this open to gauge interest - if anyone else wants to see this, let us know here.

minimul:
I have 2 projects using administrate and the first thing you/user/client is thinking after doing what they need to do in administrate is "How do I get back to the main app". So basically overriding the view is the first thing I am going to do  everytime.

[...]

PR #1376, opened 2019-07-12
Add a "back" link to the navigation

nickcharlton:
This solves half of #271 by implementing a link to take the user back to
the main application. It's assumed that `root_url` without a namespace
returns the main application's root.

### DIFF
--- a/app/views/administrate/application/_navigation.html.erb
+++ b/app/views/administrate/application/_navigation.html.erb
@@ -8,6 +8,8 @@ as defined by the routes in the `admin/` namespace
 %>
 
 <nav class="navigation" role="navigation">
+  <%= link_to "Back to app", root_url, class: "button button--alt" %> 
+
   <% Administrate::Namespace.new(namespace).resources.each do |resource| %>
     <%= link_to(
       display_resource_name(resource),
--- a/app/assets/stylesheets/administrate/components/_buttons.scss
+++ b/app/assets/stylesheets/administrate/components/_buttons.scss
@@ -41,3 +41,11 @@ input[type="submit"],
     }
   }
 }
+
+.button--alt {
+  background-color: transparent;
+  border: $base-border;
+  border-color: $blue;
+  color: $blue;
+  margin-bottom: $base-spacing;
+}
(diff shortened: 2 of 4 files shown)


nickcharlton:
@tysongach, as someone who's worked on the design here quite a bit, I'd love some input on how to improve this link.

Currently, it looks like: 

Screenshot 2019-07-12 at 15 32 09

But it feels to me that the location of the text here is wrong. What do you think?

LkeMitchll:
> @tysongach, as someone who's worked on the design here quite a bit, I'd love some input on how to improve this link.
> 
> Currently, it looks like:
> 
> Screenshot 2019-07-12 at 15 32 09
> 
> But it feels to me that the location of the text here is wrong. What do you think?

@nickcharlton If it were me I would more clearly separate this function away from the other items, as it is a contextually destructive (it tears you away from Administrate completely) action. Perhaps try a style closer to a button, vertically align it to the main heading, and then add whitespace below to clearly separate the items:

Screenshot 2020-01-03 at 11 32 34

In this image I've mocked an alternative button style that may not exist in the styling, but It's fairly important that this isn't too bold or distracting, in my opinion. I'd also like to see a 'back arrow' button inside it, but that's a nice-to-have,  for sure.

nickcharlton:
I've made those changes (trying to make it fit as well as I can into Administrate). What do you think?  https://administrate-demo-prer-pr-1376.herokuapp.com/admin

app/views/administrate/application/_navigation.html.erb
```diff
@@ -8,6 +8,8 @@ as defined by the routes in the `admin/` namespace
 %>
 
 <nav class="navigation" role="navigation">
+  <%= link_to "Back to main app", root_url, class: "button button--alt" %>
```
sedubois: How about just "Back to app"?

nickcharlton: Yeah, good idea. I prefer that.

sedubois:
NB: the README's screenshot does not include this change.

nickcharlton:
Oh thanks for pointing that out! I did that here: https://github.com/thoughtbot/administrate/pull/1523

PR #1386, opened 2019-07-31
Left join in order to support searching both a belongs_to and a field…

motdotla:
… on the root entity

First, I love Administrate. ♥ Thank you for creating such a straightforward yet reasonably customizable dashboard. It hits the perfect sweet spot.

This PR is to fix a search issue I ran up against.

In the scenario, where I have configured search to both a belongs_to `searchable_field` AND some `other_field` in the root entity, the following happens:

1. I can successfully search the belongs_to `belongs_to.searchable_field` when the root entity's belongs_to is assigned
2. I can successfully search the belongs_to `belongs_to.searchable_field` when the root entity's belongs_to is null
3. I can successfully search the `other_field` when the root entity's belongs_to is assigned
4. I can NOT successfully search the `other_field` when the root entity's belongs_to is null

This PR fixes problem #4 (caused by an inner join) by using a left join.

Is this PR useful?

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -13,7 +13,7 @@ module Administrate
       if @term.blank?
         @scoped_resource.all
       else
-        @scoped_resource.joins(tables_to_join).where(query, *search_terms)
+        @scoped_resource.left_joins(tables_to_join).where(query, *search_terms)
       end
     end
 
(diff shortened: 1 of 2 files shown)


motdotla:
Darn, specs are failing because left_joins was not added till 5.x Rails. Suggestions? 

I could rewrite to use joins with a query string but that will get a bit messy.

Or I could use left_joins for Rails 5.x and greater and fall back to standard inner `.join` for < 5.

Thoughts?

nickcharlton:
Sorry, we didn't see this before commenting on #1581. I'm going to close this one, mostly as there's more conversation there and no conflict.

Thank you for opening this PR!

Issue #1339, opened 2019-05-06
Autoloading punditize fails with Zeitwerk + Rails 6 (rc1) if not using punditize


SleeplessByte:
If you look at the explanation of https://github.com/fxn/zeitwerk/issues/45, it is clear why this happens:

- given `pundit` is not defined,
  - it tries to load the file `/administrate-0.11.0/app/controllers/concerns/administrate/punditize.rb`
  - it (correctly) does NOT define `Administrate::Punditize`
  - it errors with `eager_load`: `expected <file> to define constant Administrate::Punditize`

A solution would be:
- don't auto require/load this
- always define `Administrate::Punditize`, but keep it empty if `Pundit` doesn't exist

skyksandr:
as a workaround - add `gem 'pundit'` to `Gemfile` and it will work.

purinkle:
@SleeplessByte, thanks for raising this issue. Would you be able to provide specific steps to reproduce the issue so it is easier isolate the bug and fix it?

SleeplessByte:
In your `Gemfile` (new rails project) add:

```ruby
gem 'rails', '6.0.0.rc2'
gem 'administrate', '~>0.11'
```

Now add any administrate controller, and run `rails` in production mode (so that it definitely autoloads).

Given the tagged project in this issue (not mine), `git checkout https://github.com/captproton/workcation_green/commit/6f1c3b12646dfea1707ca8045afdf60dbae8e8ff` and running should also work as they added the workaround in the commit after this.

ksoderstrom:
Another workaround for this is to not use zeitwerk, but instead the old autoloader. As per the [blog post about zeitwerk integration](https://weblog.rubyonrails.org/2019/2/22/zeitwerk-integration-in-rails-6-beta-2/) add:

```
config.autoloader = :classic
```

to `config/application.rb`.

PR #1394, opened 2019-08-02
Fix loading of `Punditize`

purinkle:
We were only defining the `Punditize` module if we had already defined `Pundit`. Setting `Punditize` like this broke autoloading in Rails when `Pundit` was not specified. We fixed the problem by always setting `Punditize` but keeping it empty if` Pundit` does not exist.

https://github.com/thoughtbot/administrate/issues/1339

### DIFF
--- a/app/controllers/concerns/administrate/punditize.rb
+++ b/app/controllers/concerns/administrate/punditize.rb
@@ -1,6 +1,6 @@
-if Object.const_defined?("Pundit")
-  module Administrate
-    module Punditize
+module Administrate
+  module Punditize
+    if Object.const_defined?("Pundit")
       extend ActiveSupport::Concern
       include Pundit
 
--- a/spec/controllers/concerns/administrate/punditize_spec.rb
+++ b/spec/controllers/concerns/administrate/punditize_spec.rb
@@ -1,0 +1,41 @@
+require "rails_helper"
+
+module Administrate
+  RSpec.describe Punditize do
+    it "exists when 'Pundit' is not defined" do
+      unload_constants do
+        expect { build_dummy }.not_to raise_error
+      end
+    end
+
+    def unload_constants
+      original = Pundit
+      Object.send(:remove_const, "Pundit")
+      Administrate.send(:remove_const, "Punditize")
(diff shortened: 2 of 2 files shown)


spec/controllers/concerns/administrate/punditize_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+module Administrate
+  RSpec.describe Punditize do
+    it "exists when 'Pundit' is not defined" do
+      unload_constants do
+        expect { build_dummy }.not_to raise_error
```
benfurber: Do we want to be as general as testing for all errors? Could this be more specific to the error type being experienced?

i.e. Does this catch legitimate Punditize errors?

app/controllers/concerns/administrate/punditize.rb
```diff
@@ -1,6 +1,6 @@
-if Object.const_defined?("Pundit")
-  module Administrate
-    module Punditize
+module Administrate
+  module Punditize
+    if Object.const_defined?("Pundit")
```
benfurber: Would it be worth warning that `Pundit` is not defined so we don't just silently swallow this issue?

spec/controllers/concerns/administrate/punditize_spec.rb
purinkle: This is an interesting question and it is a line of thinking I went down myself.

My original test was something similar to the following.

```ruby
expect { build_dummy }.not_to raise_error LoadError
```

Rubocop was not happy with the above solution because it is too generic.

It is reasonable to expect initialising a class should not raise an error of any kind.

app/controllers/concerns/administrate/punditize.rb
purinkle: As `Pundit` is not a dependency of the library, but is instead part of a configuration of the engine, I would feel uncomfortable raising a warning at this point. The implementor has made a specific choice in their application to not include `Pundit`. We shouldn't feel the need to remind them that they chose not to use it.

PR #1398, opened 2019-08-13
Support association search for other types of association fields

josephktcheung:
*This issue and the proposed fix is very similar to https://github.com/thoughtbot/administrate/pull/1176*

Right now association search doesn't work for new types of association fields.

`Administrate::Search` hard-codes the possible types of association fields in a list to determine if association search should be used. This means that new types of fields cannot perform association search and things break.

My proposed solution (see `lib/administrate/search.rb`) does a programmatic search for field classes that inherit from `Administrate::Field::Associative`. This allows plugin authors not to worry about adding their new type to a list or any other sort of setup.

A potential problem with this approach is that it includes `Administrate::Field::Polymorphic` on the list. I don't know if this can cause problems down the line or not.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -69,11 +69,13 @@ module Administrate
     def association_search?(attribute)
       return unless attribute_types[attribute].respond_to?(:deferred_class)
 
-      [
-        Administrate::Field::BelongsTo,
-        Administrate::Field::HasMany,
-        Administrate::Field::HasOne,
-      ].include?(attribute_types[attribute].deferred_class)
+      association_classes.include?(attribute_types[attribute].deferred_class)
+    end
+    
+    def association_classes
+      @association_classes ||=
+        ObjectSpace.each_object(Class).
+          select { |klass| klass < Administrate::Field::Associative }
     end
 
     attr_reader :resolver, :term
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -71,7 +71,7 @@ module Administrate
 
       association_classes.include?(attribute_types[attribute].deferred_class)
     end
-    
+
     def association_classes
       @association_classes ||=
         ObjectSpace.each_object(Class).


jshum:
^ any updates on this? I'm seeing this issue as well on my setup

lib/administrate/search.rb
```diff
@@ -69,11 +69,13 @@ def tables_to_join
...
-        Administrate::Field::HasMany,
-        Administrate::Field::HasOne,
-      ].include?(attribute_types[attribute].deferred_class)
+      association_classes.include?(attribute_types[attribute].deferred_class)
+    end
+
+    def association_classes
+      @association_classes ||=
+        ObjectSpace.each_object(Class).
+          select { |klass| klass < Administrate::Field::Associative }
```
seanpdoyle: As an alternative to mapping over `ObjectSpace` (and maybe even skipping the memoization), have you considered using [`Class#descendants`](https://api.rubyonrails.org/classes/Class.html#method-i-descendants)?

```ruby
class C; end
C.descendants # => []

class B < C; end
C.descendants # => [B]

class A < B; end
C.descendants # => [B, A]

class D < C; end
C.descendants # => [B, A, D]
```

egreer: Could also use `ancestors` from the `deferred_class` to see if it includes `Administrate::Field::Associative`. This would allow for any plugins that extend the `BelongTo`, `HasMany` etc to also be included.

This could be:
```
  def association_search?(attribute)
      return unless attribute_types[attribute].respond_to?(:deferred_class)
      attribute_types[attribute].deferred_class.ancestors.include?(Administrate::Field::Associative)
  end
```

pablobm: In defence of this solution, I'd say it just does the same thing that already happens at [`lib/administrate/base_dashboard.rb`](https://github.com/thoughtbot/administrate/blob/acada0585b35aee532116fe04d65dc5263bba179/lib/administrate/base_dashboard.rb#L76) (and I myself introduced at https://github.com/thoughtbot/administrate/pull/1176/files).

I'd be happy with merging this, then having another PR to refactor both instances.

Issue #1369, opened 2019-06-28
No route changes or models views on administrate:install with Rails 6.rc1


zdrummond:
**Steps**
With Rails 6.rc1 installed, create a new app

`rails new administrates --skip-coffee --database=postgresql`

and new model

`rails g scaffold user name email`

then after bundling the administrate gem, run install

```
❯ rails  generate administrate:install
Running via Spring preloader in process 16421
      create  app/controllers/admin/application_controller.rb
```

**Expected**
Getting started guide says
> The installer adds some new routes to your config/routes.rb, and creates a controller at app/controllers/admin/application_controller.rb

>In addition, the generator creates a Dashboard and a Controller for each of your ActiveRecord resources:

So I would expect new routes for the admin pane plus a dashboard and controller for my model

**Actual**
One Admin app controller, but nothing else

grantsayer:
Likewise - found this repeatable and tested other scenarios of 

**Scenario-1**
1. Created vanilla Rails 6rc1 app with no new models
2. Added administrate to Gemfile
3. Executed 'rails generate administrate:install'.  Resulted in creation of /admin/application_controller.rb
4. Added sample model/view via 'rails generate scaffold Post name:string
5. Executed 'rails generate administrate:routes'. Resulted in no change - e.g, no admin routes namespace added.
6. Tried to re-execute the 'rails generate administrate:dashboard Post'  - no change

Looking into the Gem code - guess it's somewhere in generators/routes/routes_generator.rb

grantsayer:
Further investigation appears that the issue is with the method 

    def database_models
            ActiveRecord::Base.descendants.reject(&:abstract_class?)
    end

located in **lib/generators/administrate/routes/routes_generator.rb**

Basically, in Rails6.rc1 the Model classes are derived from ApplicationRecord, for e.g

    class Post < ApplicationRecord
    end

The issue is that the `ActiveRecord::Base.descendants` returns `ApplicationRecord` and the `reject` method, passed `&:abstract_class` simply rejects the abstract class `ApplicationRecord`.  The result then is an empty list of database models and no change to the `config/routes.rb` file

nickcharlton:
Closing as this is fixed with #1406.

PR #1406, opened 2019-08-23
Use Zeitwerk for loading models in Rails 6

nickcharlton:
In #1369, it's noted that under Rails 6, model loading is no longer
working. This is because Zeitwerk has replaced the original code
loading.

Here, we load the models using Zeitwerk, if it is available.


nickcharlton:
This does successfully restore auto-loading functionality (unfortunately we can only validate this manually), so I'm going to merge this.

We need to further address "models" in Rails 6.0 which aren't really models.

PR #1407, opened 2019-08-27
Install latest bundler in CI

okuramasafumi:
This fixes failing CI broken by `BUNDLED_WITH` mismatch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ commands:
           key: administrate-{{ checksum "Gemfile.lock" }}
 
       # Install latest bundler
-      - run: gem install bundler -v '2.0.2'
+      - run: gem install bundler -v '2.1.0.pre.1'
 
       # Bundle install dependencies
       - run: bundle install --path vendor/bundle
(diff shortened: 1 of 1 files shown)


okuramasafumi:
https://github.com/thoughtbot/appraisal/issues/152
The issue linked above prevents us from going with Bundler 2. Should we go back to Bundler 1 until  this bug is fixed?

nickcharlton:
Oh! That's why. Thanks for looking into this.

I _think_ we've ended up using Bundler 2 because of @dependabot-bot rather than deliberately, but I'm not sure.

I'm not quite sure what the best course of action here is. If it is @dependabot-bot, we'll keep hitting the same issue. If it's a current issue with bundler, we're a bit stuck. What do you think?

okuramasafumi:
https://rubygems.org/gems/bundler/versions/2.1.0.pre.1
Will this new release solve the issue?

nickcharlton:
I think it might, it's solving the things we care about. Do you want to give it a go on this PR and see what happens?

okuramasafumi:
I tried Bundler `2.1.0.pre.1` but it didn't work. I think `Appraisal` gem needs to be upgraded.

nickcharlton:
Thanks! I'll dig into Appraisal tomorrow and see what I can come up with.

nickcharlton:
I'm taking the easy option (mostly so I can get an administrate release out the door for Rails 6.0) and reverting back to Bundler 1.17.3 (see: #1420). We should revisit this again once Appraisal is working correctly with Bundler 2.0.0 and up.

okuramasafumi:
I see. Hope to revisit it soon!

PR #1439, opened 2019-10-10
Fix collection filters example in dashboard template

swamp09:
If implemented according to the current template, it will be `NoMethodError`.
Fix the template as in the following link.
https://github.com/thoughtbot/administrate/blob/master/spec/example_app/app/dashboards/customer_dashboard.rb#L36
http://administrate-prototype.herokuapp.com/customizing_dashboards


nickcharlton:
Thanks for doing this!

PR #1458, opened 2019-11-04
Adds options for quick search filter

moofkit:
Extend quick search feature with `strict` strategy that gives possibility
to search entities for strictly equality to search pattern.
For backward compatibility default strategy is `fuzzy` that works as before this changes.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -4,9 +4,9 @@ module Administrate
 
     def index
       search_term = params[:search].to_s.strip
-      resources = Administrate::Search.new(scoped_resource,
+      resources = Administrate::Search.run(scoped_resource,
                                            dashboard_class,
-                                           search_term).run
+                                           search_term)
       resources = apply_collection_includes(resources)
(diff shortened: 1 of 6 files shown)


pablobm:
I'm glad that people are contributing PRs with aim to improve the search functionality. However I'm concerned that what may work for some won't be good for others, and we need to balance different options in order to better serve our users.

Over at https://github.com/thoughtbot/administrate/pull/913, on a similar discussion, I suggested that search perhaps should be handled by a separate module that could be pluggable. This way, users would have the ability to create their own search parsers and strategies. I'm not entirely sure of how feasible this would be though. For example, how would it interface with fields provided by third-party gems?

All in all, I'm resistant to make any changes to the search functionality at the moment, but I'm aware that it should be improved, or at least it should be easily improvable to match each user's custom needs.

moofkit:
@pablobm yeah it seems reasonable.
It will be perfect to make searching extendable for different strategies

moofkit:
Closing MR in favor of better approach https://github.com/thoughtbot/administrate/pull/1458#issuecomment-636786631

PR #1473, opened 2019-11-22
Migrate to use System Tests

seanpdoyle:
This is a WIP Pull Request to kick the tires on CI


nickcharlton:
@seanpdoyle, is this something you'd like to continue with? I realise it's been a while and there's a few conflicts but I'm not adverse to switching to use the new system tests.

seanpdoyle:
@nickcharlton wow, I've forgotten about this. I don't have enough context on the codebase to estimate the effort involved in resolving these conflicts.

If you don't see a direct path toward resolution, we could close this.

That being said, I _do_ think that adopting System Tests and Chromedriver would be valuable.

PR #1475, opened 2019-11-29
It should be possible to run every spec individually

pablobm:
I noticed that some specs can't be run individually, as their dependencies are not correct. This can become rather inconvenient while debugging Administrate. This PR fixes this issue.

A way to check that specs run individually is with the following one-liner:

```
$ find spec | grep _spec.rb | while read spec; do bundle exec rspec $spec; done
```

This includes some changes that are not really necessary, but I think make the specs make more sense. I'm referring mainly to things like `spec/generators/assets/javascripts_generator_spec.rb`, where I make sure that `support/generator_spec_helpers` is required before the tested code, because I feel that the helpers should be loaded first (like `spec_helper` or `rails_helper`) are.

### DIFF
--- a/spec/generators/assets/javascripts_generator_spec.rb
+++ b/spec/generators/assets/javascripts_generator_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
-require "generators/administrate/assets/javascripts_generator"
 require "support/generator_spec_helpers"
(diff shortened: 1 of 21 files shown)


nickcharlton:
Checks out! Thank you.

PR #1478, opened 2019-12-02
Don't convert nil to '' (empty string)

santib:
If I have a record wil `nil` values, when updating anything else in the record, it will change my `nil` values for `''`. This PR fixes that

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -145,7 +145,7 @@ module Administrate
           raise "Unrecognised param data: #{data.inspect}"
         end
       else
-        data
+        data.presence
       end
     end
 


pablobm:
Hello @santib, and thank you for contributing to Administrate!

I'm unsure about this change. It could be either way, but I feel that the current behaviour (converting nil strings to empty strings) would be the expected behaviour for users.

A bit of context first: when a text field is submitted to the backend and there's no value, there's no right or wrong in interpreting it as null/nil or as an empty string. That's the framework's choice, and at the same time it's legitimate for users to want it the other way around.

With Rails, I think an empty string is the expected value. For example, if I create an scaffold (I checked with Rails 6.0.1), I get the same behaviour you are seeing: empty values are interpreted as empty string rather than nils. Therefore, I think Administrate should behave the same way by default.

Fortunately, there's a way to override this default without having to change Administrate: you can define how to interpret submitted values by overriding `resource_params`. This is an example:

```ruby
# Put this in your Admin::ApplicationController,
# or on individual admin controllers
private

def resource_params
  params.require(resource_name)
    .permit(dashboard_class::FORM_ATTRIBUTES)
    .transform_values do |value|
      value == "" ? nil : value
    end
end
```

I'm realising that `resource_params` is not documented. I'll fix that.

Does this make sense? Did I misunderstand you? Let me know.

santib:
@pablobm Thanks Pablo for the explanation, that sounds good to me 👍

sedubois:
@pablobm on a similar note if one has a nullable boolean field, there is no way (using `Field::Boolean`) to allow keeping the value as `nil` when editing the record, it will be forced to `false`. 

AFAIK Administrate does not allow manipulating records that should be able to store both `nil` and empty field values.

pablobm:
@sedubois - Not sure if I understand correctly. Is this what you are trying to do?

| Persisted value | Input | Should persist as |
| ----------------|-------|------------------ |
|     nil         |  "0"  |        nil        |
|     nil         |  "1"  |        true       |
|     false       |  "0"  |        false      |
|     false       |  "1"  |        true       |
|     true        |  "0"  |        false      |
|     true        |  "1"  |        true       |

If that's what you want, you may be able to achieve it by overwriting `resource_params` too. Something like this (quick implementation, no warranty!):

```ruby
private

def resource_params
  params.require(resource_class.model_name.param_key).
    permit(dashboard.permitted_attributes).
    to_h.
    map { |k, v| [k, cast_input_value(k, v)] }.
    to_h
end

def cast_input_value(attr, value)
  if dashboard_class::ATTRIBUTE_TYPES[attr.to_sym] == Administrate::Field::Boolean
    current_value = requested_resource[attr]
    if current_value.nil? && value == "0"
      nil
    else
      value
    end
  else
    value
  end
end
```

If the truth table you are looking for is different, it should be achievable with a variation of this.

I have to admit this is not the prettiest sight. I'm thinking perhaps each field class should be able to tell how it will cast its values. I'll have to discuss that with other maintainers. But it should work! :-)

Let me know if that solves your problem.

sedubois:
@pablobm that could work but it offers no way to set the value back to `nil` once it's `true`/`false`. Alternately I thought the UI could allow specifying `nil`/`false`/`true` directly, if the field is nullable (instead of the only 2 choices currently possible), using 3 radio buttons instead of a checkbox. 

As far as strings are concerned, my workaround so far has indeed been to set `nil` if the string is blank:
```rb
# app/controllers/admin/application_controller.rb
module Admin
    ...

    # Hack: Store nil instead of empty String values when submitting forms
    # https://github.com/thoughtbot/administrate/issues/441
    def read_param_value(data)
      data.is_a?(String) and data.blank? ? nil : super
    end
  end
end
```

I haven't yet implemented/chosen how to fix the situation for booleans.

pablobm:
I see. In that case you need to create a new field type, as per http://administrate-prototype.herokuapp.com/adding_custom_field_types. This will have its own field template, and you can make it look like radio buttons, or anything you like.

PR #1479, opened 2019-12-05
Document controller APIs

pablobm:
When reviewing https://github.com/thoughtbot/administrate/pull/1478, it became apparent to me that controller APIs such as `resource_params` are not documented.

I think they should be part of Administrate's interface, and as such should be documented. Here's some documentation.

Additionally:

* Changed the example for `update` so that it's more useful. The current example doesn't render or redirect, so it may cause confusion to someone trying to emulate it.
* Mentioned `requested_resource` too.
* Fixed the indent for the `scoped_resource` example.


sedubois:
NB: still needs to be deployed to https://administrate-prototype.herokuapp.com.

PR #1482, opened 2019-12-06
Cache Bundle across Gemfiles in CI

seanpdoyle:
The `appraisal install` command does not (yet?) accept flags in the same
way that `bundle install` does.

Prior to this commit, `bundle install --path vendor/bundle` is invoked
between restoring from and saving to the cache. In theory, this
relies on prior installation work.

Unfortunately, since the test suite is run through Appraisals, these
caches are unused.

This commit replaces the `appraisal` install with the underlying
behavior, called explicitly by setting `BUNDLE_GEMFILE` directly for
each supported Rails version.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,11 @@ commands:
 
       # Bundle install dependencies
       - run: bundle install --path vendor/bundle
+      - run: BUNDLE_GEMFILE=gemfiles/rails42.gemfile bundle install --path vendor/bundle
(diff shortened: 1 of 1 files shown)


seanpdoyle:
On further inspection, this might not be installing and caching in the way I predicted.

Each Appraisal install depends on a different Gemfile, separate from the `Gemfile.lock` mentioned in the cache key.

nickcharlton:
I think I'd rather fix this in Appraisal itself (which has been a little under-loved). What do you think?

pablobm:
I understand from the conversation that this can be now closed. Please re-open if I misunderstand!

PR #1484, opened 2019-12-09
Double-click to select attribute text

croaky:
Each show page displays a description list of the fields.
When the field is text, it is rendered inside an inline `<span>`.

I've found it common to want to double-click this text
to copy-paste an identifier from our admin app into other systems.

Previously, double-clicking would not select the field's value.
The bug can be seen on the show page in the demo app. For example,
try selecting the city or zip code values from:

https://administrate-prototype.herokuapp.com/admin/orders/198962

Changing the markup to be a block `<div>` fixes the bug.
Alternatively, we could change the `preserve-whitespace` CSS rule
to be `display: block;`.

### DIFF
--- a/app/views/fields/string/_show.html.erb
+++ b/app/views/fields/string/_show.html.erb
@@ -4,7 +4,7 @@
 This partial renders a string attribute,
 to be displayed on a resource's show page.
 
-By default, the attribute is rendered as an unformatted string.
+By default, the attribute is rendered as text with whitespace preserved.
 
 ## Local variables:
 
@@ -15,4 +15,4 @@ By default, the attribute is rendered as an unformatted string.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/String
 %>
 
-<%= field.data %>
+<div class="preserve-whitespace"><%= field.data %></div>
--- a/app/views/fields/text/_show.html.erb
+++ b/app/views/fields/text/_show.html.erb
@@ -16,4 +16,4 @@ whitespace preserved.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Text
 %>
 
-<span class="preserve-whitespace"><%= field.data %></span>
+<div class="preserve-whitespace"><%= field.data %></div>
--- a/app/views/fields/text/_show.html.erb
+++ b/app/views/fields/text/_show.html.erb
@@ -4,8 +4,7 @@
 This partial renders a text attribute,
 to be displayed on a resource's show page.
 
-By default, the attribute is rendered as text with
-whitespace preserved.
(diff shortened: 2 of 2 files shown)


pablobm:
@croaky - What browser are you using? I'm unable to reproduce. I have tried Firefox, Chrome and Safari on macOS, and Firefox on Linux. With and without the change, I can select the whole field with a triple click.

Incidentally, I think a better example is a Product description. The Order fields you mention are `Field::String` rather than `Field::Text`.

croaky:
Chrome. Just tried triple-clicking, that works. Double-clicking doesn't.

https://administrate-prototype.herokuapp.com/admin/products/blokus

Try to select the name value "Blokus" with a double-click. The result looks like this:

Screen Shot 2019-12-13 at 3 57 26 PM

pablobm:
OK, I was able to reproduce what your screenshot shows on macOS Chrome. However `Product#name` is not a `Field::Text`, but a `Field::String`. Therefore your PR doesn't change its behaviour. Also, I tried surrounding it with a `<span>` (same as text fields) and this made it work as you desire.

How about changing this to add a `<span>` wrapper to string fields? Please let me know if I'm missing something.

croaky:
@pablobm Good call on `Field::Text`. It looks like we've been mostly using `Field::Text`s in our Administrate instance and rarely `Field::String`s.

I tried using a `<span>` but saw the same issue as the PR description and screenshot. I pushed a follow-up change to make `Field::Text` act the same way as `Field::String` using a `<div>` and am seeing consistent double-click selection the way I'd expect it to behave. Kind of nice that the defaults for the two string-y types are the same?

For further background: we put Administrate in front of an existing production (follower, read-only) Postgres database, whose primary API server is written in Go. We generated some lightweight ActiveRecord models from the schema to make it all work correctly. Our particular choices for database schema or how we generated it might have resulted in some inconsistency in `Field::Text` vs `Field::String` use.

For example, some of our `Field::Text` fields were single-block strings like `pmt123456789` or `user123456789` where double-clicking would be the expectation. Doesn't visually look or feel like a paragraph to triple-click.

Our case might be a little edge so understand if you don't want to support or go with another solution. Easy enough to port this approach into our own app using the generators (already done). This has been the most reliable approach I've found so far, though, so wanted to report back upstream!

croaky:
Another potential approach in `admin/app/views/fields/string/_show.html.erb` to avoid duplication:

```erb
<%= render 'fields/text/show', field: field %>
```

pablobm:
On second inspection, I must have done something wrong on the first inspection, because now I can reproduce the behaviour exactly as you describe. I think the change makes sense, and that the old behaviour was a bit weird. Merging!

Thank you :+1:

Issue #1464, opened 2019-11-08
HasMany relationships load all objects and then count on collection page


RCulpepp:
* What were you trying to do?
I was attempting to load an index of resources that had hundreds of thousands of associated resources through a `has_many` relationship.  When querying, it is attempting to load all objects in memory, before counting.  Thus slowing down the app (unnecessarily and unexpectedly) since all that displays is a count for those associated objects.

I have something like this:

```
resource_dashboard.rb

ATTRIBUTE_TYPES = {
  associated_resources: Field::HasMany
}.freeze

COLLECTION_ATTRIBUTES = [
  :associated_resources,
].freeze
```

* What did you end up with (logs, or, even better, example apps are great!)?
A very slow response!

I've come up with a solution that'll I'll submit as a PR.

* What versions are you running?
  - Rails 5.0.7.2
  - administrate 0.11.0

pablobm:
I think this is the same as https://github.com/thoughtbot/administrate/issues/1464. I'm closing this one as the other one has more discussion and detail.

gabriel:
You referenced this issue (#1464), did you mean to point to a different one?

pablobm:
Ooops. And not the first time I do this either :sweat:

I'm not sure which issue I meant to link to. It might be https://github.com/thoughtbot/administrate/issues/1697, where I offer a potential solution that I'll quote here:

> Perhaps this should be solved outside Administrate? I'm thinking of using the counter_cache option to belongs_to, and show this in the index page, instead of the HasMany field. See: https://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache

PR #1485, opened 2019-12-10
On index pages for has_many fields, only use the size of the association

cselmer:
This PR addresses #1464


cselmer:
This PR was premature - closing until I've fixed the specs.

PR #1488, opened 2019-12-12
Remove 'LOWER' in search query and cast only for numbers

netelix:
In order to speed up the search :

* the LOWER sql method has been replaced by an ILIKE search
* the CAST sql method is only done for `Number` fields

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -24,8 +24,23 @@ module Administrate
         table_name = query_table_name(attr)
         attr_name = column_to_query(attr)
 
-        "LOWER(CAST(#{table_name}.#{attr_name} AS CHAR(256))) LIKE ?"
+        query_for_attribute(attribute_types[attr], "#{table_name}.#{attr_name}")
       end.join(" OR ")
+    end
+
+    def query_for_attribute(attribute, field)
+      attribute_class =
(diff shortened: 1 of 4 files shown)


lib/administrate/search.rb
```diff
@@ -1,4 +1,28 @@
...
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
+require "active_support/core_ext/module/delegation"
```
pablobm: Uh, what is this? Perhaps an accident from an automated tool of some sort?

lib/administrate/search.rb
```diff
@@ -13,23 +37,52 @@ def run
       if @term.blank?
         @scoped_resource.all
       else
+        puts search_terms
+        puts search_terms.count
```
pablobm: This should be removed before we can merge.

nickcharlton:
I'm going to close this as it's been nearly a year since the last comments and there's now conflicts.

PR #1507, opened 2019-12-23
Avoid triggering unpermitted params on JS link

pablobm:
With https://github.com/thoughtbot/administrate/pull/1457, I introduced a bug. Reproduction:

1. Visit the index page of a model with a has_many relationship (eg: customers).
2. Click on a table header to trigger some sorting. The current URL will gain query params.
3. Click on a row to see the show page of a record. Don't click on actual data on the table (eg: customer name), but instead click on a "blank" area of the row. This is so that the JS click handler kicks in and performs the "link" behaviour, as opposed to clicking on an actual link.
4. You'll get an error `ActionController::UnpermittedParameters`.

This is triggered by the following:

1. At the index page, the URL is something like `/admin/customers`
2. When sorting the table, the current URL becomes something like `/admin/customers?customer%5Bdirection%5D=desc&customer%5Border%5D=name`
3. Clicking on the row, the JS gets us a URL like so `/admin/customers/123?customer%5Bdirection%5D=desc&customer%5Border%5D=name`
4. The partial that renders the has_many relationship in the show page can receive query params. It checks that the current ones are permitted. Turns out those aren't (they are in the index page, but not here), so it raises an exception.

I'm not sure about the spec example I have introduced. I feel that someone who finds it in the future is going to wonder where that came from. I guess they can check the git history...? Thoughts?

### DIFF
--- a/app/assets/javascripts/administrate/components/table.js
+++ b/app/assets/javascripts/administrate/components/table.js
@@ -13,7 +13,7 @@ $(function() {
       var dataUrl = $(event.target).closest("tr").data("url");
       var selection = window.getSelection().toString();
       if (selection.length === 0 && dataUrl) {
-        window.location.pathname = dataUrl;
+        window.location = window.location.protocol + '//' + window.location.host + dataUrl;
       }
     }
   };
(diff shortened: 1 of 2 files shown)


nickcharlton:
I'm happy with relying on the Git history for the explanation of this, so I'm going to merge it. Thanks!

PR #1512, opened 2020-01-01
Avoid confusion with class methods and protected/private modifiers

pablobm:
Working on something else, I saw the following:

```ruby
protected

# ...

def self.field_type
  to_s.split("::").last.underscore
end
```

This code can lead to misunderstanding, as the `protected` modifier doesn't apply to class methods like `self.field_type`.

On this PR, I move two such class methods behind the modifier, near the top of the class, to avoid this confusion.

I also move three other class methods to the top of the class, as I feel class methods should go together in general, in order to clarify interfaces and avoid future instances of this anti-pattern.

### DIFF
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -20,16 +20,16 @@ module Administrate
         to_s.split("::").last.underscore
       end
 
+      def self.permitted_attribute(attr, _options = nil)
+        attr
+      end
+
       def initialize(attribute, data, page, options = {})
         @attribute = attribute
         @data = data
         @page = page
         @resource = options.delete(:resource)
         @options = options
-      end
-
-      def self.permitted_attribute(attr, _options = nil)
-        attr
       end
 
       def html_class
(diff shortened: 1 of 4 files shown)


lib/administrate/field/base.rb
```diff
@@ -16,6 +16,14 @@ def self.searchable?
         false
       end
 
+      def self.field_type
+        to_s.split("::").last.underscore
+      end
+
+      def self.permitted_attribute(attr, _options = nil)
+        attr
+      end
```
nickcharlton: When there's two, what do you think about doing:

```ruby
class << self
  def field_type
    to_s.split("::").last.underscore
  end
end
```

Do you think that helps or hinders readability?

pablobm: I think that pattern is more difficult to read. I feel that only the indent gives a hint of those methods being special in some way, particularly if there's a long list and the `class << self` line is not on screen.

nickcharlton: Yeah, I'm happy with that!

PR #1514, opened 2020-01-02
Rewrite the contributing guide

nickcharlton:
The goal behind this document should be to explain to new and potential
contributors what they need to do (and follow) to get involved and it
wasn't really doing that.

It should also be a document that we can link to when closing or explaining
reasoning behind decisions. Broadly, it should aim to ensure decisions
aren't just arbitrary.

* Link to the contributing guide in the docs,
* Move repository/architecture discussion to the Contributing Guide,
* Document the use of labels (this matches those setup in GitHub),
* Document how releases happen and how it fits the hosted applications.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,7 +49,7 @@ This will start the application defined in `spec/example_app`.
 * The demo app is nested within `spec/example_app`.
 * The guides as seen at
   [https://administrate-demo.herokuapp.com][docs] live as
-  markdown files in the `docs` subdirectory.
+  Markdown files in the `docs` subdirectory.
 
 Rails configuration files have been changed
 to recognize the app in the new location,
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,61 +1,110 @@
-# Contributing
+# Contributing Guide
 
-We love pull requests from everyone.
-By participating in this project,
-you agree to abide by the thoughtbot [code of conduct].
+## Code of Conduct
 
-We expect everyone to follow the code of conduct
-anywhere in thoughtbot's project codebases,
-issue trackers, chatrooms, and mailing lists.
+We welcome pull requests from everyone. By participating in this project, you
+agree to abide by the thoughtbot [code of conduct].
 
-1. Fork the repo.
+We expect everyone to follow the code of conduct anywhere in thoughtbot's
+project codebases, issue trackers, chat-rooms, and mailing lists.
 
-1. Run `./bin/setup`.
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -1,70 +1,130 @@
...
 
-1. Make the test pass.
+## Repository Structure
 
-1. Write a [good commit message][commit].
+* The gem's source code lives in the `app` and `lib` subdirectories.
+* The demo app is nested within `spec/example_app`.
+* The guides as seen at
+  [https://administrate-demo.herokuapp.com][docs] live as
+  markdown files in the `docs` subdirectory.
```
mike-burns: Pardon my pedantry:

```suggestion
  Markdown files in the `docs` subdirectory.
```

CONTRIBUTING.md
```diff
@@ -1,70 +1,130 @@
...
 
-[code of conduct]: https://thoughtbot.com/open-source-code-of-conduct
-[commit]: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html
-[Rack MiniProfiler]: https://github.com/MiniProfiler/rack-mini-profiler
 [PGP key]: https://thoughtbot.com/thoughtbot.asc
 
+## Releasing
+
+New releases (and the time period between them) is arbitrary, but usually
+motivated by a new Rails release or enough bug fixes and/or features that
```
mike-burns: I'm a member of the anti-"and/or" squad, here to suggest that "or" would suffice!

```suggestion
motivated by a new Rails release or enough bug fixes or features that
```

CONTRIBUTING.md
```diff
@@ -1,70 +1,130 @@
...
+[docs]: https://administrate-prototype.herokuapp.com
 
-## Performance Improvements
+## Front-end Architecture
 
-Improving our users' experience should be the primary goal of any optimization.
+This project uses:
 
-If you contribute a performance improvement,
-you must submit performance profiles
```
mike-burns: Apparently this was added because [people were trying to micro-optimize in this codebase](https://github.com/thoughtbot/administrate/pull/293). Is it worth keeping something like this around?

CONTRIBUTING.md
nickcharlton: Mmm, yes, agreed!

CONTRIBUTING.md
nickcharlton: This isn't something I've seen for a long time, so I think it's okay that it's no longer there.

mike-burns: Sounds good. We can bring it back if it pops up again.

Issue #1213, opened 2018-09-06
Values not vertically aligned with field names on show page


eliblocks:
https://administrate-prototype.herokuapp.com/admin/customers/55286

Field values (dd elements) are slightly below the labels (dt elements) because of the smaller font size on the labels. It's very noticeable and hurts readability, in my opinion.

pablobm:
Fixed at https://github.com/thoughtbot/administrate/pull/1522

PR #1522, opened 2020-01-10
Align label and data on show page

davi-tapajos-codeminer42:
Issue: https://github.com/thoughtbot/administrate/issues/1213

Captura de Tela 2020-01-10 às 17 29 42

### DIFF
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -1,8 +1,6 @@
-th.cell-label {
-  padding-top: $top-space;
-}
-
 .cell-label {
+  padding-top: $top-space;
+
   &:hover {
     a {
       color: $action-color;
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -1,3 +1,7 @@
+th.cell-label {
+  padding-top: $top-space;
+}
+
 .cell-label {
   &:hover {
     a {
--- a/app/assets/stylesheets/administrate/components/_cells.scss
+++ b/app/assets/stylesheets/administrate/components/_cells.scss
@@ -1,5 +1,5 @@
 .cell-label {
-  padding-top: $top-space;
+  padding-top: 0.15em;
 
   &:hover {
     a {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/administrate/components/_cells.scss
```diff
@@ -1,4 +1,6 @@
 .cell-label {
+  padding-top: $top-space;
+
```
pablobm: Can you provide an example (screenshot) of how cell labels should also receive this padding? I can't recall from memory right now 😓

app/assets/stylesheets/administrate/library/_variables.scss
```diff
@@ -15,6 +15,7 @@ $heading-line-height: 1.2 !default;
 $base-border-radius: 4px !default;
 $base-spacing: $base-line-height * 1em !default;
 $small-spacing: $base-spacing / 2 !default;
+$top-space: 0.39em !default;
```
pablobm: I feel that `top-space` it a bit generic a name. Should this be something else? I guess it'll depend on whether `cell-label` needs to be altered (see my comment above).

app/assets/stylesheets/administrate/components/_cells.scss
davi-tapajos-codeminer42: Sure

cell-label

pablobm: Thank you, I get it now :+1:

pablobm:
OK @davi-tapajos-codeminer42, let's see if we can see this through!

I think an opinion from a designer may take a bit to arrive, so let's see what we can fix for now. What do you think of my comment about baseline vs centre alignment?

Also, I think the name `$top-space` is not very descriptive. Also it may be that we don't use the same measure in both places in the end, depending on how the baseline vs centre alignment question is resolved. What do you think?

davi-tapajos-codeminer42:
I think that centre alignment will looks nice! 
and yep, probably we will get rid of this $top-space :)

pablobm:
Cool. Would you be able to make those changes, and I'll review again? Thank you!

davi-tapajos-codeminer42:
It is kind hard to align with "dt" and "dd" tags
Maybe in the future we could change those tags and use vertical-align to get better results

Captura de Tela 2020-01-23 às 11 49 51

pablobm:
I think this is good enough for now. Thank you again @davi-tapajos-codeminer42 

Fixes https://github.com/thoughtbot/administrate/issues/1213

Issue #1517, opened 2020-01-06
HasOne without association routes shouldn't create link


bekicot:
* What were you trying to do?
Showing `HasOne` `association` in Index page. Where `association` doesn't have the corresponding path.
e.g:
`User` has_one `Profile`. But `profile` doesn't have an admin page. 
* What did you end up with (logs, or, even better, example apps are great!)?
It raises undefined method error like `Undefined method admin_profile_path`
* What versions are you running?
  - Rails 6.0
  - Administrate 12.0

I'd like to make a fix on this. If no one has.

Labels:
`bug`, `fields`

pablobm:
@bekicot - good spot! I can reproduce it indeed.

If you provide a fix, I'll be happy to review it. It'll probably involve using `valid_action?(:show, resource)` in the right place. I'm happy to answer questions if anything is not clear.

bekicot:
@pablobm Sure, please assign this to me, I will work on it tommorow night. Pretty occupied today.

pablobm:
No rush!

nickcharlton:
@bekicot, is this something we ended up solving? I just walked through the other PRs but I can't quite be sure.

pablobm:
I think this should have been fixed by https://github.com/thoughtbot/administrate/pull/1941 Happy to reopen if that's not the case

PR #1524, opened 2020-01-13
_navigation.html.erb: Hide resources without index

bekicot:
This will hide link on the navigation bar for resources without an index page.

Add interface `actions` to Administrate::Namespace::Resource  that can be a solution to #1517.

> This is related to #1517 but not directly fix it.

### DIFF
--- a/app/views/administrate/application/_navigation.html.erb
+++ b/app/views/administrate/application/_navigation.html.erb
@@ -15,6 +15,6 @@ as defined by the routes in the `admin/` namespace
       display_resource_name(resource),
       [namespace, resource_index_route_key(resource)],
       class: "navigation__link navigation__link--#{nav_link_state(resource)}"
-    ) %>
+    ) if valid_action? :index, resource  %>
   <% end %>
 </nav>
--- a/spec/example_app/config/routes.rb
+++ b/spec/example_app/config/routes.rb
@@ -5,7 +5,7 @@ Rails.application.routes.draw do
     resources :log_entries
     resources :orders
     resources :products
-    resources :product_meta_tags
+    resources :product_meta_tags, except: [:index]
     resources :payments, only: [:index, :show]
     resources :series
 
--- a/spec/features/navigation_spec.rb
+++ b/spec/features/navigation_spec.rb
@@ -36,4 +36,10 @@ describe "navigation" do
       expect(page).to have_header("Users")
     end
   end
+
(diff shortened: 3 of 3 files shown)


bekicot:
@pablobm please review this one. It simpler problem than #1517.

pablobm:
Thank you for this @bekicot! I think you are on the right track, but this perhaps could be simpler? Would it be possible to use the existing `valid_action?` instead of `resource.actions.include?`. See my comment at https://github.com/thoughtbot/administrate/pull/1418#pullrequestreview-343781330

bekicot:
> Thank you for this @bekicot! I think you are on the right track, but this perhaps could be simpler? Would it be possible to use the existing `valid_action?` instead of `resource.actions.include?`. See my comment at [#1418 (review)](https://github.com/thoughtbot/administrate/pull/1418#pullrequestreview-343781330)

I think valid action only work on the page that the action is available.
e.g:
supose we use`valid_action? :index` to show orders admin link in navigation.

- on products page `/admin/products` && `/admin/products/:id` will work, since ProductsController have have `:index` . And it will show on navigation.

- on line items page `/admin/product_meta_tags` && `/admin/product_meta_tags/:id` wont work, since ProductMetaTagsController doesn't have `:index`, It wont show on navigation

pablobm:
Sorry, I don't understand your example :-( Did you mean "products" when you said "orders"?

In your spec, you make Product Meta Tags not have an index page. Therefore it doesn't show on the navigation. I expect that it will not show in the navigation in any page: it will not show in products/index, products/show, product_meta_tags/show, etc.

The navigation partial is the same in all pages, regardless of what resource we are looking at.

Also, I have tried implementing it with `if valid_action?(:index, resource)`, and it does what I expect. Perhaps your expectation is different?

bekicot:
I was wrong, I wasn't carefully read the source code and made some wrong assumtion.

pablobm:
I think that your solution was much closer, and it will be easier for you to adapt it to what I'm asking. If you do it, I'll approve this PR.

bekicot:
> I think that your solution was much closer, and it will be easier for you to adapt it to what I'm asking. If you do it, I'll approve this PR.

I've already used `valid_action? :index, resource` instead in this commit #8f21f7b5e9f4e.

pablobm:
Oh sorry, I didn't notice you had pushed changes!

PR #1531, opened 2020-01-17
Remove project level rubocop.yml

nickcharlton:
Rubocop brings with it a certain amount of maintenence burden and
keeping project level config files means they're always somewhat out of date.
Additionally, we already have [thoughtbot/guides][one] which is the
source of truth.

This removes that to see if we get the right outcome.

[one]: https://github.com/thoughtbot/guides/blob/master/style/ruby/.rubocop.yml


jmeinerz:
@nickcharlton will this change mean hound won't be able to post suggestions to ruby code? Would that be a desirable outcome given that this project receives quite a lot of PRs?

nickcharlton:
Thankfully, no. Hound supports an organisation wide configuration:

Screenshot 2020-01-18 at 13 26 14

The one thing I did notice that we do do in our local `.rubocop.yml` is:

```yaml
AllCops:
  Exclude:
    - spec/example_app/db/schema.rb
```

…which means that now we'd likely get complaints by Rubocop as it won't be ignored. But I think this is a small price to pay for not having another configuration file.

jmeinerz:
Thanks for the insight, that was a great explanation. I completely agree with you.

PR #940, opened 2017-07-19
Add documentation of how to disable some actions on dashboard

rozhok:
#930

Add instructions of how to disable some actions.
Add IntelliJ-generated files to .gitignore.

PR #1534, opened 2020-01-20
Move documentation of how to disable some actions on controller

aitorlb:
### Problem

Instructions from #940  of how to disable some actions on controller can be hard to find.

### Solution

Move instructions:

- from `docs/customizing_dashboards.md`
- to `docs/customizing_controller_actions.md`


pablobm:
I think you are completely right :+1: Thank you!

PR #1536, opened 2020-01-23
Declare doctype, language and charset

pablobm:
Noticed these missing on the doc pages. Surely an oversight?


nickcharlton:
Almost certainly. Thanks! I'll merge it.

PR #1548, opened 2020-02-08
Test against Ruby 2.7 on CircleCI

nickcharlton:


pablobm:
Had a look today. I cannot reproduce the issue locally either. I think it might be due to a bug in Chromium/Chromedriver v85.0.4183.102, but I can't confirm as my Linux machine runs Chromium v83, and getting v85 is a bit of a faff. For comparison, the CircleCI Ruby 2.6 image (which doesn't have this problem) runs ChromeDriver v76.0.3809.126.

Fortunately, SSH'ing into CircleCI allowed me to reproduce the issue. It appears that it arises when certain other specs are run before this one. For example, if you run this:

```
circleci@1b3969b0b512:~/administrate$ for spec_file in spec/features/*; do bundle exec rspec --order defined --format documentation $spec_file spec/features/log_entries_index_spec.rb:51 ; done
```

Then you'll see that the troublesome example will fail if one of the following specs runs first:

  - `order form`
  - `Search`
  - `customer show page`

I tried a few other things. It looks like after running a few times, Chromium/Chromedriver enters a state where this error happens. So there might be a compound effect going on... Not helpful!

An idea would be to try some arbitrary ChromeDriver options, such as fullscreen or window size. Who knows what could help.

nickcharlton:
Ah nice! The last time I tried this, I couldn't even connect to the running instance, so I'll try again and see.

nickcharlton:
It's a browser size issue:

screenshot_2020-10-06-18-46-25 705

I'm not quite sure how it's only coming up now with this configuration, that seems odd to me.

PR #1549, opened 2020-02-11
Cache Appraisal dependencies across CI builds

nickcharlton:
Previously, `appraisal install` didn't pass through a `path` argument.
[This is in progress][1]. [This continues the work started
previously.][2]

So far, this duplicates the existing cache restore/set directives to see
if we get the behaviour intended.

[1]: https://github.com/thoughtbot/appraisal/pull/165
[2]: https://github.com/thoughtbot/administrate/pull/1482


gemfiles/rails42.gemfile
```diff
@@ -28,7 +28,7 @@ end
 
 group :test do
   gem "ammeter"
-  gem "capybara", "3.30.0"
```
pablobm: Is the Capybara update relevant?

nickcharlton: No. I'll remove it before I merge it. I'm most interested here on whether or not it makes a build difference / actually works

nickcharlton:
This fails on Ruby 2.6 on CI (only) and I cannot work out why.

pablobm:
The build appears to be passing now. Has anything changed?

nickcharlton:
I wonder if we were getting the reasonably regular failure here and I rebased it?

I'm going to do that again, see if it still passes and look back at what I did: it's possible it's just working now because the feature it's intended to do is just disabled.

nickcharlton:
Okay, coming back to this now by undoing it so it goes back to the feature working as it was supposed to.

nickcharlton:
Closing this, to potentially revisit in future.

It'd be great to cache the additional gems too but it already was quite a bit of work to get going. We can revisit it after GitHub Actions is up and running.

PR #1551, opened 2020-02-13
More readable warnings

pablobm:
During normal operation, the routes generator can print a few warnings. When there's more than one reason to trigger a warning for a given resource, these are printed out separately, often with other warnings intereleaved.

For example, these are 8 warning about 4 different resources:

```
WARNING: Unable to generate a dashboard for ActionText::RichText.
         Administrate does not yet support namespaced models.
WARNING: Unable to generate a dashboard for ActionMailbox::InboundEmail.
         Administrate does not yet support namespaced models.
WARNING: Unable to generate a dashboard for ActiveStorage::Blob.
         Administrate does not yet support namespaced models.
WARNING: Unable to generate a dashboard for ActiveStorage::Attachment.
         Administrate does not yet support namespaced models.
WARNING: Unable to generate a dashboard for ActionText::RichText.
         Make sure your database migrations are up to date.
WARNING: Unable to generate a dashboard for ActionMailbox::InboundEmail.
         Make sure your database migrations are up to date.
WARNING: Unable to generate a dashboard for ActiveStorage::Blob.
         Make sure your database migrations are up to date.
WARNING: Unable to generate a dashboard for ActiveStorage::Attachment.
         Make sure your database migrations are up to date.
```

This is a bit difficult to read and can be confusing. Instead, I propose that we group together the warnings related to each given resource, like so:

```
WARNING: Unable to generate a dashboard for ActionText::RichText.
       - Administrate does not yet support namespaced models.
       - It is not connected to a database table.
         Make sure your database migrations are up to date.
WARNING: Unable to generate a dashboard for ActionMailbox::InboundEmail.
       - Administrate does not yet support namespaced models.
       - It is not connected to a database table.
         Make sure your database migrations are up to date.
WARNING: Unable to generate a dashboard for ActiveStorage::Blob.
       - Administrate does not yet support namespaced models.
       - It is not connected to a database table.
         Make sure your database migrations are up to date.
WARNING: Unable to generate a dashboard for ActiveStorage::Attachment.
       - Administrate does not yet support namespaced models.
       - It is not connected to a database table.
         Make sure your database migrations are up to date.
```

This PR implements this behaviour.

### DIFF
--- a/lib/generators/administrate/routes/routes_generator.rb
+++ b/lib/generators/administrate/routes/routes_generator.rb
@@ -20,15 +20,15 @@ module Administrate
       end
 
       def warn_about_invalid_models
-        namespaced_models.each do |invalid_model|
-          puts "WARNING: Unable to generate a dashboard for #{invalid_model}."
-          puts "         Administrate does not yet support namespaced models."
-        end
-
-        models_without_tables.each do |invalid_model|
-          puts "WARNING: Unable to generate a dashboard for #{invalid_model}."
-          puts "         It is not connected to a database table."
-          puts "         Make sure your database migrations are up to date."
+        invalid_dashboard_models.each do |model|
+          puts "WARNING: Unable to generate a dashboard for #{model}."
+          if namespaced_models.include?(model)
+            puts "       - Administrate does not yet support namespaced models."
+          end
+          if models_without_tables.include?(model)
+            puts "       - It is not connected to a database table."
+            puts "         Make sure your database migrations are up to date."
+          end
         end
 
         unnamed_constants.each do |invalid_model|
@@ -49,15 +49,15 @@ module Administrate
       end
 
       def valid_dashboard_models
-        database_models - invalid_database_models
+        database_models - invalid_dashboard_models
       end
 
       def database_models
         ActiveRecord::Base.descendants.reject(&:abstract_class?)
       end
 
-      def invalid_database_models
-        models_without_tables + namespaced_models + unnamed_constants
+      def invalid_dashboard_models
+        (models_without_tables + namespaced_models + unnamed_constants).uniq
       end
 
       def models_without_tables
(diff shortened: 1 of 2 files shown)


lib/generators/administrate/routes/routes_generator.rb
```diff
@@ -20,15 +20,15 @@ def insert_dashboard_routes
...
-        end
-
-        models_without_tables.each do |invalid_model|
-          puts "WARNING: Unable to generate a dashboard for #{invalid_model}."
-          puts "         It is not connected to a database table."
-          puts "         Make sure your database migrations are up to date."
+        invalid_dashboard_models.each do |model|
+          puts "WARNING: Unable to generate a dashboard for #{model}."
+          if namespaced_models.include?(model)
+            puts "       - Administrate does not yet support namespaced models."
```
nickcharlton: Perhaps a different question to ask is: Do we still need this warning? In general, we do support namespaced models. What's missing from the current support which means we should still have this warning?

lib/generators/administrate/routes/routes_generator.rb
```diff
@@ -49,15 +49,15 @@ def dashboard_resources
...
+        database_models - invalid_dashboard_models
       end
 
       def database_models
         ActiveRecord::Base.descendants.reject(&:abstract_class?)
       end
 
-      def invalid_database_models
-        models_without_tables + namespaced_models + unnamed_constants
+      def invalid_dashboard_models
```
nickcharlton: This is a nice change!

lib/generators/administrate/routes/routes_generator.rb
pablobm: Exactly. I got here because I was looking into that. The error messages were confusing, so I thought of fixing that first, then looking into generating those namespaces. My thinking is that getting this to work properly may (perhaps) take long, so meanwhile let's at least make this more bearable, which is low-hanging fruit.

lib/generators/administrate/routes/routes_generator.rb
pablobm: Yeah, the inconsistent naming was bothering me too.

lib/generators/administrate/routes/routes_generator.rb
nickcharlton: Alright, let's do it!

Issue #1182, opened 2018-06-26
Demo error - sorting customers by country


MartinMcDermid:
Hi.
Filtering by country on the demo produces an error:
https://administrate-prototype.herokuapp.com/admin/customers?customer[direction]=asc&customer[order]=country

Thanks!

pablobm:
Took the liberty to update the URL in the description, as the given one was deprecated (format of the query string has changed since). The bug is still valid.

PR #1552, opened 2020-02-13
Use the correct foreign key when sorting belongs_to associations

pablobm:
`Administrate::Order` makes the assumption that a relation's foreign key will be `#{relation_name}_id`, when instead it could be anything.

This currently breaks our demo site, if you go to the [customers list](https://administrate-prototype.herokuapp.com/admin/customers) and try to order by country. This is as described by https://github.com/thoughtbot/administrate/issues/1182

### DIFF
--- a/lib/administrate/order.rb
+++ b/lib/administrate/order.rb
@@ -62,7 +62,7 @@ module Administrate
     end
 
     def order_by_id(relation)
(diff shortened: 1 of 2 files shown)


nickcharlton:
Nice, thanks! I'll merge this.

PR #1561, opened 2020-02-26
Fix very narrow nested fields

sedubois:
I am not especially talented at CSS, but here is a fix for the very narrow layout which makes editing bigger nested fields such as textareas very impractical.

Before:
before

After:
after

The problem can also be seen on http://administrate-prototype.herokuapp.com/admin/products when editing some product (product meta tag meta title and meta description are very narrow, which would be problematic if those fields were textareas).

Screenshot 2020-02-26 at 18 47 43

### DIFF
--- a/app/assets/stylesheets/administrate/components/_field-unit.scss
+++ b/app/assets/stylesheets/administrate/components/_field-unit.scss
@@ -17,21 +17,23 @@
 .field-unit__field {
   float: left;
   margin-left: 2rem;
-  width: 45%;
+  width: 100%;
+  max-width: 50rem;
 }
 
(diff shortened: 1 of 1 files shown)


pablobm:
I like your approach of max-widths based on rems rather than percentages. Thank you.

pablobm:
Hm, I just noticed that the resulting commit of my merge was attributed to me instead of you. I guess it's because it was a squash commit...? Apologies for that. I'll ask.

sedubois:
Thanks @pablobm, yes a squash would be a new commit, which seems correct to me. Thanks for merging 👍

Issue #585, opened 2016-05-17
Can you choose when a list appears as a normal dropdown vs a searchable one?


kieraneglin:
I have a long list of `locations`.  I'm going to be bulk-adding data for a while, so I'd like to have a searchable dropdown instead of a normal one.

Eg, turn [this ](http://puu.sh/oUHXz/b04fa742e9.png) into [this](http://puu.sh/oUI52/8e200aab27.png).  I tried adding `searchable: true` as an option, but that didn't work.

Both are one-many relationships.

miksago:
I'm thinking of just doing up a plugin for this; not sure yet.

miksago:
I copied the template and added a javascript include that was the contents:

```
$(function(){
  $('.field-unit--belongs-to select').selectize();
});
```

carlosramireziii:
Hey @ThisIsMissEm! 

I think you've got the right idea-- a custom field is the way to go if you want to have special handling for really long lists. If you would like to package up your solution as a plugin, that'd be great!

It seems like you got this taken care of so I'm going to close this issue. Thanks!

ThisIsMissEm:
@carlosramireziii actually, this is a bug; See https://github.com/thoughtbot/administrate/blob/master/lib/administrate/field/select.rb#L6-L8 — that value "searchable?" doesn't actually do anything in the view: https://github.com/thoughtbot/administrate/blob/master/app/views/fields/select/_form.html.erb

(granted, I suspect this could be a language thing, as what you mean by `searchable?` upon re-reading is that the Collection on the index page can be searched / filtered by that field, rather than that field is able to be searched with autocomplete)

It'd be a pretty small change to core to actually support this like you do with HasMany fields. The reason I needed this is because I was building an administration panel for a microservice, and the data for that value "locations" actually came from an upstream API call — So I had like a "Artists Adminstration panel" and a "Locations API Service", and I wanted to connect the two.

I feel like this is a pretty common use case as Microservices catch on with Rails API now being really solid.

carlosramireziii:
@ThisIsMissEm Ah okay, tell me if I'm understanding correctly... You'd like to have the same "searchable" behavior that's implemented for `HasMany` fields using the `selectize` plugin to be possible for `BelongsTo` fields as well? 

If so, does the [`administrate-field-belongs_to_search`](https://github.com/fishbrain/administrate-field-belongs_to_search) plugin implement what you're looking for?

[...]

PR #1566, opened 2020-03-09
Enable Selectize for BelongsTo

sedubois:
Allow to input a model's BelongsTo relation using keyboard strokes instead of having to scroll a potentially long list of records with the mouse.

Screenshot 2020-03-09 at 12 37 14

### DIFF
--- a/app/assets/javascripts/administrate/components/associative.js
+++ b/app/assets/javascripts/administrate/components/associative.js
@@ -1,3 +1,4 @@
 $(function() {
+  $('.field-unit--belongs-to select').selectize({});
   $(".field-unit--has-many select").selectize({});
 });


pablobm:
Good call!

G-Rath:
Just FYI this is actually a pretty breaking change because there's no way to disable it, and so any existing `select` based styles are wholly incompatible.

IMO this sort of functionality should be delivered by a custom field, as mentioned in #585 (and for which actually a custom field already exists: https://github.com/fishbrain/administrate-field-belongs_to_search)

pablobm:
@G-Rath - Would you be able to provide a PR?

G-Rath:
@pablobm you mean a PR reverting this change?

pablobm:
Sorry, no: I mean something that allows disabling it, while still keeping the functionality.

G-Rath:
That's sort of part of my point and problem - I'm not sure how you can really build a clean way of disabling it, given that "it" is just a single line of plain javascript (i.e it's not even in an ERB template that would let you conditionally render it).

I guess the best way I can think of would be to add an extra css class that gets targeted by this, and so this would be replaced by a  `$('.field-unit--belongs-to selectize-please')` call instead.

But this is why I feel it should be done via a custom field, as it strikes me as exactly what custom fields are for.

pablobm:
I like the solution of the CSS class. Perhaps with a `data-` attribute instead.

I understand that this feels to you like it should be a separate field type. It's a bit on the fence for me, but I lean towards it being the same field. It's the same functionality in all cases except for a slightly different presentation on the form. In contrast, the "administrate-field-belongs_to_search" that you mention is an entirely different field, as it issues Ajax requests to search for options.

Issue #1179, opened 2018-06-19
Error on Search on belongs_to with class_name option


sobanakram:
Error:
![screenshot from 2018-06-19 17-58-41](https://user-images.githubusercontent.com/10559510/41598769-a67b80b0-73ea-11e8-89a5-21ce1b48f289.png)

Dashboard:

```ruby
ATTRIBUTE_TYPES = {
  sender: Field::BelongsTo.with_options(
    class_name: "User",
    searchable: true,
    searchable_field: :phone_number,
  )
}
```

Model:

```ruby
belongs_to :sender, class_name: 'User', foreign_key: :sender_id
```

andrepcg:
Having the same issue.

```
ATTRIBUTE_TYPES = {
    owner: Field::BelongsTo.with_options(class_name: 'Person', searchable: true, searchable_field: :first_name),
}
```

`PG::UndefinedTable: ERROR:  missing FROM-clause entry for table "owners"
LINE 1: ...name" AS CHAR(256))) LIKE '%simão%' OR LOWER(CAST("owners".."first_name"`

It's trying to use the `owners` table but I set the class to be `Person` so it should use `people.first_name` in the query

Rails 5.1.7
Administrate 0.11.0

andrepcg:
It might be fixed in https://github.com/thoughtbot/administrate/pull/1264

bjgaynor:
This is still not working to the best of my knowledge and it's a shame because I have a lot of custom class names that now can't be searchable.

ksearfos:
Agreed -- I am still having this issue

pablobm:
Should be fixed now. Please let us know if not.

andrepcg:
@pablobm Still having the same issue on v0.13.

https://github.com/thoughtbot/administrate/issues/1179#issuecomment-522488557

It still tries searching in the table named `owners` instead of `people`

sedubois:
@andrepcg release 0.13.0 was on March 13 but that PR was merged on March 21. Could you try master?

PR #1569, opened 2020-03-13
Tables won't always be named after models

pablobm:
Alternative to https://github.com/thoughtbot/administrate/pull/1264.
Fixes https://github.com/thoughtbot/administrate/issues/1179.

Currently, `Administrate::Search::Query` makes the assumption that an association's table name will have the same name as the association attribute, ignoring a `:class_name` option that signals otherwise.

This change uses the `:class_name` option when given.

Changing the example app also provides a way to test this as part of the existing feature specs.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -117,7 +117,12 @@ module Administrate
 
     def query_table_name(attr)
       if association_search?(attr)
-        ActiveRecord::Base.connection.quote_table_name(attr.to_s.pluralize)
+        provided_class_name = attribute_types[attr].options[:class_name]
+        if provided_class_name
+          provided_class_name.constantize.table_name
+        else
+          ActiveRecord::Base.connection.quote_table_name(attr.to_s.pluralize)
+        end
       else
         ActiveRecord::Base.connection.
           quote_table_name(@scoped_resource.table_name)
--- a/spec/example_app/app/dashboards/customer_dashboard.rb
+++ b/spec/example_app/app/dashboards/customer_dashboard.rb
@@ -12,9 +12,10 @@ class CustomerDashboard < Administrate::BaseDashboard
     log_entries: Field::HasManyVariant.with_options(limit: 2, sort_by: :id),
     updated_at: Field::DateTime,
     kind: Field::Select.with_options(collection: Customer::KINDS),
-    country: Field::BelongsTo.with_options(
+    territory: Field::BelongsTo.with_options(
       primary_key: :code,
       foreign_key: :country_code,
+      class_name: "Country",
       searchable: true,
       searchable_field: "name",
     ),
@@ -28,7 +29,7 @@ class CustomerDashboard < Administrate::BaseDashboard
     :email,
     :email_subscriber,
     :kind,
-    :country,
+    :territory,
     :password,
   ].freeze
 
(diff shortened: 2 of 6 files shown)


pablobm:
@sedubois @ksearfos @bjgaynor - Would you be able to check if this fixes your issues? Thank you.

sedubois:
@pablobm I tried to load the branch in my project as follows:
```rb
# Gemfile
gem 'administrate', git: 'https://github.com/pablobm/administrate.git', branch: 'association-search-table-name'
```

But I get
```
$ bundle 
Fetching https://github.com/pablobm/administrate.git
fatal: Could not parse object 'b2cb529f710915bb12ade19b2b9cd6ca30f5496a'.
Revision b2cb529f710915bb12ade19b2b9cd6ca30f5496a does not exist in the repository https://github.com/pablobm/administrate.git. Maybe you
misspelled it?
```

Any idea what I might be doing wrong?

pablobm:
That's strange. It looks to me like a GitHub issue. I upgraded from Administrate 0.12 to 0.13 on a dummy app and it installed the SHA that your setup claims to be missing:

```
Using administrate 0.12.0 (was 0.13.0) from https://github.com/pablobm/administrate.git (at association-search-table-name@b2cb529)
```

That SHA is also present on Github at https://github.com/pablobm/administrate/commit/b2cb529f710915bb12ade19b2b9cd6ca30f5496a

Perhaps try going back to 0.12, bundle, then 0.13 again? Or wiping out the gem from your machine and trying again. For example, in my machine it's at `~/.asdf/installs/ruby/2.6.5/lib/ruby/gems/2.6.0/bundler/gems/administrate-b2cb529f7109`. I use [asdf](https://github.com/asdf-vm/asdf), so it will be different with rbenv, RVM, etc.

sedubois:
@pablobm no luck but as a workaround I cloned your branch and used `gem 'administrate', path: '~/Repos/administrate-pablobm'`.

So now a first issue I observe is that `association_search?` expects the field to be HasOne, HasMany or BelongsTo, but my field is NestedHasMany. I think it can be fixed as follows:

```rb
# lib/administrate/search.rb

    def association_search?(attribute)
      return unless attribute_types[attribute].respond_to?(:deferred_class)
      
      attribute_types[attribute].deferred_class.is_a? Administrate::Field::Associative
    end
```

Secondly there appears to be a namespacing issue again, the class_name is Post::Translation but I get `ActionView::Template::Error: PG::UndefinedColumn: ERROR:  column posts.translations does not exist. LINE 1: ..."post_translations"."locale" = $3 AND (LOWER(CAST("posts"."t...`. It is constructing the table name `posts.translations` instead of `post_translations`.

sedubois:
@pablobm I updated previous comment, with that chane now the search works for me 🎉

sedubois:
@pablobm also this PR seems partly duplicate of #1203. Would it be possible to integrate the changes it contains to support multiple association search fields (`searchable_fields` array)?

pablobm:
Uh, I thought I had merged a fix for this some time ago. Turns out I hadn't: https://github.com/thoughtbot/administrate/pull/1398 Merged now.

That other PR I just merged takes a bit more of a heavy-handed approach. I like your solution using `<` instead. Would you be up for a separate MR that refactors that method, and also the following (mentioned on the PR) to use your approach?

https://github.com/thoughtbot/administrate/blob/acada0585b35aee532116fe04d65dc5263bba179/lib/administrate/base_dashboard.rb#L76-L80

As for this PR here, I just rebased it. Would you be able to check that the problem is actually fixed now?

sedubois:
@pablobm I confirm the search now works for my `Post::Translation` associations of type `NestedHasMany` without any extra modifications.

I've also created PR #1576.

pablobm:
Thank you @sedubois!

Issue #1573, opened 2020-03-16
Undefined local variable or method `root_url' for Class


rafaelgonzalez:
* What were you trying to do?

  Upgrade administrate from 0.12.0 to 0.13.0.

* What did you end up with (logs, or, even better, example apps are great!)?

  - Error
    
    ```
    Admin::EdgeCases#index: undefined local variable or method 'root_url' for #<#<Class:0x00007f9853a6bd08>:0x00007f9851407e50>
    ```
   
    ![Screen Shot 2020-03-17 at 10 31 59](https://user-images.githubusercontent.com/456921/76808464-f9b57780-683b-11ea-8aae-6b6b5f11c4ef.png)

* What versions are you running?
  - Ruby 2.6.3
  - Rails 5.2.4.1
  - administrate 0.13.0

* Solution

  It looks like this is due to not having a `root` specified in `routes.rb`. Doing that fixes the issue ([see Rails' docs](https://guides.rubyonrails.org/routing.html#using-root)).

  `root` in `routes.rb` is not required by Rails however, so likely to happen to a few apps using administrate.

pablobm:
Yup, this is a known issue. Would you be up for an MR?

rafaelgonzalez:
@pablobm I don't have much time at the moment, sorry! Glad to see someone else has jumped on it though! 😄

pablobm:
Looks like there are people on the case. Closing this!

adrianotadao:
Hey, could you generate a new version including the fix?

thx

nickcharlton:
We are working on this, it's just held up with a more serious regression we got with trying to manage sprockets/webpacker.

PR #1574, opened 2020-03-18
Fix non defined root_url bug

skibox:
Fix for https://github.com/thoughtbot/administrate/issues/1573


skibox:
Just a follow up: this feature was probably implemented for web applications, but it's unnecessary for applications that are API only.

rafaelgonzalez:
@skibox Thanks for this. You comment is spot on, I use administrate for an API and hence don't need the back button.

pablobm:
Thank you!

PR #1589, opened 2020-03-26
Dynamic select field

### LINKED ISSUE #1535: Dynamic select field
This should be the simpler case
```ruby
Field::Select.with_options(collection: [Time.current])
```
another case is if we want to keep the list up to date. e.g via database record

I think by adding another parameter type, it should be something like lambda would be enough to solve the problem.

No other approach. as to my knowledge.

bekicot:
first of all, sorry for the late PR :)

allow collection to accept proc as value.
usage:
```
release_year: Field::Select.with_options(collection: -> { (Time.current.year - 10)..Time.current.year })
```
closes #1535

cc @nickcharlton @pablobm

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -200,7 +200,8 @@ objects to display as.
 
 **Field::Select**
 
-`:collection` - Specify the array or range to select from. Defaults to `[]`.
+`:collection` - Specify the options shown on the select field. It accept either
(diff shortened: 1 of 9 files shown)


nickcharlton:
Thanks!

Could you add something to the documentation to show that this new feature exists?

nickcharlton:
Thanks! I'll go ahead and merge this now.

PR #1591, opened 2020-03-28
Check if routes can be shown in navigation

sedubois:
Honor the [authorization policy](http://administrate-prototype.herokuapp.com/authorization) when rendering side navigation index links.

### DIFF
--- a/spec/features/navigation_spec.rb
+++ b/spec/features/navigation_spec.rb
@@ -42,16 +42,4 @@ describe "navigation" do
     navigation = find(".navigation")
     expect(navigation).not_to have_link("Product Meta Tags")
   end
-
-  it "shows link to resource for which index? is authorized" do
-    visit admin_customers_path
-    navigation = find(".navigation")
-    expect(navigation).to have_link("Products")
-  end
-
-  it "hides link to resource for which index? is not authorized" do
-    visit admin_customers_path
-    navigation = find(".navigation")
-    expect(navigation).not_to have_link("Orders")
-  end
 end
--- a/spec/features/navigation_spec.rb
+++ b/spec/features/navigation_spec.rb
@@ -42,4 +42,16 @@ describe "navigation" do
     navigation = find(".navigation")
     expect(navigation).not_to have_link("Product Meta Tags")
   end
+
+  it "shows link to resource for which index? is authorized" do
+    visit admin_customers_path
+    navigation = find(".navigation")
+    expect(navigation).to have_link("Products")
+  end
+
+  it "hides link to resource for which index? is not authorized" do
+    visit admin_customers_path
+    navigation = find(".navigation")
+    expect(navigation).not_to have_link("Orders")
+  end
 end
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -38,6 +38,11 @@ module Administrate
       "#{resource_name.to_s.singularize}_dashboard".classify.constantize
     end
 
+    def model_from_resource(resource_name)
+      dashboard = dashboard_from_resource(resource_name)
+      dashboard.try(:model) || resource_name.to_sym
+    end
+
     def display_resource_name(resource_name)
       dashboard_from_resource(resource_name).resource_name(
         count: PLURAL_MANY_COUNT,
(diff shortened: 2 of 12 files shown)


nickcharlton:
Can you add a test for this?

sedubois:
@nickcharlton I've added a test and fixed the others, could you have a look? 

NB: to make the tests green I had to add some Pundit policies which were missing. In my own project I avoid having to add empty policies by falling back to the ApplicationPolicy when they don't exist. Here is my punditize.rb:

```rb
# Re-implement Administrate's Punditize which does not support namespaced models.
# See https://github.com/thoughtbot/administrate/issues/1332.
module Admin
  module Punditize
    include Pundit

    def policy_scope(scope)
      super([:admin, scope])
    rescue ::Pundit::NotDefinedError
      # Fall back to base scope if none exists for this particular record
      super(scope, policy_scope_class: Admin::BasePolicy::Scope)
    end

    def scoped_resource
      policy_scope super
    end

    def authorize_resource(resource)
      authorize resource
    end

    def pundit_user
      current_administrator
    end

    def policy(record)
      super([:admin, record])
    rescue ::Pundit::NotDefinedError
      # Fall back to base policy if none exists for this particular record
      Admin::BasePolicy.new(pundit_user, record)
    end

      policy(resource).send("#{action}?".to_sym)
    end
  end
end
```

sedubois:
@nickcharlton I've added a test and fixed the others, could you have a look? 

NB: to make the tests green I had to add some Pundit policies which were missing (*). 

NB2: the tests seem to be a bit messed up as some of them are loading the Pundit policies whereas others aren't. E.g. `orders_controller_spec.rb` does `include Administrate::Punditize` and as a consequence depends on all the other policies being present. On the other hand the test I added doesn't make use of Pundit policies as I didn't know how to load them in the navigation feature test; so I just tweaked the example app's application controller directly.

(*) In my own project I avoid these empty policies by falling back to the ApplicationPolicy when they don't exist. Here is my punditize.rb:

```rb
# Re-implement Administrate's Punditize which does not support namespaced models.
# See https://github.com/thoughtbot/administrate/issues/1332.
module Admin
  module Punditize
    include Pundit

    def policy_scope(scope)
      super([:admin, scope])
    rescue ::Pundit::NotDefinedError
      # Fall back to base scope if none exists for this particular record
      super(scope, policy_scope_class: Admin::BasePolicy::Scope)
    end

    def scoped_resource
      policy_scope super
    end

    def authorize_resource(resource)
      authorize resource
    end

    def pundit_user
      current_administrator
    end

    def policy(record)
      super([:admin, record])
    rescue ::Pundit::NotDefinedError
      # Fall back to base policy if none exists for this particular record
      Admin::BasePolicy.new(pundit_user, record)
    end

      policy(resource).send("#{action}?".to_sym)
    end
  end
end
```

nickcharlton:
Re: empty polices: Ah interesting, do you think we'd be able to avoid adding them here too? Or do you think that might be more complex?

spec/features/navigation_spec.rb
```diff
@@ -42,4 +42,16 @@
...
+    visit admin_customers_path
+    navigation = find(".navigation")
+    expect(navigation).to have_link("Products")
+  end
+
+  it "hides link to resource for which index? is not authorized" do
+    visit admin_customers_path
+    navigation = find(".navigation")
+    expect(navigation).not_to have_link("Orders")
+  end
```
pablobm, line 56: Nitpick: how about moving these to `spec/example_app/spec/features/authorization_spec.rb`? I think that the authorization feature, since it's optional, is differentiated enough that deserves having its own file.

nickcharlton: Mmm, I'd agree with this. It'll stop the file becoming too big now that we know already it's a bit different.

sedubois:
I've extracted the authorization tests as requested.

@nickcharlton to answer your question about using Pundit in tests, I guess the simplest would be to update Administrate's Punditize with the example I gave? https://github.com/thoughtbot/administrate/pull/1591#issuecomment-609077261

nickcharlton:
Thanks!

I guess it would be, yeah. What do you think about doing that in another PR, we'll hold off on merging this, sort that out and then rebase this on top?

I haven't used Pundit in quite a while, so I don't know if @pablobm has any other thoughts here?

spec/features/navigation_spec.rb
pablobm: Uh, now I realize that I initially created this file in the wrong path :sweat: @sedubois created it in the right place. I'll move mine in a separate PR.

pablobm:
Argh, sorry. I merged too quickly. The idea to hold off and merge #1663 first was good; but we'll have to do it differently... now :facepalm:

PR #1602, opened 2020-04-04
Update Ruby version

sedubois:


nickcharlton:
I'm dragging my feet a bit on this, but I'd like to do #1548, before we upgrade it project wide.

nickcharlton:
…and finally. Thanks for opening this way back when!

PR #1603, opened 2020-04-05
Add Field::CollectionSelect

sedubois:
This is in answer to https://github.com/thoughtbot/administrate/issues/1535#issuecomment-608354608.

TODO:
- [ ] test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@
 
 ## Changes
 
+* [FEATURE] Add a `CollectionSelect` field for displaying a drop-down menu of
+  options on form pages. For details and options, refer to the documentation.
+
 ### 0.13.0 (March 13, 2020)
 
 **NOTE:** This release contains the fix for [`CVE-2020-5257`][cve-5257], which
--- a/app/assets/javascripts/administrate/components/associative.js
+++ b/app/assets/javascripts/administrate/components/associative.js
@@ -1,4 +1,3 @@
 $(function() {
-  $('.field-unit--belongs-to select').selectize({});
-  $(".field-unit--has-many select").selectize({});
+  $('.field-unit--selectize select').selectize({});
 });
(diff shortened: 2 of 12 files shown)


pablobm:
Just had a cursory look. I'm not sure about Selectize being referred to directly on the Ruby code. Although it helps remove some duplication, I think it tightens the dependency on this specific solution, where users may prefer others in the future.

This is only a half-formed thought at the moment, but I feel that a more flexible way for users to enable/disable Selectize would be to create custom fields, derived from the existing ones, which override the partials and set different class names. Otherwise, we create a solution where it's more difficult to opt out of Selectize.

Does this make sense?

sedubois:
@pablobm understood about the naming. How about using another name, e.g.
```rb
def self.advanced_editing?
  options.fetch(:advanced_editing, true)
end
```

As for using a partial, I felt that passing an option to the field might in general be easier. In particular, it would enable inheritance, e.g. a field extending, say, `Field::BelongsTo` would automatically inherit the default value from its parent. Unless I'm overlooking something with partials?

sedubois:
@pablobm could you also try to run the branch to see why I get the error `uninitialized constant Administrate::Field::CollectionSelect`? It's probably something silly, I'm just a bit new to this.

pablobm:
@sedubois - I'm still not convinced about that option. The first thing is that I don't like adding that method to `Field::Base`, as it encumbers its interface with detail that is only relevant for some subclasses.

Also at a more "idealistic" level, I like how currently the forms are JS-enhanced without the Ruby playing any part. It feels to me more extensible, although at this point I admit that I'm working more with gut feel and platonic ideals of progressive enhancement.

pablobm:
@sedubois - To fix that error, add:

```ruby
### /lib/administrate/base_dashboard.rb
require "administrate/field/collection_select"
```

Admittedly, that's probably a less clear part of creating a new field type :-/

pablobm:
And by all means: please add a test. A field in the example app would be ideal, as it also acts as an example and a quick way to play with it. Thank you!

pablobm:
Closing due to lack of activity. Also I'm not sure this adds something that cannot be done with `Field::Select`? Another thing is that doing `collection: Customer.all` could backfire quickly as it would be resolved on boot and not be updated as the table changes. Instead passing `collection: ->{ Customer.all }` to `Field::Select` would be safer.

PR #1609, opened 2020-04-09
[QoL fix] Added SCSS imports

albertpalka:
When running `rails administrate:assets:stylesheets` in a Rails project (6.0.2) and updating my application.scss file, rails app throws multiple CSS errors which are caused by the lack of imports across multiple files. 

Just a minor QoL improvement.

### DIFF
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -1,3 +1,5 @@
+@import '../library/variables';
+
 fieldset {
   background-color: transparent;
   border: 0;
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -1,4 +1,4 @@
-@import '../library/variables';
+@import "../library/variables";
 
 fieldset {
   background-color: transparent;
--- a/app/assets/stylesheets/administrate/base/_layout.scss
+++ b/app/assets/stylesheets/administrate/base/_layout.scss
@@ -1,4 +1,4 @@
-@import '../library/variables';
+@import "../library/variables";
 
 html {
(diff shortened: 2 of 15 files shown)


albertpalka:
@pablobm Thank for the ✅ I should've fixed those hound warnings earlier... 

Regarding the "frontend part of Administrate", I am not a frontend developer myself so I'd rather not work on any gem updates as a part of this PR. Once agreed on the scope though, I'm more than happy to contribute in the future 😄

pablobm:
Yeah, no worries. My experience with Rails's asset handling is limited, so I just want to make sure that there wasn't an important reason why those `@import`s were not there. Perhaps related to a change to the way assets are handled in Rails?

I just had another look and I'm thinking that perhaps those imports are not there because it's the ones at `app/assets/stylesheets/administrate/application.scss` that should kick in instead...? I'm confused now. I need to have a closer look.

pablobm:
OK, so it appears to be an issue with the asset pipeline, conflicting with Webpacker. If I remove `/app/assets/stylesheets/application.css`, then the imports in `application.scss` will start working across SCSS files.

Not sure how best to proceed. I'll have to look into this more deeply. I haven't been much in touch with these features since before Webpacker was introduced, so I'm a bit out of my depth.

nickcharlton:
It sounds like this is an issue most related to how we're not using (but support) Webpacker and come a change there, this would be solved. Does that sound reasonable?

pablobm:
Yes. Currently I have no idea of what the state of frontend Rails is, regarding Pipeline vs. Webpacker. We may want to still support both, and there may be ways to do it. In any case, this needs some research. For a start creating apps, with different versions of Rails, using either option, and see how they work.

pablobm:
Related conversation on Sprockets vs. Webpacker https://github.com/thoughtbot/administrate/issues/1617

PR #1618, opened 2020-04-22
Remove jquery ujs

nhippenmeyer:
`rails-ujs` has been moved into the rails project and using both `jquery-ujs` together with `rails-ujs` causes issues when attempting to programatically submit events (eg calling `Rails.fire`):

Screen Shot 2020-04-22 at 6 02 12 PM

### DIFF
--- a/app/assets/javascripts/administrate/application.js
+++ b/app/assets/javascripts/administrate/application.js
@@ -1,5 +1,4 @@
 //= require jquery
-//= require jquery_ujs
 //= require selectize
 //= require moment
 //= require datetime_picker
--- a/spec/example_app/app/assets/javascripts/application.js
+++ b/spec/example_app/app/assets/javascripts/application.js
@@ -11,5 +11,4 @@
 // about supported directives.
 //
 //= require jquery
-//= require jquery_ujs
 //= require_tree .


pablobm:
Thank you for the PR @nhippenmeyer. I'm trying to think if this could cause an unintended incompatibility; for example with Rails 5.0, as this change you mention only took place in 5.1 (if I have this right).

I ran a quick test with Rails 5.0, and it appears to work. I see that this version of Rails includes `//= require jquery_ujs` in its `application.js`, so that may be enough... but I'm not sure, as removing it (and using Administrate with your changes) doesn't break JS in an obvious way either.

This may require a bit more of investigation. I wonder if @nickcharlton has any thoughts?

nhippenmeyer:
Thanks for taking a look @pablobm. Could we add a deprecation warning for users on an older version of rails that this could potentially cause a breaking change? Happy to do more testing as well if that would be helpful.

nickcharlton:
I'm happy to go ahead with removing a dependency as long as it doesn't have any compatibility drawbacks.

I think at this point of Rails releases that if we had a warning for the last old version of Rails we support which helps users solve their problem easily it would also be a good tradeoff.

pablobm:
I feel confident enough to merge this :crossed_fingers: 

@nickcharlton - Regarding your second point, I have created this issue to clarify Administrate's compatibility: https://github.com/thoughtbot/administrate/issues/1631

PR #1620, opened 2020-04-23
Remove unused variables in scss files

weapp:
I notice that `$flash-colors` is unused in any files, and `$base-spacing` is previously defined in [_variable.css](https://github.com/thoughtbot/administrate/blob/master/app/assets/stylesheets/administrate/library/_variables.scss#L16).
I also moved the current `$flashes` variable to the `_variables.sccs` with the rest of the variables.


nickcharlton:
Thanks! Tried this out and it looks good.

PR #1626, opened 2020-04-28
Allow translating resource names in CRUD messages

### LINKED ISSUE #1242: Resource names are not being translated properly on notice messages
The resource model name isn't being translated when you execute a DB transaction (create, for example). The whole message is translated according to the `yml` file but the resource model name.

I believe the problem relies on `ResourceReolver#resource_title`, which just parses the current path and returns a `joined` string. The `#translate_with_resource` (the method that translated the message itself) should use `ResourceReolver#resource_class` combined with ActiveRecord locale methods to return the resource model name already localized to the message _translator_ method

This is the message that I'm getting
image

And my `yml` file does have that model translation entry
```
pt-BR:
  activerecord:
    models:
      transaction_category:
        one: 'Natureza de Transação'
        other: 'Naturezas de Transação'
```

This problem is happening with other resources around my admin section

betelgeuse:
Fixes #1242

### DIFF
--- a/spec/features/edit_page_spec.rb
+++ b/spec/features/edit_page_spec.rb
@@ -65,13 +65,23 @@ describe "customer edit page" do
     new_email = "example@example.com"
     customer = create(:customer)
 
-    visit edit_admin_customer_path(customer)
-    fill_in "Name", with: new_name
-    fill_in "Email", with: new_email
-    click_on "Update Customer"
+    translations = {
+      activerecord: {
+        models: {
+          customer: "Custom name",
+        },
+      },
+    }
+
+    with_translations(:en, translations) do
+      visit edit_admin_customer_path(customer)
+      fill_in "Name", with: new_name
+      fill_in "Email", with: new_email
+      click_on "Update Custom name"
+    end
 
     expect(page).to have_text(new_name)
     expect(page).to have_text(new_email)
-    expect(page).to have_flash("Customer was successfully updated.")
+    expect(page).to have_flash("Custom name was successfully updated.")
   end
 end
--- a/spec/lib/administrate/resource_resolver_spec.rb
+++ b/spec/lib/administrate/resource_resolver_spec.rb
@@ -1,7 +1,4 @@
-require "spec_helper"
-require "active_support/core_ext/string/inflections"
-require "support/constant_helpers"
-require "administrate/resource_resolver"
+require "rails_helper"
 
 describe Administrate::ResourceResolver do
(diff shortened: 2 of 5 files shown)


betelgeuse:
How do I found out the issue that @hound is still reporting?

sedubois:
@betelgeuse seems like all warnings are fixed now

spec/features/edit_page_spec.rb
```diff
@@ -65,13 +65,23 @@
...
     customer = create(:customer)
 
-    visit edit_admin_customer_path(customer)
-    fill_in "Name", with: new_name
-    fill_in "Email", with: new_email
-    click_on "Update Customer"
+    translations = {
+      activerecord: {
+        models: {
+          customer: "Custom"
```
nickcharlton, line 71: What do you think of naming the overridden name something like "Custom name"? At first glance, I thought this was a mistake in translations, rather than it being set by you

spec/lib/administrate/resource_resolver_spec.rb
```diff
@@ -61,17 +63,39 @@ module Library; class Book; end; end
   end
 
   describe "#resource_title" do
+    # rubocop:disable Rails/ApplicationRecord
     it "handles global-namepsace models" do
+      class User < ActiveRecord::Base; self.abstract_class = true; end
```
nickcharlton, line 68: What's stopping us from using `ApplicationRecord` here?

betelgeuse: It’s not in scope when just running the file. Could be required of course.

spec/features/edit_page_spec.rb
betelgeuse: Sure

betelgeuse:
Concerning the labels this is more of a bug fix than a feature.

nickcharlton:
Thanks for opening this! I've merged it now.

Re: labels: I do think anything with relating to translations is an issue which needs to be solved, but in this case, we weren't explicitly trying to translate the flash messages. If we were and it were broken, it would've been a bug. Is it a bit arbitrary? yeah, but I came up with them!

flipsasser:
Ahoy hoy! Hi my name is Flip; I'm a big fan of Administrate.

@nickcharlton out of curiosity - what is the release cycle? I'd really love this patch to make its way to rubygems.org.

PS. Thanks @betelgeuse

nickcharlton:
I'm expecting that we'll do a release next before Christmas.

PR #1627, opened 2020-04-29
Fix display resource name plural singular & localisation

### LINKED ISSUE #1072: Text in links not fully translated
In [/app/views/administrate/application/index.html.erb](https://github.com/thoughtbot/administrate/blob/master/app/views/administrate/application/index.html.erb) every links is not translated fully.

For example, when generating a link to create a new resource from a model named SubCategory:
```
  <div>
    <%= link_to(
      t(
        "administrate.actions.new_resource",
        name: page.resource_name.titleize.downcase
      ),
      [:new, namespace, page.resource_path],
      class: "button",
    ) if valid_action?(:new) && show_action?(:new, new_resource) %>
  </div>
```

in french, this will output something like this:
`<a class="button" href="/admin/sub_categories/new">Création sub category</a>`

The word create will be translated correctly, but not the resource name sub category.
The very useful `display_resource_name` helper, which calls `human` which itself calls `translate` could help achieve that.

```
      t(
        "administrate.actions.new_resource",
        name: display_resource_name(resource_name).downcase
      )
```
output in french:
`<a class="button" href="/admin/sub_categories/new">Création sous catégorie</a>`
Hooray!

That way, if our resource name is translated in `your_locale.activerecord.models.your_model`, it will be translated correctly and we can still handle format in `your_locale.administrate.actions.new_resource`.

I haven't had time to check the other views yet, but would you be interested in a pull request? I'm running a similar fix using 0.8.1 in production, it probably won't introduce breaking change on master.

Have a nice day and thanks for this gem, it works great!

betelgeuse:
Fixes two bugs:
- In a couple places plural gets used when it should be singular
- New page link model name can't be translated

Fixes #1072

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -38,10 +38,10 @@ module Administrate
       "#{resource_name.to_s.singularize}_dashboard".classify.constantize
     end
 
-    def display_resource_name(resource_name)
+    def display_resource_name(resource_name, plural:)
       dashboard_from_resource(resource_name).resource_name(
-        count: PLURAL_MANY_COUNT,
-        default: default_resource_name(resource_name),
+        count: plural ? PLURAL_MANY_COUNT : 1,
+        default: default_resource_name(resource_name, plural),
       )
     end
 
@@ -76,8 +76,12 @@ module Administrate
 
     private
 
-    def default_resource_name(resource_name)
-      resource_name.to_s.pluralize.gsub("/", "_").titleize
+    def default_resource_name(resource_name, plural)
+      base = resource_name.to_s
+      if plural
+        base = base.pluralize
+      end
+      base.gsub("/", "_").titleize
     end
   end
 end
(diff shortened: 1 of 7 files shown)


betelgeuse:
Fixes #1072

app/helpers/administrate/application_helper.rb
```diff
@@ -38,10 +38,10 @@ def dashboard_from_resource(resource_name)
       "#{resource_name.to_s.singularize}_dashboard".classify.constantize
     end
 
-    def display_resource_name(resource_name)
+    def display_resource_name(resource_name, plural:)
```
betelgeuse, line 41: This won't be API compatible so when writing CHANGELOG it needs to be noted. Should pull requests already make entries?

nickcharlton: Is there a reasonable way we could make it backwards compatible? Or would that cause more problems than it solves?

betelgeuse: The naming is unexpected at the moment as the return value is plural but indicated in the signature so in my opinion it needs to go eventually. Deprecation can start with just a warning but the error message procuded to the call site with my addition should be clear enough.

nickcharlton: I think that's a good strategy. Do you think you'd be able to add something using [`ActiveSupport::Deprecation`](https://api.rubyonrails.org/classes/ActiveSupport/Deprecation.html)

betelgeuse: Probably

RaeRachael:
Ideas made here for plural/singular option are included in PR1777 (where is it backwards compatible)

Issue #442, opened 2016-01-29
Default order/sort of lists


gylaz:
It would be nice if there was an option to configure the default order that record collection uses when displaying a list on the dashboard.

Something like:

``` rb
class UserDashboard < Administrate::BaseDashboard
  ...
  def default_order
    resources.order(name: :desc)
  end
end
```

FelipePLima:
@gylaz The user would choose the attribute used in default_order method ?

avk:
Is there a way to change the default order now, for all dashboard/models by default?

avk:
I found another way that worked for me, across all dashboards:

```ruby   
module Admin
  class ApplicationController < Administrate::ApplicationController
    before_filter :default_params

    def default_params
      params[:order] ||= "created_at"
      params[:direction] ||= "desc"
    end
  end
end
```

ianagne:
👍 That is a way better way to set a default order.

allthesignals:
I'm using a gem that handles ordering through a method #roots_and_descendants_preordered (specifically, the closure_tree gem, which has preordering https://github.com/mceachen/closure_tree/issues/38). Setting this as the default scope seemed to throw a SQL error from `collection_presenter.attributes_for(resource)`.
```
default_scope -> { roots_and_descendants_preordered }
```

I wondered if there were a way to set a default order in administrate by calling a class method. Any thoughts? Thank you!

danielricecodes:
very nice @avk.  

It would be nice to be able to sort nulls also.  I tried "DESC NULLS LAST" which is fine in Postgres but I got an error from Administrate. 

```
Direction "DESC NULLS LAST" is invalid. Valid directions are: [:asc, :desc, :ASC, :DESC, "asc", "desc", "ASC", "DESC"]
```

Fair enough Administrate.

rozhok:
In rails 5.1 I've had to change `before_filter` into `before_action`:
```ruby
module Admin
  class ApplicationController < Administrate::ApplicationController
    before_action :default_params

    def default_params
      params[:order] ||= 'id'
      params[:direction] ||= 'desc'
    end
  end
end
```

rozhok:
All of sudden, @avk solution works for one dashboard, but didn't work for another. Sorting by clicking on field names also doesn't working. Dashboards are almost the same.

nickcharlton:
[...]

PR #1635, opened 2020-05-04
Document configuration for default ordering

sedubois:
Ref: https://github.com/thoughtbot/administrate/issues/442#issuecomment-623361129


sedubois:
@pablobm should this be closed in favor of #1644?

nickcharlton:
Yeah, I'm going to close this in favour of #1644 as it adds more and solves the slight weirdness we'd otherwise have with overriding bits which aren't everywhere.

Thanks!

PR #1644, opened 2020-05-12
Add options for default sorting

jbennett:
Provide methods to override the default ordering attribute and direction.

Reference: [#442](https://github.com/thoughtbot/administrate/issues/442#issuecomment-564940944)

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -78,6 +78,22 @@ module Administrate
       redirect_to action: :index
     end
 
+    def default_sorting_attribute
+      :id
+    end
+
+    def default_sorting_direction
+      :asc
+    end
+
+    def sorting_attribute
+      params.fetch(resource_name, {}).fetch(:order, default_sorting_attribute)
+    end
+    
+    def sorting_direction
+        params.fetch(resource_name, {}).fetch(:direction, default_sorting_direction)
+    end
+
     private
 
     helper_method :nav_link_state
@@ -101,10 +117,7 @@ module Administrate
     end
 
     def order
-      @order ||= Administrate::Order.new(
-        params.fetch(resource_name, {}).fetch(:order, nil),
-        params.fetch(resource_name, {}).fetch(:direction, nil),
-      )
+      @order ||= Administrate::Order.new(sorting_attribute, sorting_direction)
     end
 
     def dashboard
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -87,11 +87,17 @@ module Administrate
     end
 
     def sorting_attribute
-      params.fetch(resource_name, {}).fetch(:order, default_sorting_attribute)
+      params.fetch(resource_name, {}).fetch(
+        :order, 
+        default_sorting_attribute
+      )
     end
-    
+
     def sorting_direction
-        params.fetch(resource_name, {}).fetch(:direction, default_sorting_direction)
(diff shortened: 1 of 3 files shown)


spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -37,6 +37,30 @@
...
+      customer1 = create(:customer)
+      customer2 = create(:customer)
+      customers = [customer1, customer2]
+
+      locals = capture_view_locals { get :index }
+      expect(locals[:resources].map(&:id)).to eq customers.map(&:id).sort
+    end
+
+    it "supports default sorting overrides" do
+      allow_any_instance_of(Admin::CustomersController).to(receive_messages(
```
nickcharlton, line 51: Is there a way we can do this without using `allow_any_instance_of`? Perhaps by creating another controller subclass inside the test?

jbennett: I replaced it with an anonymous class. Does this help clarify the test or are you thinking something else?

app/controllers/administrate/application_controller.rb
```diff
@@ -78,6 +78,28 @@ def destroy
...
+      )
+    end
+
+    def sorting_direction
+      params.fetch(resource_name, {}).fetch(
+        :direction,
+        default_sorting_direction,
+      )
+    end
+
```
pablobm, line 102: I think these methods should all be private. In a controller, only actions should be public. Making these private still allows developers to override them in their controllers.

jbennett: Woah, you just blew my mind a bit. I'm pretty sure every language I've worked with keeps private as private even from subclasses. There are some completely unnecessary hoops I have jumped through… /me adds Ruby access control research to TODO list

Yes, I agree, private is preferred, will push an update

pablobm: Yeah, I had a similar one back when I first learned of this. Can't blame you! Thank you for the change.

pablobm:
Uh, I didn't realise that the tests were failing. Any idea of what's going on?

jbennett:
Figured out what was going on. 

Replacing the `@controller` with an anonymous class to stub out the sorting methods with alternative values leaked into other tests since the controller is recycled, not re-instantiated between tests. Moving my test into a `context` and using the `controller()` rspec method isolated my test from the rest and stopped everything from blowing up.

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -37,6 +37,35 @@
...
+        def default_sorting_attribute
+          :name
+        end
+
+        def default_sorting_direction
+          :desc
+        end
+      end
+
+      it "retreives resources in the correct order" do
```
betelgeuse, line 61: typo retreives --> retrieves

pablobm: Good catch!

pablobm, line 61: In fact we can use the new "suggestion" feature :-)

```suggestion
      it "retrieves resources in the correct order" do
```

jbennett:
CI failures are unrelated to this change. It's regarding CVE security issue with gems.

pablobm:
@jbennett - Do you mind rebasing on top of master? It should fix the CVE warning.

jbennett:
Rebased and passing 👍🏻

nickcharlton:
Thanks for your work here!

PR #1646, opened 2020-05-13
Allow for label/value setting in Field::Select collection

efatsi:
This mimics the default behavior in Rails `select` helpers, where you can supply an array of length-2-arrays, and each collection item is understood as `[label, value]`.

### DIFF
--- a/spec/features/products_form_spec.rb
+++ b/spec/features/products_form_spec.rb
@@ -37,13 +37,13 @@ describe "product form has_one relationship" do
 
   it "respects setting of label/value in Field::Select" do
     old_release_year = ProductDashboard::ATTRIBUTE_TYPES[:release_year]
-
-    ProductDashboard::ATTRIBUTE_TYPES[:release_year] = Administrate::Field::Select.with_options(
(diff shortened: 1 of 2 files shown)


spec/features/products_form_spec.rb
```diff
@@ -35,6 +35,24 @@
...
 
+  it "respects setting of label/value in Field::Select" do
+    old_release_year = ProductDashboard::ATTRIBUTE_TYPES[:release_year]
+
+    ProductDashboard::ATTRIBUTE_TYPES[:release_year] = Administrate::Field::Select.with_options(
+      collection: [
+        ["Last Year", 2019],
+        ["This Year", 2020]
+      ],
+    )
```
efatsi, line 46: this is a bit of a hacky way to test out the new functionality. Let me know if you'd prefer I add some arbitrary functionality to the example app to avoid the use of temporarily resetting this attribute type.

bbbyyyy: ["Last Year", 2019],

PR #1648, opened 2020-05-14
Recursively process nested parameters for polymorphism

efatsi:
- see details in https://github.com/nickcharlton/administrate-field-nested_has_many/issues/38

Not sure what the best way to test this would be. I could manually build a parameter hash to simulate that which comes in from NestedHasMany? Or perhaps this is straightforward enough to not require specific testing?

### DIFF
--- a/spec/controllers/admin/log_entries_controller_spec.rb
+++ b/spec/controllers/admin/log_entries_controller_spec.rb
@@ -26,6 +26,52 @@ describe Admin::LogEntriesController, type: :controller do
         post_create
         expect(response).to redirect_to([:admin, LogEntry.last])
       end
+
+      it "parses nested polymorphic resources" do
+        customer = create(:customer)
+
+        resource_params = attributes_for(:log_entry).merge(
+          arbitrarily: {
+            nested: {
+              params: {
+                logeable: {
+                  type: "Administrate::Field::Polymorphic",
+                  value: customer.to_global_id.to_s,
+                },
+              },
+            },
+          },
+        )
+
+        allow_any_instance_of(
+          LogEntryDashboard,
+        ).to receive(:permitted_attributes).and_return(
+          [
+            arbitrarily: {
+              nested: {
+                params: {
+                  logeable: [
+                    :type,
+                    :value,
+                  ],
+                },
+              },
+            },
+          ],
+        )
+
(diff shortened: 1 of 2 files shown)


nickcharlton:
It'd be good to have a test here. I'd suggesting going ahead with a manually built parameter hash — I can't think of a better plan.

efatsi:
@nickcharlton added a spec here, verified that it failed prior to the change in ApplicationController and passes now. Let me know what you think.

spec/controllers/admin/log_entries_controller_spec.rb
```diff
@@ -26,6 +26,55 @@ def post_create(action: "create", logeable: create(:customer))
...
+              }
+            }
+          ]
+        )
+
+        begin
+          post :create, log_entry: resource_params
+        rescue ActiveModel::UnknownAttributeError
+          # Ignore errors that arise because the parameters are contrived to
+          # test that nested polymorphic parameters are properly converted
```
nickcharlton, line 67: Does this happen with this test? I wonder if there's a way around it if so

efatsi: This does happen, because the params that are being used to attempt to create a LogEntry record are fictitious and don't align with viable attributes. I'll try tossing spinning up an `attr_accessor` on the model, my first thought was to mock in the entire NestedHasMany logic to make those params valid but now that you ask I think there's probably a quick workaround.

efatsi: @nickcharlton `attr_accessor` worked great, pushed a commit up for that + hound fixes. Only hound stuff left looks ignorable (I think).

spec/controllers/admin/log_entries_controller_spec.rb
```diff
@@ -26,6 +26,52 @@ def post_create(action: "create", logeable: create(:customer))
...
+                  ],
+                },
+              },
+            },
+          ],
+        )
+
+        LogEntry.attr_accessor :arbitrarily
+
+        post :create, log_entry: resource_params
```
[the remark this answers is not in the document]
nickcharlton: I'm happy to ignore this one.

nickcharlton:
Thanks for your work on this!

PR #1649, opened 2020-05-14
Use singular resource name for error explanation

efatsi:
Just a little thing I noticed.

Was:
Screen Shot 2020-05-14 at 11 40 06 AM

Now:
Screen Shot 2020-05-14 at 11 43 17 AM

<hr>

Not sure if this is too much of a bandaid approach here. Was imagining potentially passing in the `count` option to `display_resource_name` from the error partial view code, but if you haven't defined a translation, then the `default` string that's constructed is what ends up being returned. Given that the default comes from `default_resource_name` (which pluralizes the model name), the count is often ignored for the plural version.

Thoughts on if this is good enough or if a larger refactor around how `display_resource_name` is used is called for?


nickcharlton:
Yeah — I think this is fine for now and ensures it reads properly. Thanks!

PR #1651, opened 2020-05-18
Adding 'Back to app' label to internationalization

### LINKED ISSUE #1622: Back to app not translated
* What were you trying to do?
  Use administrate with locale different from English
  Not being able to translate "Back to app"
  master of administrate

The code added in #1376 is missing i18n keys.

edimossilva:
'Back to app' label was hardcoded and now is on internationalization files

fix #1622


pablobm:
@nickcharlton - You were looking into i18n a bit more than me. Do you have any further comments?

nickcharlton:
Thanks!

PR #1655, opened 2020-05-18
Detect enum fields as Selects rather than as Strings

### LINKED ISSUE #1629: Enum fields should be generated as Field::Select instead of Field::String
* What were you trying to do?
  Run generator for a ActiveRecord model with an enum column.
  Dashboard with Field::String
  - Rails: 6.0
  - administrate: master

The generated field should be like this

```ruby
enum_attribute: Field::Select.with_options(collection: Model.enum_attributes.keys)
```

With Field::String submitting to the update action with invalid value results in a Server Error of ArgumentError in Admin::ModelController#update

tadeusz-niemiec:
Fixes #1629
Improve enum field definition in the dashboard generator

### DIFF
--- a/lib/generators/administrate/dashboard/dashboard_generator.rb
+++ b/lib/generators/administrate/dashboard/dashboard_generator.rb
@@ -1,13 +1,12 @@
 require "rails/generators/named_base"
-
 module Administrate
   module Generators
     class DashboardGenerator < Rails::Generators::NamedBase
       ATTRIBUTE_TYPE_MAPPING = {
+        # enum is handled in enum_type_definition method
         boolean: "Field::Boolean",
         date: "Field::Date",
         datetime: "Field::DateTime",
-        enum: "Field::String",
         float: "Field::Number",
         integer: "Field::Number",
         time: "Field::Time",
@@ -16,7 +15,7 @@ module Administrate
       }
 
       ATTRIBUTE_OPTIONS_MAPPING = {
-        enum: { searchable: false },
+        # enum is handled in enum_type_definition method
         float: { decimals: 2 },
       }
 
@@ -39,11 +38,17 @@ module Administrate
         destination = Rails.root.join(
           "app/controllers/#{namespace}/#{file_name.pluralize}_controller.rb",
         )
-
         template("controller.rb.erb", destination)
       end
 
       private
+
+      def enum_type_definition(attr)
+        options = ["searchable: false",
+                   "collection: #{klass.name}.#{attr.pluralize}.keys"]
+
+        "Field::Select.with_options(#{options.join(', ')})"
+      end
 
       def namespace
         options[:namespace]
@@ -76,8 +81,9 @@ module Administrate
 
       def field_type(attribute)
         type = column_type_for_attribute(attribute.to_s)
-
-        if type
+        if type == :enum
+          enum_type_definition(attribute)
+        elsif type
           ATTRIBUTE_TYPE_MAPPING.fetch(type, DEFAULT_FIELD_TYPE) +
             options_string(ATTRIBUTE_OPTIONS_MAPPING.fetch(type, {}))
         else
--- a/lib/generators/administrate/dashboard/dashboard_generator.rb
+++ b/lib/generators/administrate/dashboard/dashboard_generator.rb
@@ -147,7 +147,7 @@ module Administrate
       def proc_string(value)
         source = value.source_location
         proc_string = IO.readlines(source.first)[source.second - 1]
-        proc_string[/(->.*?){(.*?)}/]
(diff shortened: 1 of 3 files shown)


tadeusz-niemiec:
This fails due to the vulnerabilities found in 6.0.2.2 version of some action gems, not sure if it's ok to bump them in this PR.

stevenbuccini:
Thanks for putting this up, just ran into this problem and it was really nice to see that someone had taken the time to make it better.

pablobm:
Don't worry about the vulns. We just upgraded some gems, so a rebase might be enough.

pablobm:
Regarding the PR, I think it could do the trick, although here's another potential idea.

Since https://github.com/thoughtbot/administrate/pull/1589 (not get released in the gem) it's possible to pass a `call`-able as `collection`, for example a lambda. Therefore I wonder if it would be possible to use `ATTRIBUTE_TYPE_MAPPING`, and then `ATTRIBUTE_OPTIONS_MAPPING` with something that generates the values.

There's one problem I see: at the moment, `call` doesn't pass any arguments, so it's not possible to know anything about the field inside the lambda. I think it would be reasonable to pass the field instance (at https://github.com/thoughtbot/administrate/blob/204e0ab225491b0197732764582856374c35aff3/lib/administrate/field/select.rb#L18), and then have the field be something like this:

```ruby
state: Field::Select.with_options(
  collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys) },
)
```

OK, I'm not 100% sure of what I'm saying... The good thing about your approach is that it impacts only the generator, so no interfaces have to change. I have to think about this.

Thoughts anyone? /cc @nickcharlton

tadeusz-niemiec:
@pablobm alright, I think I figured it out. The only problem here is that it's pretty tricky to fetch Proc's definition as a String. The easiest way is to read the definition line and parse it, but this forces us to define Procs in one line (or write a really good parser for multi-lines). The other solution would be to add external gem for it, but it's additional dependency.

Please let me know what you think about this before I proceed with satisfying the hound bot :D

spec/generators/dashboard_generator_spec.rb
```diff
@@ -130,28 +130,52 @@ class InventoryItem < ApplicationRecord
...
-          expect(attrs[:status]).
-            to eq(Administrate::Field::String.with_options(searchable: false))
-        ensure
-          remove_constants :Shipment, :ShipmentDashboard
+        expect(attrs[:status].deferred_class).to eq(Administrate::Field::Select)
+      ensure
+        remove_constants :Shipment, :ShipmentDashboard
+      end
+
+      it "properly handles collection procs option in the 'Select' field" do
```
nickcharlton, line 154: We can drop "properly" on this descriptor.

lib/generators/administrate/dashboard/dashboard_generator.rb
```diff
@@ -16,7 +15,9 @@ class DashboardGenerator < Rails::Generators::NamedBase
       }
 
       ATTRIBUTE_OPTIONS_MAPPING = {
-        enum: { searchable: false },
+        # procs must be defined in one line!
+        enum: {  searchable: false,
+                 collection: ->(field) { field.resource.class.send(field.attribute.to_s.pluralize).keys } },
```
nickcharlton, line 20: It's a little thing, but I wonder if we can use the `lamdba do; end` syntax here to achieve the same goal?

lib/generators/administrate/dashboard/dashboard_generator.rb
```diff
@@ -1,13 +1,12 @@
 require "rails/generators/named_base"
-
```
nickcharlton, line 2: Can you add the the deleted line back in?

lib/generators/administrate/dashboard/dashboard_generator.rb
tadeusz-niemiec: It all depends on how good the parser will be, because I'm fetching raw file lines there and check them using regexp. We'll see how this idea works out and I'll try to implement parser for all possible syntaxes if it's the way to go.

tadeusz-niemiec:
> OK, I think I prefer your approach to what I mentioned. Your change only affects generators and don't change the internal APIs. It'll safer particularly as we are starting to think about changing those APIs and we'll want to be careful about it.

Do you mean that I should revert this to the previous version? Because actually it does what you mentioned, but imho it's pretty safe - the only difference in the select field is one ternary operator to check if proc requires any arguments.

btw some vulns were found

nickcharlton:
Kaminari's update was merged in #1659 so if you rebase against current `master` it'll solve that vulnerability.

PR #1660, opened 2020-05-29
Add Field::RichText

sedubois:
This is required for models with ActionText `has_rich_text`. I thought it would be good to have this in Administrate itself as it's a core Rails feature and it's so small it's maybe not something for a separate gem?


nickcharlton:
I don't have a good rule on what should be in Administrate and what shouldn't be as it comes to field plugins. I've generally felt like it should cover the basic data types and push you to pull in others from other gems, but that perhaps is and has been a bit arbitrary based on what _I_ consider a basic data type.

Anyway, I guess this is needed and like you say, it is small.

Can you add a some tests for using the field?

lib/administrate/field/rich_text.rb
```diff
@@ -0,0 +1,11 @@
+require "administrate/field/base"
+
+module Administrate
+  module Field
+    class RichText < Administrate::Field::Base
```
nickcharlton, line 5: Would we get _any_ benefit from inheriting from the plain text plugin here, or would that just add more complexity for no benefit?

pablobm:

kaka-ruto:
I think this was helpful to have @pablobm. Can we add it to https://github.com/thoughtbot/administrate/issues/2360 ?

pablobm:
I think this is different from that epic, which is about the general visual design and layout of Administrate, so I wouldn't put it there.

However if you think this is something you think you can pick up, rebase, and add a spec, I would be happy to review it.

littleforest:
I'm going to see if I can finish up this PR.

littleforest:
I'm trying to write a spec for `Administrate::Field::RichText`, and within it trying to create a RichText object that I can pass: `action_text = ::ActionText::RichText.new(body: "<trix><p>Foo</p></trix>")`, but I get `uninitialized constant ActionText::RichText`. I have tried to add `require "action_text/engine"` (among many other library attempts), but nothing works. Anyone have any ideas?

kaka-ruto:
@littleforest thanks for taking this up! I don't think Administrate has defined any `ActionText::RichText` class anywhere, so unless you have the class somewhere, calling it will not work, you will want to create that class if it is needed. `require "action_text"` won't work as well as Adminstrate doesn't know what that is or where to get it from

littleforest:
@pablobm thanks for your thoughts on this! I got it to work by putting `require "action_text/engine"` in `lib/administrate/engine.rb`.

PR #1663, opened 2020-06-03
Punditize add support for namespaced models and base policy fallbacks

### LINKED ISSUE #1332: Authorization with Pundit doesn't allow Policy Namespacing

sedubois:
See https://github.com/thoughtbot/administrate/pull/1591#issuecomment-638218493

Fixes #1332.


pablobm:
I think this is good. I haven't used this that much, but I think that `policy_admin` should be removed, as the same functionality can be achieved more naturally from a controller. Unless I'm missing something...?

One problem though is that some people may rely on `policy_admin`. We should still support it, with a warning, and then remove it in a future version (not the next version).

pablobm:
@sedubois - As mentioned above, would you be able to add back `policy_scope_admin`, making it issue a deprecation warning on usage?

nickcharlton:
@sedubois, re: @pablobm's comment above, would you be able to do that? I think rebasing too will get rid of the CI errors as those are things I've seen before. Then we should be good to go.

sedubois:
@pablobm @nickcharlton I'm still happy to get back to this and a couple of other PRs but I'm a bit swamped ATM. I'd suggest leaving on my plate unless of course someone feels like taking over.

nickcharlton:
That's perfect, thanks! I'll leave it to you to get back to when you can.

pablobm:
Looks like there's renewed interest on this PR? If someone picks this up and implements the proposed tweaks above, that would be great. For an example of how to issue warnings within Administrate, see: https://github.com/thoughtbot/administrate/blob/master/lib/administrate.rb

pablobm:

Issue #1671, opened 2020-06-09
Time fields default to 20:00:00


Pezmc:
* What were you trying to do?

Due to datetimepicker, time fields in Administrate currently default to 20:00:00

* What did you end up with (logs, or, even better, example apps are great!)?

How to reproduce:

```
rails new AdministrateTimeBug
```

Add administrate to `Gemfile`

```
gem 'administrate', git: 'https://github.com/thoughtbot/administrate', branch: 'master'
```

Then:

```
bundle install
rails generate model ObjectWithTime start_time:time end_time:time other_time:time
bundle exec rake db:migrate
rails generate administrate:install
```

Use administrate to create a new instance of the model with different times.
Observe that the show page shows correct times.
Try to edit the model, note they all show 20:00:00.

Example app: https://github.com/Pezmc/AdministrateTimeBug

![image](https://user-images.githubusercontent.com/507155/84131245-36a89900-aa3c-11ea-900b-2cd4450eaf23.png)

![image](https://user-images.githubusercontent.com/507155/84131264-3c9e7a00-aa3c-11ea-8712-0af94491fa6a.png)

* What versions are you running?
  - Rails (6.0.3.1)
  - administrate/maser@246175b5ee49fc6b059b100079898db853976a3c (0.13.0)

Pezmc:
Checking the source, you can see that server-side, the time fields values are set correctly:

```
<div class="field-unit field-unit--time field-unit--optional">
      
<div class="field-unit__label">
  <label for="object_with_time_start_time">Start time</label>
</div>
<div class="field-unit__field">
  <input data-type="time" type="text" value="2000-01-01 05:00:00 UTC" name="object_with_time[start_time]" id="object_with_time_start_time" />
</div>

    </div>
    <div class="field-unit field-unit--time field-unit--optional">
      
<div class="field-unit__label">
  <label for="object_with_time_end_time">End time</label>
</div>
<div class="field-unit__field">
  <input data-type="time" type="text" value="2000-01-01 06:00:00 UTC" name="object_with_time[end_time]" id="object_with_time_end_time" />
</div>

    </div>
    <div class="field-unit field-unit--time field-unit--optional">
      
<div class="field-unit__label">
  <label for="object_with_time_other_time">Other time</label>
</div>
<div class="field-unit__field">
  <input data-type="time" type="text" value="2000-01-01 07:00:00 UTC" name="object_with_time[other_time]" id="object_with_time_other_time" />
</div>
```

[...]

PR #1672, opened 2020-06-09
Fix time fields default to 8pm

Pezmc:
Fixes https://github.com/thoughtbot/administrate/issues/1671
Root cause explained on the issue: https://github.com/thoughtbot/administrate/issues/1671#issuecomment-641168322

For the same app as shown in that issue, the time fields now display and update as you'd expect:

image

image

You can check that out here: https://github.com/Pezmc/AdministrateTimeBug/tree/bugfix

Tests: Given this is effectively a front-end bug, and there are no `Field::Time`'s in the example-app (and few tests against that app), I wasn't clear where to add test coverage. I could add a `spec/administrate/views/fields/time/_form_spec.rb` that asserts the value is formatted as `HH:MM:SS`, but it doesn't feel like that's asserting the right thing.

### DIFF
--- a/app/views/fields/time/_form.html.erb
+++ b/app/views/fields/time/_form.html.erb
@@ -18,5 +18,5 @@ By default, the input is a select field for the time attributes.
   <%= f.label field.attribute %>
 </div>
 <div class="field-unit__field">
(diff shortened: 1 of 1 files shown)


pablobm:
I think this is good. Thank you!

Issue #1661, opened 2020-05-31
Always render flashes div so you can insert HTML via Javascript


stevenbuccini:
### Scenario
Within my table row, I have a button that fires an asynchronous JS request. I want to be able to notify the user about the result of this request. 

Currently, this div is only rendered into the DOM when flashes exist. However, you might have a scenario in which no flash messages are sent down from the server but you still would like to prompt flashes programmatically from the frontend.

### Proposal
Always render in the flash partial into the DOM. When empty, it doesn't seem to affect the page layout at all so even conditionally adding `display: none` based on flash message absence shouldn't be necessary. Clients can then hook into this to insert their own HTML flashes as desired.

Happy to submit a PR for this if you're okay with this change.

pablobm:
I think your use case is valid. I wonder though: would it be possible to do this by providing your own `app/views/administrate/application/_flashes.html.erb`? (I haven't tried; just asking to probe the usability of different approaches).

stevenbuccini:
It's totally possible. I'm currently using that work around right now. My thinking was that that it seemed like overkill to generate an `application.html.erb` layout just to insert this empty div when one was already (sometimes) available in the layout. Additionally, putting flashes in the same place would allow for styling consistency. Agree it's a minor nit in the grand scheme of things which is why I wanted to see what other folks thought.

pablobm:
Oh, right. I think it's not necessary to generate the whole thing. Try provide only a `app/views/administrate/application/_flashes.html.erb`. No need for the other files if you only want to change the flash partial. I just had a quick look and was able to make it work.

Please let us know if this works for you.

stevenbuccini:
That's a fair solution. Thanks.

pablobm:
I think that the docs could explain this a bit better. I have created a PR to address this: https://github.com/thoughtbot/administrate/pull/1679

PR #1679, opened 2020-06-18
Any view can be replaced, not only from generators

pablobm:
Inspired by https://github.com/thoughtbot/administrate/issues/1661, I thought this part of the docs could explain a bit more.

### DIFF
--- a/docs/customizing_page_views.md
+++ b/docs/customizing_page_views.md
@@ -16,7 +16,7 @@ Administrate's own.
 Figuring out which views are available and where can be repetitive. You can
 spare yourself some effort by using the built-in view generators.
 
-[1](https://github.com/thoughtbot/administrate/tree/master/app/views).
+[1]: https://github.com/thoughtbot/administrate/tree/master/app/views
 
 ## Customizing for all resources
 
--- a/docs/customizing_page_views.md
+++ b/docs/customizing_page_views.md
@@ -7,7 +7,7 @@ This way you can change the appearance of any page or element of
 the interface.
 
 In general, you can override any of the views under Administrate's
-[/app/views](https://github.com/thoughtbot/administrate/tree/master/app/views).
+[/app/views][1].
 For example, say that you want to customize the template used for flash
 messages. You can provide your own as
 `/app/views/administrate/application/_flashes.html.erb`, and it will replace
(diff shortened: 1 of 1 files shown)


docs/customizing_page_views.md
```diff
@@ -2,12 +2,22 @@
...
+
+In general, you can override any of the views under Administrate's
+[/app/views](https://github.com/thoughtbot/administrate/tree/master/app/views).
+For example, say that you want to customize the template used for flash
+messages. You can provide your own as
+`/app/views/administrate/application/_flashes.html.erb`, and it will replace
+Administrate's own.
+
+Figuring out which views are available and where can be repetitive. You can
+spare yourself some effort by using the built-in view generators.
```
nickcharlton, line 17: What do you think about using a Markdown reference link in this section to make it a bit easier to read?

```suggestion
[/app/views][1].
 For example, say that you want to customize the template used for flash
 messages. You can provide your own as
 `/app/views/administrate/application/_flashes.html.erb`, and it will replace
 Administrate's own.

 Figuring out which views are available and where can be repetitive. You can
 spare yourself some effort by using the built-in view generators.
 
 [1]: https://github.com/thoughtbot/administrate/tree/master/app/views
```

pablobm: Not sure if the specific suggestion (as in: the diff) that you propose is the one you intended? The idea is good though. I just pushed a change. Let me know whether I understood correctly or not.

nickcharlton: Ah yeah, sorry the diff ended up being really hard to understand. That is what I meant!

docs/customizing_page_views.md
```diff
@@ -2,12 +2,24 @@
...
+[/app/views][1].
+For example, say that you want to customize the template used for flash
+messages. You can provide your own as
+`/app/views/administrate/application/_flashes.html.erb`, and it will replace
+Administrate's own.
+
+Figuring out which views are available and where can be repetitive. You can
+spare yourself some effort by using the built-in view generators.
+
+[1](https://github.com/thoughtbot/administrate/tree/master/app/views).
```
nickcharlton, line 19: For reference links we don't need the brackets, ala:

```suggestion
[1]: https://github.com/thoughtbot/administrate/tree/master/app/views
```

pablobm: Oopsie. Fixed now :+1:

nickcharlton:
Thanks! I'll merge this now.

Issue #1643, opened 2020-05-11
Destroy link does not work


sedubois:
## What were you trying to do?

I want to destroy a record using the red "Destroy" link on the index view.

* What did you end up with (logs, or, even better, example apps are great!)?

Instead it simply shows the record, as if a GET was sent instead of DELETE, and so there is no way to delete the record. 

* What versions are you running?
  - Rails 6.0.3
  - administrate 3a438ebefae7d342cbf4b6b7e2d367987573b87e

nickcharlton:
Is this perhaps related to removing jquery-ujs in https://github.com/thoughtbot/administrate/pull/1618?

My understanding is that that's what ensures the correct request occurs, but I'm sort of surprised it doesn't fallback

sedubois:
@nickcharlton I confirm that by switching back to 34686446c54ba296e03f67adf44eb828a4c02d1e the issue disappears.

nickcharlton:
Damn. If you make sure `rails-ujs` is loaded for the Administrate Engine, does that help? I'm wondering if it's a Rails Engine compatibility issue.

I'm also surprised we didn't catch this on CI, so that's something to look at too.

sedubois:
@nickcharlton I'm quite illiterate with this rails-ujs thing and rails engines, would you mind giving a bit more guidance?

sedubois:
BTW (TL = Turbolinks),

> The need for UJS is gone with TL6

> Plan is to have TL6 out ~late summer/early fall

https://discuss.rubyonrails.org/t/turbolinks-doesnt-handle-render/74772/3
https://discuss.rubyonrails.org/t/turbolinks-doesnt-handle-render/74772/5

Maybe the change to rails UJS imports can be reverted and then upgrade to TL6 when available?

nickcharlton:
I think we might just about know as much as each other here, but it seems like we expected the `ujs` removal to not be a problem as it was pulled in elsewhere so no longer needed. As for Rails engines, this sounds very similar to issues we've been having with the asset pipeline/webpacker and needing to do extra work to pull in dependencies.

Can you try on Rails 6.1? (or, perhaps replicate it with a failing test so we can see it on CI?)

sedubois:
I don't have time to dig deeper with Rails 6.1 right now but for what it's worth I've commented out `turbolinks.start()` in my admin JS pack and the issue persists.

pablobm:
I looked into why this test passes with rack-test. I really didn't expect what I found...

[...]

PR #1688, opened 2020-06-25
WIP: fix "destroy" links

pablobm:
Fixes https://github.com/thoughtbot/administrate/issues/1643

When https://github.com/thoughtbot/administrate/pull/1618 was merged, some JS broke. One important piece of functionality that stopped working was "destroy" links.

We were using the JS-driven destroy links that are common in Rails. These work on links with the attribute `data-method="delete"`, and show a `confirm()` box with text extracted from the attribute `data-confirm`.

You might be wondering why the specs didn't break... Well, it turns out that Capybara tries to be clever about these special attributes, even when using Rack::Test. As a result, tests passed even though the JS to drive the functionality wasn't working. To avoid this happening again, I have told Capybara to `respect_data_method: false`.

OK, so how can we fix this? We can fix the JS, but there are two problems:

1. #1618 wasn't merged entirely arbitrarily. It was the product of our trying to figure out how to best integrate with both the asset pipeline and webpacker, two competing alternatives for handling frontend assets. We are still trying to figure this one out, and I don't think there's going to be a simple answer.
2. I don't like JS-driven delete buttons anyway! I'm ok with progressive enhancement, but having them depend on JS leads to issues... like this one we had.

So I'm proposing that we change the way that destroy links work, so that they use a "traditional" approach, driven by the backend:

1. Click on "destroy" link.
2. View separate confirmation page.
3. Click "confirm" button.
4. Redirect back to index page.

This is a first draft, which perhaps should be styled a bit. The confirmation page looks like this at the moment:

Ask if user is sure, provide destroy button

Pending:

* [ ] Discuss: how do we feel generally about this approach?
* [ ] Discuss: this introduces new i18n strings. How do we feel about this? I suspect they should be more more, as the reuse I do here may not work well across languages.
* [ ] Style the new destroy confirmation page.
* [ ] Provide i18n strings for all supported languages.


nickcharlton:
Ahh this is an interesting way to approach this problem!

Is it just deletions that have this problem? We don't do anything else which was dependent on `ujs`?

My initial thought would be that this will make deleting items really slow, and having _something else_ to make mass deletions faster isn't something I think we should worry about right now.

pablobm:
I agree: the UX of this solution is a step backwards compared to the pre-bug solution.

Now, I have these two thoughts:

1. I don't like relying on ujs. Perhaps I'm a bit old-fashioned, but I do appreciated progressive enhancement.
2. I think the JS should still be fixed, precisely to provide progressive enhancement and fix the UX, and even improve from what we had before.

Having said that, perhaps we should fix the JS issue first, and consider this PR as an additional enhancement that should come later.

nickcharlton:
Is the best option — in the interest of unblocking the release — to just revert the removal? I'd like to add the Capybara change from this too though.

pablobm:
Yeah, I think you are right. I'd still like to provide a JS-independent solution, but let's solve the problem at hand first. So here's a PR that includes the breaking test and the commit reversal: https://github.com/thoughtbot/administrate/pull/1691

pablobm:
Oops, didn't notice you already had created a PR. I closed mine now.

nickcharlton:
Sorry! I should've mentioned it!

Given other priorities, what do you think of closing this one for now and revising it later on?

pablobm:
Yes. I'll come back to this in the future, but I'm ok with closing it for now.

PR #1694, opened 2020-06-30
I had a name change.

thecalliopecrashed:


nickcharlton:
Oh! Exciting!

We also reference your deadname in the Gemspec: https://github.com/thoughtbot/administrate/blob/master/administrate.gemspec#L8

I imagine you have some references in some of the fields too, I'll take a look at the ones I keep an eye on.

PR #1698, opened 2020-07-06
Allow fetching special files in the documentation

### LINKED ISSUE #1693: Unable to bundle on Windows 10 because of a symlink.
**Pre-conditions** (platform setup):

1. OS: Windows 10 Pro
2. IDE: Atom 2020 06 29 12
3. Rails version: Rails 6.0.3.2
4. Ruby version: ruby 2.6.6p146 (2020-03-31 revision 67876) [x64-mingw32]
5. Administrate version: administrate 0.13.0

**Steps to repro:**
1. Create a rails app. (having one or more models), open the app., then
2. Open its Gem file
3. Insert the line: gem ‘Administrate’ to the last line of the file
4. Open terminal and run: bundle install
    
**Expected behaviour:**
1. Administrate Gem is successfully bundled.
    
**Actual behaviour:**
1. User is presented with the error-message: 

Installing administrate 0.13.0
_Errno::EACCES: Permission denied @ rb_file_s_symlink - (../CONTRIBUT
TING.md,
C:/Ruby26-x64/lib/ruby/gems/2.6.0/gems/administrate-0.13.0/docs/cont
tributing.md)
An error occurred while installing administrate (0.13.0), and Bundle
er cannot
continue.
Make sure that `gem install administrate -v '0.13.0' --source
'https://rubygems.org/'` succeeds before bundling.

In Gemfile:
  administrate_
(see youtube link: https://youtu.be/n-80J7F1spk )

EWright212:
In #1514, we added a symlink so that the docs would show up on the demo app …but
symlinks aren't supported on Windows and was causing a problem when trying to bundle.

This instead adds a lookup table of files not in docs/ which allows us to render those.

This partially fixes #1693, in allowing Administrate itself to be bundled without a symlink but
other dependencies have the same problem.

### DIFF
--- a/spec/example_app/app/controllers/docs_controller.rb
+++ b/spec/example_app/app/controllers/docs_controller.rb
@@ -16,7 +16,7 @@ class DocsController < ApplicationController
   end
 
   def show
-    if find_special_file != nil 
+    if !find_special_file.nil?
       render_page find_special_file[:file]
     else
       render_page "docs/#{params[:page]}"
--- a/spec/example_app/app/controllers/docs_controller.rb
+++ b/spec/example_app/app/controllers/docs_controller.rb
@@ -27,7 +27,7 @@ class DocsController < ApplicationController
 
   def render_correct_page
     if !find_special_file.nil?
-      render_page find_special_file[:file]
+      render_page(find_special_file[:file])
     else
       render_page "docs/#{params[:page]}"
     end
--- a/spec/example_app/app/controllers/docs_controller.rb
+++ b/spec/example_app/app/controllers/docs_controller.rb
@@ -15,10 +15,6 @@ class DocsController < ApplicationController
     autolink: true,
   }.freeze
 
-  # def index
-  #   render_page "README"
-  # end
-
   def show
     render_correct_page
   end
@@ -35,14 +31,8 @@ class DocsController < ApplicationController
 
   def render_correct_page
     if !find_special_file.nil?
-      p find_special_file
       render_page find_special_file[:file]
-    # elsif params[:page] == nil
-    #     p params
-    #   render_page "README"
     else
-      p "here"
-      p params
       render_page "docs/#{params[:page]}"
     end
   end
--- a/spec/example_app/app/controllers/docs_controller.rb
+++ b/spec/example_app/app/controllers/docs_controller.rb
@@ -4,7 +4,10 @@ class DocsController < ApplicationController
     {
       file: "CONTRIBUTING",
       page: "contributing",
-    }.freeze
+    }, {
+      file: "djsfhkjdh",
+      page: "ksfhalk",
+    }
 
   ]
 
@@ -18,9 +21,11 @@ class DocsController < ApplicationController
   end
 
   def show
-    # First check if page in special files, if so
-    if (params[:page] == SPECIAL_FILES[0][:page])
-      render_page SPECIAL_FILES[0][:file]
+
+    p  SPECIAL_FILES.select { |page| page[:page] == params[:page]}
+
+    if SPECIAL_FILES.select { |page| page[:page] == params[:page]}.length > 0 
+      render_page SPECIAL_FILES.select { |page| page[:page] == params[:page]}[0][:file]
     else
       render_page "docs/#{params[:page]}"
     end
--- a/spec/example_app/app/controllers/docs_controller.rb
+++ b/spec/example_app/app/controllers/docs_controller.rb
@@ -16,11 +16,7 @@ class DocsController < ApplicationController
   end
 
   def show
-    if !find_special_file.nil?
-      render_page find_special_file[:file]
-    else
-      render_page "docs/#{params[:page]}"
-    end
+    render_correct_page
   end
 
   def find_special_file
@@ -28,6 +24,14 @@ class DocsController < ApplicationController
   end
 
   private
+
+  def render_correct_page
+    if !find_special_file.nil?
+      render_page find_special_file[:file]
+    else
+      render_page "docs/#{params[:page]}"
+    end
+  end
 
   def render_page(name)
     path = full_page_path(name)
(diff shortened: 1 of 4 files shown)


spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,7 +16,15 @@ def index
...
-    render_page "docs/#{params[:page]}"
+    if check_special_file.length > 0 
+      render_page check_special_file[0][:file]
+    else
+      render_page "docs/#{params[:page]}"
+    end
+  end
+
+  def check_special_file
+    SPECIAL_FILES.select { |page| page[:page] == params[:page] }
```
nickcharlton, line 27: If we were to use [`first`][1], rather than `select`, we'd get back the first result only, which would avoid us needing to call `render_page check_special_file[0]` up on line 20. What do you think?

[1]: https://ruby-doc.org/core-2.6.3/Array.html#method-i-first

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,7 +17,15 @@ def index
   end
 
   def show
-    render_page "docs/#{params[:page]}"
+    if check_special_file.length > 0
```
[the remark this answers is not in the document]
nickcharlton: I'm not sure where the trailing whitespace issue is here 🤔 .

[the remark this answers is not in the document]
nickcharlton: We might want to extract the `if` here into another (private) method so we can do something like:

```ruby
if has_special_file
  render_page
else
  ...
end
```

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,7 +16,15 @@ def index
...
   def show
-    render_page "docs/#{params[:page]}"
+    if check_special_file.length > 0 
+      render_page check_special_file[0][:file]
+    else
+      render_page "docs/#{params[:page]}"
+    end
+  end
+
+  def check_special_file
```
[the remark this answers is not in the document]
EWright212: agreed, changed

spec/example_app/app/controllers/docs_controller.rb
EWright212: Agreed, used `select` then `first` to return one item only

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,11 +16,23 @@ def index
...
+  end
+
+  def find_special_file
+    SPECIAL_FILES.select { |page| page[:page] == params[:page] }.first
   end
 
   private
 
+  def render_correct_page
+    if !find_special_file.nil?
```
nickcharlton, line 29: Instead of the double negative, we could go:

```suggestion
    if find_special_file
```

as if there is a special file, we'll render it and if it's `nil` we'll drop into the `else`.

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,7 +16,15 @@ def index
   end
```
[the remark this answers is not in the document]
nickcharlton: (We decided to do this in another PR.)

spec/example_app/app/controllers/docs_controller.rb
EWright212: Oh obviously, thanks!

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,11 +16,23 @@ def index
...
 
   def show
-    render_page "docs/#{params[:page]}"
+    render_correct_page
+  end
+
+  def find_special_file
+    SPECIAL_FILES.select { |page| page[:page] == params[:page] }.first
```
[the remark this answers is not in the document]
EWright212: I see, thanks!

pablobm:
As mentioned, I've been trying to get this working on Windows, but have failed so far :-/ How did you install things on Windows yourself? I have tried this so far:

* RailsInstaller: latest version of Ruby there is 2.3, and Administrate requires at least 2.4.
* WSL+Ubuntu on Windows: symlinks appear to work.
* RubyInstaller: I can't get `bundle install` to complete, first for lack of symlinks, then due to the gem kgio not being compatible.

So it appears that you installed things in some way that allowed you to run `bundle install`, but yet failed later due to lack of symlinks. Am I correct?

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,11 +16,23 @@ def index
...
 
   private
 
+  def find_special_file
+    SPECIAL_FILES.select { |page| page[:page] == params[:page] }.first
+  end
+
+  def render_correct_page
+    if !find_special_file.nil?
+      render_page find_special_file[:file]
```
nickcharlton, line 30: A tiny one: I have a preference to wrapping method call arguments with a `()` to make it explicit what it's doing:

```suggestion
      render_page(find_special_file[:file])
```

i.e.: here I now know that `find_special_file` is returning a hash of some kind and `render_page` is a method call

spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -9,11 +16,23 @@ def index
...
+    else
+      render_page "docs/#{params[:page]}"
```
nickcharlton, line 32:
```suggestion
      render_page("docs/#{params[:page]}")
```

pablobm:
Right, I had misunderstood something. I thought this was about running the Administrate dev environment, which would require running `bundle install` in the folder of the Administrate gem. However, this is about using Administrate in a different project. I can reproduce this now.

OK, let's try this. I can't really test it properly because the problem doesn't appear if I install the gem from the GitHub repo (`gem 'administrate', git: '...'`) as opposed to from the gem proper (just `gem 'administrate'`). The error does point at that symlink though, so I think it's worth trying this.

PR #1703, opened 2020-07-08
Update Portuguese translation

tiagotex:
Translated the form key from English in the Portuguese and Portuguese from Brazil translation files.
 
back_to_app was using an expression used in Portuguese from Brazil that is not used in Portuguese from Portugal.

### DIFF
--- a/config/locales/administrate.pt-BR.yml
+++ b/config/locales/administrate.pt-BR.yml
@@ -23,7 +23,7 @@ pt-BR:
         none: Nenhum
     form:
       error: erro
-      errors: "%{pluralized_errors} não permitiram guardar este %{resource_name}:"
+      errors: "%{pluralized_errors} impediram %{resource_name} de ser gravado:"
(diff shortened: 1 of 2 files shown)


config/locales/administrate.pt-BR.yml
```diff
@@ -22,8 +22,8 @@ pt-BR:
         more: "Exibindo %{count} de %{total_count}"
         none: Nenhum
     form:
-      error: error
-      errors: "%{pluralized_errors} prohibited this %{resource_name} from being saved:"
+      error: erro
```
nickcharlton, line 25: Is dropping the `r` here deliberate?

tiagotex: yeah, `erro` is the correct translation for `error`

nickcharlton: Oh huh! Cool, thanks!

PR #1718, opened 2020-07-21
Ew replace phantomjs with selenium 2

EWright212:
Updated the discontinued phantomjs with selenuim webdirvers and fixed related tests.
Update of pull request to be based off master branch instead of unmerged pull requests.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -30,7 +30,7 @@ group :test do
   gem "formulaic"
   gem "launchy"
   gem "pundit"
-  gem 'selenium-webdriver'
+  gem "selenium-webdriver"
   gem "shoulda-matchers"
   gem "timecop"
   gem "webmock"
--- a/Gemfile
+++ b/Gemfile
@@ -29,8 +29,8 @@ group :test do
   gem "database_cleaner"
   gem "formulaic"
   gem "launchy"
-  gem "poltergeist"
   gem "pundit"
+  gem 'selenium-webdriver'
   gem "shoulda-matchers"
   gem "timecop"
   gem "webmock"
--- a/spec/features/log_entries_index_spec.rb
+++ b/spec/features/log_entries_index_spec.rb
@@ -53,9 +53,8 @@ feature "log entries index page" do
 
     visit admin_log_entries_path
     accept_confirm do
-      click_on t("administrate.actions.destroy")
+      click_on t('administrate.actions.destroy')
     end
-    
     expect(page).to have_flash(
       t("administrate.controller.destroy.success", resource: "LogEntry"),
     )
--- a/spec/features/log_entries_index_spec.rb
+++ b/spec/features/log_entries_index_spec.rb
@@ -53,7 +53,7 @@ feature "log entries index page" do
 
     visit admin_log_entries_path
     accept_confirm do
-      click_on t('administrate.actions.destroy')
+      click_on t("administrate.actions.destroy")
     end
     expect(page).to have_flash(
       t("administrate.controller.destroy.success", resource: "LogEntry"),
--- a/spec/features/log_entries_index_spec.rb
+++ b/spec/features/log_entries_index_spec.rb
@@ -52,8 +52,10 @@ feature "log entries index page" do
     create(:log_entry)
 
     visit admin_log_entries_path
-    click_on t("administrate.actions.destroy")
-
+    accept_confirm do
+      click_on t("administrate.actions.destroy")
+    end
+    
     expect(page).to have_flash(
       t("administrate.controller.destroy.success", resource: "LogEntry"),
     )
--- a/spec/features/orders_index_spec.rb
+++ b/spec/features/orders_index_spec.rb
@@ -62,7 +62,7 @@ feature "order index page" do
 
     visit admin_orders_path
     accept_confirm do
-      click_on t('administrate.actions.destroy')
+      click_on t("administrate.actions.destroy")
     end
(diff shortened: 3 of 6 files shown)


spec/rails_helper.rb
```diff
@@ -31,14 +31,22 @@ module Features
...
-  Capybara::RackTest::Driver.new(app, respect_data_method: false)
+Capybara.register_driver :headless_chrome do |app|
+  capabilities = Selenium::WebDriver::Remote::Capabilities.chrome(
+    chromeOptions: {
+      args: %w[headless enable-features=NetworkService,NetworkServiceInProcess]
+    }
+  )
+  Capybara::Selenium::Driver.new app,
+                                 browser: :chrome,
+                                 desired_capabilities: capabilities
```
pablobm, line 47: I find this style of indent problematic. It breaks the indent for all lines if the first one is replaced with one of different length. How about writing it like this?

```suggestion
  Capybara::Selenium::Driver.new(
    app,
    browser: :chrome,
    desired_capabilities: capabilities,
)
```

To be honest, I'm not sure if we have a style rule for this, and Hound appears not to be enabled at the moment? @nickcharlton, do you have any thoughts?

nickcharlton: I'm curious about the `tmp/pids/.keep`. I tried to fix the first clone problem you'd get in #1695, do we need to ignore that file too so if you delete it locally it doesn't matter?

Hound did run with this PR. It might've been done when the configuration in `thoughtbot/guides` was deleted. It got put back in again, so we might be alright now. A new push might highlight the changes.

Gemfile
```diff
@@ -29,8 +29,8 @@ group :test do
   gem "database_cleaner"
   gem "formulaic"
   gem "launchy"
-  gem "poltergeist"
   gem "pundit"
+  gem 'selenium-webdriver'
```
nickcharlton, line 33:
```suggestion
  gem "selenium-webdriver"
```

spec/features/log_entries_index_spec.rb
```diff
@@ -52,8 +52,9 @@
     create(:log_entry)
 
     visit admin_log_entries_path
-    click_on t("administrate.actions.destroy")
-
+    accept_confirm do
+      click_on t('administrate.actions.destroy')
```
nickcharlton, line 56:
```suggestion
      click_on t("administrate.actions.destroy")
```

spec/features/orders_index_spec.rb
```diff
@@ -49,8 +49,9 @@
     create(:order)
 
     visit admin_orders_path
-    click_on t("administrate.actions.destroy")
-
+    accept_confirm do
+      click_on t('administrate.actions.destroy')
```
nickcharlton, line 53:
```suggestion
      click_on t("administrate.actions.destroy")
```

spec/features/orders_index_spec.rb
```diff
@@ -60,8 +61,9 @@
     create(:payment, order: create(:order))
 
     visit admin_orders_path
-    click_on t("administrate.actions.destroy")
-
+    accept_confirm do
+      click_on t('administrate.actions.destroy')
```
nickcharlton, line 65:
```suggestion
      click_on t("administrate.actions.destroy")
```

spec/rails_helper.rb
```diff
@@ -6,7 +6,7 @@
 
 require "rspec/rails"
 require "shoulda/matchers"
-require "capybara/poltergeist"
+require 'selenium/webdriver'
```
nickcharlton, line 9:
```suggestion
require "selenium/webdriver"
```

EWright212: Ah thanks for these changes! I actually didn't have them originally but these were changes houndbot wanted so I thought that was the way to go. Thanks for these!

spec/rails_helper.rb
```diff
@@ -6,7 +6,7 @@
 
 require "rspec/rails"
 require "shoulda/matchers"
-require "capybara/poltergeist"
+require "selenium/webdriver"
```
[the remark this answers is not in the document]
nickcharlton: Oh godamnit Hound.

spec/rails_helper.rb
EWright212: @pablobm This indentation style was suggested by Houndbot, as was many of the changes you and Nick asked me to undo in this commit. I like yours better though!

pablobm: Oh, sorry about that. Normally we follow Hound's opinion to the letter, but a couple of weeks ago there was an issue with it. The set of rules changed and I'm not sure they are back to normal yet.

Don't worry then. Go with your preference and we'll come around to fixing Hound in the future.

pablobm:
Mind just adding back that `tmp/pids/.keep` file then?

spec/rails_helper.rb
nickcharlton: Yeah, it’s been really annoying. I don’t much like commenting on style things manually, as even though consistency is good, nit-picking to the nth degree is not enjoyable for anyone!

I think there’s only one or two things left here: everything we can fix on GitHub apart from the `tmp/pids/.keep` file missing.

EWright212: Okay, I think I've fixed all the issues now from the houndbot to the tmp/pids/.keep (which was in the gitignore so wasn't getting pushed). Thanks @nickcharlton

Issue #1709, opened 2020-07-13
Is there a workaround for a model named "Page"?


milkfarm:
I didn't find any documentation about reserved names for the administrate gem. Is there a workaround for a model named `Page`? When I create a standard dashboard (`rails g administrate:dashboard Page`), I encounter a variety of errors. For example, when trying to access the pagination links on the `index` view.

```
NoMethodError in Admin::PagesController#index
undefined method `fetch' for "2":String

def order
  @order ||= Administrate::Order.new(
    params.fetch(resource_name, {}).fetch(:order, "created_at"),
    params.fetch(resource_name, {}).fetch(:direction, "desc")
  )
end
```

The issue relates to the paging/sorting information being stored under the same param key as the resource.

Rails: 6.0.3.2
administrate: 0.13.0

pablobm:
Uh, that's indeed a bug. I have put together some code that might help. Mind trying out branch https://github.com/thoughtbot/administrate/compare/master...pablobm:fix-page and tell us how it went? Thank you.

milkfarm:
The paging links for unsorted results now work. Clicking a table column header to sort records, however, generates the following error message: 

```
NoMethodError in Admin::PagesController#index
undefined method `to_i' for #<ActionController::Parameters:0x00007fec239ddbd8> Did you mean? to_s to_h

Extracted source (around line #17):

15  def self.#{Kaminari.config.page_method_name}(num = nil)
16    per_page = max_per_page && (default_per_page > max_per_page) ? max_per_page : default_per_page
17    limit(per_page).offset(per_page * ((num = num.to_i - 1) < 0 ? 0 : num)).extending do
18      include Kaminari::ActiveRecordRelationMethods
19      include Kaminari::PageScopeMethods
20  end
```

I'm ok renaming my model if you don't think this can be remedied in an elegant manner. I'm also here to help if you'd like to continue investigating. In any regard, thanks!

pablobm:
Thank you again for that diagnosis. I have pushed a new change to that branch. Would you have time to try it out?

The bug should definitely be fixed, and I don't expect it to be a particularly difficult fix.

milkfarm:
You've squashed another part it -- well done. The only problem I see is the paging when the records are sorted. I don't encounter an exception but the behavior isn't correct. The paging links don't retain the sorting. I didn't look at the code but I imagine the "page[:order]" and "page[:direction]" are being overwritten by the "page" param that contains the page number.

Example:
[...]

PR #1725, opened 2020-07-30
Fix pagination of "Page" models

pablobm:
Fixes https://github.com/thoughtbot/administrate/issues/1709

Using the param `:page` for top-level pagination (ie: index pages as opposed to has_many lists in show pages) conflicts with paginating resources whose type happens to be "Page".

I'd rather have written a unit spec for this, instead of a whole new dashboard for the example app. However the pagination code is all over the place making this quite tricky. Ideally, we should move all the pagination code into a single module that can be tested in isolation. This in turn is not easy, as Kaminari doesn't make it easy. I'm leaving that for another day.

I have created a new feature spec called `pagination_spec.rb`. Into this I have moved other pre-existing pagination examples, to help organise things, and also to take advantage of the helper method `expect_to_appear_in_order` in the new example.


stadia:
https://github.com/kaminari/kaminari#the-page-parameter-is-in-paramspage

pablobm:
@stadia - Care to elaborate on what you mean by providing that link?

PR #1731, opened 2020-08-03
Pid inclusion on setup

KOlofinmoyin:
observer that the tmp/pid/server.pid keeps getting omitted on startup of a new repo/ pulling from origin master so included a rule in the gitignore to include the folder whenever there's a pull from master.


nickcharlton:
Did #1726 not help at all?

EWright212:
> Did #1726 not help at all?

Unfortunately no, this new PR works for me as the file missing was tmp/pids/server.pid so we have included that one

nickcharlton:
Closing this as we concluded that whilst there's still an issue around this, we'll not take this approach.

Some of the problem being faced here was a messy clone, which was the origin of this particular issue.

PR #1739, opened 2020-08-13
Adds ability to set the first sort direction on a dashboard.

connor:
I wanted this functionality in a project I'm working on, so decided to submit a little PR to be considered.

Basically, in my project we want the first sort order when sorting a dashboard's column to be `:desc` instead of the default (`:asc`).

This PR allows you to override the `default_first_sort_direction` in an `application_controller` to be `:desc`.

I'm relatively inexperienced with Ruby, so if there is a better way to do some of the things this does, I would love to hear ways to improve it!

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -21,7 +21,6 @@ group :development, :test do
   gem "factory_bot_rails"
   gem "i18n-tasks", "0.9.31"
   gem "pry-rails"
-  gem 'pry-byebug'
 end
 
 group :test do
--- a/Gemfile
+++ b/Gemfile
@@ -21,6 +21,7 @@ group :development, :test do
   gem "factory_bot_rails"
   gem "i18n-tasks", "0.9.31"
   gem "pry-rails"
+  gem 'pry-byebug'
 end
 
 group :test do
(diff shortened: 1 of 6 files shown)


pablobm:
Thank you for this PR. I've been trying to wrap my head about all this section of the code, and I see it's tricky.

Initially, your change would work, but I'm feeling uneasy about `Administrate::Order` getting too many responsibilities. I feel that it shouldn't know how to order as well as what the next option should be, and instead this second concern should be handled... perhaps by the collection...? This concern is already there though, it's not like your PR adds it.

I'm not sure though. A new question here is how to configure this behaviour in tables for fields of type HasMany. The moment we consider this, perhaps the controller shouldn't have been the place for this in the first place.

I need to think about this a bit. I wonder if someone else may have some ideas. Perhaps @nickcharlton?

connor:
Totally understand!

I'm not familiar with the administrate plugin api, but maybe this could be a plugin instead?

pablobm:
Theoretically, it could be done with a plugin, but it would require significant monkeypatching and it could break easily when the affected code changes within Administrate.

nickcharlton:
I'm with @pablobm on it not quite feeling right in `Order`. This feels like maybe it's a Dashboard responsibility, do you have any thoughts on how it could instead go there?

PR #1762, opened 2020-09-13
Better error message if key is not specified in ATTRIBUTE_TYPES

inem:
I was getting:

```
NoMethodError: undefined method `associative?' for nil:NilClass
```

After the change:

```
KeyError: key not found: :created_at
```

### DIFF
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -91,7 +91,7 @@ module Administrate
 
     def attribute_includes(attributes)
       attributes.map do |key|
-        field = self.class::ATTRIBUTE_TYPES.fetch(key)
+        field = attribute_type_for(key)
 
         key if field.associative?
       end.compact
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -91,7 +91,7 @@ module Administrate
 
     def attribute_includes(attributes)
       attributes.map do |key|
-        field = self.class::ATTRIBUTE_TYPES[key]
+        field = self.class::ATTRIBUTE_TYPES.fetch(key)
 
         key if field.associative?
       end.compact


inem:
@nickcharlton @pablobm  @sedubois guys, can someone take a look please? It is a small change but I find it very useful - helps you debug thing when you messed up dashboard by not declaring some attributes.

harley:
i got so used to this error that didn't pay attention, but i wanted to say this is an elegant change @inem - hope it gets merged soon as we're seeing more activities on master again :)

lib/administrate/base_dashboard.rb
```diff
@@ -88,7 +88,7 @@ def attribute_not_found_message(attr)
 
     def attribute_includes(attributes)
       attributes.map do |key|
-        field = self.class::ATTRIBUTE_TYPES[key]
+        field = self.class::ATTRIBUTE_TYPES.fetch(key)
```
pablobm, line 91: Good call; thank you. Since we are at it, how about using `attribute_type_for` instead? That way we also get a more descriptive error than just a `KeyError`.

caljess599:
+1 for this change.

lib/administrate/base_dashboard.rb
inem: Ok, I didn't know what is `attribute_type_for`

Now googled and found that it is an [internal administrate method](https://www.rubydoc.info/gems/semantic-administrate/0.1.0/Administrate/BaseDashboard:attribute_type_for). 

[Kind of done](https://github.com/thoughtbot/administrate/pull/1762/commits/0a050c9014a8ee423bbd528eb8ec0604223f72db), I guess.

inem: Ok, blind commit was not a good idea

pablobm:
Looks good! Could you please rebase so that the build passes? The errors should just be about security advisories in dependencies, but it's worth seeing it green before merging.

pablobm (approved):

PR #1765, opened 2020-09-15
Do not try to convert time if value is not there

Fodoj:


pablobm:
Uh, good catch, thank you. I see that this guard exists for the `DateTime` field but not for this one.

Would you be able to add a spec? I think a view spec would do nicely in this case. You can see an example at https://github.com/thoughtbot/administrate/blob/4207c9d1fe40410738a04a5f40502b14ccbdc7a0/spec/administrate/views/fields/url/_show_spec.rb

nickcharlton:
Thanks for your contribution!

I happened to come across, rebase and merge #1226 just earlier today so I'm going to close this as we've done this now.

PR #1769, opened 2020-09-18
Nederlands translation update

RaeRachael:


nickcharlton:
Thanks!

I think if you rebase against current master the failure will go away; it's a vulnerability warning.

pablobm (approved):

PR #843, opened 2017-04-29
Support for decorating resources

pustomytnyk:
Allows use of decorator pattern (e.g. [Draper](https://github.com/drapergem/draper)). One can define `#decorate_resource` on a dashboard.

- [x] Decoration of resource and collection (pagination friendly)

- [x] Add an example to example app

- [x] Tests for classes which use new methods

PR #1772, opened 2020-09-18
Remove autoprefixer-rails as a gem dependency

nickcharlton:
In #1749, we removed `autoprefixer-rails` from the usual `Gemfile` but it
didn't get removed from the `gemspec`.

PR #1773, opened 2020-09-20
Support for decorating resources - revised

dgmstuart:
Currently it's possible to modify how different types of field  are presented (with [attribute partials](https://administrate-prototype.herokuapp.com/customizing_attribute_partials)), and to modify the title used when displaying a resource (with [`display_resource`](https://administrate-prototype.herokuapp.com/customizing_dashboards)), BUT there's no straightforward way to modify how an individual field is displayed, beyond creating a [custom field type](https://administrate-prototype.herokuapp.com/adding_custom_field_types) for each field (which feels like a bit of a hack).

One use case for this is when a resource has a field which stores some kind of numeric code, and it's desirable to display a text description of the code instead of the code itself. 

This PR is an attempt to resurrect https://github.com/thoughtbot/administrate/pull/843 (big credit to @pustomytnyk). It started with a rebase of that PR, but has been heavily modified - in particular the refactorings required to move responsibility for resources into the `Collection` page have been separated out into their own commits.

The overall change is kind of complicated and touches a lot of files, but **please see the individual commit messages** for a description of the what's and why's of each step.

There are a number of bits of remaining awkwardness and awkward naming which I'd love to get some input on.

TODO: 

- [ ] Document internal API changes
- [ ] Check the first page of reverse dependencies for any breaking changes:
  - [ ] administrate-field-password
  - [ ] administrate-field-active_storage
  - [ ] administrate-field-enum
  - [ ] administrate-field-nested_has_many
  - [ ] administrate-field-belongs_to_search
  - [ ] administrate-field-image
  - [ ] administrate-field-carrierwave
  - [ ] administrate-field-paperclip
  - [ ] administrate-field-date_picker
  - [ ] administrate_exportable
  - [ ] administrate-field-ckeditor
  - [ ] administrate-field-jsonb
  - [ ] administrate-field-money
  - [ ] administrate-field-country
  - [ ] administrate-field-lat_lng
  - [ ] administrate-field-json
  - [ ] administrate-field-simple_markdown
  - [ ] administrate-field-boolean_to_yes_no
  - [ ] administrate-base_controller
  - [ ] administrate-field-select
  - [ ] administrate-field-boolean_emoji
  - [ ] administrate-field-lazy_belongs_to
  - [ ] administrate-field-jsontable
  - [ ] administrate-field-hidden
  - [ ] administrate-field-time
  - [ ] administrate-field-refile
  - [ ] administrate-field-froala
  - [ ] administrate-field-state_machine
  - [ ] administrate-field-collection_select
  - [ ] administrate-field-globalize-string

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -10,10 +10,13 @@ module Administrate
       resources = apply_collection_includes(resources)
       resources = order.apply(resources)
       resources = resources.page(params[:_page]).per(records_per_page)
-      page = Administrate::Page::Collection.new(dashboard, order: order)
+      page = Administrate::Page::Collection.new(
+        dashboard,
+        resources,
+        order: order,
+      )
 
       render locals: {
-        resources: resources,
         search_term: search_term,
         page: page,
         show_search_bar: show_search_bar?,
--- a/app/views/administrate/application/index.html.erb
+++ b/app/views/administrate/application/index.html.erb
@@ -11,10 +11,6 @@ It renders the `_table` partial to display details about the resources.
   An instance of [Administrate::Page::Collection][1].
   Contains helper methods to help display a table,
   and knows which attributes should be displayed in the resource's table.
-- `resources`:
-  An instance of `ActiveRecord::Relation` containing the resources
-  that match the user's search criteria.
-  By default, these resources are passed to the table partial to be displayed.
 - `search_term`:
   A string containing the term the user has searched for, if any.
 - `show_search_bar`:
@@ -58,9 +54,8 @@ It renders the `_table` partial to display details about the resources.
     collection_presenter: page,
     collection_field_name: resource_name,
     collection_names: [resource_name],
-    resources: resources,
     table_title: "page-title"
   ) %>
 
-  <%= paginate resources, param_name: '_page' %>
+  <%= paginate page.resources, param_name: '_page' %>
 </section>
--- a/app/views/administrate/application/index.html.erb
+++ b/app/views/administrate/application/index.html.erb
@@ -57,7 +57,7 @@ It renders the `_table` partial to display details about the resources.
     "collection",
     collection_presenter: page,
     collection_field_name: resource_name,
-    page: page,
+    collection_names: [resource_name],
     resources: resources,
     table_title: "page-title"
   ) %>
--- a/app/views/administrate/application/index.html.erb
+++ b/app/views/administrate/application/index.html.erb
@@ -57,5 +57,5 @@ It renders the `_table` partial to display details about the resources.
     table_title: "page-title"
   ) %>
 
-  <%= paginate page.resources, param_name: '_page' %>
+  <%= paginate page.resources_for_pagination, param_name: '_page' %>
 </section>
--- a/lib/administrate/field/has_many.rb
+++ b/lib/administrate/field/has_many.rb
@@ -11,8 +11,12 @@ module Administrate
         { "#{attr.to_s.singularize}_ids".to_sym => [] }
       end
 
-      def associated_collection(resources, order = self.order)
-        Administrate::Page::Collection.new(associated_dashboard, resources, order: order)
+      def associated_collection(page = 1, order = self.order)
+        Administrate::Page::Collection.new(
+          associated_dashboard,
+          resources(page, order),
+          order: order,
+        )
       end
 
       def attribute_key
@@ -39,11 +43,6 @@ module Administrate
         self.class.permitted_attribute(attribute)
       end
 
-      def resources(page = 1, order = self.order)
-        resources = order.apply(data).page(page).per(limit)
-        includes.any? ? resources.includes(*includes) : resources
-      end
-
       def more_than_limit?
         data.count(:all) > limit
       end
@@ -64,6 +63,11 @@ module Administrate
       end
 
       private
+
+      def resources(page, order)
+        resources = order.apply(data).page(page).per(limit)
+        includes.any? ? resources.includes(*includes) : resources
+      end
 
       def includes
         associated_dashboard.collection_includes
--- a/lib/administrate/field/has_many.rb
+++ b/lib/administrate/field/has_many.rb
@@ -11,8 +11,8 @@ module Administrate
         { "#{attr.to_s.singularize}_ids".to_sym => [] }
       end
 
-      def associated_collection(order = self.order)
-        Administrate::Page::Collection.new(associated_dashboard, order: order)
+      def associated_collection(resources, order = self.order)
+        Administrate::Page::Collection.new(associated_dashboard, resources, order: order)
       end
 
       def attribute_key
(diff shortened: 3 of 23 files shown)


dgmstuart:
The test failure on ruby 2.5 seems to be a result of CI environment configuration?
https://app.circleci.com/pipelines/github/thoughtbot/administrate/956/workflows/5cf2976f-fc78-410c-bc98-0de2ef5257fb/jobs/6869/parallel-runs/0/steps/0-110

     ActionView::Template::Error:
       Autoprefixer doesn’t support Node v8.10.0. Update it.

Looks like https://github.com/thoughtbot/administrate/pull/1772 might resolve that?

dgmstuart:
Rebasing

dgmstuart:
@pablobm @nickcharlton sorry to bother you: I know everyone is probably super busy with regular work, but let me know if I can do anything to make this easier to review. 

I guess this should have a "feature" label on it - is that something I can do, or does it have to be done by someone with permissions on the repo?

pablobm:
Thank you for this PR. At the moment, I think it's a matter of leaving it in our hands. This is a big PR, and although the individual commits are useful, we'll still need some time to wrap our heads around it.

One potential cause of concern is the changes to internal interfaces. Will it break third-party plugins? I recommend having a look at https://rubygems.org/gems/administrate/reverse_dependencies (particularly 
administrate-field-nested_has_many), to see if they are compatible. If these break, it's not necessarily KO, but we'll have to see how users can upgrade.

app/controllers/administrate/application_controller.rb
```diff
@@ -10,10 +10,13 @@ def index
       resources = apply_collection_includes(resources)
       resources = order.apply(resources)
       resources = resources.page(params[:_page]).per(records_per_page)
-      page = Administrate::Page::Collection.new(dashboard, order: order)
+      page = Administrate::Page::Collection.new(
+        dashboard,
+        resources,
+        order: order,
+      )
```
pablobm, line 17: Moving the resources into the `Page::Collection` object is probably a good call, and it bring it in line with the others page objects :+1:

app/views/administrate/application/index.html.erb
```diff
@@ -57,10 +53,9 @@ It renders the `_table` partial to display details about the resources.
     "collection",
     collection_presenter: page,
     collection_field_name: resource_name,
-    page: page,
-    resources: resources,
+    collection_names: [resource_name],
```
pablobm, line 56: How about renaming `collection_names` as `allowed_order_params`. At the end of the day, that's exactly what it's used for: telling `#sanitised_order_params` which other params can be permitted.

The mere existence of this param is awkward, but it's not the fault of your PR. Search params are sanitized at the bottom (templates/helpers), with information about what's allowed flowing top-to-bottom
  1. Dashboard/template figures out which params it allows.
  2. It renders partial and tells it "please allow these params along with those you allow"

I wonder if this should be sanitized a the top (the controller), with this info flowing bottom-to-top:
  1. Dashboards need to publish which params they allow
  2. Controller asks for lists, then sanitises params before rendering the top template.
  3. Templates don't need to know about sanitising.

But this is an issue for another day, out of scope for this PR.

lib/administrate/field/has_many.rb
```diff
@@ -11,8 +11,12 @@ def self.permitted_attribute(attr, _options = nil)
         { "#{attr.to_s.singularize}_ids".to_sym => [] }
       end
 
-      def associated_collection(order = self.order)
-        Administrate::Page::Collection.new(associated_dashboard, order: order)
+      def associated_collection(page = 1, order = self.order)
```
pablobm, line 14: I'm thinking that given that we are changing the method signature, we may as well change it to use keyword arguments. Positional arguments look a bit arbitrary to me. (I see it's the same as the old private `#resources`).

lib/administrate/page/collection.rb
```diff
@@ -3,6 +3,11 @@
 module Administrate
   module Page
     class Collection < Page::Base
+      def initialize(dashboard, resources, options = {})
+        super(dashboard, options)
+        @resources = resources
+      end
```
pablobm, line 9: Again, agree with changing this initializer :+1: We may want to go one step further and change it at `Page::Base`, but let's leave that for now.

lib/administrate/page/show.rb
```diff
@@ -8,10 +8,12 @@ def initialize(dashboard, resource)
...
       end
 
-      attr_reader :resource
+      def resource
+        dashboard.decorate_resource(@resource)
+      end
 
       def page_title
-        dashboard.display_resource(resource)
+        dashboard.display_resource(@resource)
```
pablobm, line 16: How about having a private getter with name `#undecorated_resource`? That may help clearing confusion between `@resource` and `#resource`.

lib/generators/administrate/dashboard/templates/dashboard.rb.erb
```diff
@@ -65,4 +65,11 @@ class <%= class_name %>Dashboard < Administrate::BaseDashboard
   # def display_resource(<%= file_name %>)
   #   "<%= class_name %> ##{<%= file_name %>.id}"
   # end
+
+  # Implement this method if you want to use Decorator pattern.
+  # Fields will access the returned object instead of the resource.
+  #
+  # def decorate_resource(<%= file_name %>)
+  #   <%= file_name %>
+  # end
```
pablobm, line 74: Thank you for adding docs!

pablobm (commented):
I have had a first review of this PR. I think it's a desirable feature. Ideally, I'd like Administrate not to need these hooks, and allow users to change this behaviour by more conventional means. However there are are a few things that would need to change for that first. I think it's ok to add this feature as a stopgap.

The PR description confused me a bit though: is it describing the same thing as this implements? When it comes to displaying fields, I would definitely prefer custom field types. However this PR enables a use case that is actually hacky to achieve currently: displaying new fields derived from other fields (or methods) of the model.

A small problem is that it introduces new fields that users can't sort by. But we have this issue already elsewhere, so I think it's not a big deal.

januszm:
This is a great topic to discuss, I just started using Administrate and I am impressed with its simplicity and ease of implementation. Thank you maintainers!
However, I immediately noticed that for the Index and Show views, I may need extended presentation layer configuration options. It is about the possibility of introducing additional methods in Presenter, e.g. aliasing attribute names, or when a resource has the UUID or URL attribute, I would like a link to the resource outside Administrate (in App) to be generated in the presentation layer. Of course, Administrate is supposed to be used for data management and when we talk about new / edit actions it should use the original model, not the presenter. I am referring here only to a limited set of views, which facilitates the work of non-technical people

januszm:
Please disregard my comment above, I think I can answer my own questions or clarify doubts, it looks like Custom Field Types serve as a "kind of" attribute presenters in Administrate https://github.com/thoughtbot/administrate/blob/main/docs/adding_custom_field_types.md

pablobm:

PR #1776, opened 2020-09-21
Improve how documentation pages are rendered

nickcharlton:
Taking @EWright212's work from #1705, as I don't seem to be able to push to that branch.

---

* Extracts an object to encapsulate rendering a given page (which also
   allows us to explicitly test it),
* Allows us to render pages without the YAML frontmatter, which was
   making a mess of the README and Contributing Guide,

### DIFF
--- a/spec/example_app/app/controllers/docs_controller.rb
+++ b/spec/example_app/app/controllers/docs_controller.rb
@@ -1,97 +1,32 @@
 class DocsController < ApplicationController
-  SPECIAL_FILES = [
-    {
-      file: 'CONTRIBUTING',
-      page: 'contributing'
-    }
-  ].freeze
-
-  REDCARPET_CONFIG = {
-    fenced_code_blocks: true,
-    autolink: true,
-  }.freeze
-
   def index
-    render_page "README"
+    render_page("README")
   end
 
   def show
-    render_correct_page
-  end
-
-  private
-
-  def find_special_file
-    SPECIAL_FILES.select { |page| page[:page] == params[:page] }.first
-  end
-
-  def render_correct_page
-    if !find_special_file.nil?
-      render_page(find_special_file[:file])
+    case params[:page]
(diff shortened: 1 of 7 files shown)


spec/example_app/app/controllers/docs_controller.rb
```diff
@@ -87,7 +96,7 @@ def title
     end
 
     def title_suffix
-      metadata["home"] ? "" : " - Administrate"
+      metadata["home"] ? "Administrate" : " - Administrate"
```
pablobm, line 99: Am I correct that this `metadata["home"]` doesn't exist any more for any page? If so, this line probably needs removing the reference to it.

pablobm: Oh, I see it's added in the constructor. I feel that's a bit too indirect. Would it make sense to directly the following instead, and do away with the "artificial" metadata?

```suggestion
      front_matter_parsed.front_matter.empty? ? "Administrate" : " - Administrate"
```

nickcharlton: Yeah, I was thinking something like that. So far I've just rebased it, but I sketched out a potentially different way of doing it the other day, which I should be able to get back to soon.

nickcharlton:
I ended up completely visiting this one, as I wasn't happy with the direction the `DocsController` was going in. We seemed to be going down a path which was just awkward (from how I'd explained how we might do this in the first place, to the implementation we had to ended up with because of it).

I also backfilled some feature specs here: I did this to drive me getting this working but maybe they could be better or just go.

nickcharlton:
It turns out that the mysterious CI failure was related to case sensitive file systems: CI, being Linux, is case sensitive and so `CONTRIBUTING` and `contributing` aren't the same thing!

Finally, paired with @iftheshoefritz to go over this, so I'm going to merge this now!

PR #1782, opened 2020-10-05
Adding an error message during 'administrate' initialize when there are no models in db.

KOlofinmoyin:
This PR fixes the behaviour observed, during the administrate install on rails apps without models, where the admin route isn't created, resulting in an application error and no warning to the user. 

The change displays a friendly error message, when this happens.

### DIFF
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -1,3 +1,9 @@
+if defined?(Zeitwerk)
+  Zeitwerk::Loader.eager_load_all
+else
+  Rails.application.eager_load!
+end
+
 require "rails/generators/base"
 require "administrate/generator_helpers"
 require "administrate/namespace"
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -18,15 +18,9 @@ module Administrate
 
       def model_check
         if database_models.none?
-          puts "ERROR: Add models before installing Administrate."
-        else
-          run_routes_generator
-          create_dashboard_controller
-          run_dashboard_generators
+          raise Error, "Add models before installing Administrate."
         end
       end
-
-      private
 
       def run_routes_generator
         if dashboard_resources.none?
@@ -48,6 +42,9 @@ module Administrate
             "--namespace", namespace
         end
       end
+ 
+      private
+
 
       def namespace
         options[:namespace]
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -62,7 +62,7 @@ module Administrate
       end
 
       def database_models
-        ActiveRecord::Base.descendants.reject(&:abstract_class?)
+        ActiveRecord::Base.descendants.reject(&:abstract_class?).reject { |d| d.name == d.to_s }
       end
 
       def unnamed_constants
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -18,7 +18,7 @@ module Administrate
 
       def model_check
         if database_models.none?
-          say_status :conflict, :red
+          say_status(:conflict, :red, "Add models before installing Administrate.")
           raise Error, "Add models before installing Administrate."
         end
       end
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -17,6 +17,7 @@ module Administrate
       class_option :namespace, type: :string, default: "admin"
 
       def model_check
+        puts database_models, database_models.none?
         if database_models.none?
           raise Error, "Add models before installing Administrate."
         end
--- a/lib/generators/administrate/install/install_generator.rb
+++ b/lib/generators/administrate/install/install_generator.rb
@@ -17,11 +17,8 @@ module Administrate
       class_option :namespace, type: :string, default: "admin"
 
       def model_check
-        puts database_models, database_models.none?
-        puts models_without_tables, "w/o"
-        puts unnamed_constants, "unnamed"
         if database_models.none?
-          raise Error, "Add models before installing Administrate."
+          raise IOError, "Add models before installing Administrate."
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/generators/administrate/install/install_generator.rb
```diff
@@ -10,6 +10,12 @@ class InstallGenerator < Rails::Generators::Base
 
       class_option :namespace, type: :string, default: "admin"
 
+      def model_check
+        if database_models.none?
+          puts "ERROR: Unable to generate a dashboard without models."
```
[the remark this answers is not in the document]
RaeRachael: I think that is a better phrase, also as the check doesn't stop the rest of the generator trying to run (which could be done) maybe adding blank lines above and below to make it standout?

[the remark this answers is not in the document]
nickcharlton: We could `exit` after the error. Is there anything in Rails' existing generators which behaves in a similar way?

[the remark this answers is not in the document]
nickcharlton: I was initially wondering if we should actually be writing to standard error, but a quick search around didn't seem to find that being used for generators (but I might've just looked in bad places.)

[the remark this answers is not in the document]
RaeRachael: I can't find any other example of a generator exiting... so `exit` could be the best solution. 

I think a puts error would be fine, especially if its the last output on the CL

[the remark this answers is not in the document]
nickcharlton: Yeah, sounds good!

lib/generators/administrate/install/install_generator.rb
```diff
@@ -10,6 +16,18 @@ class InstallGenerator < Rails::Generators::Base
...
 
+      def model_check
+        if database_models.none?
+          puts "ERROR: Add models before installing Administrate."
+        else
+          run_routes_generator
+          create_dashboard_controller
+          run_dashboard_generators
+        end
+      end
```
pablobm, line 27: I'm not sure about this implementation. I have these concerns:

1. A lack of models will display an error on `stdout` instead of `stderr`, and it will not result in a non-zero [exit status](https://en.wikipedia.org/wiki/Exit_status) returned to the shell.
2. Also this won't be an error internally to Ruby, which will probably cause it not to play well if ran as part of a "parent" generator.
3. It is not idiomatic of Rails generators, bundling all steps into a single method.

Looking into how this should be done in the realm of Thor scripts or Rails generators... I don't see much information! I had to go check Rails's source code, and I've seen a couple of examples that raise [`Rails::Generators::Error`](https://github.com/rails/rails/blob/master/railties/lib/rails/generators/base.rb#L14), like this one:

https://github.com/rails/rails/blob/e44b3419d44b5fa6e66305fe703f34bb63932593/railties/lib/rails/generators/actions/create_migration.rb#L62-L65

It might even be worth adding a `say_status` line like in that example, where it generates an output line like this one (but with colours):

```
    conflict    db/migrate/20201008102310_create_things.rb
Another migration is already named create_things: /home/pablobm/Downloads/deletemes/rails-deleteme/db/migrate/20201008102256_create_things.rb. Use --force to replace this migration or --skip to ignore conflicted file.
```

In practical terms, what I suggest means:

1. Make `model_check` a method that just checks if there are models and does `raise Error, message` if not.
2. Leave the other methods where they were: public and not called directly. If I recall correction, the generator will call them in order if there aren't previous errors.

RaeRachael: interesting, I didn't know how a raised error would work here, I'll give it a go tomorrow or friday

pablobm: Looks good now!

pablobm (changes requested):
This looks almost ready now. Just two things:

1. The tests are failing because of a security advisory on a dependency. They should pass if you rebase.
2. Make sure to correct the linter warnings raised by Hound.

lib/generators/administrate/install/install_generator.rb
```diff
@@ -31,6 +37,13 @@ def run_dashboard_generators
         end
       end
 
+      def model_check
+        if valid_dashboard_models.none?
+          puts "** ERROR: Add models before installing Administrate. **"
```
pablobm, line 42: If this is not going to stop the generator, then I think that this should be a warning, not an error. Mind changing it to be similar to existing warnings? For example those at https://github.com/thoughtbot/administrate/blob/a5d695da773f15e1709e6be06c0a526156f6c1d6/lib/generators/administrate/routes/routes_generator.rb#L26

lib/generators/administrate/install/install_generator.rb
```diff
@@ -31,6 +37,13 @@ def run_dashboard_generators
...
+          puts "**        Consider removing 'app/controllers/admin'. **"
```
pablobm, line 43: Why this message?

RaeRachael: because, trying to avoid the other methods by an error didn't seem possible. and that is the file that the install generator creates if no valid models are present

pablobm: I feel that it can confuse more than help. I think a reader would think that there's something special in that file that needs to be refreshed every time they run the generator. Additionally, if they are running the generator now, they will later when the have models, so I think it's safe to assume that they'll want the file.

RaeRachael: maybe, but there is the potential that they would like to use the admin namespace for something specific to their rails app, and now we have potentially forced them to use a different name for it (unless they are comfortable deleting the file). 

I don't know, I feel it is useful to make people comfortable resetting their app to a state before call administrate install. But I trust your opinion more on how the message reads (gets interpreted)

pablobm: I see what you mean. The problem I see is lack of consistency. If you run any other generator and there's an error, any files created so far will stay, and it will be on the developer what to do. Typically you'd use Git, which would point out the changes and provide means for rolling back.

As for the actual message, "consider removing this file", to me is an encouragement strong enough that can send the wrong message, and lead to believe that there's something special to the file. In my case, I think that I would understand that the file is different in every run, perhaps based on the models at the time of running, and perhaps even that it should be re-generated if we add new models.

I'd prefer to remove it.

RaeRachael: That makes sense, I'll make those changes now

pablobm (approved):
LGTM :+1:

lib/generators/administrate/install/install_generator.rb
```diff
@@ -30,9 +42,10 @@ def run_dashboard_generators
             "--namespace", namespace
         end
       end
-
+
```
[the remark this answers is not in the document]
pablobm: Make sure to fix these, please

pablobm:
Thank you!

PR #1788, opened 2020-10-08
Pass page local when rendering fields

bobcats:
Hi there!

We live dangerously and run on edge for administrate, and seems like a recent change (#1259) has caused `HasOne`'s to not render in our app with a `undefined local variable or method 'page'`

The proposed change in the PR seems to resolve the error.

Let me know if I need to change anything, or if this is too naive of a solution.

Thanks!

### DIFF
--- a/spec/administrate/views/fields/has_one/_show_spec.rb
+++ b/spec/administrate/views/fields/has_one/_show_spec.rb
@@ -23,6 +23,14 @@ describe "fields/has_one/_show", type: :view do
     it "renders a link to the record" do
       product = create(:product)
       product_path = polymorphic_path([:admin, product])
+      nested_show = instance_double(
+        "Administrate::Page::Show",
+        resource: double(
+          class: ProductMetaTag,
+        ),
+        attributes: [],
+        resource_name: "Product Tag",
+      )
       has_one = instance_double(
         "Administrate::Field::HasOne",
         display_associated_resource: product.name,
@@ -43,15 +51,54 @@ describe "fields/has_one/_show", type: :view do
       expect(rendered.strip).to include(link)
     end
 
-    def nested_show
-      instance_double(
+    it "renders nested attribute relationships" do
+      template.extend Administrate::ApplicationHelper
+
+      product = create(:product)
+      page = create(:page, product: product)
+
+      nested_has_many = instance_double(
+        "Administrate::Field::HasMany",
+        associated_collection: [page],
+        attribute: :page,
+        data: [page],
+        resources: [page],
+        html_class: "has-many",
+        name: "Page",
+        to_partial_path: "fields/has_many/index",
+        order_from_params: {},
+      )
+
+      nested_show = instance_double(
         "Administrate::Page::Show",
         resource: double(
           class: ProductMetaTag,
         ),
-        attributes: [],
+        attributes: [nested_has_many],
         resource_name: "Product Tag",
       )
+
+      has_one = instance_double(
+        "Administrate::Field::HasOne",
+        display_associated_resource: product.name,
+        data: product,
+        nested_show: nested_show,
+      )
+
+      page_double = instance_double("Administrate::Page::Show")
+
+      render(
+        partial: "fields/has_one/show.html.erb",
+        locals: {
+          field: has_one,
+          namespace: "admin",
+          page: page_double,
+          resource_name: "product_meta_tag",
+        },
+      )
+
+      has_many_count = "1 page"
+      expect(rendered.strip).to include(has_many_count)
     end
   end
 end
--- a/app/views/fields/has_one/_show.html.erb
+++ b/app/views/fields/has_one/_show.html.erb
@@ -32,7 +32,7 @@ All show page attributes of has_one relationship would be rendered
         ) %>
       </dt>
       <dd class="attribute-data attribute-data--<%= attribute.html_class %>">
-        <%= render_field attribute %>
+        <%= render_field attribute, { page: page } %>
       </dd>
       </div>
     <% end -%>


nickcharlton:
Ah nice, thanks! It's fairly likely that me merging that caused this.

To stop this happening again, can you think of a way we can test this? Ideally the ideal solution to the problem.

bobcats:
Totally! I think the reproduction ended up being trying to render `has_one`s with `has_one`s. Setting up Payment with a `has_one` without passing the page local reproduced the error.

I'm used to using `render_views`, but it seems like this isn't used much in the codebase. Is there a different approach you'd like to take with the test? Just checking for not raising an error seems not ideal. Also, trying to figure out how to get the Hound rule to pass, as with positional args it seems to cause the test to fail.

pablobm:
Issue also reported at https://github.com/thoughtbot/administrate/issues/1792

spec/controllers/admin/payments_controller_spec.rb
```diff
@@ -0,0 +1,12 @@
+require "rails_helper"
+
+describe Admin::PaymentsController, type: :controller do
+  render_views
+  describe "GET index" do
+    it "hides the search bar" do
+      payment = create(:payment)
+
+      expect { get(:show, id: payment.id) }.not_to raise_error
```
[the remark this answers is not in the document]
pablobm: Looking at [the docs for this rule](https://docs.rubocop.org/rubocop-rails/cops_rails.html#railshttppositionalarguments), I think all this needs is to use the `params:` keyword argument:

```suggestion
      expect { get(:show, params: { id: payment.id }) }.not_to raise_error
```

pablobm:
@bobcats - I agree that testing for the page not to error is not ideal. I wonder if this could be done with a view spec instead? Eg: trying to render something at https://github.com/thoughtbot/administrate/blob/master/spec/administrate/views/fields/has_one/_show_spec.rb and checking that the expected text is there.

bobcats:
Thanks for the feedback folks! I've tried testing the scenario as a view spec, with albeit a lot of double-ifying. Let me know what you all think.

spec/administrate/views/fields/has_one/_show_spec.rb
```diff
@@ -23,6 +23,14 @@
     it "renders a link to the record" do
       product = create(:product)
       product_path = polymorphic_path([:admin, product])
+      nested_show = instance_double(
+        "Administrate::Page::Show",
+        resource: double(
+          class: ProductMetaTag,
+        ),
```
pablobm, line 30: I think we can provide the real deal here, instead of a stub:

```suggestion
        resource: create(:product_meta_tag),
```

spec/administrate/views/fields/has_one/_show_spec.rb
```diff
@@ -43,15 +51,54 @@
...
+        "Administrate::Field::HasMany",
+        associated_collection: [page],
+        attribute: :page,
+        data: [page],
+        resources: [page],
+        html_class: "has-many",
+        name: "Page",
+        to_partial_path: "fields/has_many/index",
+        order_from_params: {},
+      )
```
pablobm, line 70: Similarly, I think we can provide a real `Administrate::Field::HasMany`. Here's an example:

https://github.com/thoughtbot/administrate/blob/64aaa98a79b3108fb1442025eb818d3aea8e7b3d/spec/lib/fields/has_many_spec.rb#L22-L27

spec/administrate/views/fields/has_one/_show_spec.rb
```diff
@@ -43,15 +51,54 @@
...
+        attributes: [nested_has_many],
         resource_name: "Product Tag",
       )
+
+      has_one = instance_double(
+        "Administrate::Field::HasOne",
+        display_associated_resource: product.name,
+        data: product,
+        nested_show: nested_show,
+      )
```
pablobm, line 86: Same here, let's use the real stuff if possible.

pablobm (commented):
Thank you for this. It looks pretty good. Regarding those stubs, I think let's try use the real objects (left some comments on that). I see that the original file was already using many stubs and you followed the same pattern, but I'd rather start do differently here.

spec/administrate/views/fields/has_one/_show_spec.rb
bobcats: Scratching my head on this one:

```
Failure/Error: <%= render(
     
     ActionView::Template::Error:
       Missing partial /_collection, fields/has_one/_collection with {:locale=>[:en], :formats=>[:html, :text, :js, :css, :ics, :csv, :vcf, :vtt, :png, :jpeg, :gif, :bmp, :tiff, :svg, :mpeg, :mp3, :ogg, :m4a, :webm, :mp4, :otf, :ttf, :woff, :woff2, :xml, :rss, :atom, :yaml, :multipart_form, :url_encoded_form, :json, :pdf, :zip, :gzip], :variants=>[], :handlers=>[:raw, :erb, :html, :builder, :ruby]}. Searched in:
         * "/Users/brianfield/code/buildrtech/administrate/spec/example_app/app/views"
         * "/Users/brianfield/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/administrate-field-image-1.1.0/app/views"
         * "/Users/brianfield/code/buildrtech/administrate/app/views"
         * "/Users/brianfield/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/kaminari-core-1.2.1/app/views"
```

```diff
diff --git a/spec/administrate/views/fields/has_one/_show_spec.rb b/spec/administrate/views/fields/has_one/_show_spec.rb
index 6ca4309..31d75ed 100644
--- a/spec/administrate/views/fields/has_one/_show_spec.rb
+++ b/spec/administrate/views/fields/has_one/_show_spec.rb
@@ -1,5 +1,6 @@
 require "rails_helper"
 require "administrate/field/has_one"
+require "administrate/field/has_many"
 
 describe "fields/has_one/_show", type: :view do
   context "without an associated record" do
@@ -25,9 +26,7 @@ describe "fields/has_one/_show", type: :view do
       product_path = polymorphic_path([:admin, product])
       nested_show = instance_double(
         "Administrate::Page::Show",
-        resource: double(
-          class: ProductMetaTag,
-        ),
+        resource: create(:product_meta_tag),
         attributes: [],
         resource_name: "Product Tag",
       )
@@ -52,28 +51,22 @@ describe "fields/has_one/_show", type: :view do
     end
 
     it "renders nested attribute relationships" do
-      template.extend Administrate::ApplicationHelper
+      view.extend Administrate::ApplicationHelper
 
       product = create(:product)
       page = create(:page, product: product)
+      product_meta_tag = create(:product_meta_tag, product: product)
 
-      nested_has_many = instance_double(
-        "Administrate::Field::HasMany",
-        associated_collection: [page],
-        attribute: :page,
-        data: [page],
-        resources: [page],
-        html_class: "has-many",
-        name: "Page",
-        to_partial_path: "fields/has_many/index",
-        order_from_params: {},
+      nested_has_many = Administrate::Field::HasMany.new(
+        :pages,
+        Page.all,
+        :show,
+        resource: product,
       )
 
       nested_show = instance_double(
         "Administrate::Page::Show",
-        resource: double(
-          class: ProductMetaTag,
-        ),
+        resource: product_meta_tag,
         attributes: [nested_has_many],
         resource_name: "Product Tag",
       )
```

Something simple I'm missing?

bobcats:
@pablobm Thanks for the feedback again! So, I tried using real objects, but end up getting being unable to reproduce the original error because the combinations of associations that show this issue don't seem to exist in the sample app's models.

Should I create that particular combination of associations in the sample app and then use the real models to do so? I was trying to avoid changing the models, but since there's a fair amount of reflecting going on, it seems more straightforward. I think I need a Parent `has_one` Child `has_many` Children.

pablobm:
@bobcats - It's ok, thanks for trying. I think we can merge this as-is then. No need to make it more complicated than it needs be.

PR #1791, opened 2020-10-08
Upgrade bundler to 2.1.4

nickcharlton:
This is a bit overdue, but is needed for Ruby 2.7 onwards.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ commands:
       - checkout
 
       # Install bundler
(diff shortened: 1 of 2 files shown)


nickcharlton:
circleci: what if we made bash …but made it one hundred times worse.

🙄

nickcharlton:
Right, now it's failing because of the Rails 6.0.3.4 bump, which is fixed in #1787. So I'm going to tidy up and merge and go back to what I came here to do…

PR #1794, opened 2020-10-12
Don't show unpersisted `has_one` associations

rwehresmann:
Showing unpersisted `has_one` associations can be misleading and break the application is some cases.

Let's consider the following example:

```
class Product < ApplicationRecord
  has_one :product_meta_tag

  def product_meta_tag
    super || build_product_meta_tag
  end
end
```

That opens for a scenario where a `product.product_meta_tag` will return a non-persisted instance of a ProductMetaTag. In the `_show` partial we're just checking if `field.data` (i.e., ProductMetaTag instance) exists to display it, and a non-persisted instance would be displayed. The link displayed in the partial (`link_to( field.display_associated_resource, [namespace, field.data]`) would be pointing to the index page of product meta tags, what is not desired. Also, in the absence of an index route, the page would break.

The solution: just check if data is persisted to display it or not.

### DIFF
--- a/spec/administrate/views/fields/has_one/_show_spec.rb
+++ b/spec/administrate/views/fields/has_one/_show_spec.rb
@@ -4,10 +4,10 @@ require "administrate/field/has_one"
 describe "fields/has_one/_show", type: :view do
   context "without an associated record" do
     it "displays nothing" do
-      has_one = instance_double(
-        "Administrate::Field::HasOne",
-        display_associated_resource: "",
-        data: nil,
+      has_one = Administrate::Field::HasOne.new(
+        :product_meta_tag,
+        build(:product_meta_tag),
+        :show,
       )
 
       render(
@@ -27,6 +27,7 @@ describe "fields/has_one/_show", type: :view do
         "Administrate::Field::HasOne",
         display_associated_resource: product.name,
         data: product,
+        linkable?: true,
         nested_show: nested_show,
       )
 
--- a/spec/lib/fields/has_one_spec.rb
+++ b/spec/lib/fields/has_one_spec.rb
@@ -63,4 +63,45 @@ describe Administrate::Field::HasOne do
       expect(path).to eq("/fields/has_one/#{page}")
     end
   end
+
+  describe "#linkable?" do
+    context "when data is persisted" do
+      it "shows it" do
+        product_meta_tag = create(:product_meta_tag)
+        field = described_class.new(
+          :product_meta_tag,
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field).to be_linkable
+      end
+    end
+
+    context "when data isn't persisted" do
+      it "doesn't shows it" do
+        product_meta_tag = build(:product_meta_tag)
+        field = described_class.new(
+          :product_meta_tag,
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field).to_not be_linkable
+      end
+    end
+
+    context "when data doesn't respond to `persisted?`" do
+      it "doesn't show it" do
+        product_meta_tag = double(:product_meta_tag)
+        field = described_class.new(
+          :product_meta_tag,
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field.linkable?).to be_falsey
+      end
+    end
+  end
 end
--- a/spec/lib/fields/has_one_spec.rb
+++ b/spec/lib/fields/has_one_spec.rb
@@ -100,7 +100,7 @@ describe Administrate::Field::HasOne do
           :show,
         )
 
-        expect(field.linkable?).to be_falsey
+        expect(field).not_to be_linkable
       end
     end
   end
--- a/app/views/fields/has_one/_index.html.erb
+++ b/app/views/fields/has_one/_index.html.erb
@@ -15,7 +15,7 @@ By default, the relationship is rendered as a link to the associated object.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/HasOne
 %>
 
-<% if field.data %>
+<% if field.linkable? %>
   <%= link_to(
     field.display_associated_resource,
     [namespace, field.data],
--- a/app/views/fields/has_one/_show.html.erb
+++ b/app/views/fields/has_one/_show.html.erb
@@ -15,7 +15,7 @@ All show page attributes of has_one relationship would be rendered
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/HasOne
 %>
 
-<% if field.data %>
+<% if field.linkable? %>
   <fieldset class="attribute--nested">
     <legend>
       <%= link_to(
--- a/lib/administrate/field/has_one.rb
+++ b/lib/administrate/field/has_one.rb
@@ -31,6 +31,10 @@ module Administrate
         )
       end
 
+      def linkable?
+        data.try(:persisted?)
+      end
+
       private
 
       def resolver
--- a/spec/administrate/views/fields/has_one/_index_spec.rb
+++ b/spec/administrate/views/fields/has_one/_index_spec.rb
@@ -3,7 +3,11 @@ require "rails_helper"
 describe "fields/has_one/_index", type: :view do
   context "without an associated record" do
     it "displays nothing" do
-      has_one = double(data: nil)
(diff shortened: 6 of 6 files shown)


spec/administrate/views/fields/has_one/_show_spec.rb
```diff
@@ -2,12 +2,11 @@
 require "administrate/field/has_one"
 
 describe "fields/has_one/_show", type: :view do
-  context "without an associated record" do
+  context "when shouldn't show the record" do
```
pablobm, line 5: I think that the original test is ok here. I wouldn't change it.

spec/administrate/views/fields/has_one/_show_spec.rb
```diff
@@ -19,14 +18,15 @@
     end
   end
 
-  context "with an associated record" do
+  context "when should show the record" do
```
pablobm, line 21: Same here, I would leave the original text.

spec/lib/fields/has_one_spec.rb
```diff
@@ -63,4 +63,45 @@
       expect(path).to eq("/fields/has_one/#{page}")
     end
   end
+
+  describe "#show?" do
+    context "when data is persisted" do
+      it "allows to show" do
```
pablobm, line 69:
```suggestion
      it "shows it" do
```

spec/lib/fields/has_one_spec.rb
```diff
@@ -63,4 +63,45 @@
...
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field.show?).to be_truthy
+      end
+    end
+
+    context "when data isn't persisted" do
+      it "doesn't allows to show" do
```
pablobm, line 82:
```suggestion
      it "doesn't show it" do
```

spec/lib/fields/has_one_spec.rb
```diff
@@ -63,4 +63,45 @@
...
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field.show?).to be_falsey
+      end
+    end
+    
+    context "when data doesn't respond to `persisted?`" do
+      it "doesn't allows to show" do
```
pablobm, line 95:
```suggestion
      it "doesn't show it" do
```

pablobm (commented):
Thank you for your PR! On one hand, I agree that we shouldn't be displaying links to unpersisted resources. On the other hand, I wonder if the problem here is not so much with Administrate as with the underlying model, forcing an unpersisted association and confusing the code that enters in contact with it.

I'd like a second opinion on this one. Perhaps @nickcharlton or @c4lliope have thoughts?

rwehresmann:
@pablobm I understand your point. From what I saw until now in rails projects is quite common to find has_one relationships described in the model like in my example. However, that doesn't mean it's the best approach and I'm also not sure if this is enough reason to merge the PR (I'll keep doing these rewritings in my projects, though). In any case, I updated the commit with the test names you suggested to keep.

pablobm:
@rwehresmann - Can you also solve Hound's lint warnings, please?

c4lliope:
Can we add a check like...
```erb
<%= @product.product_meta_tag.persisted? ? link_to(@product.product_meta_tag) : @product.product_meta_tag.name %>
```

So the link is only rendered for persisted models, and otherwise we only display plain text on the page?

Nice catch here, this seems like a use case that I wouldn't want to break peoples' applications over.

rwehresmann:
@pablobm PR updated with the corrections.

@c4lliope I'm thinking if it's worthy. We actually don't have how to control the name of the model fields (we don't have the warranty that there is a `name` field, for instance), so the alternative would be to print the model instance as a string, what I think it isn't visually useful. 

What we're showing now for unpersisted models:
Screenshot from 2020-10-25 21-55-09

What we would be showing printing as a string:
Screenshot from 2020-10-25 21-59-38

c4lliope:
Could you access `product_meta_tag_dashboard.display_resource(...)` ?
...[as described here])https://github.com/thoughtbot/administrate/blob/b85abcf965ecc843ee9b25e3a2ea65dfeb1b0b3c/spec/dashboards/customer_dashboard_spec.rb#L70-L79).

Or, i believe, there is `dashboard.display_associated_resource()`, doing similar.

c4lliope:
Ah, here;
`Administrate::Field::HasOne < Administrate::Field::Associative`
and `::Associative` has...
https://github.com/thoughtbot/administrate/blob/844c27091bab1a384b260a47b0a67dbff856640d/lib/administrate/field/associative.rb#L6-L8

rwehresmann:
Oh, I see what you mean. However, `display_resource` called inside `display_associated_resource` is implemented in `Administrate::BaseDashboard` as 

```
    def display_resource(resource)
      "#{resource.class} ##{resource.id}"
    end
```
The output for the situation we're treating in this PR would be `ProductMetaTag #`, but if I haven't a `product_meta_tag` as `super || build_product_meta_tag`, it would be nil and break when trying to access `resource.id`. 

I could change the `display_resource` implementation to avoid this, of course. Do you think it's worth it?

c4lliope:
Ooh, good digging.

Hm. Hmmmm.
Changing `display_resource` wouldn't be a *breaking* change, but it would
be a visible change on peoples' dashboards.
I'm curious what our options would be for making that change; I chose
`"#{resource.class} ##{resource.id}"` as a sensible default for persisted
records; I'm not sure a similar sensible default is possible for
unpersisted.

Unless, maybe;
```ruby
def display_resource(resource)
  resource.persisted?
  ? "#{resource.class} ##{resource.id}"
  : "Unpersisted #{resource.class}"
end
```

c4lliope:
Sorry, I guess email breaks comment formatting.

rwehresmann:
@c4lliope I still think that from a user perspective display nothing at all is the best option. It's possible that they don't even understand what is "persistence" in the context of the application. We could use "unrecord" for instance, but it seems redundant to me. However, if we should display something, I think your suggestion is the way to go.

I updated the PR with the described solution. 

Just to record: I'm calling always `try(:persisted?)`. I never used another ORM then ActiveRecord, and I'm not sure if Administrate works just fine with another. In any case, if the ORM doesn't implement `persisted?` these alterations will not break the application, we'll just display text instead. If that becomes a problem, it'll be easy to change it to make it work with another ORM.

c4lliope:
Go ahead and display a blank, I think that is a fine solution also.

spec/lib/fields/has_one_spec.rb
```diff
@@ -63,4 +63,45 @@
...
+        product_meta_tag = double(:product_meta_tag)
+        field = described_class.new(
+          :product_meta_tag,
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field.linkable?).to be_falsey
```
nickcharlton, line 103: Fun thing we can do here:

```suggestion
        expect(field).not_to be_linkable
```

spec/lib/fields/has_one_spec.rb
```diff
@@ -63,4 +63,45 @@
...
+    context "when data isn't persisted" do
+      it "doesn't shows it" do
+        product_meta_tag = build(:product_meta_tag)
+        field = described_class.new(
+          :product_meta_tag,
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field.linkable?).to be_falsey
```
nickcharlton, line 90: Fun thing we can do here:

```suggestion
        expect(field).not_to be_linkable
```

spec/lib/fields/has_one_spec.rb
```diff
@@ -63,4 +63,45 @@
...
+    context "when data is persisted" do
+      it "shows it" do
+        product_meta_tag = create(:product_meta_tag)
+        field = described_class.new(
+          :product_meta_tag,
+          product_meta_tag,
+          :show,
+        )
+
+        expect(field.linkable?).to be_truthy
```
nickcharlton, line 77: Fun thing we can do here:

```suggestion
        expect(field).to be_linkable
```

nickcharlton:
We don't need to worry about other ORMs, thankfully, so `persisted?` is fine!

rwehresmann:
Sorry for the delay on this.

@c4lliope done: display a link if persisted, or blank if not.

@nickcharlton greath. Thanks for the tips!

pablobm (approved):

PR #1797, opened 2020-10-14
Add delimiter option for number fields

ianbayne:
Great gem, everyone! We're using it in a project currently and are finding it very helpful.

I realized there doesn't seem to be a delimiter option for number fields to, for example, display a number like `1000000` in the more legible format of `1,000,000`, so I whipped up this PR to add it in. Please let me know if there's anything else I can or should add. 🙇

### DIFF
--- a/lib/administrate/field/number.rb
+++ b/lib/administrate/field/number.rb
@@ -1,16 +1,19 @@
 require_relative "base"
+require "active_support/number_helper"
 
 module Administrate
   module Field
     class Number < Field::Base
       def to_s
-        data.nil? ? "-" : format_string % value
+        result = data.nil? ? "-" : format_string % value
+        result = format(result) if options[:format]
+        prefix + result + suffix
       end
 
       private
 
       def format_string
-        prefix + "%.#{decimals}f" + suffix
+        "%.#{decimals}f"
       end
 
       def prefix
@@ -29,6 +32,20 @@ module Administrate
 
       def value
         data * options.fetch(:multiplier, 1)
+      end
+
+      def format(result)
+        formatter = options[:format][:formatter]
+        formatter_options = options[:format][:formatter_options].to_h
+
+        case formatter
+        when :number_to_delimited
+          ActiveSupport::NumberHelper.number_to_delimited(
+            result, **formatter_options
+          )
+        else
+          result
+        end
       end
     end
   end
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -158,6 +158,13 @@ more results than expected. Default is `false`.
 
 `:suffix` - Suffixes the number with a string. Defaults to `""`.
 
+`:format` - Specify a hash which defines a formatter. This uses ActiveSupport
+and works by  by passing a hash that includes the formatter (`formatter`) and
+the options for the formatter (`formatter_options`). Defaults to the locale's
+delimiter when `formatter_options` does not include a `delimiter`. See the
+example below. Note that currently only
+`ActiveSupport::NumberHelper.number_to_delimited` is supported.
+
 For example, you might use the following to display U.S. currency:
 
 ```ruby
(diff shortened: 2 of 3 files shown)


lib/administrate/field/number.rb
```diff
@@ -30,6 +32,11 @@ def decimals
       def value
         data * options.fetch(:multiplier, 1)
       end
+
+      def add_delimiter(result)
+        result.reverse.gsub(/(\d{3})(?=\d)/, "\\1#{options[:delimiter]}").
```
sedubois, line 37: How about using `ActiveSupport::NumberHelper.number_to_delimited`?

ianbayne: Thanks for the comment, @sedubois. Sure, that sounds good to me. Using `ActiveSupport::NumberHelper.number_to_delimited` will make it easier to understand at a glance as well.

lib/administrate/field/number.rb
```diff
@@ -1,16 +1,19 @@
...
+require "active_support/number_helper"
 
 module Administrate
   module Field
     class Number < Field::Base
       def to_s
-        data.nil? ? "-" : format_string % value
+        result = data.nil? ? "-" : format_string % value
+        result = add_delimiter(result) if options[:delimiter]
+        prefix + result + suffix
```
ianbayne, line 10: I realized that the delimiter option works fine as originally coded, but only when there isn't a prefix or suffix. Commit b163adb ensures it works with both and adds a test to ensure that that's the case.

lib/administrate/field/number.rb
```diff
@@ -30,6 +33,11 @@ def decimals
       def value
         data * options.fetch(:multiplier, 1)
       end
+
+      def add_delimiter(result)
+        ActiveSupport::NumberHelper.
+          number_to_delimited(result, delimiter: options[:delimiter])
```
pablobm, line 39: I'm not sure about the user passing a specific delimiter explicitly. In theory, this should be provided by the locale. In fact `number_to_delimited` should by the default format as per the locale, so it shouldn't need a delimiter either.

So in my mind, `:delimiter` should be just a boolean option. Thoughts?

sedubois: It would be great to have a way to give various options that then get forwarded to `number_to_delimited`: 
- delimiter
- separator
- delimiter_pattern

However the same could be said for other ActiveSupport methods, e.g `number_to_percentage`:
- precision
- significant
- separator
- delimiter
- strip_insignificant_zeros
- format

+ `number_to_rounded`, `number_to_phone`, `number_to_currency`, `number_to_human_size`, `number_to_human`. So even if all of that is not enabled in this PR it might be good to have an API that does not prevent these other possibilities.

For example this field could accept two options:
- `conversion_type` with possible values `number_to_delimited`, `number_to_percentage`, etc
- `conversion_options` would be passed transparently to the relevant ActiveSupport method

ianbayne: Thanks for the comments, @pablobm & @sedubois. 

>  In theory, this should be provided by the locale. In fact `number_to_delimited` should by the default format as per the locale, so it shouldn't need a delimiter either. So in my mind, `:delimiter` should be just a boolean option. Thoughts?

Personally, while I imagine that in the majority of cases users will default to whatever is standard for their locale, when using Administrate I would appreciate the added flexibility of being able to choose my own delimiter. 

> It would be great to have a way to give various options that then get forwarded to `number_to_delimited`

I like this idea, and the idea of leaving the API "ready" for future updates to allow other ActiveSupport methods. On note, however, is that as there's already a `decimals` option, if we were to do so, we'd need to make sure that `separator` from `number_to_delimited` wouldn't step on its toes. 

It looks like there are three options.

1. Do as @pablobm suggests and convert `:delimiter` to just being a boolean.
2. Keep as is (i.e., explicit defining of delimiter), perhaps adding the option to pass `true` to default to the delimiter for the user's locale.
3. Update to allow passing of various options to `number_to_delimited`, as @sedubois suggests.

Thoughts? I'm happy to implement any of these the maintainers/etc. agree is the preferred direction they/you want to take Administrate.

pablobm: Good point on other available number helpers, @sedubois! This does complicate things. Still, I think it's worth looking into this issue, as `Number` is already a bit locale-unfriendly with its use of a format string like `"%.#{decimals}f"`.

Your idea of having two arguments `conversion_type` and `conversion_options` sounds appealing to me. One step further would be to allow a lambda to do the conversion, but I feel that could go too far, and `Number` could become a "universal" type of field.

Perhaps use different names, like `formatter` and `formatter_options`? I think this describes their intent better.

sedubois: > Perhaps use different names, like formatter and formatter_options

LGTM 👍

ianbayne: Thanks, @pablobm. I've now implemented this. Please take a look when you get the chance.

pablobm (approved):
Looks good! Thank you,

PR #1804, opened 2020-10-18
Enable Selectize for polymorphic fields

joanmc:
Use selectize for polymorphic fields as well as has many and belongs to fields


pablobm:
Good call! There's one tiny detail though: the group names are not styled any differently from normal items by default. As a result, I feel it can be confusing.

selectize-optgroup

Would you be able to add some styling to fix this?

pablobm:
Much better, thank you. Do you mind sorting those linter warnings, and we can then merge this?

joanmc:
> Much better, thank you. Do you mind sorting those linter warnings, and we can then merge this?

My bad, missed those - should be good now 👍

pablobm (approved):
Great; thank you!

PR #1805, opened 2020-10-21
Order by belongs_to field

sejinkim1904:
#### What does this PR do?

- Updates `Order` to be able to sort by `belongs_to` field
  - Updates `initialize` to take in `association_attribute`
  - Adds `order_by_association_attribute`
  - Adds `order_by_attribute`
  - Updates `order_by_association` to use `order_by_attribute` for `belongs_to` associations
  - Adds `association_has_attribute?`
- Adds `belongs_to_sorting_field` to `ApplicationController` to be overridden by users to select which field to sort by
- Adds `belongs_to_sorting_attribute` to `OrdersController` for testing
- Updates/adds tests to specs 
- Adds `Customizing belongs_to Sorting Field` section in `/docs/customizing_controller_actions.md`
  
#### How should this be manually tested?

- Start local server
- Visit `/admin/orders`
- Click on 'Customer'

#### Any background context you want to provide?

My team currently uses `administrate` in our application and we noticed when trying to sort by a `belongs_to` association, by default it sorts by the association's `id`. We found a workaround by sub-classing `Administrate::Order` in our `controller` to be able to sort by an association's field. Our implementation was very specific to our data model but I thought this feature could be useful for future or current devs so I decided to create a more dynamic version. 

#### Screenshots (if appropriate)

**Before:**
```ruby
class OrdersController < Admin::ApplicationController
end
```
before-admin-sorting

**After:**
```ruby
class OrdersController < Admin::ApplicationController
  def belongs_to_sorting_attribute
    :name
  end
end
```
after-admin-sorting

#### Questions:
  - Do Migrations Need to be ran? NO
  - Do Environment Variables need to be set? NO
  - Any other deploy steps? NO

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -102,7 +102,9 @@ module Administrate
 
     def order
       @order ||= Administrate::Order.new(
-        sorting_attribute, sorting_direction, belongs_to_sorting_attribute
+        sorting_attribute,
+        sorting_direction,
+        association_attribute: belongs_to_sorting_attribute,
       )
     end
 
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -101,11 +101,17 @@ module Administrate
     end
 
     def order
-      @order ||= Administrate::Order.new(sorting_attribute, sorting_direction)
+      @order ||= Administrate::Order.new(
+        sorting_attribute, sorting_direction, belongs_to_sorting_attribute
+      )
     end
 
     def sorting_attribute
       sorting_params.fetch(:order) { default_sorting_attribute }
+    end
+
+    def belongs_to_sorting_attribute
+      nil
     end
 
     def default_sorting_attribute
--- a/docs/customizing_controller_actions.md
+++ b/docs/customizing_controller_actions.md
@@ -70,3 +70,16 @@ def default_sorting_direction
   :desc
 end
 ```
+
+## Customizing belongs_to Sorting Field
+
+To set which field to use when sorting by a `belongs_to` association, you could override
+`belongs_to_sorting_field` in your dashboard controller like this:
+
+```ruby
+def belongs_to_sorting_field
+  :name
+end
+```
+
+If the field doesn't exist, the collection will be sorted by the associations `id`.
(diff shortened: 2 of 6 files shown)


pablobm (commented):
Thank you for this PR. I notice is that it's not obvious how to have more than one sorting attribute. This is useful in dashboards where there's more than one belongs-to association (for example with line items in the example app).

Playing a bit with this, I see I can use `params` to figure out what the sorting column is, but it's not very clear. Perhaps we could pass an argument to `belongs_to_sorting_attribute`?

sejinkim1904:
Requested changes made:
- Update `initialize` to take keyword argument
- Update `#order_by_association_attribute` to use reflection to find table name

In regards to dashboard with more than 1 `belongs_to` association:

I feel the association's attribute is going to be very specific to the user's database.  `#sortings_params` only provides the associations table name, so maybe we can update the README to incorporate instructions on how to handle dashboards with more than 1 `belongs_to` association.

```ruby
# With switch statement

class LineItemsController < Admin::ApplicationController
  def belongs_to_sorting_attribute
    case sorting_attribute
    when 'product'
      :name
    when 'order'
      :id
    end
  end
end
```

pablobm (commented):
I'm not sure about your proposal of using `sorting_attribute`. It works, but here we enter the difficult territory of what our interfaces communicate to developers, which has ramifications in usability and ability to understand Administrate with minimum documentation.

In my mind `belongs_to_sorting_attribute` does refer to something that can "have-many" values, and developers will have this expectation of it. Therefore it makes sense to me that the method signature does match this expectation. The sorting attribute can be read from the context as you propose, but I prefer keeping things functional where possible, as this helps communicating things better, and is less prone to mistakes derived from unexpected state.

nickcharlton:
@sejinkim1904, are @pablobm's comments something you'd be able to take a look at? It'd be great to move this one forwards!

jsrn:
Hey @nickcharlton & @sejinkim1904.

We'd love to see this feature progressed. Is there still an appetite for it? I'm happy to pick up this work if so.

Cheers.

pablobm:
@jsrn - Go for it! :slightly_smiling_face:

pablobm:
Superseeded by https://github.com/thoughtbot/administrate/pull/2208

Issue #1799, opened 2020-10-15
Fields are sometimes marked as required when they are not


pablobm:
Merging https://github.com/thoughtbot/administrate/pull/1521 gave Administrate the ability to detect "required" fields and mark them as such.

However, this detection is not always perfect. On occasion, there will be `:if`/`:unless` conditions attached to the validation rule. These are not correctly interepreted by Administrate. So for example:

```rb
  validates :amount, :currency, presence: true, if: ->{ stripe? }
```

<img width="351" alt="Screenshot 2020-10-14 at 18 11 58" src="https://user-images.githubusercontent.com/4217871/96016219-c7f3a600-0e48-11eb-8de9-dfd877de43a5.png">

_Originally posted by @sedubois in https://github.com/thoughtbot/administrate/pull/1521#issuecomment-708506703_

jmeinerz:
@pablobm this should be solved by #1808 (which you could argue open a new issue...)

pablobm:
Fixed by https://github.com/thoughtbot/administrate/pull/1808

PR #1808, opened 2020-10-23
Conditionals mark fields as optional (#1799)

jmeinerz:
We were assuming that all fields were required when a presence
validation existed. While that makes sense, it's also possible for
validations to be conditional.

Take the following validation as an example:

```ruby
validates :phone_number, presence: true, if: :egyptian?
```

Before this commit, the UI would flag phone_number as required, even for
records who were not egyptian.

We now always flag these fields as optional. This is a bit misleading
too, but it's impossible to know these things when the page is rendered,
and marking them as optional makes for a slightly better user interface,
as the user will most likely be prompted with detailed validation errors
after trying to persist an invalid item, rather than be led to fill some
fields that are, in fact, not mandatory.

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -17,22 +17,7 @@ module Administrate
     end
 
     def requireness(field)
-      required_field?(field) ? "required" : "optional"
-    end
-
-    def required_field?(field)
-      has_presence_validator?(field.resource.class, field.attribute)
-    end
-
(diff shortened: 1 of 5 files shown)


nickcharlton (approved):
Looks good!

Do you think it'd be worth having a feature spec here, or this enough to show the behaviour?

jmeinerz:
@nickcharlton I think the feature spec is a good call 👍 I'll add it as soon as I find some time. Thanks!

PR #1813, opened 2020-10-29
Document both demo apps consistently

nickcharlton:
Previously, the pre-release version of the Heroku app wasn't mentioned
consistently. Plus standardise to using the "demo", rather than
"prerelease" (as of this commit, there's no difference to each app).

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,6 +11,8 @@ home: true
 A framework for creating flexible, powerful admin dashboards in Rails.
 [Try the demo][demo].
 
+[demo]: https://administrate-demo.herokuapp.com/admin
+
 ### Warning:
 
 Administrate is still pre-1.0,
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -94,11 +99,17 @@ $ rails generate administrate:dashboard Foo
 ## Documentation
 
 To customize the appearance, behavior, and contents of the dashboard,
-see the guides at
-[https://administrate-prototype.herokuapp.com][prototype_heroku].
+we publish a set [of guides for the current release[released_docs].
```
pablobm, line 102: Is there a missing closing bracket `]` here?

pablobm (approved):
Looks good! Just one tiny detail.

README.md
nickcharlton: You are right!

PR #1815, opened 2020-10-29
generators: add views:index:actions

### LINKED ISSUE #1676: Custom additional action on actions column on index page
Custom additional action on actions column on the index page

The simplest way is to move the actions column into partial so that the user can override just the actions column part of the table. Instead of overriding the whole `collection` partial.
We also need to update the column heading count to also account for the additional custom action.

Another approach is probably to add ability to add an action button via the dashboard class. 
```
ACTIONS = [:start, :stop, :destroy, :show]

# Controller.rb
def start
resource.start!
end
```

Currently, we can override the whole `collection` partial. But it will lead to a greater incompatibility upon updating administrate.

bekicot:
This PR should close #1676

### DIFF
--- a/app/views/administrate/application/_action_headers.html.erb
+++ b/app/views/administrate/application/_action_headers.html.erb
@@ -1,0 +1,4 @@
+<% [valid_action?(:edit, collection_presenter.resource_name),
+    valid_action?(:destroy, collection_presenter.resource_name)].count(true).times do %>
+  <th scope="col"></th>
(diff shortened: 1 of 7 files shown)


bekicot:
Looks like there is an issue if there list sub resources. (e.g orders in customer show page)
Investigating it

bekicot:
Having no time to continue to this PR.

Apology to the people who have spent time in this.

Issue #1045, opened 2017-12-05
Missing data with has_many collection pagination


KidA001:
When I have a `Field::HasMany` collection in my view, the default pagination for that collection is 5. Setting the `per_page` param does not change how many are displayed and I cannot find any documentation to set the number of resources that are displayed in a `has_many` collection.

Using `administrate (0.8.1)` with rails `5.1.2` and Kaminari `1.1.1`

KidA001:
I found some PR mentioning `limit` in the ResourceDashboard::ATTRIBUTE_TYPES
`projects: Field::HasMany.with_options(limit: 20)`

When `limit` is higher than the number of resources in that collection, nothing displays.

nickcharlton:
Hi @KidA001!

Would you be able to provide a small example app so I can see this in action?

pablobm:
Closing due to lack of activity. If anyone is still experimenting this issue, and can provide more information, please let us know and we can reopen.

sedubois:
@pablobm I confirm that the issues still exists. I have a ProgramDashboard with `courses: Field::HasMany`. The program happens to have 13 courses. 5 courses are displayed and the rest can be viewed using the pagination ( 1 2 3 Next › Last » ). If I then change the config to `courses: Field::HasMany.with_options(sort_by: :position, limit: 10)`, OK now 10 courses are shown at once.

However if I change the limit to 15 (`courses: Field::HasMany.with_options(sort_by: :position, limit: 15)`), then no courses are shown at all, instead of showing the 13 courses I have.

pablobm:
Thank you @sedubois. I'm reopening then.

cernivec:
If you have a browser open on a show page that has Field::HasMany collection table and you change the limit, it is possible that the url params point to the page that doesn't exist anymore due to the collection now having less pages.

We added a [test](https://github.com/thoughtbot/administrate/pull/1840) that shows it’s not a bug, but most likely only appears when you change the limit and refresh your local dev page.

pablobm:
@sedubois - Do you think that could be the issue, changing the limit and then refreshing the page without resetting the query string?

sedubois:
Thanks @pablobm, indeed I can't reproduce this now, suggest to close!

pablobm:
Thank you!

PR #1840, opened 2020-12-04
Add customer show spec

cernivec:
For when has_many record number exceeds the limit per page for pagination

Issue: https://github.com/thoughtbot/administrate/issues/1045

Co-authored-by: David Anderson <david@awtdigital.co.uk>

### DIFF
--- a/spec/features/show_page_spec.rb
+++ b/spec/features/show_page_spec.rb
@@ -2,30 +2,50 @@ require "rails_helper"
 
 RSpec.describe "customer show page" do
   describe "paginates customers' orders" do
-    it "displays the first page by default, other pages when specified" do
-      customer = create(:customer)
-      orders = create_list(:order, 4, customer: customer)
-      order_ids = orders.map(&:id)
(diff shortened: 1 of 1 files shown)


spec/features/show_page_spec.rb
```diff
@@ -28,6 +29,22 @@
       expect(ids_in_table).to match_array(order_ids)
     end
 
+    it "displays the first page by default, other pages when specified where \
+      the total number of records don't exceed pagination per page limit" do
```
pablobm, line 33: What do you think about this phrasing? I think it reflects the situation a bit better:

```suggestion
    it "displays all results when the total number of records \
        doesn't exceed pagination per page limit" do
```

Also, since each has a context ("where exceeding (or not) the page limit"), perhaps these would read better if they used `context`?

```ruby
context "when the total number of records exceeds the pagination limit" do
  it "displays the first page by default, other pages on request"
end
context "when the total number of records does not exceed the pagination limit" do
  it "displays all records"
end
```

PR #1844, opened 2020-12-16
Select field selected value

nhippenmeyer:
When using the select field for a boolean attribute that can be either `true`, `false` or `nil`, the current behavior causes `false` and `nil` to be treated the same. If the attribute is set to `false`, the edit dashboard shows the selected value as being the `nil` option instead of `false` as expected.

### DIFF
--- a/app/views/fields/select/_form.html.erb
+++ b/app/views/fields/select/_form.html.erb
@@ -26,7 +26,7 @@ to be displayed on a resource's edit form page.
         field.selectable_options,
         :last,
         :first,
-        field.data.presence,
+        field.data,
       )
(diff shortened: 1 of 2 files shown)


pablobm:
Good catch :+1: Would it be possible to add a test? Perhaps a view spec similar to https://github.com/thoughtbot/administrate/blob/cb4a93746e919ba9b3b0db2506e1f049f8feb67e/spec/administrate/views/fields/url/_form_spec.rb

nhippenmeyer:
@pablobm done!

pablobm:
Thank you!

Issue #1837, opened 2020-12-02
Search yields "error: column thoughts.quote does not exist"


sedubois:
* What were you trying to do?

Basically all of our app's content is stored in DB and translated using [Mobility](https://github.com/shioyama/mobility). For instance the `Thought` model has `translates :quote` and `ThoughtDashboard` has `quote: Field::Text`. Listing, showing and editing the thoughts' quotes in the different languages from the admin dashboard works as expected (as in the main app, there is a language switcher so that the values are scoped to the relevant language). This works because behind the scenes, Mobility automatically generates the required getter and setter methods which it maps to ActiveRecord associations.

When using the administrate search box, thoughts which contain a quote containing the provided search term should be returned.

* What did you end up with (logs, or, even better, example apps are great!)?

It yields the error:
```
ActionView::Template::Error: PG::UndefinedColumn: ERROR:  column thoughts.quote does not exist
LINE 1: ...me" AS CHAR(256))) LIKE '%thought%' OR LOWER(CAST("thoughts"...
                                                             ^
```

It appears that because the dashboard is configured with `Field::Text`, Administrate assumes that the data is stored in a text column in the same table. However the data may well be stored in the same table but using JSON, or in a key-value polymorphic table, or (as in our case) in a belongs-to table, or why not even in other ways. Mobility has various backend strategies but provides a unified querying API (i.e. `thought.quote` "just works").

Mobility has an [integration with Ransack](https://github.com/shioyama/mobility-ransack) allowing for case-insensitive partial maches with `Thought.ransack(quote_cont: 'foo').result`. But even without Ransack, it's possible to do a basic exact-match search with `Thought.i18n.where(quote: "blah")`.

* What versions are you running?
  - Rails  6.1.0.rc1
  - administrate master 1a35ab210435995a5e3afcd1efd4163c9aff9f83

pablobm:
This is going to be a tricky one. Administrate's search currently makes some assumptions that don't hold well here. Ideally we would want to create a custom field type (eg: `Field::MobilityI18nText`) that described how to search it. However this is not currently supported, and it's not trivial to add.

An alternative would be creating your own search, instead of using Administrate's default one. Remove Administrate's search field, use your own one, send queries to a custom action, handle it, and render the result with an Administrate `index` view. I think that should work, but I haven't tried.

sedubois:
@pablobm OK below is a custom solution to get ILIKE search working for translated models. 

```diff
[...]

PR #1845, opened 2020-12-16
Add searchable option to Field::DateTime

Samemura:
Since `Field::DateTime` inherites `Field::Base`, `Field::DateTime` could be searchable if set `searchable` option to true.
I was wondering this is true or not because this document  does not describe it. I think it would be useful for this document.

I checked this behavior with
- Ruby 2.6.2
- administrate 0.11.0

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -209,6 +209,9 @@ objects to display as.
 `:timezone` - Specify which timezone `Date` and `DateTime` objects are based
 in.
 
+`:searchable` - Specify if the attribute should be considered when searching.
+Default is `false`.
+
 **Field::Date**
 
 `:format` - Specify what format, using `strftime` you would like `Date`


sedubois:
AFAIK Administrate uses the LIKE operator when performing a search, so I guess searching datetime would yield an error?

https://github.com/thoughtbot/administrate/blob/975f514673d7e091a231b7f7b68de3bea4ac8000/lib/administrate/search.rb#L86

My understanding is that Administrate's code needs to be overridden to specify how to search non-string columns or when the LIKE operator does not suffice. Ideally its API should be modified to allow this more easily or to broaden the supported scearios supported out of the box.

See related discussion: https://github.com/thoughtbot/administrate/issues/1837#issuecomment-741950596

pablobm:
I just tried myself. On the example app, I made `Order#shipped_at` searchable, and then searched with a string like `2020-11-30`.

It worked, so technically that's it. However I wonder if this is behaviour what we should document. In my mind, people will want to search in more "human" ways. For example: `30 November 2020`, which wouldn't work. Therefore it's not a great feature, and if we publicised it users would end up filing issues telling us that we should fix it.

So I think we shouldn't document it :-/

I'm going to close this, but feel free to continue the conversation.

PR #1868, opened 2021-01-14
Correct text of misleading specs

pablobm:
I was reading through these specs, and I think the descriptions are incorrect, as the condition is never actually checked. It looks like it was introduced by mistake at https://github.com/thoughtbot/administrate/pull/1808


jmeinerz (approved):
👏 This is way better phrasing! Thanks Pablo!

Issue #1865, opened 2021-01-11
fix syntax error in getting started documentation


caljess599:
```ruby
Rails.application.routes.draw do
  namespace :admin do
    # Add dashboard for your models here
    resources :customers,  # <== please remove this comma!
    resources :orders

    root to: "customers#index" # <--- Root route
  end
 end
 ```
 commas between lines not correct here

pablobm:
Thank you. Would you be able to provide a PR for this?

caljess599:
Duh, of course the documentation is in the repo! Submitted.

pablobm:
Fixed now :-) Thank you!

PR #1870, opened 2021-01-15
Update getting_started.md

caljess599:
remove comma from `Rails.application.routes.draw do` block

Solves #1865


pablobm:
Thank you @caljess599!

Issue #1860, opened 2021-01-07
Administrate doesn't read context on Model when load form for mandatory field


helmiItsavirus:
* What were you trying to do?

When I define validation inside of the context on my Model, Administrate doesn't read the context. So, Mandatory appears on the form that should not appear because of the mandatory inside of the context.

* What did you end up with (logs, or, even better, example apps are great!)?

```
class User < ApplicationRecord
  with_options on: [:update_birth_date] do
    before_validation :age_is_old_enough
    validates_presence_of :birth_date
  end
end
```
         
* What versions are you running?

- ruby "2.6.6"
- gem "rails", "~> 6.0.3", ">= 6.0.3.2"
- gem "administrate", "~> 0.14.0"

pablobm:
Thank you for your report. I don't entirely understand what your code block represents. Is it:

* A workaround?
* An example of code that doesn't work?

I'm trying to reproduce, but I'm not sure of what to do.

helmiItsavirus:
Hi, @pablobm 

Thanks for your response. Actually, I had a problem when the validation e.g "validates_precense_of" inside the context.

Administrate make the field that has validation inside the context become mandatory when I see it in the form.

I think Administrate should not make the field become mandatory in form because of the validation inside the context.

pablobm:
I see now. You are right, this is a bug. A fix is not entirely obvious as there are several ways to go about this. I have created a PR at https://github.com/thoughtbot/administrate/pull/1872. This will need a bit of discussion, and hopefully we'll be able to fix this.

pablobm:
By the way @helmiItsavirus: I'm not sure if you are aware, so I'll explain here. In your code, the problem is not the `with_options` (which I think is what you mean as "context"). Your code is equivalent to the following:

```ruby
class User < ApplicationRecord
  validates_presence_of :birth_date, on: [:update_birth_date]
end
```

It is actually the `:on` option that Administrate has an issue with. The way you asked confused me a bit, so I thought I'd explain.

helmiItsavirus:
@pablobm thanks for your correction with simple code, actually I'm using with_options to use the context even though too long looks like.

PR #1872, opened 2021-01-21
Correct "required" asterisk when using validation option `:on`

pablobm:
Fixes https://github.com/thoughtbot/administrate/issues/1860

We show a "required" asterisk when fields are required. However there are some edge cases, like when the requirement has an if/unless. We don't consider those actually required and we don't show the asterisk there.

But there are other edge cases. Issue #1860 shows how the validation option `:on` can also create cases where an apparent requirement is not actually a requirement. Normally this option has the values `:create` or `:update`, but there are other possibilities.

This PR tries to solve the issue. There are three commits, each an incremental improvement:

1. Consider `:on` to be the same as `:if` and `:unless`, and hide the asterisk.
2. Refactor specs to avoid the mystery guests involved in the examples for `:if`, `:unless`, and `:on`. This also allows for more scalable specs, should we decide to add more fine-grained handling of these options.
3. Consider the special (and most common) cases of `on: :create` and `on: :update`, and hide the asterisk for any other `:on` value.

So we could go just with the first commit, or the first two, or the whole thing. Or more than that! Opinions?

### DIFF
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -52,10 +52,25 @@ module Administrate
         return false unless resource.class.respond_to?(:validators_on)
 
         resource.class.validators_on(attribute).any? do |v|
-          v.class == ActiveRecord::Validations::PresenceValidator &&
-            !v.options.include?(:if) &&
-            !v.options.include?(:unless) &&
-            !v.options.include?(:on)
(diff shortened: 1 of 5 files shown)


nickcharlton:
I like this change! I think we should — and so I am going to — merge it in as is.

PR #1904, opened 2021-02-12
Add assets.precompile config to Engine

### LINKED ISSUE #1816: Sprockets 4 and Webpacker Config/Wiki
* What were you trying to do?

Install administrate with webpacker/sprockets 4 using the wiki

And error message about variables not being defined and about assets not being precompiled

  - Rails 6.0.3.4
  - administrate 0.14.0

It seems that the latest rails uses sass-rails 6 and sprockets 4 which creates a few issues for the current wiki instructions and webpacker. 

1) It seems the previous config/manifest.js had a line like this:
```
//= link_directory ../stylesheets .css
```
When generating the assets from administrate to modify for the webpacker js changes, it will compile all of the .css under stylesheets out of order and throw errors about things like base/form not having variables defined.

More details here https://github.com/rails/sprockets/issues/597

2) The fix for above is to change that line to just //= link application.css, but application.css by default has this line, which ultimately causes the same problem. 
```
*= require_tree .
```

In the end I got it working by having this in my manifest.js:

```
//= link_tree ../images
//= link application.css
//= link administrate/application.css
```

And removing the require_tree . from application.css

Spone:
This is a PR to apply the fix suggested by https://github.com/thoughtbot/administrate/issues/1816#issuecomment-738414440

Close #1816

### DIFF
--- a/lib/administrate/engine.rb
+++ b/lib/administrate/engine.rb
@@ -25,7 +25,7 @@ module Administrate
     initializer "administrate.assets.precompile" do |app|
       app.config.assets.precompile += [
         "administrate/application.js",
-        "administrate/application.css"
+        "administrate/application.css",
       ]
     end
 
--- a/lib/administrate/engine.rb
+++ b/lib/administrate/engine.rb
@@ -21,9 +21,12 @@ module Administrate
 
     @@javascripts = []
     @@stylesheets = []
-    
+
     initializer "administrate.assets.precompile" do |app|
-      app.config.assets.precompile += %w( administrate/application.js administrate/application.css )
+      app.config.assets.precompile += [
+        "administrate/application.js",
+        "administrate/application.css"
+      ]
     end
 
     def self.add_javascript(script)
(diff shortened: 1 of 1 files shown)


pablobm (approved):

pablobm:
This looks good to me. I tested it in the following situations:

* Several versions of Rails pre 6.0.
* Rails 6.0.
* Rails 6.0 without Sprockets.
* Rails 6.0 without Webpacker.

Rails pre-6.0 is not affected as it doesn't have Webpacker. For the other three, this PR removes the issue of having to manually update `manifest.js` when Administrate is installed, and it appears to work well.

My tests were not very comprehensive, but between this and the test suite passing (which includes some JS-dependant code), I feel confident enough that this is a good change.

@nickcharlton Any thoughts? Double checking before I merge as this might be an important change.

nickcharlton:
This was a bit fun to test! I tried with [a script I have to make demo apps][1] but couldn't replicate the original problem but tried it on my current client project where I could. My feeling is that I'm happy with that enough, because clearly this change doesn't _break_ the old implementation and does work where it should.

Thanks for this! I'm going to merge this now.

[1]: https://github.com/nickcharlton/dotfiles/blob/master/bin/make-administrate-demo

PR #1909, opened 2021-02-15
Rely on Rails' `route` method to be correct

nickcharlton:
Previously, we were searching for a `routes.rb` to verify that we'd
already added the routes or not. We don't need to do this: the `route`
method will only ever inject into the `config/routes.rb` file it knows
about — so us looking elsewhere would never make a difference to the
outcome of the generator.

On top of this, the `route` method in turn leans on `inject_into_file`
from Thor which is configured to not force the result.

So, much of this can go. We just continue to check that we don't
generate routes with no models.

https://github.com/rails/rails/blob/291a3d2ef29a3842d1156ada7526f4ee60dd2b59/railties/lib/rails/generators/actions.rb#L269
https://github.com/erikhuda/thor/blob/34df888d721ecaa8cf0cea97d51dc6c388002742/lib/thor/actions/inject_into_file.rb#L26


cernivec (approved):
👍

PR #1910, opened 2021-02-16
Fix an issue where loading in routes blew up

nickcharlton:
Rails has a method for reloading routes and it will reliably work —
loading in the file again throws an exception on more complex routing
structures.


nickcharlton:
I extracted this from #1909, as it feels better to keep this more specific.

PR #1930, opened 2021-03-04
Update instructions for controllers without models

chozandrias76:
This should include directions to use `ActionController::Base` for the controller instead of `Admin::ApplicationController`.
If you use Admin::ApplicationController, you get ActiveRecord related issues trying to access the "resource"
image

### DIFF
--- a/docs/adding_controllers_without_related_model.md
+++ b/docs/adding_controllers_without_related_model.md
@@ -34,7 +34,7 @@ end
 ```ruby
 # app/controllers/admin/stats_controller.rb
 module Admin
-  class StatsController < Admin::ApplicationController
+  class StatsController < ActionController::Base
     def index
       @stats = {
         customer_count: Customer.count,


pablobm:
I think this depends. I suspect that your controller is not overriding the `index` action? If you don't do it (even if it's by just providing an empty action `def index; end`), then Administrate's own `index` will be called, resulting in a call to `scoped_resource` and this error.

There are arguments for doing this one way or the other. The reason (I think...) why we go with `Admin::ApplicationController` is that this way you get Administrate's layout by default. You can see this in the example app: https://administrate-prototype.herokuapp.com/admin/stats

Thoughts?

nickcharlton:
I'm going to close this, as I don't think this is the right solution and the docs do mention the need to add an `index` route.

_However_: If there's anything you can think can be improved in the existing docs, we'd more than welcome improvements to the wording!

PR #1932, opened 2021-03-06
Add a basic GitHub Actions CI workflow

nickcharlton:
This is the closest absolute conversation from the original CircleCI 2.0
version as possible. Main differences:

* Actions' Postgres container requires a password or otherwise doesn't
  start (see SO answer),
* Ports are explicitly exported,
* Actions has a stronger, built-in approach to service health,
* Using the "global" environment variables stops us having to repeat the
  Postgres config multiple times,

https://docs.github.com/en/actions/guides/building-and-testing-ruby
https://docs.github.com/en/actions/guides/creating-postgresql-service-containers
https://docs.github.com/en/actions/reference/environment-variables#about-environment-variables
https://docs.github.com/en/actions/guides/creating-redis-service-containers#running-jobs-directly-on-the-runner-machine
https://stackoverflow.com/a/60618750/83386
https://www.postgresql.org/docs/9.5/libpq-envars.html

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,54 @@
+---
+name: CI
+on: [push]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        ruby: ['3.0', 3.1, 3.2]
+    env:
+      PGHOST: localhost
+      PGUSER: administrate
+      PGPASSWORD: administrate
+    services:
+      postgres:
+        image: postgres
+        env:
+          POSTGRES_USER: administrate
+          POSTGRES_DB: administrate_test
+          POSTGRES_PASSWORD: administrate
+        ports:
+           - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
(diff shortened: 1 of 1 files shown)


pablobm (approved):

nickcharlton:
This is currently blocked on us merging https://github.com/thoughtbot/appraisal/pull/174.

pablobm:
Is this something that should be incorporated? https://github.com/thoughtbot/administrate/pull/1967

nickcharlton:
I'm going to do the Dependabot migration before coming back to this (i.e.: today).

nickcharlton:
This is now back up to date again, using a similar approach to how we've been running appraisals on certain Ruby versions by skipping Rails 6.0 above Ruby 3.0.

I think string comparisons are fine here, but other ideas welcome.

---

I don't plan to merge this until I've cut a new release of Appraisal, so once that's done our CI checks will take even longer until I disable CircleCI.

nickcharlton:
Okay, so, 2 years later, I was able to get Appraisal working on GitHub Actions and then do a release. And now I've had enough time to tidy this up and finally merge it.

I'll open a PR soon to remove Circle CI, but it'll be a while before it's gone completely.

PR #1933, opened 2021-03-08
Update bin/setup

nickcharlton:
These are a couple of changes that make `bin/setup` work well again:

1. Switch PhantomJS for Chromedriver,
2. Stop joining the Heroku app, as most people won't,
3. Drop suggestion/requirement for Foreman.


jsilasbailey (approved):
LGTM

Issue #1861, opened 2021-01-07
Check if an associated model's route exists before trying to link it


pablobm:
Over at StackOverflow, a user had an issue: https://stackoverflow.com/questions/65538115/administrate-gem-does-not-allow-to-view-user

I was able to reproduce it on the example app as follows:

1. Remove `resources :orders` from the routes file.
2. Visit the "show" page of a customer.

The page breaks because Administrate is trying to link to the orders of the customer. However orders do not have routes, so there's nothing to link.

Administrate is checking with `show_action?` before rendering the link, but that only checks for authorization. It should be checking with `valid_action?`, which does check for the existence of the route.

This brings a new question: should we have only one such helper, which checks for both authorization and existence?

nickcharlton:
Heh, I saw this problem recently myself. One helper would work well, I think: if you're allowed to view something that's sort of moot if it doesn't exist in the first place!

PR #1941, opened 2021-03-11
Unify action checks

pablobm:
Fixes: https://github.com/thoughtbot/administrate/issues/1861

When referring to a route in the code, we run two checks:

* `valid_action?` is `true` if the route is defined, `false` otherwise.
* `show_action?` is expected to be overridden by developers in order to implement authorization. For example, it's implemented by `Administrate::Punditize` in order to integrate Administrate with Pundit. It should return `true` if the current user can access a given route, `false` otherwise.

These two check should (almost) always happen together. For this reason, our code is peppered with `if` statements where both are checked... and a few others where we forget one or the other.

These checks should be unified into a single method call, in order to avoid issues like the one described at https://github.com/thoughtbot/administrate/issues/1861. Here I propose such a method, which I call `accessible_action?`.

The original methods should still exist, as they do have their uses individually. The new method will delegate to the existing ones.

I have also renamed the two existing methods to something that I hope will make their intent clear:
- `valid_action?` becomes `existing_action?`
- `show_action?` becomes `authorized_action?`
- In order to provide a clear upgrade path, the old names still exist and work, but they show a deprecation warning when used. They can be removed properly at a later version of Administrate.

### Open question!
The original methods have a signature `the_old_question?(action, resource)`. I have flipped the order of parameters to be `the_new_question?(resource, action)`, as I find that more natural. But if people think the original order is better (or have some other proposal), I'm happy to oblige.

Also! The old `valid_action?` has a default parameter value (the full signature is `valid_action?(name, resource = resource_class)`). The new version doesn't have it, as I can't see that it has a good use case. Incidentally, this makes sense with the parameter order flip just described.

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -59,13 +59,13 @@ to display a collection of resources in an HTML table.
   <tbody>
     <% resources.each do |resource| %>
       <tr class="js-table-row"
-          <% if valid_action?(:show, resource.class) && show_action?(:show, resource) %>
+          <% if accessible_action?(resource, :show) %>
             <%= %(tabindex=0 role=link data-url=#{polymorphic_path([namespace, resource])}) %>
           <% end %>
           >
         <% collection_presenter.attributes_for(resource).each do |attribute| %>
           <td class="cell-data cell-data--<%= attribute.html_class %>">
-            <% if valid_action?(:show, resource.class) && show_action?(:show, resource) -%>
+            <% if accessible_action?(resource, :show) -%>
               <a href="<%= polymorphic_path([namespace, resource]) -%>"
                  tabindex="-1"
                  class="action-show"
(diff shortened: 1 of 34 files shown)


app/views/administrate/application/_collection.html.erb
```diff
@@ -45,23 +45,21 @@ to display a collection of resources in an HTML table.
           <% end %>
         </th>
       <% end %>
-      <% [valid_action?(:edit, collection_presenter.resource_name),
-          valid_action?(:destroy, collection_presenter.resource_name)].count(true).times do %>
-        <th scope="col"></th>
-      <% end %>
```
c4lliope, line 51: I'm glad you simplified this; no need to make things complex when we can render simple blank columns.

c4lliope (approved):
This change looks good; for your checklist, maybe you could rename:
* `valid_action?` -> `possible_action?`
* `show_action?` -> `permissible_action?`

...though I'm unsure because I'm barely involved here.
I'm sure you'll choose sensible names.

Much obliged for your changes, @pablobm !

pablobm:
@c4lliope @nickcharlton - After more than a year, this PR is finally out of draft!

nickcharlton (approved):
🎉

PR #1955, opened 2021-04-02
Bring Appraisal gemfiles up to date

pablobm:
I get these gemfile updates when I run `./bin/appraisal`. I don't know what the mechanism is, but I figured that this is what ultimately runs on CI, so we should keep the files up to date in the repo.


nickcharlton (approved):
Yeah — this is what I do.

I've been (slowly) working on a GitHub Action to do this for us …one day!

PR #1956, opened 2021-04-07
css: isolate spacing properties for 'button--alt' class

paulRbr:
When customising your own admin dashboards, it is quite expected to
use consistent graphical elements (such as buttons) in custom pages.

Exposing a `button--alt` CSS class is helpful when adding some custom
buttons, however the existing CSS class includes some spacing
properties which is specific to the navigation bar (because this class
is only used there in the Administrate engine).

This PR moves the spacing property (margin) into a dedicated
`.button--nav` class in order to give the `.button--alt` class a
reusable role.


nickcharlton (approved):

nickcharlton:
Thanks, seems like a good idea!

PR #1960, opened 2021-04-16
Fix deprecation warning about dots in paths

pablobm:
This is a deprecation introduced in Rails 6.1. For example:

```
DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated:
fields/time/_index.html.erb (called from block (3 levels) in
<top (required)> at $WORDIR/spec/administrate/views/fields/time/_index_spec.rb:24)
```


nickcharlton (approved):
Ooh! That's what it's referring to!

PR #1963, opened 2021-04-24
Namespace generators

RaeRachael:
link to the old PR for disscussion https://github.com/thoughtbot/administrate/pull/1812


pablobm (changes requested):
I'm going to explain a bit more the concept of separate classes that I first introduced at https://github.com/thoughtbot/administrate/pull/1812#discussion_r540294128

Currently, `dashboard_resources` returns something like this:

```ruby
["customers", "orders", "blog/posts", "countries", "line_items", "log_entries", "pages", "payments", "products", "product_meta_tags", "series"]
```

This could be more complicated: more levels of namespacing, nested resources, etc.

What my proposed design, we'd have classes that take the output of `dashboard_resources` and turn it into a data structure representing the desired route tree in the abstract.

Then this result would be read by code that would actually turn this abstract tree into strings, which will go in the routes file.

I mentioned last time how `generate_resource_routes` is a bit hidden, in that a reader needs two leaps (template to generator, then to mixin) in order to find it. My proposal is that it lives in the generator, looking something like this:

```ruby
module Administrate
  module Generators
    class RoutesGenerator < Rails::Generators::Base
      # ...

      def generate_resource_routes
        rtc = RouteTreeCalculator.new(resources)
        RouteTreePrinter.new(rtc.to_tree).to_s
      end

      # ...
    end
  end
end
```

With this design, we get these `RouteTreeCalculator` and `RouteTreePrinter`, which are classes that can be tested separately. We can then have specs that throw all manner of edge cases at them, and we can test these without having to run the entire generator.

Examples of use of these separate classes (to be tested in the spec):

```ruby
rtc = RouteTreeCalculator.new(['blog/posts', 'blog/posts/authors'])
tree = rtc.to_tree
# =>
# [
#   {
#     type: "namespace",
#     name: "blog",
#     children: [
#       type: "resources",
#       name: "posts",
#       children: [
#         {
#           type: "resources",
#           name: "authors"
#         }
#       ]
#     ]
#   }
# ]

rtp = RouteTreePrinter.new(tree)
rtp.to_s
# =>
# namespace :blog do
#   resources :posts do
#     resources :authors
#   end
# end
```

Let me know if you still have questions.

pablobm:
Closing due to lack of activity

Issue #1957, opened 2021-04-08
Pagination broken in 0.15 for customized views


mmrobins:
tl;dr the `CHANGELOG.md` should call out `* [BUGFIX] [#1725] Fix pagination of "Page" models.` as breaking pagination for customized views.

After I upgraded 0.14.0 -> 0.15.0 I started getting issues with pagination

I have `config.action_controller.action_on_unpermitted_parameters = :raise`, so started seeing errors

```
ActionView::Template::Error (found unpermitted parameter: :page):
```
 
 I tried disabling that, but ended getting the `page` param stripped, so paging didn't work.

```
[2021-04-08T16:08:52.592-07:00] INFO: Request
{
  :method             => "GET",
  :path               => "/admin/myresource",
  :format             => :html,
  :controller         => "Admin::MyResoureController",
  :action             => "index",
  :status             => 200,
  :duration           => 107.39,
  :view               => 85.79,
  :db                 => 13.58,
  :unpermitted_params => [
    "page",
    "page",
    "page",
    "page"
  ],
  :request_id         => "cf594d02-436f-4bc2-8747-1b655a95c41d",
}
```
 
 I finally tracked down the reason being https://github.com/thoughtbot/administrate/pull/1725 changing the paging parameter from `page` to `_page`.  This would be fine, but if you've customized your `index.html.erb` at all, you won't get the update to fix that rename.
 
 My fix ended up being pretty, simple, just unexpected to need it
 
 ```
 --- app/views/admin/application/index.html.erb
+++ app/views/admin/application/index.html.erb
@@ -70,5 +70,5 @@ It renders the `_table` partial to display details about the resources.
     table_title: "page-title"
   ) %>

-  <%= paginate resources %>
+  <%= paginate resources, param_name: '_page' %>
 </section>
```

I looked through the `CHANGELOG.md` but didn't see anything that would have led me to expect this, so posting here in case anyone else bangs there head against this.

It would be great if versioning for this could bump to `1.0` so that future version bumps could use semver to communicate breaking changes more clearly.  I get the project wants to be able to make breaking changes, but it seems mature enough that communicating that through semver would be pretty helpful.

Thanks for the taking the time to look through this, and for the great library.

pablobm:
Thank you for reporting @mmrobins. You are right that this is a breaking change and we should have communicated it.

[...]

PR #1968, opened 2021-04-29
Script to list recent changes to templates

pablobm:
A first stab at a script to help with writing changelogs. Currently, it lists which template files have changed since the version given as an argument (actually a Git tag):

```
$ ./bin/changelog v0.15.0
The following templates have changed since v0.15.0:

  app/views/administrate/application/_navigation.html.erb
  app/views/fields/url/_index.html.erb
  app/views/fields/url/_show.html.erb

If your application overrides any of them, make sure to review your
custom templates to ensure that they remain compatible.
```

This was inspired by https://github.com/thoughtbot/administrate/issues/1957, where a user observed that it would be worth mentioning these template updates in the changelog, as it's something that people need to look out for when upgrading Administrate.

Perhaps in the future we can add other capabilities to this script.

### DIFF
--- a/bin/changelog
+++ b/bin/changelog
@@ -1,0 +1,26 @@
+#!/usr/bin/env ruby
+
+if ARGV.count != 1
+  script = $0
+  puts "Usage:"
+  puts "  #{script} <git reference>"
+  puts
+  puts "Example:"
+  puts "  #{script} v0.15.0"
+  exit 1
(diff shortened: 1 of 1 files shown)


nickcharlton:
I used this for the 0.16.0 release and it worked well. So I'm going to merge it! Thank you!

PR #1972, opened 2021-05-05
Support Rails 6.1.3.2 & 6.0.3.7 and use symbols for polymorphic routes

### LINKED ISSUE #1971: Rails 6.1.3.2 breaks Administrate
* What were you trying to do?

Upgrade to the latest version of Rails

An `ActionView::Template::Error`  (see below)

  - Rails 6.1.3.2
  - administrate 0.15.0

https://github.com/rails/rails/commit/c4c21a9f8d7c9c8ca6570bdb82d64e2dc860e62c requires that things passed to route helpers be symbols and not strings. https://github.com/thoughtbot/administrate/blob/main/app/views/administrate/application/index.html.erb#L44 and L56 break with 

```
     ActionView::Template::Error:
       Please use symbols for polymorphic route arguments.
```

I'm not sure how prevalent this is within Administrate since I was just testing the simple show case when I tripped on this.

lostmahbles:
This potentially fixes #1971. I didn't do a thorough dive, but these changes got our admin back up and running. This also encompasses the changes in #1970.


lib/administrate/namespace/resource.rb
```diff
@@ -4,7 +4,7 @@ class Resource
       attr_reader :namespace, :resource
 
       def initialize(namespace, resource)
-        @namespace = namespace
+        @namespace = namespace.to_sym
```
cgrothaus, line 7: **question:** is this necessary? 

I have the feeling that `Resource` as an inner class of `Namespace` is only ever initialized from there. At least I found now other places when grepping the codebase. And in `Namespace`,  `namespace` is already made sure to be a symbol.

nickcharlton: Good catch! I removed this in the branch I'm trying to test and I think you're right.

nickcharlton:
I've tested this on a thoughtbot project which uses Administrate fairly extensively, plus I'm trusting the now passing test suite to be capturing the original error before the symbol change successfully. I'm going to merge this once CI passes and then sort out a new release today.

If anyone's following along and having other issues relating to these Rails releases (both 6.1.3.2 and to 6.0.3.7) I'd love to hear about them!

ELepolt:
Utilizing this commit I have the following two lines, and I am still getting this error. Any thoughts on which part is still upset?

```
- Administrate::Namespace.new(namespace).resources.each do |resource|
    = link_to(display_resource_name(resource), [namespace.to_sym, resource.path], class: "navigation__link navigation__link--#{nav_link_state(resource)}")
```

xvega:
Still having an error with this piece of code `redirect_to [namespace, resource]`

cgrothaus:
I think `resource.path` is a String, and that is causing the error in Rails routing. It must be a Symbol instead, try `resource.path.to_sym`. Also, I think that `namespace` is already a Symbol, thanks to the fix, so you can omit `.to_sym` there.

kemenaran:
You may need to update your _navigation.html.erb according to the latest template from administrate : 

```diff
  <%= link_to(
      display_resource_name(resource),
-      [namespace, resource.path],
+      resource_index_route(resource),
      class: "navigation__link navigation__link--#{nav_link_state(resource)}"
    ) %>
```

See https://github.com/betagouv/demarches-simplifiees.fr/pull/6178/files#diff-f27c4d37d2b93f892244d908e50fc93837814f8e0ad85697a8dde21076ce1df0

PR #1973, opened 2021-05-05
feat: refactor page redirect target path

saiqulhaq:
this will make us easier to override redirect path

should I make test & update the doc?

PR #1982, opened 2021-05-11
Fix redirect after delete resource without index

### LINKED ISSUE #626: question: Can I hide dashboards from the left menu but keep them in as relationships?

edimossilva:
fix #626

**Steps to reproduce**
- `Resourse A` has many `Resource B` 
- `Resourse B` has no index page
- go to show page of `Resourse A` and click on destroy of `Resource B` 
- resource is destroyed

**Behavior before this fix**
- **An error is raised**
image

**Behavior after this fix**
- **is redirected `Resourse A` show page**

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -76,7 +76,12 @@ module Administrate
       else
         flash[:error] = requested_resource.errors.full_messages.join("<br/>")
       end
-      redirect_to after_resource_destroyed_path(requested_resource)
+      after_action = after_resource_destroyed_path(requested_resource)[:action]
+      if valid_action?(after_action, resource_class)
+        redirect_to after_resource_destroyed_path(requested_resource)
+      else
+        redirect_to params[:origin_url]
+      end
     end
 
     private
@@ -104,6 +109,12 @@ module Administrate
       !!routes.detect do |controller, action|
         controller == resource.to_s.underscore.pluralize && action == name.to_s
       end
+    end
+
+    helper_method :origin_url_params
+    def origin_url_params(page)
+      is_show_page = page.instance_of?(Administrate::Page::Show)
+      { origin_url: polymorphic_path([:admin, page.resource]) } if is_show_page
     end
 
     def routes
--- a/app/views/administrate/application/_collection_item_actions.html.erb
+++ b/app/views/administrate/application/_collection_item_actions.html.erb
@@ -9,7 +9,7 @@
 <% if valid_action?(:destroy, collection_presenter.resource_name) %>
   <td><%= link_to(
     t("administrate.actions.destroy"),
-    [namespace, resource],
+    [namespace, resource, origin_url_params(page)],
     class: "text-color-red",
     method: :delete,
     data: { confirm: t("administrate.actions.confirm") }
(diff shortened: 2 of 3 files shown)


edimossilva:
@pablobm

pablobm (commented):
Thank you @edimossilva!

I see why you had to create a new model `MetaTagGroups`. The way the example app is put together is not great. I'm currently working on a branch that will bring some order to this.

A few days before you, someone created a related PR: https://github.com/thoughtbot/administrate/pull/1973. I think yours may complement it. I feel that, instead of creating this new query param `origin_url`, perhaps the same effect could be achieved by overriding `after_resource_destroyed_path`, as provided by the other PR. What do you think? If you think it makes sense, we should wait until the other PR is merged before rebasing yours on top.

edimossilva:
> Thank you @edimossilva!
> 
> I see why you had to create a new model `MetaTagGroups`. The way the example app is put together is not great. I'm currently working on a branch that will bring some order to this.
> 
> A few days before you, someone created a related PR: #1973. I think yours may complement it. I feel that, instead of creating this new query param `origin_url`, perhaps the same effect could be achieved by overriding `after_resource_destroyed_path`, as provided by the other PR. What do you think? If you think it makes sense, we should wait until the other PR is merged before rebasing yours on top.

Hi @pablobm , PR #1973 looks nice. I agree that the best way is overriding `after_resource_destroyed_path`. Let's wait for the merge of #1973 and rebase this one.

About the new model `MetaTagGroups`. Should I wait for your branch as well?

pablobm:
I'd say let's wait for #1973 first and then let's see where my branch is when it arrives, and if we can use it to simplify your spec.

pablobm:
This is the branch that reorganises the example app: https://github.com/thoughtbot/administrate/pull/1998. When it's merged, we can revisit this PR.

app/controllers/administrate/application_controller.rb
```diff
@@ -76,7 +76,12 @@ def destroy
       else
         flash[:error] = requested_resource.errors.full_messages.join("<br/>")
       end
-      redirect_to after_resource_destroyed_path(requested_resource)
+      after_action = after_resource_destroyed_path(requested_resource)[:action]
+      if valid_action?(after_action, resource_class)
+        redirect_to after_resource_destroyed_path(requested_resource)
+      else
+        redirect_to params[:origin_url]
+      end
```
pablobm, line 84: This code assumes that `after_resource_destroyed_path` always returns a hash with an `:action` key. This will probably not be the case when someone overrides it. (And it was created specifically to be overridden).

We need some other way to tell if the route exists. Perhaps we can do something with `url_for`? (No idea from the top of my head).

Alternative idea: only provide `origin_url` if the show page doesn't exist. This means checking the existence of the route when rendering the link, instead of when following the link. Then we always follow `origin_url` if present.

pablobm (changes requested):
Hello @edimossilva; apologies for the late review; I have very little time these days. I'm still trying to catch up with the Administrate backlog, but it will take some time.

I think there's an issue with your PR (sorry...). Can you see if you can improve on that? I'll try to be more responsive on this PR in the future.

pablobm:

PR #1991, opened 2021-05-25
Allow different form attributes for new/update actions

kolasss:


lib/administrate/base_dashboard.rb
```diff
@@ -50,7 +50,12 @@ def all_attributes
...
 
-    def form_attributes
+    def form_attributes(action = nil)
+      if action
+        cname = "FORM_ATTRIBUTES_#{action.upcase}"
+        return self.class.const_get(cname) if self.class.const_defined?(cname)
+      end
+
       self.class::FORM_ATTRIBUTES
     end
```
pablobm, line 60:
```suggestion
    def form_attributes(action = nil)
      specific_form_attributes_for(action) || self.class::FORM_ATTRIBUTES
    end

    def specific_form_attributes_for(action)
      return unless action

      cname = "FORM_ATTRIBUTES_#{action.upcase}"

      self.class.const_get(cname) if self.class.const_defined?(cname)
    end
```

lib/administrate/page/form.rb
```diff
@@ -10,8 +10,13 @@ def initialize(dashboard, resource)
...
       attr_reader :resource
 
-      def attributes
-        dashboard.form_attributes.map do |attribute|
+      def attributes(action = nil)
+        action = case action
+                 when "update" then "edit"
+                 when "create" then "new"
+                 else action
+                 end
```
pablobm, line 18: I have a very personal yet deep aversion to this sort of indentation :see_no_evil: How about the following?

```suggestion
        action =
          case action
          when "update" then "edit"
          when "create" then "new"
          else action
          end

```

Also added a space at the end.

pablobm (commented):
Thank you for the PR @kolasss! I think it's pretty good. If you can just have a look at my comments, I think we can get this merged soon.

Regarding your question about the Rubocop config, I think Hound uses the rules found at https://github.com/thoughtbot/guides/blob/main/ruby/.rubocop.yml

pablobm (approved):
Perfect. Thank you!

alesrosina:
This thing looks like I would need - is there any plan on making next gem release? Current `0.16.0` does not yet include this change.

PR #1992, opened 2021-05-26
Add dockerfile & docker-compose

kolasss:
Docker image for easier development


nickcharlton:
Thanks for opening this! Unfortunately, I don't think adding a `Dockerfile` is necessary or valuable for Administrate as it's a very simple Rails app and this would add to the maintenance burden so I'm going to close this.

PR #1995, opened 2021-06-03
Refactor page redirect path after 'destroy', 'create', 'update' 

edimossilva:
@pablobm , please review

This is a follow-up of PR https://github.com/thoughtbot/administrate/pull/1973/

Should I squash the commits?

hound is complaining about something but I think it is not relevant, and maybe this rule should be disabled => https://github.com/rubocop/rubocop/issues/3675

### DIFF
--- a/docs/customizing_controller_actions.md
+++ b/docs/customizing_controller_actions.md
@@ -70,3 +70,21 @@ def default_sorting_direction
   :desc
 end
 ```
+
+## Customizing Redirects after actions
+
+To set custom redirects after the actions `create`, `update` and `destroy` you can override `after_resource_created_path`, `after_resource_updated_path` or `after_resource_destroyed_path` like this:
+
+```ruby
+    def after_resource_destroyed_path(_requested_resource)
+      { action: :index, controller: :some_other_resource }
+    end
+
+    def after_resource_created_path(requested_resource)
+      [namespace, requested_resource.some_other_resource]
+    end
+
+    def after_resource_updated_path(requested_resource)
+      [namespace, requested_resource.some_other_resource]
+    end
+```
--- a/spec/controllers/admin/application_controller_spec.rb
+++ b/spec/controllers/admin/application_controller_spec.rb
@@ -1,0 +1,41 @@
+require "rails_helper"
+
+RSpec.describe Admin::OrdersController, type: :controller do
+  controller(Admin::OrdersController) do
+    def after_resource_destroyed_path(_requested_resource)
+      { action: :index, controller: :customers }
+    end
+
+    def after_resource_created_path(requested_resource)
+      [namespace, requested_resource.customer]
+    end
+
+    def after_resource_updated_path(requested_resource)
+      [namespace, requested_resource.customer]
+    end
+  end
+
+  it "redirect to custom route after destroy" do
+    order = create(:order)
+
+    delete :destroy, id: order.to_param
+    expect(response).to redirect_to(admin_customers_path)
+  end
+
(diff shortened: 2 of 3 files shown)


docs/customizing_controller_actions.md
```diff
@@ -69,4 +69,22 @@ end
 def default_sorting_direction
   :desc
 end
-```
\ No newline at end of file
+```
+
+## Customizing Redirects after actions
+
+To change redirect after the actions `create`, `update` and `destroy` you could override `after_resource_created_path` or `after_resource_updated_path` or `after_resource_destroyed_path` like this
```
pablobm, line 76:
```suggestion
To set custom redirects after the actions `create`, `update` and `destroy` you can override `after_resource_created_path`, `after_resource_updated_path` or `after_resource_destroyed_path` like this:
```

spec/controllers/admin/application_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def after_resource_created_path(requested_resource)
+      [namespace, requested_resource.customer]
+    end
+
+    def after_resource_updated_path(requested_resource)
+      [namespace, requested_resource.customer]
+    end
+  end
+end
```
pablobm, line 17: Creating modules like this in spec files can create issues, as the modules stay around afterwords, leaving the state modified. For example, a spec that runs after this one would get this implementation of the hooks.

A better way to implement this is to use RSpec's "anonymous controller" feature, which allows you to create a disposable child class of the controller you are testing: https://relishapp.com/rspec/rspec-rails/docs/controller-specs/anonymous-controller

spec/controllers/admin/application_controller_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+RSpec.describe Admin::OrdersController, type: :controller do
```
pablobm, line 19: Following my comment above, I think you can implement the same effect here, like this:

```suggestion
RSpec.describe Admin::OrdersController, type: :controller do
  controller do
    def after_resource_destroyed_path(_requested_resource)
      { action: :index, controller: :customers }
    end

    def after_resource_created_path(requested_resource)
      [namespace, requested_resource.customer]
    end

    def after_resource_updated_path(requested_resource)
      [namespace, requested_resource.customer]
    end
  end

```

pablobm (commented):
Thank you for taking over this one! I agree that those Hound comments are not important. I also get them and I ignore them happily.

I have left a couple of comments. After that we should be good to go.

spec/controllers/admin/application_controller_spec.rb
edimossilva: we still have to use controller(Admin::OrdersController), but thanks a lot, good to learn that

spec/controllers/admin/application_controller_spec.rb
edimossilva: @pablobm , thanks a lot for the explanation, really nice!

edimossilva:
> Thank you for taking over this one! I agree that those Hound comments are not important. I also get them and I ignore them happily.
> 
> I have left a couple of comments. After that we should be good to go.

I've done the suggested changes, please review again @pablobm

pablobm:
Thank you @edimossilva! Incidentally, I found what was wrong with those Hound errors on `get`, `post`, etc. It was because of a shim in our code. I'm fixing that on a different PR: https://github.com/thoughtbot/administrate/pull/2001 I'll rebase on top of your code to also fix the ones that you introduce here.

PR #1996, opened 2021-06-06
Opt-out of FLoC: https://amifloced.org/

pablobm:
Instruct Chrome not to use the Administrate example site for its FLoC calculations.

FLoC is a browser-based[1] technology that Google is currently trialling. It's a means of tracking users without using third-party cookies. See: https://amifloced.org/

[1] Chrome-only at the moment


nickcharlton:
Thank you! Yes, definetly.

Issue #1778, opened 2020-09-24
Create a "how to" section in the documentation


pablobm:
I think that our documentation is pretty good for experienced Rails developers, who know the conventions well and just want to get going. However I feel that it falls short for those who are a bit newer to the framework. Many things are left unsaid because they are implied in the conventions of the community. Newcomers are less aware of these and will struggle.

A way to solve this could be to create a "how to" section in our documentation, similar to the one offered by Devise: https://github.com/heartcombo/devise/wiki/How-Tos. These are some initial ideas of articles there, taken from actual questions from users:

* How to do file processing / parsing? https://github.com/thoughtbot/administrate/issues/1756
* How to hide dashboards from the sidebar: https://github.com/thoughtbot/administrate/issues/1587
* How to override only specific parts of templates: https://github.com/thoughtbot/administrate/issues/1661 https://github.com/thoughtbot/administrate/issues/1315
* How to scope `HasMany` tables: https://stackoverflow.com/questions/59059701/rails-administrate-customize-has-many

For clarity, I would add this section to the current documentation files, as opposed to a GitHub wiki like Devise does.

nickcharlton:
To copy an idea from another community (as I've been doing a bunch of React Native recently), I found React Navigation's set of "Guides" to be pretty helpful: https://reactnavigation.org/docs/getting-started#!

nickcharlton:
Some related issues which are related to this area:

* Hiding Dashboards from the Sidebar: https://github.com/thoughtbot/administrate/issues/1587
* Providing a good plugin example to build new ones from: https://github.com/thoughtbot/administrate/issues/1601

c4lliope:
Looks like @nickcharlton is making good progress;
could someone add a link to his page in the documentation sidebar, here?
https://github.com/thoughtbot/administrate/blob/a4dcaf339989c14498a921d9a1dc40ac84576d1a/spec/example_app/app/views/layouts/docs.html.erb
that way it would show up for beginners who are seeking guidance on https://administrate-prototype.herokuapp.com/.

I also like the name "Guides", "Common Guides", or "Road Map", instead of "How To",
and perhaps instead of "Getting Started" - which implies your experience could also be ending soon,
we could change that page's name to "Day One" - implying you may be using our library for years to come.

c4lliope:
Hm... https://administrate-prototype.herokuapp.com/how_to is broken; has there been a recent deploy on heroku?

nickcharlton:
[...]

PR #1998, opened 2021-06-10
Reformulate authorization in example app

pablobm:
Working on https://github.com/thoughtbot/administrate/pull/1941, I'm having issues with the way authorization is currently tested. I'm trying to solve this problem first before I proceed with the linked PR.

Sorry, this is a big PR :-( To make reviews easier, I have put some effort into splitting it into commits with individual descriptions that make as much sense as possible in isolation. I haven't created individual PRs for each because I think that would leave things in a weird intermediate state.

Although there are some changes to Administrate proper, this PR is actually about the specs and the example app. The commits can be grouped as follows:

1. Actually get Punditize working in the example app (one commit)
2. Changes to authorization (two commits):
    * Allow override of fragments of the collection partial.
    * Reformulate authorization in example app.
3. Changes to navigation (two commits):
    * Remove line items from navigation.
    * Bring product meta tags into navigation.
4. Minor tweaks (three commits):
    * Destroy payments before customers or they'll stay.
    * Create some payments on seed.
    * Less confusing example of authorization.

The only change to Administrate proper is in the first commit of point 2 above, where I extract some parts of the collection partial into smaller partials, allowing them to be overriden more easily.

The example app gets new a new feature where users can "become" a customer with a single click. This provides a simple form of authentication that we can use to test authorization. This feature is visible in the Customers collection:

Links to become a customer and back an admin


nickcharlton (approved):
I think this looks great. It's a much better example and provides a use case which is definitely needed.

I appreciate you splitting it out into a set of commits that are easy to understand in isolation — I think it'd be best to not squash these so that we maintain them.

Do you think we should document (this could be separate from this) how this example of authorization should work? A bit like a "here's something you probably will need to do" guide?

pablobm:
OK, so about adding authentication+authorization in general. I agree that we should document it. I'm adding it to the list at https://github.com/thoughtbot/administrate/issues/1778

PR #2003, opened 2021-06-17
Only allow HTTP(S) URLs in example app

pablobm:
We may want to clean the example app's DB after merging this.

### DIFF
--- a/spec/example_app/spec/models/product_spec.rb
+++ b/spec/example_app/spec/models/product_spec.rb
@@ -1,0 +1,19 @@
+require "rails_helper"
+
+RSpec.describe Product do
+  describe "validations" do
+    it { should allow_value("http://example.com/foo/bar").for(:image_url) }
+    it { should allow_value("https://example.com/foo/bar").for(:image_url) }
+    it { should_not allow_value("ftp://example.com/foo/bar").for(:image_url) }
+  end
+
+  describe "#image_url" do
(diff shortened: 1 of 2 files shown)


spec/example_app/spec/models/product_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    it { should_not allow_value("ftp://example.com/foo/bar").for(:image_url) }
+  end
+
+  describe "#image_url" do
+    it "is trimmed on save" do
+      attrs = FactoryBot.attributes_for(
+        :product,
+        image_url: "\n https://example.com/foo/bar  \n",
+      )
+      product = Product.create(attrs)
```
nickcharlton, line 16:
```suggestion
    it "is trimmed on save" do
      product = create(
      :product,
      image_url: "\n https://example.com/foo/bar  \n",
    )
```

I don't seem to be able to get the suggestion to work out, but I hope you get the idea 😅 

What do you think about collapsing all of these together?

nickcharlton (approved):
One thought, but otherwise looks good!

spec/example_app/spec/models/product_spec.rb
pablobm: Yeah, I don't know why I went with that. It's something about FactoryBot not doing anything clever behind my back, but it's a bit silly. OK, I'll fix it :+1:

PR #947, opened 2017-07-22
Collection filters

koppen:
This adds the option to define a set of filters that can be triggered in order to reduce the number of resources listed on the index page of a dashboard. Currently the only way to trigger the filters is via the search field, using a syntax similar to the one used here on GitHub.

## Why

Often we need to expose more detailed/involved queries than those currently capable via simple searches. Previously we'd have to add special actions that query the database correctly and customize the views with links to trigger those actions.

## Example

For example, if we want to be able to find "inactive" customers (for some definition of inactive), we can add

    COLLECTION_FILTERS = {
      inactive: ->(resources) { resources.where("login_at < ?", 1.week.ago) }
    }

to a dashboard. This way, we can query the resources of that dashboard by typing `bob inactive:` in the search field to find users named "bob" who hasn't logged in the last week.

## Credits

This is basically a reimplementation of @nando's work from #276, I merely took some of his thoughts and ideas and repackaged them in a way that makes sense for Administrate anno 2017. 

Also in an effort to keep an otherwise huge pull request as small as possible, I haven't gone all the way that he did in the original pull request, so we're missing automatic filtering setup and exposing available filters in the UI.

Nor are the filters implemented here as flexible or powerful as those proposed in #276.

## Further details

If you already had the `inactive` scope you could define the filter like the following to take advantage of existing ActiveRecord scopes (and other class methods on the resource class).

    COLLECTION_FILTERS = {
      inactive: ->(resources) { resources.inactive }
    }

While the chosen hash-based syntax is a bit more verbose than simply exposing already defined scopes like so:

    # app/dashboards/customer_dashboard.rb
    COLLECTION_FILTERS = [:inactive]

it allows us to define filters for use in Administrate without having to clutter the resource classes with scopes.

## Future proofing

The chosen syntax should allow us to add the above mentioned simpler syntax in a backwards compatible way at some point down the line if we feel the need. For example it could end up looking like:

    COLLECTION_FILTERS = {
      vip: :vip, # This could call the method `vip` on resources
      inactive: ->(resources) { resources.where("login_at < ?", 1.week.ago) }
    }

but that's a step I am not ready to take just yet.

It also opens up for a way to have filters that take arguments, ie we could define:

    COLLECTION_FILTERS = {
      before: ->(resources, args) {
                resources.where("created_at < ?", args)
              }
    }

and query it as `before:2016-01-12` which, again, is a step I am not ready to take here and now.

PR #2005, opened 2021-06-21
add params to collection filter

aka47:
To filter a collection you can define COLLECTION_FILTER and use them like "filter:".

With this PR you can also params to those filter and use them like "filter:param". The COLLECTION_FILTER will be called with the params a second argument.

This ideas was already mentioned in the initial PR for COLLECTION_FILTERS, but not implemented. (#947)


pablobm:
Thank you for your PR @aka47 - I think it's good. Generally I'm inclined not to enhance the search, as I think it should be completely redesigned, but your change is a small increment of what we already have, so I think it's fine.

Would you be able to make the following two changes before I merge this?

1. This PR creates a situation where searching for anything with a colon in the middle (eg: "order:12") will stop working if it's intended as a string search. Before it only broke things if the colon was at the end, which I think was a much less common case. Would you be able to tweak your code to avoid this? I think you could change it so that `filter?` not only does the regexp match, but also sees if the filter name exists in `COLLECTION_FILTERS`. Does that make sense?
2. Fix that Hound warning (or any other that may come later).

aka47:
@pablobm thanks for your feedback - sure filters and terms should not get confused. It needed a bit more changes then I thought, as the Query class does not have access to the Search Class and the valid_filters from the dashboard. 
If you need anything more - please tell me.

mmrobins:
FWIW this is great and I would love to see it released.  I was running a hacky version of this in my local code to do the same.  I get that search needs an overhaul, but until then something small like this would be a great incremental improvement in the meantime.

pablobm:
Thank you @aka47!

PR #2013, opened 2021-06-27
Fix nav styles for namespaced resources

n-studio:
```ruby
resource_name # => "blog__post"
resource # => "blog/posts"
```


nickcharlton:
Thanks!

Before/After:

Screenshot 2021-06-28 at 16 37 42

Issue #2009, opened 2021-06-24
Template error on html_safe in flash when Devise OAuth2 session expired


januszm:
* What were you trying to do?
I'm using Administrate in an app that authenticates users with Devise+Omniauth. I get this error after the session has expired:

```
An ActionView::Template::Error occurred in ...#index:

  undefined method `html_safe' for true:TrueClass
Did you mean?  html_safe?
```

```
data: {
  "session_id"=>"...",
  "user_return_to"=>"/admin",
  "flash"=> {
    "discard"=>[],
    "flashes"=>{
      "alert"=>"Your session expired. Please sign in again to continue.",
      "timedout"=>true
    }
},
```

in `app/views/administrate/application/_flashes.html.erb` around line 17 we have

```
<% flash.each do |key, value| -%>
  <div class="flash flash-<%= key %>"><%= value.html_safe %></div>
<% end -%>
```

since it loops on `flash`, it ends up calling .html_safe on `"timedout"`'s value which is `true` 

* What did you end up with (logs, or, even better, example apps are great!)?
No permanent solution yet. Visiting the sign_out URL forces new login and resolves the issue temporarily.

* What versions are you running?
  - Rails 6.0.4
  - administrate 0.16

januszm:
UPDATE: see https://github.com/heartcombo/devise#configuring-controllers the bottom of that section says:

> In some circumstances, Devise adds a :timedout key to the flash hash, which is not meant for display. Remove this key from the hash if you intend to print the entire hash

We could only iterate over "typical" for Rails flash keys like `notice`, `alert`, `error`, or, before calling `.html_safe`, verify that the value responds to this method. Without touching administrate code, it can also be done on the application side, a patch for Devise to remove this key and timedout value

pablobm:
Thank you for reporting. A workaround would be to provide a custom template at `app/views/administrate/application/_flashes.html.erb` for your app. However remember to check that this template hasn't changed in Administrate when you update to a new version.

PR #2015, opened 2021-07-02
Only call html_safe on flash message that responds to it

januszm:
This one line change is already covered by existing feature specs.

Resolves: https://github.com/thoughtbot/administrate/issues/2009


januszm:
While setting up the project locally I had some issues with selenium webdriver gem.
Perhaps we should consider switching to `gem 'webdrivers'` ?

PR #2018, opened 2021-07-14
Add :include_blank option to Field::Select

efatsi:
- copy/pasted pretty directly from Field::BelongsTo
- defaults to true similar to BelongsTo configuration

### DIFF
--- a/spec/administrate/views/fields/select/_edit_spec.rb
+++ b/spec/administrate/views/fields/select/_edit_spec.rb
@@ -9,6 +9,7 @@ describe "fields/select/_form", type: :view do
       attribute: :email_subscriber,
       data: false,
       selectable_options: [true, false, nil],
+      include_blank_option: false,
     )
 
     render(
--- a/app/views/fields/select/_form.html.erb
+++ b/app/views/fields/select/_form.html.erb
@@ -27,7 +27,8 @@ to be displayed on a resource's edit form page.
         :last,
         :first,
         field.data,
-      )
+      ),
+      include_blank: field.include_blank_option
     ) %>
   <% else %>
     <%= f.select(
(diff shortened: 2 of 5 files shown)


efatsi:
Had a thought: perhaps a better move would be to set the default to `false` for the new `Fields::Select#include_blank` option. That would mean this PR is not a breaking change, and users could opt-in to including a blank option.

Anyone on the maintenance team have an opinion here?

pablobm:
Thank you for this @efatsi. I agree that we want to avoid unexpected changes of behaviour, and therefore the default should be the same as it is now. Would you be able to make this change?

efatsi:
Can do!

efatsi:
@pablobm just pushed an update. Looks like ruby-25 failed on Circle, although the failure seems unrelated to the changes I've made.

spec/administrate/views/fields/select/_edit_spec.rb
```diff
@@ -9,6 +9,7 @@
       attribute: :email_subscriber,
       data: false,
       selectable_options: [true, false, nil],
+      include_blank_option: false,
```
pablobm, line 12: Wrong option name?

```suggestion
      include_blank: false,
```

efatsi: 🤦 yeah, thanks

efatsi: just kidding, this is correct. The option itself is named `include_blank`, but the getter on the field is `include_blank_option`. Similar to BelongTo field: https://github.com/thoughtbot/administrate/blob/main/lib/administrate/field/belongs_to.rb#L30-L32

pablobm: Ah, you are right. This is a double and not the real thing, therefore this is a literal method name to stub, as opposed to a constructor argument. Sorry!

pablobm (approved):
I think I have figured out the issue with CI (see https://github.com/thoughtbot/administrate/pull/2023). Let me double-check it works, merge the fix, and I'll merge this.

PR #2020, opened 2021-07-20
Render extra columns from COLLECTION_ATTRIBUTES on index page dynamically during runtime without code deployment 

malaya7:
Hi, I'm using this library for a project working on, and would like to be able to have extra columns show up on the index page without making code changes and deployment 

Example: http://localhost:PORT/?extra_columns=id
expected: column id will show up on the subsequent requests 

PS: this is my first time working with ruby and rails, open for feedback and comments 

Thanks.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -14,7 +14,7 @@ module Administrate
       page = Administrate::Page::Collection.new(dashboard, order: order)
 
       col =  params[:extra_columns]&.split(",")&.map(&:to_sym) || []
-      page.attribute_names_append(col)
+      page.attribute_names_append(col) unless col.empty?
 
       render locals: {
         resources: resources,
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -13,6 +13,9 @@ module Administrate
       resources = resources.page(params[:_page]).per(records_per_page)
       page = Administrate::Page::Collection.new(dashboard, order: order)
 
+      col =  params[:extra_columns]&.split(",")&.map(&:to_sym) || []
+      page.attribute_names_append(col)
+
       render locals: {
         resources: resources,
         search_term: search_term,
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -56,8 +56,14 @@ module Administrate
       association_params = collection_names.map do |assoc_name|
         { assoc_name => %i[order direction page per_page] }
       end
-      params.permit(:search, :id, :_page, :per_page,
-        association_params, :extra_columns)
+      params.permit(
+        :search,
+        :id,
+        :_page,
+        :per_page,
+        association_params,
+        :extra_columns
+      )
     end
 
     def clear_search_params
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -56,11 +56,11 @@ module Administrate
       association_params = collection_names.map do |assoc_name|
         { assoc_name => %i[order direction page per_page] }
       end
-      params.permit(:search, :id, :_page, :per_page, association_params)
+      params.permit(:search, :id, :_page, :per_page, association_params, :extra_columns)
     end
 
     def clear_search_params
-      params.except(:search, :_page).permit(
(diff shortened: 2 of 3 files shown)


pablobm:
Hello @malaya7, thank you for your PR.

Unfortunately, the way this is implemented it will not work as you expect. You modify `dashboard.collection_attributes` in memory, and assume that this modification will persist across requests. This works in development, but will not work in production in many cases. For example, when there is more than one server process, your modification will be in memory in one of them, but not in others. Also, it will affect other users who get their request served by the process whose memory you altered.

To make this work, you should store this knowledge somewhere else. An idea would be to use the session cookie, or you could use the DB and make the change persistent across sessions.

However there's a bigger problem: this is a new feature that would require careful planning and future maintenance work. We can't just add it like this. New features need to make sure that the interface is right, they don't interfere with current functionality, they respond to the needs of the community, and they are consistent with Administrate's vision. For this reason, we cannot accept this PR.

One thing that is interesting about this PR though is the concept of "dynamic attributes" for a page. You are right that it may make sense to make the list of attributes more dynamic than it is now. Perhaps `Administrate::Page::Base` should take an additional argument in the constructor to allow for this, or something like that. However we shouldn't rush into implementing this, as there are other changes in the horizon that may play into how this should be ultimately implemented.

For now, I think you should keep this as a custom change, instead of adding it to Administrate officially. The change to the controller is the sort of thing that I expect customisations to have. The change to the helper... I'm not sure that query params are the right way to go; perhaps you should try with a form and a custom action. As for the change to the collection, this one is trickier, as Administrate right now doesn't allow for dynamic attributes. Perhaps a change to your controller could also make it use a custom page class (eg: `Administrate::Page::DynamicCollection`) that does accept the additional constructor argument, and work from there.

malaya7:
Hi @pablobm, 
Thank you for the thorough feedback.

Would addressing the concerns about persistence (either creating ```Administrate::Page::DynamicCollection``` class or making it non persistent across different requests) helps getting it merged or maybe this can be revisited after the other changes (mentioned above) are planed out.

pablobm:
For clarification, when I mentioned `Administrate::Page::DynamicCollection` I wasn't really referring to persistence, but rather to how you could have this customization in your specific app while minimising changes to Administrate's code. I'll try to explain a bit better, let's see...

Normally, your application will have it's own `AdminController`, inheriting from `Administrate::ApplicationController`. In yours, you could override the `index` action to be exactly like Administrate's original (see at https://github.com/thoughtbot/administrate/blob/main/app/controllers/administrate/application_controller.rb), except for the difference that it would use `Administrate::Page::DynamicCollection` instead of `Administrate::Page::Collection`.

This `Administrate::Page::DynamicCollection` would be something that you add to your app and inherits from `Administrate::Page::Collection`, while adding a way to handle dynamic attributes.

The persistence would be handled either in the controller (also in `index`) or on `DynamicCollection`.

This way you can implement this in your own app by adding a new class (`DynamicCollection`) and overriding a controller action (`index`), which is a medium-friction customisation, as opposed to the higher friction of changing Administrate itself. Just make sure to check changes to Administrate when you update, as `index` may have changed and you may have to update it.

Anyway! That's just trying to explain a bit how to work with Administrate. But getting to the question of what it would take to get this merged: I think we would not be merging a change like this for the time being. In the future, as other issues in our backlog are solved, perhaps we would be in a better position to see if and how something like this can be added. To be honest, in my mind I'm thinking of having (still in the future) a facility that makes it easier for people to add this feature as a customisation, instead of adding the feature itself to Administrate.

malaya7:
Hi @pablobm Thank you for the detailed explanations! 
Overriding the index method is giving me the intended behavior, but had to override other models to allow query params like this
```
module Administrate
  module ApplicationHelper
    def sanitized_order_params(page, current_field_name)
      collection_names = page.item_includes + [current_field_name]
      association_params = collection_names.map do |assoc_name|
        {assoc_name => %i[order direction page per_page]}
      end
      params.permit(
        **:extra_col,**
        :search,
        :id,
        :_page,
        :per_page,
        association_params
      )
    end

    def clear_search_params
      params.except(:search, :_page).permit(**:extra_col,**
        :per_page, resource_name => %i[order direction])
    end
  end
end
```
is this the best approach or is there a better way to go about that ?

pablobm:
You are correct, I forgot about that. Unfortunately there isn't a more flexible way to handle those strong params at the moment.

PR #2027, opened 2021-07-30
Drop support for Ruby <2.6

nickcharlton:
Ruby 2.5 reached EOL with the 2.5.9 release on 2021-04-05, we're now
starting to see many deprecations of other libraries and so it's time
for us to drop it too.

http://www.ruby-lang.org/en/news/2021/04/05/ruby-2-5-9-released/


pablobm (approved):
:shipit:

Issue #1227, opened 2018-10-15
Why Time doesn't have a option `:format` ?


rnitta:
Like Datetime, Time should have `:format` option, too.

If this is acceptable, I will be code it, after https://github.com/thoughtbot/administrate/pull/1226 is dealt with.

nickcharlton:
Hi @rnitta, it's obviously been a long time since #1226 but I've just merged that if this is something you'd like to pick up?

nickcharlton:
Closing this as we merged in #2029.

PR #2029, opened 2021-08-05
Add format option to time field

amymariewall:
Previously there was no way to specify format for a time field, but you can specify a format for date and datetime fields. This commit adds the ability to pass a format option to time fields.

Because I18n [uses the time locale key for for all objects with a time](https://github.com/ruby-i18n/i18n/blob/0888807ab2fe4f4c8a4b780f5654a8175df61feb/lib/i18n/backend/base.rb#L82), retain the hard-coded time field format where no format option is passed to avoid a breaking change.

Addresses #1227.

### DIFF
--- a/spec/lib/fields/time_spec.rb
+++ b/spec/lib/fields/time_spec.rb
@@ -13,4 +13,29 @@ describe Administrate::Field::Time do
       expect(path).to eq("/fields/time/#{page}")
     end
   end
+
+  describe "#time" do
+    it "formats the time according to the passed format option" do
+      time = DateTime.new(2021, 3, 26, 16, 38)
+      formats = {
+        time: {
+          formats: { short: "%H:%M" },
+        },
+      }
+
+      options_field = Administrate::Field::Time.with_options(format: :short)
+      field = options_field.new(:time, time, :index)
+
+      with_translations(:en, formats) do
+        expect(field.time).to eq("16:38")
+      end
+    end
+
+    it "it defaults to the expected format if no format option is passed" do
+      time = DateTime.new(2021, 3, 26, 16, 38)
+      field = Administrate::Field::Time.new(:time, time, :index)
+
+      expect(field.time).to eq("04:38PM")
+    end
+  end
 end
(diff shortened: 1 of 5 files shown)


tjvc:
@pablobm @amymariewall I think this is ready for another review. I've addressed the linter errors and another issue I spotted: because I18n uses the same key in the locales file for all objects with a time (`time`), we cannot set separate default formats for our time and datetime fields. I've resolved this by applying a hard-coded format to the time field where a format option is not passed (as we were previously) so as not to introduce a breaking change, but definitely open to other ideas here!

spec/lib/fields/time_spec.rb
```diff
@@ -13,4 +13,30 @@
...
+
+  describe "#time" do
+    let(:time) { DateTime.new(2021, 3, 26, 16, 38) }
+    let(:formats) do
+      {
+        time: {
+          formats: { short: "%H:%M" },
+        },
+      }
+    end
```
nickcharlton, line 25: A minor point: we typically avoid using `let` in this codebase and prefer to either inline a variable in each `it` or extract it into a normal Ruby method.

nickcharlton (commented):
I think this looks good; just one comment about use of `let`.

spec/lib/fields/time_spec.rb
```diff
@@ -13,4 +13,30 @@
       expect(path).to eq("/fields/time/#{page}")
     end
   end
+
+  describe "#time" do
+    before do
+      @time = DateTime.new(2021, 3, 26, 16, 38)
+    end
```
nickcharlton, line 20: And `before`! I'd just put the `DateTime` definition in each `it` block.

nickcharlton (commented):
One more comment; if you change that, feel free to merge it whenever!

spec/lib/fields/time_spec.rb
tjvc: Sorry, I totally missed the second part of your original comment! 🤦‍♂️

PR #2046, opened 2021-10-06
remove password field from Extending Administrate Section

ewertoncodes:
Remove password field from Extending Administrate section. Because password field already exist in the gem administrate.

### DIFF
--- a/docs/extending_administrate.md
+++ b/docs/extending_administrate.md
@@ -5,12 +5,13 @@ title: Extending Administrate
 Apart from the configuration described in these pages, it is possible to
 extend Administrate's capabilities with the use of plugins. There are a
 number of plugins available, many of which can be found at [RubyGems.org].
-At the time of writing, these appear to be the most popular ones:
+These are some popular examples:
 
 1. [ActiveStorage support](https://github.com/Dreamersoul/administrate-field-active_storage)
 2. [Enum field](https://github.com/Valiot/administrate-field-enum)
 3. [Nested has-many forms](https://github.com/nickcharlton/administrate-field-nested_has_many)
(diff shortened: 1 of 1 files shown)


getaaron:
I think the `belongs_to_search` could also be removed?

docs/extending_administrate.md
```diff
@@ -8,10 +8,9 @@ number of plugins available, many of which can be found at [RubyGems.org].
 At the time of writing, these appear to be the most popular ones:
```
pablobm, line 8:
```suggestion
These are some popular examples:
```

pablobm (commented):
Hello @ewertoncodes, and thank you for the PR. Apologies for the delay.

I think the change makes sense, although it would need to also change the phrasing in the sentence right before the list, as these are not the "most popular ones" any more, but rather a hand picked selection. (See my suggestion).

Perhaps also add a new fifth one, like the jsonb plugin?

Regarding @getaaron's suggestion, I'm not sure it's the same, as the `belongs_to_search` plugin is different from the built-in `belongs_to` field. However the plugin is quite tricky to use (or was last time I checked) so perhaps it's worth removing it too.

Thoughts?

ewertoncodes:
@pablobm and @nickcharlton. Looks like the circle ci needs an update on line:

https://github.com/thoughtbot/administrate/blob/main/.circleci/config.yml#L10.
Can I solve that in this PR?

pablobm:
@ewertoncodes - I just fixed the CI: it required upgrading Rails. Could you please rebase now and I'll merge?

ewertoncodes:
Rebased

pablobm:
Merged! Thank you @ewertoncodes 🙂

PR #2067, opened 2021-11-25
Refactor search_attributes to be defined in BaseDashboard

livedo:
`BaseDashboard` offers a clear easily overrideable interface for customizing which attributes are used for the index, show, and edit/update actions in `#collection_attributes`, `#show_page_attributes` and `#form_attributes`, respectively. This PR brings search into the same pattern and adds `BaseDashboard#search_attributes` allowing the dashboard customizer to easily change the logic which is used for filtering attributes for searching. 

A typical use case might entail implementing an app-specific field option and using that to customize the field filtering.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -6,7 +6,7 @@ module Administrate
       authorize_resource(resource_class)
       search_term = params[:search].to_s.strip
       resources = Administrate::Search.new(scoped_resource,
-                                           dashboard_class,
+                                           dashboard,
                                            search_term).run
       resources = apply_collection_includes(resources)
       resources = order.apply(resources)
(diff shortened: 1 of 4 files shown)


livedo:
@pablobm @nickcharlton does this look like a sensible direction to you?

pablobm:
I think this might be a good idea:

1. I prefer `Administrate::Search` receiving an dashboard instance rather than a class. In general we should pass instances around, not classes.
2. I prefer `Administrate::Search` asking the dashboard for what's searchable, rather than asking individual fields. It reduces the contact surface and opens the possibility for more complex definitions for `search_attributes` in the future.
3. As you say, since we are listing index/show/edit attributes, it makes sense to list the search ones too.

Let me sleep over it for a couple more days, but I think I'll merge it.

livedo:
@pablobm did you sleep well? 😴☺️

pablobm:
[The Magi](https://en.wikipedia.org/wiki/Epiphany_(holiday)#Spain) have come bearing gifts and merges :-)

livedo:
@pablobm thanks for the merge! Are there any plans for chopping a new release?

PR #2074, opened 2021-12-03
gemspec: Drop unused directive test_files

olleolleolle:
This directive is not used by RubyGems.org.


pablobm:
Thank you @olleolleolle!

For the benefit of future readers, the directive `test_files` is not currently listed at https://guides.rubygems.org/specification-reference/, plus here we pointing it at directory `test/`, which doesn't exist in this project anyway.

olleolleolle:
@pablobm Thanks for adding that context!

Issue #2063, opened 2021-11-17
Search by Polymorphic Field


jeremiahlukus:
* What were you trying to do?
I have a Subscription Dashboard that has a field `owner: Field::Polymorphic` I would like to be able to search by owner attributes 

* What versions are you running?
  - Rails  `gem "rails", "~> 6.1.0"`
  - administrate latest

pablobm:
Thank you for your report. Indeed, polymorphic fields are tricky and our search doesn't currently support them.

Fortunately there are ways around this. It's possible to provide your own controllers that override Administrate's behaviour. This is the basis of many customisations. In this case, overriding `Administrate::ApplicationController#index` would be a way to do it.

Having said that, I had a look and realised that the way things are hooked up, this is not so evident in the case of customising the search. To fix this, I have opened a PR that introduces a small change that should make this easier to do, as well as provides an example of how to do it (in the specs): https://github.com/thoughtbot/administrate/pull/2096

pablobm:
Closing in favour of https://github.com/thoughtbot/administrate/issues/2076, which is more general of this whole issue.

PR #2096, opened 2021-12-25
Make search easier to override and adapt to custom use cases

pablobm:
Seeing https://github.com/thoughtbot/administrate/issues/2063 made me think about quick ways to customise Administrate's search.

I realised that `Administrate::ApplicationController#index` doesn't make it easy to override the way that search is performed, so this seems like a first "obvious" port of entry.

By factoring out lines 8-10 into their own method, now we can do things like shown in the provided spec. This should be enough for everyone to adapt the search to their own use case.

In the future, we can try rework `Administrate::Order` into something more configurable and "better". However we all know this is long term future, so this PR should get us going, for a long way, until then.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -5,9 +5,7 @@ module Administrate
     def index
       authorize_resource(resource_class)
       search_term = params[:search].to_s.strip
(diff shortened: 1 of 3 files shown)


nickcharlton (approved):
This looks good!

PR #2101, opened 2021-12-28
Limit the highest momentjs-rails version to 2.20.1

shouichi:
v2.29.1 causes "export declarations may only appear at top level of a
module" error.

See https://github.com/derekprior/momentjs-rails/issues/61.


pablobm:
Thank you @shouichi

PR #2107, opened 2022-01-04
Relax momentjs-rails version constraint

shouichi:
The problem caused by v2.29.1 is now solved in v2.29.1.1.

See https://github.com/derekprior/momentjs-rails/pull/62.


pablobm:
Thank you for this @shouichi

PR #2108, opened 2022-01-04
Hide link if user is not authorized to access resource

nhippenmeyer:
When using pundit we should not display links to associated resources unless the user has access to those resources.


pablobm (approved):
Thank you for this. I've been meaning to provide a more comprehensive solution for a while, but this will do for now.

PR #2112, opened 2022-01-13
Change simplified words to traditional words at back_to_app

jhang-jhe-wei:
# Change words
The original character was simplified, and now it is replaced by traditional.


pablobm:
Thank you @jhang-jhe-wei

PR #2113, opened 2022-01-13
WIP: Test in all currently supported Ruby versions

pablobm:
As it says on the tin. List taken from https://en.wikipedia.org/wiki/Ruby_(programming_language)#Table_of_versions

I thought that in preparation for supporting Rails 7 we should also update the list of tested rubies. Interestingly, Rails 7 requires 2.7+, but 2.6 is not deprecated just yet (will be soon on 2021-04-05).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,9 +73,39 @@ jobs:
           POSTGRES_DB: ruby27
           POSTGRES_PASSWORD: ""
 
+  ruby-30:
+    <<: *default_job
+    docker:
+      - image: circleci/ruby:3.0-node-browsers-legacy
+        environment:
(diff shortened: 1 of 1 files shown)


pablobm:
Looks like `circleci/ruby:3.1-node-browsers-legacy` doesn't exist. I had a mind of stop using `node-browsers-legacy` at some point, so this looks like a good opportunity. Also I'll have to see what's wrong with Ruby 3.0.

nickcharlton:
Thanks for doing this. I'm going to close this as I merged in #2127 already and #2129 should solve Ruby 3.1.

Annoyingly, there's some other incompatibilities (that those have taken into account) and once again, Circle is creating busy work.

PR #2114, opened 2022-01-14
Add field has_many i18n

jhang-jhe-wei:
# Field::HasMany i18n
## How to Use?
If you have ActiveRecord i18n translation of the model you don't need to do anything extra, if not, make an ActiveRecord i18n translation of the model.

## Example
The recipe has many recipe items.

> app/dashboards/recipe_dashboard.rb
```ruby
  ATTRIBUTE_TYPES = {
    recipe_items: Field::HasMany,
    id: Field::Number,
    name: Field::String,
    package_spec: Field::String,
    remark: Field::String,
    created_at: Field::DateTime,
    updated_at: Field::DateTime,
  }.freeze

```
> zh-TW.yml
```yaml
zh-TW:
  activerecord:
    models:
      recipe: 配方表
      recipe_item: 材料
```
> Result

image

> `材料種類` is defined in helper.label.recipe.recipe_items, it's not the same, don't confuse.


app/views/fields/has_many/_index.html.erb
```diff
@@ -16,4 +16,4 @@ as a count of how many objects are associated through the relationship.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/HasMany
 %>
 
-<%= pluralize(field.data.size, field.attribute.to_s.humanize.downcase.singularize) %>
+<%= pluralize(field.data.size, t("activerecord.models.#{field.attribute.to_s.singularize}", default: field.attribute.to_s.humanize.downcase.singularize)) %>
```
pablobm, line 19: One detail: it's possible to provide plurals for translations like this one. In this case, it would look like this:

```yaml
es:
  activerecord:
    models:
      order:
        one: Pedido
        other: Pedidos
```

In this case, your code would fail with `NoMethodError undefined method `pluralize' for {:one=>"Patronisation", :other=>"Patronisations"}:Hash`. To avoid this I think we can just add a `count`, like follows:

```suggestion
<%= pluralize(field.data.size, t("activerecord.models.#{field.attribute.to_s.singularize}", default: field.attribute.to_s.humanize.downcase.singularize, count: field.data.size)) %>
```

What do you think?

pablobm (commented):
Thank you for this :-) Just one thing that I think can be improved.

app/views/fields/has_many/_index.html.erb
jhang-jhe-wei: Oh! I know what do you mean, that's the one thing I haven't considered because Chinese doesn't have singular and plural problems XD.  It looks good to me, thanks.

pablobm:
@jhang-jhe-wei - Could you please rebase? The specs (I don't know why) are failing and I can't merge, but it looks like a rebase should fix it.

nickcharlton:
Thanks for opening this PR! I rebased and tidied up the commit so we could get it merged.

PR #2115, opened 2022-01-14
Remove inline style from icon svg

### LINKED ISSUE #2087: Inline Style doesn't work with Content Security Policy
Hello,

after enabling CSP I get this error in the browser console when opening administrate:

```
Refused to apply a stylesheet because its hash, its nonce, or 'unsafe-inline' does not appear in the style-src directive of the Content Security Policy.
```

Referencing [this](https://github.com/thoughtbot/administrate/blob/main/app/views/administrate/application/_icons.html.erb#L1) line:

```xml
<svg xmlns="http://www.w3.org/2000/svg" style="display: none;">
```

Maybe this simple style could be transferred to the administrate css?

Greetings,
Michael

michaelbaisch:
Fixes #2087


app/views/administrate/application/_icons.html.erb
```diff
@@ -1,4 +1,4 @@
-<svg xmlns="http://www.w3.org/2000/svg" style="display: none;">
+<svg hidden xmlns="http://www.w3.org/2000/svg">
```
pablobm, line 1: Uh, I didn't know of the `hidden` attribute. Cool, compatibility seems wide enough according to https://caniuse.com/?search=hidden. Only problematic with IE before 11, so it should be fine, I think.

Just one question: any reason that you went for this instead of a CSS-based solution? (As in: putting it in a stylesheet).

pablobm (commented):
Thank you for this :-) Just one question!

app/views/administrate/application/_icons.html.erb
michaelbaisch: With just a brief look around in the administrate css, there wasn't an obvious place and appropriate naming for me.
So this was simple, quick and elegant enough :)

PR #2117, opened 2022-01-20
Use `camelize` rather than `classify` when supplying a namespace

edwinwills:
`classify` will singularise a namespace, and that feels like unintended behaviour. E.g. if I use `--namespace=operations`, with `classify`, the generated module name is `Operation`, but I would expect this to be `Operations`.

Swapping `classify` for `camelize` solves this issue for me, but this may not be what you intended with this, so no problem if this isn't relevant.

Thanks for a great gem!


pablobm:
I think you are right :-) Thank you

PR #2118, opened 2022-01-20
Handle Selenium "options" deprecation

jordan-brough:
Silences this warning while running "bundle exec rspec":

    WARN Selenium [DEPRECATION] [:browser_options] :options as a parameter for
    driver initialization is deprecated. Use :capabilities with an Array of
    value capabilities/options if necessary instead.

### DIFF
--- a/spec/support/webdrivers.rb
+++ b/spec/support/webdrivers.rb
@@ -5,19 +5,16 @@ Capybara.register_driver :chrome do |app|
 end
 
 Capybara.register_driver :headless_chrome do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+  options.headless!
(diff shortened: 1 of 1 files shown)


jordan-brough:
@nickcharlton thanks for merging #2124! I've rebased this on latest `main` to resolve the merge conflicts.

I see you also merged #2064 (cc @pablobm). This PR is another way to avoid the warnings, while still maintaining the perhaps "nicer"(?) api for configuring the options.  I don't think there's anything different other than the syntax, so feel free to close this PR if you want to just stick with what's in `main` now.

nickcharlton:
I think on balance, I prefer the way this looks and I often find I (and presumably others) copy out the config so if we've got a better option I'm going to go with it!

Thanks!

PR #2125, opened 2022-01-20
Fix rspec invocation in CircleCI

jordan-brough:
The tests weren't actually running for Rails >= v6. Only the "bundler:audit" task was.

E.g. see the output and the test time (only 3s) in [this CircleCI run](https://app.circleci.com/pipelines/github/thoughtbot/administrate/1604/workflows/b1c92739-9bd6-4746-8200-ee0d42c64aaa/jobs/8676):

image

See also the `bundle exec appraisal rake` part of that same test run. The tests only ran for Rails < v6.

PR #2127, opened 2022-01-21
Add Ruby 3 to Circle tests

jordan-brough:
Start testing the code on Ruby 3, without making it the default yet.

If we get to a point where there are gem conflicts for `ruby5x` that won't even _install_ under Ruby 3 then we'd need to do something more complex, but this should work for now I think, and get Ruby 3 under test coverage at least?

Notes:

- We have to exclude Rails < 6.0 in `.circleci/config.yml` when running tests on Ruby 3, since only Rails 6 and above support Ruby 3.
- ✅ ~~**This won't actually _run_ any of the rspec tests on Ruby 3 until https://github.com/thoughtbot/administrate/pull/2125 is merged**, since the tests aren't currently actually running on Circle for Rails >= 6 and Ruby 3 only runs on Rails >= 6.~~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,9 +82,30 @@ jobs:
           POSTGRES_DB: ruby27
           POSTGRES_PASSWORD: ""
 
+  ruby-30:
+    <<: *default_job
+    steps:
+      - shared_steps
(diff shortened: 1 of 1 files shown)


jordan-brough:
Sounds great! Thanks for the quick responses!

PR #2129, opened 2022-01-21
Add Ruby 3.1 to Circle tests

jordan-brough:
This PR builds on #2127 and #2128 to add test coverage for Ruby 3.1.
~~i.e.: **Ignore the first 2 commits in this PR.**~~ (update: Those PRs were merged and I've rebased this)

A couple things:

- Remove usage of YAML aliases in the test `example_app` to work around a Psych 4 incompatibility with Rails 6. (See the commit message on that commit for more details).
- Add Circle configs for testing Ruby 3.1, and limit it to testing Rails 6.1. I don't think Rails 6.0 is compatible with Ruby 3.1 (e.g. see [here](https://www.fastruby.io/blog/ruby/rails/versions/compatibility-table.html)) and I was seeing some test failures that seemed to confirm that, iirc.  When Rails 7 test coverage is added then Ruby 3.1 can cover that as well of course.
- This uses CircleCI's new `cimg/ruby` docker image, which replaces the old `circleci/ruby` docker image, because the Ruby 3.1 hasn't been added to the latter and it sounds like it won't be. We could probably update everything to `cimg/ruby` if we want. Or we could test it out a bit on Ruby 3.1 for a while first.

Does that all seem reasonable?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 version: 2.1
 
 orbs:
-  browser-tools: circleci/browser-tools@1.2.3
+  browser-tools: circleci/browser-tools@1.2.4
 
 commands:
   shared_steps:
@@ -86,7 +86,7 @@ jobs:
     <<: *default_job
     steps:
       - shared_steps
-      # Run the tests against the versions of Rails that support Ruby 3
+      # Run the tests against the versions of Rails that support Ruby 3.0
       - run: bundle exec appraisal install
       - run: bundle exec appraisal rails60 rspec
       - run: bundle exec appraisal rails61 rspec
@@ -102,10 +102,30 @@ jobs:
           POSTGRES_DB: ruby30
           POSTGRES_PASSWORD: ""
 
+  ruby-31:
+    <<: *default_job
+    steps:
+      - shared_steps
+      # Run the tests against the versions of Rails that support Ruby 3.1
+      - run: bundle exec appraisal install
+      - run: bundle exec appraisal rails61 rspec
+    docker:
+      - image: cimg/ruby:3.1-browsers
+        environment:
+          PGHOST: localhost
+          PGUSER: administrate
+          RAILS_ENV: test
+      - image: postgres:10.1-alpine
+        environment:
+          POSTGRES_USER: administrate
+          POSTGRES_DB: ruby31
+          POSTGRES_PASSWORD: ""
+
 workflows:
   version: 2
   multiple-rubies:
     jobs:
+      - ruby-31
       - ruby-30
       - ruby-27
       - ruby-26
--- a/spec/example_app/config/secrets.yml
+++ b/spec/example_app/config/secrets.yml
@@ -1,15 +1,11 @@
-default: &default
-  secret_key_base: <%= ENV["SECRET_KEY_BASE"] %>
-
 development:
-  <<: *default
   secret_key_base: 1a022e4f335d24af3d6bd622b9daef5a44808afbe16d4f1c8bed03675ab91ecd9c76ddc1a30f3fbb668b02c00abcba2ecc3714c95943b8f4af86289a2a46d5e1
 
 test:
   secret_key_base: test_secret
 
 staging:
-  <<: *default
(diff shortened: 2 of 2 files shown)


jordan-brough:
@nickcharlton @pablobm The other 2 PRs were merged (thank you!) so I rebased this, and the tests are green so I'm marking this "ready for review". There are a few thoughts/questions in the PR description that might merit feedback/changes.

.circleci/config.yml
```diff
@@ -102,10 +102,30 @@ jobs:
...
 
+  ruby-31:
+    <<: *default_job
+    steps:
+      - shared_steps
+      # Run the tests against the versions of Rails that support Ruby 3.1
+      - run: bundle exec appraisal install
+      - run: bundle exec appraisal rails61 rspec
+    docker:
+      - image: cimg/ruby:3.1-browsers
```
nickcharlton, line 113: Do we need the browsers image here, or is the standard `ruby:3.1` with the orb enough?

jordan-brough: If I understand correctly, the `browsers` variant supplies the dependencies needed for the orb to work:

> The browsers variant is designed to work in conjunction with the CircleCI Browser Tools orb. You can use the orb to install a version of Google Chrome and/or Firefox into your build.

https://circleci.com/developer/images/image/cimg/ruby#browsers

But I'll push up a quick test commit just to confirm.

jordan-brough: Ah, actually I forgot for a bit that this PR only uses `cimg/ruby` in the 3.1 tests.  And there is a difference of 7s (with `-browsers`) vs 23s (without `-browsers) for "Install Google Chrome".  We could try testing each version a few more times to confirm the speed difference.  I've also opened that ticket w/ Circle to confirm.
What do you prefer to do at this point?

jordan-brough: @nickcharlton I went ahead and did some speed testing and it does seem like the difference is probably that some stuff is preinstalled on `-browsers`, which makes the browser installation faster.  However, I'd bet it probably makes the image download longer also.

What I saw:

| test run | "-browsers"? | env spin up (incl download) | chrome install | total |
| - | - | - | - | - |
| [link](https://app.circleci.com/pipelines/github/rinsed-org/administrate/33/workflows/bca36891-68a7-4603-9275-29eee46fef38/jobs/101) | yes | 30s | 9s | 39s |
| [link](https://app.circleci.com/pipelines/github/rinsed-org/administrate/36/workflows/8857a2fc-5a3f-4ce6-a3f2-648dc71bf1d2/jobs/114) | yes | 6s | 7s | 13s |
| [link](https://app.circleci.com/pipelines/github/rinsed-org/administrate/38/workflows/d18672dd-3987-4c20-9103-7550f7ce0288/jobs/119) | yes | 13s | 7s | 20s |
| [link](https://app.circleci.com/pipelines/github/rinsed-org/administrate/34/workflows/bd77b117-6010-4612-b642-88948c808e39/jobs/105) | no | 5s | 18s | 23s |
| [link](https://app.circleci.com/pipelines/github/rinsed-org/administrate/37/workflows/1b3b74a7-8307-48ff-9774-10fbc8cae36e/jobs/117) | no | 12s | 17s | 29s |
| [link](https://app.circleci.com/pipelines/github/rinsed-org/administrate/39/workflows/6ea45fc7-8e97-45f5-b5c0-7f5bb123a290/jobs/126) | no | 3s | 19s | 22s |

Lmk what you think and which you prefer. Circle has enough variability in timing that it's hard to really get a firm idea of things 😕

nickcharlton: Oh I see. Right, it is intentional on Circle's part. I didn't think to even look at that, so thanks for highlighting it.

I think we should go with what they recommend — it's a pity your table shows how variable the runs are (thanks for doing that!).

jordan-brough: 👍 sounds good. I've force-pushed to remove the commit that tried out the non-`-browsers` variant.

nickcharlton:
Thanks for this! I'm going to go ahead and merge it now!

PR #2132, opened 2022-01-25
Add Rails 7 to tests

jordan-brough:
And refactor ".circleci/config" a bit.  With all the current variations & limitations it seems like it's probably better to just list
the supported configurations explicitly rather than trying to share config blocks.

**NOTE**: I haven't actually tried running administrate on a Rails 7 app! But I intend to upgrade to Rails 7 and I saw that tests pass for it here so I thought it'd be worth opening a PR at least.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,15 +47,19 @@ commands:
 
 default_job: &default_job
   working_directory: ~/administrate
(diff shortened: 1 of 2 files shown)


nickcharlton (approved):
Thanks!

PR #2133, opened 2022-01-26
Sort dashboard attributes

nhippenmeyer:
Sorting the fields listed in the `ATTRIBUTE_TYPES` hash of the dashboard file will make these files easier to maintain. For dashboards that contain a large number of attributes, it can be time-consuming/inefficient to find the correct line when making a change.


nickcharlton (commented):
Thanks for this!

I think if we're going to sort one of them, we should also sort the rest of the dashboard attributes. What do you think?

nhippenmeyer:
> I think if we're going to sort one of them, we should also sort the rest of the dashboard attributes. What do you think?

I'm happy to sort the attributes in `COLLECTION_ATTRIBUTES`, `SHOW_PAGE_ATTRIBUTES` and `FORM_ATTRIBUTES` as well, however I feel less strongly about that since those lists are typically reordered manually anyway based on the requirements of each dashboard.

> I feel that id should go first by default. Thoughts?

I agree that `id` is more of special-case attribute and could make sense to put it first in the list. How do you feel about where `created_at` and `updated_at` should appear in the list? Should they be at the bottom or sorted alphabetically with the rest of the attributes?

nickcharlton:
Good point! We should put `created_at`/`updated_at` at the bottom, I think.

pablobm:
Agreed: `id` first, standard timestamps last.

nhippenmeyer:
@nickcharlton @pablobm mind taking another look?

nickcharlton:
Ah hah, yeah, that works as I was hoping. I'm pondering how we can simplify the implementation, but I'm not getting anywhere today, any thoughts, @pablobm?

lib/generators/administrate/dashboard/dashboard_generator.rb
```diff
@@ -53,9 +53,16 @@ def namespace
...
+        columns = klass.columns.map(&:name)
+
+        (klass.reflections.keys +
+          columns -
+          redundant_attributes -
+          %w[id created_at updated_at]).sort.tap do |result|
+            result.unshift(klass.primary_key) if klass.primary_key.present?
+            result.push("created_at") if columns.include?("created_at")
+            result.push("updated_at") if columns.include?("updated_at")
+          end
```
pablobm, line 65: Responding to @nickcharlton's comment, yeah, this sort of stuff is always tricky and open to personal preference. This code doesn't offend me, but I do wonder if it's possible to make it more readable... How about the following?

```suggestion
        unsorted_attrs = (klass.reflections.keys +
          klass.columns.map(&:name) -
          redundant_attributes)

        primary_key = unsorted_attrs.delete(klass.primary_key)
        updated_at_key = unsorted_attrs.delete("updated_at")
        created_at_key = unsorted_attrs.delete("created_at")

        (
          [primary_key] +
          unsorted_attrs.sort +
          [created_at_key, updated_at_key]
        ).compact
```

I'm not married to my proposed solution either, just checking what others think :slightly_smiling_face:

spec/generators/dashboard_generator_spec.rb
```diff
@@ -61,6 +61,31 @@ class Foo < ApplicationRecord
         end
       end
 
+      it "sorts the attributes" do
+        begin
+          ActiveRecord::Schema.define do
+            create_table(:foos) do |t|
```
pablobm, line 67: How about adding a custom primary key here, providing an example for your use of `primary_key` in the exercised code?

```suggestion
            create_table(:foos, primary_key: :code) do |t|
```

spec/generators/dashboard_generator_spec.rb
```diff
@@ -61,6 +61,31 @@ class Foo < ApplicationRecord
...
+
+          class Foo < ApplicationRecord
+            reset_column_information
+          end
+
+          run_generator ["foo"]
+          load file("app/dashboards/foo_dashboard.rb")
+          attrs = FooDashboard::ATTRIBUTE_TYPES.keys
+
+          expect(attrs).to eq(%i[id col_1 col_2 col_3 created_at updated_at])
```
pablobm, line 83: Companion to my comment above about a primary key:

```suggestion
          expect(attrs).to eq(%i[code col_1 col_2 col_3 created_at updated_at])
```

spec/generators/dashboard_generator_spec.rb
```diff
@@ -33,8 +33,8 @@ class Foo < ApplicationRecord
           load file("app/dashboards/foo_dashboard.rb")
           attrs = FooDashboard::ATTRIBUTE_TYPES
 
-          expect(attrs[:id]).to eq(Administrate::Field::Number)
           expect(attrs[:created_at]).to eq(Administrate::Field::DateTime)
+          expect(attrs[:id]).to eq(Administrate::Field::Number)
```
pablobm, line 37: Why this change?

lib/generators/administrate/dashboard/dashboard_generator.rb
nhippenmeyer: @pablobm thanks for the suggestion. I've taken what you proposed and modified it slightly. Let me know what you think .

spec/generators/dashboard_generator_spec.rb
nhippenmeyer: Great call - added.

spec/generators/dashboard_generator_spec.rb
nhippenmeyer: Updated.

spec/generators/dashboard_generator_spec.rb
nhippenmeyer: I'm not sure what I originally made this change. I've undone it and the tests are still passing.

nhippenmeyer:
@nickcharlton @pablobm I've made your suggested improvements. Does this look good now?

pablobm:
Nice, let's merge this bad boi! :shipit:

Issue #2134, opened 2022-01-27
Support opening url in a new browser tab


shouichi:
* What would you like to be able to do? Can you provide some examples?

By using `Field::Url` we can display URLs as links. In some cases, it would be convenient to open links in new browser tabs.

* How could we go about implementing that?

Though the implementation would be straightforward, we may want to add `rel="noopener"` or `rel="noreferrer`.

https://docs.rubocop.org/rubocop-rails/cops_rails.html#railslinktoblank

* Can you think of other approaches to the problem?

No.

nickcharlton:
It's not something we currently support. I wondered if we'd exposed enough of the HTML options to do it, but apparently not. `content_tag` accepts a hash of options, so you could do similar things too. It'd be nice to be able to do this on all fields, I'd think, if you or someone was able to open a PR to do so.

You _could_ do something like this, but it'd mean overriding the partials in your project:

```ruby
# dashboard
ATTRIBUTE_TYPES = {
  image_url: Field::Url.with_options(target: "_blank"),
}
```

```erb
# field partial
<%= content_tag :a, href: field.data, target: field.options[:target] do %>
  <%= field.data %>
<% end %>
```

PR #2139, opened 2022-02-08
Enable to pass html options to URL field

shouichi:
Close https://github.com/thoughtbot/administrate/issues/2134.

### DIFF
--- a/lib/administrate/field/url.rb
+++ b/lib/administrate/field/url.rb
@@ -11,6 +11,10 @@ module Administrate
         data.to_s[0...truncation_length]
       end
 
+      def html_options
+        @options[:html_options] || {}
+      end
+
       private
 
       def truncation_length
--- a/app/views/fields/url/_index.html.erb
+++ b/app/views/fields/url/_index.html.erb
@@ -15,6 +15,6 @@ By default, the value is rendered as an `a` element.
 [1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/Url
 %>
 
-<%= content_tag :a, href: field.data do %>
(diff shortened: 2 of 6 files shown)


lib/administrate/field/url.rb
```diff
@@ -11,6 +11,10 @@ def truncate
         data.to_s[0...truncation_length]
       end
 
+      def html_options
+        @options[:html_options] || {}
+      end
```
shouichi, line 16: I have two questions.

1. Is the method name OK?
2. Should we automatically add `rel="noopener"` for security?

nickcharlton: 1. I think the method name is okay — it's the same as what Rails calls it.
2. No, I think that'd be surprising for the end user.

nickcharlton (commented):
This looks good!

One thought though: Do you think we could move the options up to the `Base` class so that it applies to all fields?

shouichi:
> One thought though: Do you think we could move the options up to the Base class so that it applies to all fields?

Sounds cool! I searched the codebase (`git grep content_tag`) and found that only `Fields::Url` uses `content_tag`. The templates of some fields are somewhat non-trivial. It might not be obvious where the `html_options` applies to. I believe we can gradually add `html_options` to certain fields, then introduce it to the base class when we find it useful (and we can do that in a backward-compatible way I guess). What do you think?

By the way, I realized that I also needed to change `_index.html.erb` template too. I'll fix it tomorrow.

Thanks!

nickcharlton (approved):
Thanks for updating!

Yes, I think we can leave it as just the URL field for now and expand it over time.

PR #2141, opened 2022-02-14
Fix Pundit >2.2.0 include

nickcharlton:
In #2140, we updated Pundit but it also needs the include changing.

https://github.com/varvet/pundit/commit/4d9b5849f41ae540cd48b323e100fe74d76ded92

---

@sedubois, a fix for your comment. Is this what you meant?


nickcharlton:
@dependabot rebase

PR #2154, opened 2022-03-03
Trying out GitHub's code scanning tool

pablobm:
As described at https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning

Just playing with this. No idea of whether it's useful. @nickcharlton, any thoughts?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,36 @@
+---
+name: "CodeQL"
+
+on:
+  push:
+  schedule:
+    - cron: '44 6 * * 4'
+
+jobs:
+  analyze:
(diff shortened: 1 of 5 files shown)


nickcharlton:
This is neat! I tidied up the default workflow so that it's a lot smaller.

It's highlighted one issue we should fix in: spec/example_app/app/models/doc_page.rb:7

https://github.com/thoughtbot/administrate/security/code-scanning/1

pablobm:
@nickcharlton - Fixed the security warning. I had to go out of my way a bit for the tool to accept it, as well as to be able to write a spec for it.

nickcharlton:
Oh nice, thanks!

PR #2156, opened 2022-03-07
Use locale for form label

nhippenmeyer:
This allows us to override the default form label for `BelongsTo` fields with a value in the locales file.


pablobm:
Always tricky with i18n stuff, but I think this might not be necessary. I was able to change the label in the demo app by specifying the foreign key as attribute, as opposed to the association name:

```
en:
  helpers:
    label:
      order:
        customer_id: "Esteemed patron"
```

Would you be able to check if this works for you?

nhippenmeyer:
You're right! Thanks for the reminder about how the I18n stuff works. I'll close this PR.

PR #2164, opened 2022-03-24
Use `field.name` rather than `field.nested_form.resource_name` for `Field::HasOne` labels

edwinwills:
If a model has multiple has_one relationships to the same model (e.g. two has_one associations with different scopes), we can't override the field label for the two associations, we can only change the translation for both at the same time, as the translation is based on the associated class name, not the association name.

This change uses the association name as the lookup field and default value for the label for has_one relationships, allowing multiple relations to the same class to have their own translated labels.

This is a potential breaking change for apps that have set a translation label based on the class name, but have a custom-named association (e.g. if a model has the association `:external_user` pointing to a `User` class, the translated label would be listed under `user` and would need to be changed to `external_user`.


pablobm:
Yup, I think this is correct :+1: The whole i18n setup needs a good lookover, but these incremental fixes may get us there one step at a time. Thank you!

PR #2180, opened 2022-04-13
Call .selectize() on all select fields

nhippenmeyer:
This enables using selectize with custom fields that inherit from `Administrate::Field::Select` without having to declare them in `associative.js`.

### DIFF
--- a/app/assets/javascripts/administrate/components/associative.js
+++ b/app/assets/javascripts/administrate/components/associative.js
@@ -1,5 +1,3 @@
 $(function() {
-  $('.field-unit--belongs-to select').selectize({});
-  $(".field-unit--has-many select").selectize({});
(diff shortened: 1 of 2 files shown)


pablobm:
Do you mean "fields that inherit from `Administrate::Field::Associative`" instead? `Select` fields don't use selectize (I think...?)

That aside, this might be a good idea, but I'm worried about cases where people create fields with a `select` field but for some reason don't want selectize on them.

Perhaps we could change the way that `html_class` works and make it `html_classes` instead. Then associative fields (and select fields) could use this to add a `selectize` class, and target those instead. This way a class that inherits from a selectizable one could opt out.

But perhaps I'm overthinking this :stuck_out_tongue: What do you think?

pablobm:

PR #2182, opened 2022-04-20
Change ApplicationController's routes's class to Set to speed up "valid_action?"

Shell-Wataru:
I changed ApplicationController's routes's class to Set to speed up `valid_action?`.

Let `N` be number of routes.
`valid_action?` checks all routes in current version. So the time complexity is `O(N)` .
When we store `routes` as Set, the time complexity is `O(log N)` .

My application has hundreds of routes. `valid_action?` took a long time.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -107,13 +107,11 @@ module Administrate
 
     helper_method :valid_action?
     def valid_action?(name, resource = resource_class)
-      !!routes.detect do |controller, action|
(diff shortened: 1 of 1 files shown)


pablobm:
Good call. Thank you!

Incidentally: is it right that the complexity of searching within a Set is `O(log(n))`? I would have expected it to be `O(1)`, using a hash table or something like that.

Shell-Wataru:
Oh、 You are right. I misunderstood Set's data structure. Thanks.

PR #2183, opened 2022-04-22
only include locales

jubilee2:
only include locales for plugin setting


pablobm:
Sounds good to me :+1: Thank you!

PR #2184, opened 2022-04-28
Consistent custom dashboards

pablobm:
Working on https://github.com/thoughtbot/administrate/pull/1941, I noticed that model-less dashboards (also known as "custom dashboards") do not behave very consistently. Eventually I found myself having to fix this before I could work on the former PR.

There are some twists and turns in the code that handles them:

1. There doesn't seem to be a way to have the navigation show a singular name for the dashboard. So for instance in the example app we have "Stats", but I couldn't create one with the name "Home".
2. When used with Punditize, it requires the policy to use a plural name (`HomesPolicy`) instead of singular like for any other resource.
3. The example `StatsDashboard` declares `resource "Stats"`, which is a weird mix of singular and plural (in the end it doesn't seem to matter which one you use here).
4. Internally this complicates things when working with authorization.

This PR introduces consistency, not fixing point 1 but at least addressing the other points and allowing me to proceed with my work at #1941. Original I was going to bundle it as part of the larger PR, but I thought it deserved scrutiny enough to have its own one.

As for the issue with singular names, this looks like a larger problem that should be looked into with more time, and it's probably related to how we currently can't generate dashboards for singular resources.

### DIFF
--- a/docs/adding_controllers_without_related_model.md
+++ b/docs/adding_controllers_without_related_model.md
@@ -9,7 +9,7 @@ To do that, you must define an `index` route, as only controllers with index
 routes are displayed in the sidebar and then add a custom dashboard:
 
 ```erb
-# app/views/admin/stats/index.html.erb 
+# app/views/admin/stats/index.html.erb
 
 <div style="padding: 20px">
   <h1>Stats</h1>
@@ -25,7 +25,7 @@ routes are displayed in the sidebar and then add a custom dashboard:
 require "administrate/custom_dashboard"
 
 class StatDashboard < Administrate::CustomDashboard
-  resource "Stats" # used by administrate in the views
+  resource "Stat" # used by administrate in the views
 end
 ```
 
--- a/docs/adding_controllers_without_related_model.md
+++ b/docs/adding_controllers_without_related_model.md
@@ -11,11 +11,10 @@ routes are displayed in the sidebar and then add a custom dashboard:
 ```erb
 # app/views/admin/stats/index.html.erb
 
-<div style="padding: 20px">
+<div>
   <h1>Stats</h1>
   <br>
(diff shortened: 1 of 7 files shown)


spec/example_app/app/views/admin/stats/index.html.erb
```diff
@@ -1,7 +1,6 @@
 <div style="padding: 20px">
   <h1>Stats</h1>
   <br>
-  <p><b>Total Customers:</b> <%= @stats[:customer_count] %></h1>
-  <br>
-  <p><b>Total Orders:</b> <%= @stats[:order_count] %></h1>
+  <p><b>Total Customers:</b> <%= @stats[:customer_count] %></p>
+  <p><b>Total Orders:</b> <%= @stats[:order_count] %></p>
```
pablobm, line 5: Fixed markup.

docs/adding_controllers_without_related_model.md
```diff
@@ -6,16 +6,15 @@ Sometimes you may want to add a custom controller that has no resource
...
 To do that, you must define an `index` route, as only controllers with index
-routes are displayed in the sidebar and then add a custom dashboard:
+routes are displayed in the sidebar, and then add a custom dashboard:
 
 ```erb
-# app/views/admin/stats/index.html.erb 
+# app/views/admin/stats/index.html.erb
 
-<div style="padding: 20px">
+<div>
```
pablobm, line 14: The example is copied straight from `spec/example_app/app/views/admin/stats/index.html.erb`. I have removed the padding because it's not relevant here.

pablobm:
Bringing this back to draft as I have to rethink it.

pablobm:
Looks like I don't need any of this in the end. Closing.

PR #2186, opened 2022-05-05
Correct grammar on German error messages

5minpause:
There are two changes, the first is simply adding a translation for the word 'error'.
The second improves on the grammar for notifying the user of errors.

The string "%{pluralized_errors} haben das Speichern dieses %{resource_name} verhindert:" is not always grammatically correct. That would depend on the amount of errors. The word "haben" has a singular and a plural form. The singular, for one error, needs to be "hat".
Also the `resource_name` would need to have a different grammatical case => "Genitiv".

To prevent adding logic to solve this, I rewrote the sentence to be simpler and always fit.

### DIFF
--- a/config/locales/administrate.de.yml
+++ b/config/locales/administrate.de.yml
@@ -21,8 +21,8 @@ de:
         more: "%{count} von %{total_count}"
         none: Keine
     form:
-      error: error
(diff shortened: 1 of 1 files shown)


pablobm:
Thank you!

PR #2195, opened 2022-05-19
Forgot this line when dropping Ruby 2.6

pablobm:
This time I'm waiting for this build to pass.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,4 +121,3 @@ workflows:
       - ruby-31
       - ruby-30
       - ruby-27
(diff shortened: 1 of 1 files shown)


pablobm:
Hm... why isn't CircleCI running? What did I break this time?

pablobm:
Oh, ok. It's running now.

pablobm:
OK. We are good now. The build is failing, but it's for a different reason that was ocurring already before updating Capybara. I'll merge this and will fix the other issue at some other point.

PR #2198, opened 2022-05-23
Dart-sass compatibility

brent-yearone:
When updating to dartsass-rails (to replace the deprecated sassc-rails), and get my hands on some SCSS module hawtness, SASS issues deprecation warnings that look like this:

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($base-spacing, 2) or calc($base-spacing / 2)
```

The easy fix, while maintaining backwards compatibility, is multiply by 0.5 instead of dividing by two -- that syntax works effectively in dart-sass as well as sassc.

### DIFF
--- a/app/assets/stylesheets/administrate/base/_forms.scss
+++ b/app/assets/stylesheets/administrate/base/_forms.scss
@@ -85,7 +85,7 @@ textarea {
 [type="checkbox"],
 [type="radio"] {
   display: inline;
-  margin-right: $small-spacing / 2;
+  margin-right: $small-spacing * 0.5;
 }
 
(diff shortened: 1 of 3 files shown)


brent-yearone:
The existing tests cover this change, which is effectively neutral with regards to the compiled CSS output.

pablobm:
Indeed, I don't think this needs to wait for the other situation to resolve. Thank you for the simple PR!

brent-yearone:
You're very welcome! Thank you for all the work you do on administrate! 😀

PR #2200, opened 2022-05-26
Has one translations

### LINKED ISSUE #2185: I18n translations for nested resources on custom designs not working
Thank you for this wonderful gem.

* What were you trying to do?
I customized the designs to use TailwindCSS. Everything is working, but the translation for nested records does not work. The same models have their translations on their own index and show pages.

We are looking at a custom show page.
Bildschirmfoto 2022-05-02 um 13 54 31
The nested `Address` record has no localized labels. The same view for the same record on its own SHOW page:
Bildschirmfoto 2022-05-02 um 13 57 55

And the `Address`es INDEX page:
Bildschirmfoto 2022-05-02 um 13 58 04

The code to render the attributes on the SHOW page:

I believe it is almost the same as the original one.
```ruby
<% page.attributes.each do |attribute| %>
  <div class="py-4 sm:py-5 sm:grid sm:grid-cols-3 sm:gap-4 sm:px-6">
    <dt class="text-sm font-medium text-gray-500" id="<%= attribute.name %>">
      <%= t(
        "helpers.label.#{resource_name}.#{attribute.name}",
        default: page.resource.class.human_attribute_name(attribute.name),
      ) %>
    </dt>
    <dd class="mt-1 text-sm text-gray-900 sm:mt-0 sm:col-span-2">
      <%= render_field attribute, page: page %>
    </dd>
  </div>
<% end %>
```
Do I need to edit some other files to achieve this?

  - Rails 7.0.2.4
  - administrate 0.17.0

pablobm:
Fixes #2185.

The template `app/views/fields/has_one/_show.html.erb` wasn't using the correct i18n key to translate the field names of the associated record.

This PR includes a heavy revamp of `spec/administrate/views/fields/has_one/_show_spec.rb`, which needed some TLC in order to work with it.

The diff for `lib/administrate/field/associative.rb` looks a bit misleading. The actual change is that I moved the definition of `associated_class_name` to be above the `private` declaration.

### DIFF
--- a/app/views/fields/has_one/_show.html.erb
+++ b/app/views/fields/has_one/_show.html.erb
@@ -28,7 +28,7 @@ All show page attributes of has_one relationship would be rendered
       <div>
       <dt class="attribute-label">
         <%= t(
-          "helpers.label.#{resource_name}.#{attribute.name}",
+          "helpers.label.#{field.associated_class_name.underscore}.#{attribute.name}",
           default: attribute.name.titleize,
         ) %>
       </dt>
--- a/lib/administrate/field/associative.rb
+++ b/lib/administrate/field/associative.rb
@@ -31,12 +31,6 @@ module Administrate
         end
       end
 
-      private
-
-      def associated_dashboard
-        "#{associated_class_name}Dashboard".constantize.new
-      end
-
       def associated_class_name
         if option_given?(:class_name)
           deprecated_option(:class_name)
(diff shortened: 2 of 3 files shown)


nickcharlton:
@pablobm, do you think you could rebase this? I merged in #1941 and I can't work out how to combine them!

nickcharlton (approved):

pablobm:
@nickcharlton I think this is ready now. That was tricky! :sweat: There has to be a better way to test this stuff...

nickcharlton:
Yeah, they're pretty rough! I didn't even know where to start when I tried myself!

PR #2208, opened 2022-06-07
Enable ordering the BelongsTo fields by using order option.

patrickgramatowski:
Enable ordering the BelongsTo fields by using order option. The change let the gem users to be able to order the fields by the column they'll specify, instead of ordering by `id`.

I've also updated the query to solve the possible issues with `PG::AmbiguousColumn` errors - for now it's not handled in any way, so if the user is using the tables that use the same columns names, then the error will apear and the only way to fix that is to monkey patch the gem. To solve that, I've added the column name as the prefix for both possible forms of the query.

Based on #1215
And kinda related to #1805 the result is exactly as presented on videos attached to this task's description.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -77,8 +77,9 @@ which are specified through the `.with_options` class method:
 
 **Field::BelongsTo**
 
-`:order` - Specifies the order of the dropdown menu, can be ordered by more
-than one column. e.g.: `"name, email DESC"`.
+`:order` - Specifies the column used to order the records. It will apply both in
+the table views and in the dropdown menu on the record forms.
+You can set multiple columns as well with direction. E.g.: `"name, email DESC"`.
 
 `:scope` - Specifies a custom scope inside a callable. Useful for preloading.
 Example: `.with_options(scope: -> { MyModel.includes(:rel).limit(5) })`
--- a/lib/administrate/order.rb
+++ b/lib/administrate/order.rb
@@ -1,9 +1,9 @@
 module Administrate
   class Order
-    def initialize(attribute = nil, direction = nil, field = nil)
+    def initialize(attribute = nil, direction = nil, association_attribute: nil)
       @attribute = attribute
       @direction = sanitize_direction(direction)
-      @field = field
+      @association_attribute = association_attribute
     end
 
     def apply(relation)
@@ -33,7 +33,7 @@ module Administrate
 
     private
 
-    attr_reader :attribute, :field
+    attr_reader :attribute, :association_attribute
 
     def sanitize_direction(direction)
       %w[asc desc].include?(direction.to_s) ? direction.to_sym : :asc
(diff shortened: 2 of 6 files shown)


lib/administrate/order.rb
```diff
@@ -68,7 +73,30 @@ def order_by_count(relation)
...
     def order_by_id(relation)
-      relation.reorder("#{foreign_key(relation)} #{direction}")
+      relation.reorder(Arel.sql(order_by_id_query(relation)))
+    end
+
+    def order_by_attribute(relation)
+      return order_by_id(relation) unless field
+
+      relation.joins(
+        attribute.to_sym,
```
patrickgramatowski, line 83: Is that required? It'll remove the records with no value in the specified column.

lib/administrate/order.rb
```diff
@@ -12,11 +13,15 @@ def apply(relation)
...
       return relation.reorder(Arel.sql(order)) if
-        relation.columns_hash.keys.include?(attribute.to_s)
+        column_exist?(relation, attribute)
 
       relation
     end
 
+    def column_exist?(table, column_name)
+      table.columns_hash.key?(column_name.to_s)
+    end
```
patrickgramatowski, line 23: TODO: It should be a private method.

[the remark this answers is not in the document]
patrickgramatowski: Done! ✅

pablobm (commented):
Looking good! I have left comments on minor points. Let me know what you think.

spec/example_app/app/dashboards/order_dashboard.rb
```diff
@@ -10,7 +10,7 @@ class OrderDashboard < Administrate::BaseDashboard
     address_city: Field::String,
     address_state: Field::String,
     address_zip: Field::String,
-    customer: Field::BelongsTo,
+    customer: Field::BelongsTo.with_options(order: "name"),
```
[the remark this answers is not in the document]
pablobm: Sorry, only now I realize that the option already exists :sweat_smile: In which case, ok, leave the name as it is.

Would you be able to correct the documentation at https://github.com/thoughtbot/administrate/blob/main/docs/customizing_dashboards.md? Currently it says it's about the dropdown menu (in the form page) only.

patrickgramatowski:
AFAIU the failing CI tests are not related to my changes. I think I've done everything that you asked me to update/change/fix. 👍

docs/customizing_dashboards.md
```diff
@@ -77,8 +77,8 @@ which are specified through the `.with_options` class method:
 
 **Field::BelongsTo**
 
-`:order` - Specifies the order of the dropdown menu, can be ordered by more
-than one column. e.g.: `"name, email DESC"`.
+`:order` - Specifies the order of either the dropdown menu and the table,
+can be ordered by more than one column. e.g.: `"name, email DESC"`.
```
pablobm, line 81: I think we can use a better phrasing here:

```suggestion
`:order` - Specifies the column used to order the records. It will apply both in
the table views and in the dropdown menu on the record forms. 
You can set multiple columns as well with direction. E.g.: `"name, email DESC"`.
```

pablobm (commented):
Great 🙂 Just one more thing and I think we can merge this.

docs/customizing_dashboards.md
patrickgramatowski: Done ✅

pablobm (approved):
Thank you!

Issue #1697, opened 2020-07-04
Can't avoid under/overfetching


sedubois:
## What were you trying to do?

I added bullet to my project, which shows that I should add eager loading for my dashboard relationships to avoid N+1 queries when showing the index. For instance I have a `Course` with `program: Field::BelongsTo`, and this `program` is visible on the course index.

Bullet throws:
```
Bullet::Notification::UnoptimizedQueryError: user: sdubois
GET /admin/courses
USE eager loading detected
  Program => [:translations]
  Add to your query: .includes([:translations])
```

The issue is gone if I remove `program` from the Course's `COLLECTION_ATTRIBUTES`.

So following the [documentation](http://administrate-prototype.herokuapp.com/customizing_dashboards), I changed `program:` to `Field::BelongsTo.with_options(scope: -> { Program.eager_load(:translations) })`. 
 
## What did you end up with (logs, or, even better, example apps are great!)?

The option has no effect. In fact even if I change to `Field::BelongsTo.with_options(scope: -> { raise "I got called!" })`, it does not raise. It seems that this `scope` option is not applied.

## What versions are you running?
  - Rails 6.0.3.2
  - administrate 0.14.0

pablobm:
Good catch! I think that what you describe should have been caught by `BaseDashboard#collection_includes`. Maybe worth looking into there to see what's going on?

As for `:scope`, I'm entertaining the possibility of removing it altogether. I think that the same effect can be reached by other means, using ActiveRecord's own features. I'm not 100% sure yet though.

sedubois:
## Issue 1: BelongsTo in course collection

@pablobm based on your hint, the following workaround in `Course` dashboard allows to get rid of the N+1 when viewing my course index with a BelongsTo association:
```rb
  def collection_includes
    super + [program: :translations]
  end
```

(The above is necessary because `ProgramDashboard#display_resource(program)` calls program.title where `title` is loaded through a `translations` association by Mobility.)

OK so now the course index works. 

## Issue 2: HasMany/BelongsTo circular reference in course show

Now I try to view a course. I have a data over-fetching of Course because Course has an associative attribute of type HasMany: `course_modules: Field::HasMany`, included in `SHOW_PAGE_ATTRIBUTES`, and CourseModule has a `course: Field::BelongsTo` attribute included in its `COLLECTION_ATTRIBUTES` (i.e. the Course and CourseModule show each other).

<details>

```rb
[...]

PR #2211, opened 2022-06-14
Index eager load performance

jamie:
Attempted fix for [Can't avoid under/overfetching](https://github.com/thoughtbot/administrate/issues/1697#)

Hey @pablobm - I took your suggestion from [point 1 here](https://github.com/thoughtbot/administrate/issues/1697#issuecomment-1004606731) and ran it a little further to get the tests passing.

- Update `attribute_includes` to restrict eager loading to BelongsTo and HasOne association fields
- Also created `item_associations` and `attribute_associated` methods with the old behaviour so that `sanitized_order_params` in ApplicationHelper still works correctly - it was needed to fix a few specs I think related to column sorting.
- No new specs, as there should be no behavioural change to users.

---

Old request logs from the test app (excluding renders):

```
Started GET "/admin/customers?search=jesus" for 127.0.0.1 at 2022-06-13 20:38:22 -0700
   (1.2ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
Processing by Admin::CustomersController#index as HTML
  Parameters: {"search"=>"jesus"}
  Customer Load (2.4ms)  SELECT "customers".* FROM "customers" LEFT OUTER JOIN "countries" ON "countries"."code" = "customers"."country_code" WHERE "customers"."hidden" = $1 AND (LOWER(CAST("customers"."email" AS CHAR(256))) LIKE '%jesus%' OR LOWER(CAST("customers"."name" AS CHAR(256))) LIKE '%jesus%' OR LOWER(CAST("customers"."kind" AS CHAR(256))) LIKE '%jesus%' OR LOWER(CAST("countries"."name" AS CHAR(256))) LIKE '%jesus%') LIMIT $2 OFFSET $3  [["hidden", false], ["LIMIT", 20], ["OFFSET", 0]]
  Order Load (0.4ms)  SELECT "orders".* FROM "orders" WHERE "orders"."customer_id" = $1  [["customer_id", 1]]
  LogEntry Load (0.4ms)  SELECT "log_entries".* FROM "log_entries" WHERE "log_entries"."logeable_type" = $1 AND "log_entries"."logeable_id" = $2  [["logeable_type", "Customer"], ["logeable_id", 1]]
  Country Load (0.6ms)  SELECT "countries".* FROM "countries" WHERE "countries"."code" = $1  [["code", "CN"]]
  LineItem Load (0.4ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 1]]
  ↳ app/models/order.rb:16:in `map'
  LineItem Load (0.2ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 2]]
  ↳ app/models/order.rb:16:in `map'
  LineItem Load (0.2ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 3]]
  ↳ app/models/order.rb:16:in `map'
  LineItem Load (0.3ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 4]]
  ↳ app/models/order.rb:16:in `map'
Completed 200 OK in 1502ms (Views: 1442.8ms | ActiveRecord: 22.1ms | Allocations: 803206)
```

Same request, with these changes:

```
Started GET "/admin/customers?search=jesus" for 127.0.0.1 at 2022-06-13 20:39:38 -0700
   (0.4ms)  SELECT "schema_migrations"."version" FROM "schema_migrations" ORDER BY "schema_migrations"."version" ASC
Processing by Admin::CustomersController#index as HTML
  Parameters: {"search"=>"jesus"}
  Customer Load (2.4ms)  SELECT "customers".* FROM "customers" LEFT OUTER JOIN "countries" ON "countries"."code" = "customers"."country_code" WHERE "customers"."hidden" = $1 AND (LOWER(CAST("customers"."email" AS CHAR(256))) LIKE '%jesus%' OR LOWER(CAST("customers"."name" AS CHAR(256))) LIKE '%jesus%' OR LOWER(CAST("customers"."kind" AS CHAR(256))) LIKE '%jesus%' OR LOWER(CAST("countries"."name" AS CHAR(256))) LIKE '%jesus%') LIMIT $2 OFFSET $3  [["hidden", false], ["LIMIT", 20], ["OFFSET", 0]]
  Country Load (0.2ms)  SELECT "countries".* FROM "countries" WHERE "countries"."code" = $1  [["code", "CN"]]
  Order Load (0.6ms)  SELECT "orders".* FROM "orders" WHERE "orders"."customer_id" = $1  [["customer_id", 1]]
  ↳ app/models/customer.rb:24:in `map'
  LineItem Load (1.0ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 1]]
  ↳ app/models/order.rb:16:in `map'
  LineItem Load (0.3ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 2]]
  ↳ app/models/order.rb:16:in `map'
  LineItem Load (0.1ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 3]]
  ↳ app/models/order.rb:16:in `map'
  LineItem Load (0.1ms)  SELECT "line_items".* FROM "line_items" WHERE "line_items"."order_id" = $1  [["order_id", 4]]
  ↳ app/models/order.rb:16:in `map'
   (0.6ms)  SELECT COUNT(*) FROM "log_entries" WHERE "log_entries"."logeable_id" = $1 AND "log_entries"."logeable_type" = $2  [["logeable_id", 1], ["logeable_type", "Customer"]]
  ↳ app/views/fields/has_many_variant/_index.html.erb:6
Completed 200 OK in 1394ms (Views: 1348.5ms | ActiveRecord: 18.2ms | Allocations: 799880)
```

This change converted the log_entries from an eager load to just a count. The orders association would also have converted, except that there's that silly `lifetime_value` calculation that does some manual math on the orders, so it gets loaded in for that anyway (and is its own N+1 problem, but that's not Administrate's fault).

### DIFF
--- a/app/helpers/administrate/application_helper.rb
+++ b/app/helpers/administrate/application_helper.rb
@@ -52,7 +52,7 @@ module Administrate
     end
 
     def sanitized_order_params(page, current_field_name)
-      collection_names = page.item_includes + [current_field_name]
+      collection_names = page.item_associations + [current_field_name]
       association_params = collection_names.map do |assoc_name|
         { assoc_name => %i[order direction page per_page] }
       end
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -17,7 +17,7 @@ module Administrate
       end
 
       def self.eager_load?
-        self <= BelongsTo || self <= HasOne
+        false
       end
 
       def self.searchable?
--- a/lib/administrate/field/base.rb
+++ b/lib/administrate/field/base.rb
@@ -16,6 +16,10 @@ module Administrate
         self < Associative
       end
 
+      def self.eager_load?
+        self <= BelongsTo || self <= HasOne
+      end
+
       def self.searchable?
         false
       end
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Administrate
 
-[![CircleCI](https://img.shields.io/circleci/project/github/thoughtbot/administrate.svg)](https://circleci.com/gh/thoughtbot/administrate/tree/master)
+[![CircleCI](https://img.shields.io/circleci/project/github/thoughtbot/administrate.svg)](https://circleci.com/gh/thoughtbot/administrate/tree/main)
 [![Gem Version](https://badge.fury.io/rb/administrate.svg)](https://badge.fury.io/rb/administrate)
 [![Code Climate](https://codeclimate.com/github/thoughtbot/administrate/badges/gpa.svg)](https://codeclimate.com/github/thoughtbot/administrate)
 [![Reviewed by Hound](https://img.shields.io/badge/Reviewed_by-Hound-8E64B0.svg)](https://houndci.com)
--- a/lib/administrate.rb
+++ b/lib/administrate.rb
@@ -20,4 +20,14 @@ module Administrate
       "if you think otherwise.",
     )
   end
+
+  def self.warn_of_deprecated_method(klass, method)
+    ActiveSupport::Deprecation.warn(
+      "The method #{klass}##{method} is deprecated. " +
+      "If you are seeing this message you are probably " +
+      "using a dashboard that depends explicitly on it. " +
+      "Please make sure you update it to a version that " +
+      "does not use a deprecated API"
+    )
+  end
 end
--- a/lib/administrate.rb
+++ b/lib/administrate.rb
@@ -27,7 +27,7 @@ module Administrate
       "If you are seeing this message you are probably " +
       "using a dashboard that depends explicitly on it. " +
       "Please make sure you update it to a version that " +
-      "does not use a deprecated API"
+      "does not use a deprecated API",
     )
   end
 end
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -94,6 +94,8 @@ module Administrate
     end
 
     def item_includes
+      # Deprecated, internal usage has moved to #item_associations
+      Administrate.warn_of_deprecated_method(self.class, :item_includes)
       attribute_includes(show_page_attributes)
     end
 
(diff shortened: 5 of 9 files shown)


app/helpers/administrate/application_helper.rb
```diff
@@ -52,7 +52,7 @@ def resource_index_route(resource_name)
     end
 
     def sanitized_order_params(page, current_field_name)
-      collection_names = page.item_includes + [current_field_name]
+      collection_names = page.item_associations + [current_field_name]
```
pablobm, line 55: If I have this right, it looks like `item_includes` won't be used any more, so I think it can be removed. Perhaps not straightaway, but starting with a deprecation warning like we do at https://github.com/thoughtbot/administrate/blob/fd83fbbb125446f782d9713229c69dcc1d826843/lib/administrate/field/associative.rb#L40-L42.

lib/administrate/field/base.rb
```diff
@@ -16,6 +16,10 @@ def self.associative?
         self < Associative
       end
 
+      def self.eager_load?
+        self <= BelongsTo || self <= HasOne
+      end
```
pablobm, line 21: I think a better way to do this would be to have subclasses implement it. So the default here in `Field::Base` would be always `false`, but then `Field::BelongsTo#eager_load?` would always return `true` (same for `HasOne`). What do you think?

jamie: Good OO 👍

pablobm:
Thank you!

PR #2212, opened 2022-06-14
Add correct hyperlinks on nested models in collections

paulRbr:
## Context

E.g. In the following routes, `Order` model can only be shown (or
edited) when it's accessed via it's nested `/customers/:customer_id/orders` route. If we imagine order ids to be dependent of customers (`Order#1` for `Customer#1` and `Order#1` for `Customer#2`):
```ruby
  namespace(:admin) do
    resources :customers do
      resources :orders, only: [:show, :edit]
    end
  end
end
```

Until now, administrate would not link the associated models in collections when their actions can only work correctly with the existence of a parent resource (by defining a `Admin::OrdersController#find_resource` method to find the nested resource depending on URL `params`).

## Suggestion

This PR tries to allow models which are nested within a parent related model to be accessed via a nested URL.

I'm not 100% happy with the current codebase (it has lots of duplication) but I wanted to share the main idea which works for us @bump-sh.

Please let me know if this feature could see life in Administrate and how I can make this suggestion better?
Thanks!

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -106,12 +106,20 @@ module Administrate
     end
 
     helper_method :valid_action?
-    def valid_action?(name, resource = resource_class)
-      routes.include?([resource.to_s.underscore.pluralize, name.to_s])
+    def valid_action?(name, resource = resource_class, parent_resource = nil)
+      required_parts = routes[[resource.to_s.underscore.pluralize, name.to_s]]
+
+      return false if required_parts.nil?
+
+      if parent_resource.nil?
+        required_parts.difference(["id"]).empty?
+      else
+        required_parts.include?("#{parent_resource.to_s.underscore}_id")
+      end
     end
 
     def routes
-      @routes ||= Namespace.new(namespace).routes.to_set
+      @routes ||= Namespace.new(namespace).routes
     end
 
     def records_per_page
--- a/app/views/administrate/application/_collection_header_actions.html.erb
+++ b/app/views/administrate/application/_collection_header_actions.html.erb
@@ -1,4 +1,6 @@
-<% [valid_action?(:edit, collection_presenter.resource_name),
-    valid_action?(:destroy, collection_presenter.resource_name)].count(true).times do %>
+<% [
+    valid_action?(:edit, collection_presenter.resource_name, page.try(:resource).presence),
+    valid_action?(:destroy, collection_presenter.resource_name, page.try(:resource).presence)
+].count(true).times do %>
   <th scope="col"></th>
 <% end %>
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -59,13 +59,22 @@ to display a collection of resources in an HTML table.
   <tbody>
     <% resources.each do |resource| %>
       <tr class="js-table-row"
-          <% if valid_action?(:show, resource.class) && show_action?(:show, resource) %>
-            <%= %(tabindex=0 role=link data-url=#{polymorphic_path([namespace, resource])}) %>
+          <% if page.try(:resource).present? && valid_action?(:show, resource.class, page.resource.class) && show_action?(:show, resource) %>
+          <%= %(tabindex=0 role=link data-url=#{polymorphic_path([namespace, page.resource, resource])}) %>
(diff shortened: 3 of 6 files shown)


app/controllers/administrate/application_controller.rb
```diff
@@ -106,8 +106,19 @@ def nav_link_state(resource)
     end
 
     helper_method :valid_action?
-    def valid_action?(name, resource = resource_class)
-      routes.include?([resource.to_s.underscore.pluralize, name.to_s])
+    def valid_action?(name, resource = resource_class, parent_resource = nil)
+      !!routes.detect do |controller, action, required_parts|
```
pablobm, line 110: This change from using `routes.include?` to using `routes.detect`. This brings us back to a O(n) algorithm, which is slow for apps with many routes: https://github.com/thoughtbot/administrate/pull/2182.

app/views/administrate/application/_collection_header_actions.html.erb
```diff
@@ -1,4 +1,8 @@
-<% [valid_action?(:edit, collection_presenter.resource_name),
-    valid_action?(:destroy, collection_presenter.resource_name)].count(true).times do %>
+<% [
+    page.try(:resource).present? && valid_action?(:edit, collection_presenter.resource_name, page.resource) ||
+    valid_action?(:edit, collection_presenter.resource_name),
+    page.try(:resource).present? && valid_action?(:destroy, collection_presenter.resource_name, page.resource) ||
+    valid_action?(:destroy, collection_presenter.resource_name)
```
pablobm, line 5: I think it's possible to simplify these two:

```suggestion
    valid_action?(:edit, collection_presenter.resource_name, page.try(:resource).presence),
    valid_action?(:destroy, collection_presenter.resource_name, page.try(:resource).presence)
```

pablobm (commented):
Hello @paulRbr. Thank you for taking the time to work on this.

This is unfortunately a complex issue. Finding a proper solution would probably require going a bit deeper into Administrate, I think cleaning other parts before it can be addressed, and potentially altering some design decisions. Your concern about duplication is quite valid, and I think it stems from the general difficulty of implementing this on Administrate as the design doesn't quite allow for it.

In summary, I am against this approach as it is a bit shoehorned. We should solve the problems within Administrate first, so that this is possible in the future.

An additional detail: changes are coming that will affect this change. The main one is that `valid_action?` and `show_action?` are due to change with https://github.com/thoughtbot/administrate/pull/1941. When this happens, your changes will have to adapt.

app/controllers/administrate/application_controller.rb
paulRbr: Indeed. Thanks for the heads up.

I've tried to change the data structure to a hash, to keep the O(1) access of the route in the routes list and be able to store “required action parts” too.

I guess it would require more memory, but would keep the runtime access complexity similar, do you think that would be suitable?

app/views/administrate/application/_collection_header_actions.html.erb
paulRbr: Well spotted, thanks!

paulRbr:
Hi @pablobm thanks for the first review, I've adapted the code with your remarks.

I understand the solution is not ideal and it would probably require some more in depth work in administrate itself. However for now we don't have so much energy (at my company), nor do I have it (personally) to spend on solving this in administrate. I'll try to maintain this branch up-to-date while administrate codebase evolves because we need this solution for our needs. Let's see if we can take it to a better direction while iterating!

If you have some pointers of where I could be of any help to find a better solution, don't hesitate to mention it. 
Thanks again

app/controllers/administrate/application_controller.rb
pablobm: A hash would probably work too, yes.

pablobm:
@paulRbr - That's fair enough. You maintaining your own branch is very possibly the best way forward at the moment.

Regarding what would need to change in order to fix this properly, it's difficult to tell as I haven't looked in depth. An idea (just now from the top of my mind) that might work is a wrapper to `link_to` (eg: `link_to_dashboard(resource, :edit)` or something like that) that uses an abstraction of the resource can be used to tell its path without the assumptions that we have currently. But getting there is going to require other changes: to page objects, dashboards, probably the resolver...  Potentially a lot :sweat_smile:

For now, I'm going to close this PR for the sake of housekeeping. Feel free to keep making changes and leaving comments if you want to try your hand at take it further.

PR #2215, opened 2022-06-21
Fix typos and formatting in hiding dashboard docs

dharamgollapudi:
Fixes the typo and adds some basic formatting.

### DIFF
--- a/docs/guides/hiding_dashboards_from_sidebar.md
+++ b/docs/guides/hiding_dashboards_from_sidebar.md
@@ -2,7 +2,8 @@
 title: Hiding Dashboards from the Sidebar
 ---
 
-Resources can be removed form the sidebar by removing their index action from the routes. For example:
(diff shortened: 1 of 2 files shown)


docs/guides/hiding_dashboards_from_sidebar.md
```diff
@@ -16,4 +16,4 @@ Rails.application.routes.draw do
 end
 ```
 
-In this case, only Orders and Products will appear in the sidebar, while Line Items can still appear as an association.
+In this case, only `Orders` and `Products` will appear in the sidebar, while `Line Items` can still appear as an association.
```
pablobm, line 19: Nitpick: these are not class names (eg: the class is `Order`, not `Orders`) so I'm not sure about formatting them as code. How about using double quotes instead?

```suggestion
In this case, only "Orders" and "Products" will appear in the sidebar, while "Line Items" can still appear as an association.
```

PR #2216, opened 2022-06-22
Move pagination into private method for overriding

dwilkie:
Allows you to do this:

```ruby
module Admin

    private

    def paginate_resources(resources)
      super.without_count
    end
  end
end
```


pablobm:
I'm generally in favour of moving those things to private methods. So I'm in! :+1: Thank you.

PR #2225, opened 2022-07-16
Document how to customize Field::Select option labels

cabe56:
Customizing select dropdowns helps keep the UI friendly without having to store lengthy values in a database column.

After looking through the `customizing_dashboard` docs I did not find a reference on how to do this. These resources helped me figure out how the select field's `_form` view handled option labels:
- https://github.com/thoughtbot/administrate/blob/main/app/views/fields/select/_form.html.erb
- https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/options_from_collection_for_select

These are before/after screenshots of this PR:
Screen Shot 2022-07-15 at 7 13 08 PM
Screen Shot 2022-07-15 at 7 12 53 PM

Documenting support for this would imply supporting `[[value, label]]` as an input for `:collection` in `Field::Select.with_options`. If this is a concern I can write some tests to ensure this is accounted for in future releases.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -221,6 +221,17 @@ objects to display as.
 `:collection` - Specify the options shown on the select field. It accept either
 an array or an object responding to `:call`. Defaults to `[]`.
 
+To customize option labels, pass an array of pairs where the first element is the value submitted with the form and the second element is the label shown to the user.
+
+For example:
+
+```ruby
+  currency = Field::Select.with_options(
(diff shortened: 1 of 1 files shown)


pablobm:
Uh, good point. I agree we should document this. I'm thinking that perhaps we should mention that it's the same options that would be passed to `options_from_collection_for_select` (as you correctly point out) with a link to its documentation. What do you think?

pablobm:
Actually, I'm wrong. The code does assume things, with that `:first`/`:last` in the array case... OK. Let's leave it as you propose :+1:

PR #2227, opened 2022-07-21
Update Rails out of CVE-2022-32224

pablobm:
Rails 6.1.6 fails the gem audit due to a security advisory. Let's upgrade out of it.

Ideally I should upgrade to Rails 7, but there's something tricky about it. I'm going to stick to 6.1.x for now as the security advisory breaks the build, and therefore it is more important.


pablobm:
I was upgrading the example app to Rails 7 here, and it turns out that we need some changes to make it work with Rails 7. And that the Appraisal-driven Rails 7 app was actually running Rails 6.1 :sweat: So I'll fix that first.

pablobm:
Giving up on the Rails 7 update and sticking to 6.1.* for now. Updated description to reflect this.

PR #2237, opened 2022-08-08
Move bundle-audit to GitHub Actions

nickcharlton:
Following the GitHub Actions pattern of having one check per service,
rather than one big check for faster feedback.

This also means we no longer need to bundle `bundler-audit`.


pablobm (approved):
Nice!

PR #2238, opened 2022-08-08
Pass associated_class to collection from show

stormmaster42:
This fixes an issue where the embedded collection is using the parent's class to define i18n headers.

Example: Ressource order has_many order_items.
Order dashboard declares `order_items: Field::HasMany`

When the show which includes the `order_items` association is rendered within the collection partial it uses the parent's resource (`Order`) instead of `OrderItem` to determine the header label default.

Not sure how to add a test for this behavior.

### DIFF
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -28,6 +28,7 @@ from the associated resource class's dashboard.
     page: page,
     resources: field.resources(page_number, order),
     table_title: field.name,
+    resource_class: field.associated_class,
   ) %>
   <% if field.more_than_limit? %>
     <%= paginate field.resources(page_number), param_name: "#{field.name}[page]" %>


pablobm:
Thank you for the MR :slightly_smiling_face: I'm trying to reproduce the issue locally, but I'm not having any luck. Would you be able to explain it in terms of the example app?

For instance, in the example app we have `Customer.has_many :orders`, and this is reflected in `CustomerDashboard`. I have tried to replicate the issue by adding this translation:

```yaml
---
en:
  # ...
  helpers:
    label:
      order:
        address_state: "State code"
```

When I visit a customer show page, I do see the translation:

Customer dashboard showing the translation

Perhaps I'm looking at the wrong thing?

jubilee2:
https://github.com/thoughtbot/administrate/blob/main/app/views/administrate/application/_collection.html.erb#L37
i believed related here

stormmaster42:
Oh yes. Sorry forgot to add that detail. Yes it's related to the active record fallback.
So if you declare the translation on attribute level like so:
```
en:
  activerecord:
    attributes:
      order_item:
        address_state: "State code"
```
it would be picked up by the order item dashboard, but not the embedded order item list in the order show.

pablobm:
Oh, I see! Thank you for explaining. The fact that `resource_class` can be either a method or a variable was confusing in particular. I get it now :+1: 

Don't worry about testing. This is too tricky (I already spent a lot of time trying to test something similar elsewhere and I'm not sure that there's much return of investment to be had). Merging.

PR #2239, opened 2022-08-09
Move pagination into partial

dwilkie:
Related to #2216, allows the user to override the pagination partial without having to override the entire [index.html.erb](https://github.com/thoughtbot/administrate/blob/main/app/views/administrate/application/index.html.erb) template.

Useful for using custom pagination options e.g:
https://github.com/kaminari/kaminari#paginating-without-issuing-select-count-query


pablobm (approved):
:shipit:

PR #2243, opened 2022-08-10
Add guide on how to scope has_many relations

cabe56:
Documents third point in #1778

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -2,5 +2,6 @@
 title: Guides
 ---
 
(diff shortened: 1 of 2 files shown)


cabe56:
@pablobm took your [SO answer](https://stackoverflow.com/questions/59059701/rails-administrate-customize-has-many) as inspiration for a guide as you suggested in the issue

pablobm (approved):
Amazing, thank you!

PR #2245, opened 2022-08-11
remove redundant ARIA roles from elements with implicit role

### LINKED ISSUE #2244: Some HTML elements have redundant ARIA roles
* What were you trying to do?
I noticed that the `header` in the [index view](https://github.com/thoughtbot/administrate/blob/f10b5567916703a2ac242c2e9808cfafcfc7fbff/app/views/administrate/application/_index_header.html.erb#L5) has a redundant `role` property. There are other `header` elements that have the same issue, as well as a `main` with a redundant `role="main"`, a `nav` with `role="navigation"` and a `th` that has a `role="columnheader"`.

```
<header class="main-content__header" role="banner">
```

While this doesn't cause any technical issues, it's [bad practice and it encourages other developers to follow the same pattern](https://www.w3.org/TR/html-aria/#don-t-add-redundant-roles).

A list of all implicit roles is available here: [ARIA in HTML](https://www.w3.org/TR/html-aria/)

N/A

  - Rails: 7.0.2.3
  - administrate: 0.17.0

metamoni:
### What does this do?
Fixes #2244 by removing the `role` attribute from all HTML elements that have an implicit role.

### Why?
Because using the `role` attribute on HTML elements with implicit roles is bad practice.

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -27,7 +27,6 @@ to display a collection of resources in an HTML table.
         cell-label--<%= collection_presenter.ordered_html_class(attr_name) %>
         cell-label--<%= "#{collection_presenter.resource_name}_#{attr_name}" %>"
         scope="col"
-        role="columnheader"
         aria-sort="<%= sort_order(collection_presenter.ordered_html_class(attr_name)) %>">
         <%= link_to(sanitized_order_params(page, collection_field_name).merge(
           collection_presenter.order_params_for(attr_name, key: collection_field_name)
--- a/app/views/administrate/application/_index_header.html.erb
+++ b/app/views/administrate/application/_index_header.html.erb
@@ -2,7 +2,7 @@
   <%= display_resource_name(page.resource_name) %>
 <% end %>
 
-<header class="main-content__header" role="banner">
+<header class="main-content__header">
   <h1 class="main-content__page-title" id="page-title">
     <%= content_for(:title) %>
   </h1>
--- a/app/views/administrate/application/_navigation.html.erb
+++ b/app/views/administrate/application/_navigation.html.erb
@@ -7,7 +7,7 @@ for all resources in the admin dashboard,
 as defined by the routes in the `admin/` namespace
 %>
 
-<nav class="navigation" role="navigation">
+<nav class="navigation">
   <%= link_to(t("administrate.navigation.back_to_app"), root_url, class: "button button--alt button--nav") if defined?(root_url) %>
 
   <% Administrate::Namespace.new(namespace).resources_with_index_route.each do |resource| %>
--- a/app/views/administrate/application/edit.html.erb
+++ b/app/views/administrate/application/edit.html.erb
@@ -17,7 +17,7 @@ It displays a header, and renders the `_form` partial to do the heavy lifting.
 
 <% content_for(:title) { t("administrate.actions.edit_resource", name: page.page_title) } %>
 
-<header class="main-content__header" role="banner">
+<header class="main-content__header">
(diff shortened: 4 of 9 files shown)


cabe56:
@metamoni thanks for bringing this up, made me more aware of accessibility bad practices. I found more [references like the ones you edited in this PR in the codebase](https://github.com/thoughtbot/administrate/search?p=1&q=role%3D), do you think we should get rid of those too? 

From my limited understanding [[1]](https://www.accessibility-developer-guide.com/knowledge/aria/bad-practices/) [[2]](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role), it seems the only appropriate use of ARIA in Administrate is `search`. Do you think that's the case or are there other instances that adhere to best practices in the codebase?

metamoni:
> We could create another issue about removing this link role + styling as it takes more work and this PR is already a good addition to the codebase.

That sounds like a good idea. It might require some rethinking/redesign. It sounds like you're just looking to create links inside a table. As long as you make sure the links are either simple `a` elements or `link_to` elements, the `role` should not be needed. Either way, it sounds like there's no need for the rows themselves to be focusable/clickable, so you'll then be able to remove the `tabindex=0` on them too.

Happy to take a look at the issue/PR when it's ready, so feel free to tag me. If you let me know what the steps are to get to the right page, I can pull the branch and test the behaviour with a keyboard and screen reader.

pablobm:
Thank you so much for this!

Re: that `tabindex=0 role=link`, it was introduced in this PR: https://github.com/thoughtbot/administrate/pull/1557. The idea was to make the whole row clickable when possible, leaving links in the individual cell contents are a fallback. The reason is that the table row is a block element (or whatever table rows are) as opposed to the content of each cell being an inline element, more difficult to click on. Thoughts?

cabe56:
@pablobm thanks for the context. 

IMO this PR is good to go as @metamoni did not remove the `link` role which seems to be the only usage that has unintended consequences in other Administrate features (making rows look clickable as you mention). 

We can deal with that specific `link` role in a separate PR as it requires more thought (as @metamoni mentions here https://github.com/thoughtbot/administrate/pull/2245#issuecomment-1212996364).

With this PR merged and a new PR for `link` we can close #2244.

cabe56:
Was about to write an update to the issue regarding my suggestion about another PR, but it seems that I overcomplicated this whole thing: could we just change [the CSS selector](https://github.com/thoughtbot/administrate/blob/f91c6c4f7172fd3d64a6fa34f2b6f519813afb4b/app/assets/stylesheets/administrate/base/_tables.scss#L26) from `role=link` to something else?

For example, if we just replace `role=link` with a reference to `.js-table-row` we get the same styling without using a role.

metamoni:
@pablobm Hmm thanks for the context, it's really useful to understand why this change was introduced. I get the intention behind it, but I don't think that behavior of tabbing through rows and cells that way is an established pattern. [This article by Leonie Watson](https://tink.uk/how-screen-readers-navigate-data-tables/) explains how real assistive tech users would navigate a table. Unfortunately, I think making non-interactive elements focusable would actually make for a confusing experience in this case.

It's best to stick to semantic HTML and leave links as interactive elements, as assistive tech gives users various ways to navigate the page, navigate tables, find links, etc.

brunoprietog:
Hi, as a blind user here, I confirm that putting a role=link in a tr breaks the table navigation commands for screen readers.

pablobm:
Thank you all for your input! I agree that this PR is good to go. Merging.

Re: the row links, it does seem that you are all right and the current implementation is counterproductive. @metamoni - Would you be able to create an issue as you propose? Then we can agree on a suitable alternative.

PR #2250, opened 2022-08-17
Make the select box has same style as has_many box

jubilee2:
from 
image

to
image


pablobm (approved):
Makes sense :+1: Thank you

PR #2255, opened 2022-08-24
Redescribed

c4lliope:
Hello; hope all's good.
I looked back across our `docs`, and I can see much of the language is unchanged from the original,
produced around six years ago. I'm sure my speaking and manner of expression has changed much since,
and so should the composition of our code's guides.

I am going through and re-describing our code, as a means of reminding my preoccupied brain.
Once done on our docs / guides, I am hoping I can begin proposing an agenda of changes to be included in `0.19.0` and `0.20.0`;
I am hoping these couple releases can be used as a gauge of `1.0`-readiness,
and in some places break up old logic so more concise phrasings can be used.

Alongside a `1.0` (or any planned upcoming) release, I'd like to begin discussing a possible change of this library's name.
"Administrate" is nonsense based on the phrase "Administration", and has been a solid choice for many years.
Regardless, there is a chance to more clearly describe our engine.
I am hoping a change in name can spur some more clear direction in deciding on our engine's use cases.

In such direction, I booked the rubygems `enroll` and `enrolled` as possible options.
These names recommend an impression of a leader of a group or class, checking names and addresses on a clipboard.
I imagine a similar scene upon seeing any one of our dashboards, laid up in precise lines summarizing a user's records.
The purpose of a user's dashboard code is the "enrolling" of models, so our engine can handle displaying and actions.


c4lliope:
Alongside a release roadmap - I am making a summary page describing our plugins;
see https://rubygems.org/search?query=name%3Aadministrate (73 in all!)

Perhaps some more commonly-used plugins can be pulled inside our main codebase,
and repackaged as core engine use cases.

pablobm (commented):
Hello @c4lliope, fancy seeing you around :-)

Thank you for that revamp of the README. I have left some comments; in general I agree that it can be improved.

Re: a potential name change, personally I am hesitant about it. I feel that, when people go searching for engines such as this one, "administration" is one of the top (if not the top) word that comes to mind.

A piece of evidence on this would be the section [Rails Admin Interfaces](https://www.ruby-toolbox.com/categories/rails_admin_interfaces) at https://www.ruby-toolbox.com. Administrate is listed there alongside other gems that occupy a similar space, and the word "admin" is common to many of them.

Additionally, I feel that a name change would cause confusion among users, and would lose the recognition that Administrate already has as a name, which is not huge, but still sizeable enough that I have come across many people in the Rails community that knew of it.

I'm curious, what motivated you to think about a potential rename? Perhaps there are reasons that I'm failing to see.

c4lliope:
> A piece of evidence on this would be the section [Rails Admin Interfaces](https://www.ruby-toolbox.com/categories/rails_admin_interfaces) at https://www.ruby-toolbox.com/. Administrate is listed there alongside other gems that occupy a similar space, and the word "admin" is common to many of them.

@pablobm; many programmers are no longer using `master` in describing their main code branches. In a similar regard, `admin` is based on `minister`, and causes me pain upon seeing such a poorly chosen hokey phrase.

I'm also eager to begin making changes again because I had designed our code as being endlessly pre-1.0; our users recognize some changes could be coming along on the schedule, and so long as someone can easily upgrade (`rails generate ...`), I see no problems in choosing a more reliable, on-brand name. "Administrate" simply becomes a precursor library, and had helped us learn our user's needs more.

c4lliope:
Perhaps one more reason is the class of so-called "admin" engines is really bounded by prior norms;
our code's real skill is producing loosely-described pages, in a similar manner as low-code website builders.
I'd like a name encouraging us to expand our game and help people encode more complex use cases easily.

docs/permissions.md
```diff
@@ -0,0 +1,74 @@
...
+title: Permissions
+---
+
+Administrate's base logic assumes any signed-in super-user
+can access any record and run any operation.
+You can manage permissions by placing logic in your controllers.
+
+[Pundit] is a popular gem many people describe permissions rules in.
+Administrate's Pundit plugin applies your permissions across your dashboards.
+See the end of this page.
```
pablobm, line 11: I think that mentioning Pundit from the get go will lead some users to think that it is a requirement.

Or perhaps this could be rephrased to clarify that this is not the case:

```suggestion
You can write this logic yourself using your preferred tools. Alternatively, Administrate
provides an easy path to integrate with [Pundit], a popular gem for describing permissions.
For more details, read on.
```

docs/permissions.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Basic permissions
+
+Permissions can be described in `Admin::ApplicationController`
+or any subclass.
+
+Here are some examples:
```
pablobm, line 18: I think this doesn't make it clear enough that the examples are indeed the full API that users will need to override in order to implement authorization. How about the following?

```suggestion
The following are the methods that you will need to override in order to implement your own permission scheme, shown with example implementations:
```

docs/specialized_actions.md
```diff
@@ -62,9 +60,11 @@ Rails.application.routes.draw do
 end
 ```
 
-## Customizing Default Sorting
+## Specialized ordering
 
-To set the default sorting on the index action you could override `default_sorting_attribute` or `default_sorting_direction` in your dashboard controller like this:
+You can order records on index actions
```
pablobm, line 65: Ordering can be decided by the user in many cases by clicking the table headers. This piece of advice refers to the default order:

```suggestion
You can set the default order of records on index actions
```

docs/specialized_dashboards.md
```diff
@@ -0,0 +1,361 @@
...
+  ]
+end
+```
+
+You can change members displayed
+on each `index`, `show`, and `form` page,
+by adding or dropping member names as symbols in each corresponding array.
+
+Your `ATTRIBUTE_TYPES` hash should describe
+each member's shape, as dispalyed in each dashboard page.
```
pablobm, line 56:
```suggestion
each member's shape, as displayed in each dashboard page.
```

pablobm:
> admin is based on minister, and our name causes me pain upon seeing such a poorly chosen hokey phrase.

I'm curious, what is the issue with the etymology?

PR #2256, opened 2022-08-26
Fix Tailwind compatibility

geniuskidkanyi:
Migrated from sassc-rails to dartass-rails. this fixes the issue with new css specs that sassc doesnt support new css syntaxes and its been deperecated  few years now.


pablobm:
Closing in favour of #2311

PR #2252, opened 2022-08-19
rspec on github action

jubilee2:
* Create rspec.yml

PR #2260, opened 2022-09-08
Correctly detect filters with non-word params

### LINKED ISSUE #2228: Improve filter regex
I would like to be able to pass filter attribute in French, German, or other languages. Because now 
regex doesn't allow non-Latin characters.

In my case I have Mobility gem and want to search in other languages. I found solution with custom search method, but in my case it's enough to add this:
```
COLLECTION_FILTERS = {
    name: -> (resources, attr) { resources.search(attr) }
  }.freeze
```
and also I add search method to resource model with the logic that I need:
```
scope :search, -> (name_q) do
    i18n do
      name.lower.matches("%#{name_q&.downcase}%")
    end
  end
``` 

Switch \w to Unicode codes/ranges or `word.match?(/^#{filter}:.*$/)` but this way may be unsafe

Allow custom regex for filter attribute

cattekin:
Change the filter regex to return valid filter matches when the filter attribute isn't an ascii word object.

This fixes #2228 as well as allowing using comma separated values when filtering

It shouldn't matter when checking for the existence of a `filter` whether the characters to the right of the `:` in the `word` object contain things like unicode characters or punctuation.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -31,7 +31,7 @@ module Administrate
       private
 
       def filter?(word)
-        valid_filters&.any? { |filter| word.match?(/^#{filter}:.*$/) }
+        valid_filters&.any? { |filter| word.match?(/^#{filter}:/) }
       end
 
       def parse_query(query)
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -31,7 +31,7 @@ module Administrate
       private
 
       def filter?(word)
-        valid_filters&.any? { |filter| word.match?(/^#{filter}:\w*$/) }
+        valid_filters&.any? { |filter| word.match?(/^#{filter}:.*$/) }
       end
 
       def parse_query(query)


jubilee2:
can you explain more?

cattekin:
> can you explain more? I do not see related between this and #2252

sorry, somehow messed up the pasting the issue, the correct one is https://github.com/thoughtbot/administrate/issues/2228

jubilee2:
after reviewing the code,

I think the `@filters` is better to storage hash of name and param

is this good plan?
@pablobm @cattekin 
```ruby
#the def filter? return hash like {name: 'vip', param: ''}

filters=[{name: "kind", param: "vip"},{name: 'vip', param: ''}]

then we can use that filters in filter_results function
```

lib/administrate/search.rb
```diff
@@ -31,7 +31,7 @@ def to_s
       private
 
       def filter?(word)
-        valid_filters&.any? { |filter| word.match?(/^#{filter}:\w*$/) }
+        valid_filters&.any? { |filter| word.match?(/^#{filter}:.*$/) }
```
pablobm, line 34: I wonder: why do we even bother matching to the right of the colon? Would it be simpler to just do the following?

```suggestion
        valid_filters&.any? { |filter| word.match?(/^#{filter}:/) }
```

cattekin: Yeah, I agree 🤔 
also removes ambiguity as to what is being matched

pablobm:
Looks good. Thank you!

jubilee2:
it's should fixed for #2265

Issue #1931, opened 2021-03-05
Administrate::ResourceResolver always singularizes namespace. 


timobleeker:
* What were you trying to do?
   Make a dashboard for a namespaced resource: `Users::Preference`
* What did you end up with (logs, or, even better, example apps are great!)?
   `uninitialized constant User::PreferenceDashboard` 

* What versions are you running?
  - Rails 6.0.3
  - administrate 0.13.0

From reading the source of the `Administrate::ResourceResolver`, it's always going to singularize the namespace:
```
    def controller_path_parts
      controller_path.split("/")[1..-1].map(&:singularize)
    end
```

I don't understand why it would want to signularize the namespace part of the controller path. To my knowledge it is not possible to have `Foo::Bar` in `foos/bar.rb`. Would it not make more sense to just singularize the last part of the controller path?

Am I missing something, or is this a bug in administrate?

agnessa:
I am hit by the same issue in 0.15.0, it seems only the last part of the path should be singularized.

pablobm:
I agree. I can't think of a reason why namespaces should be singularized. Would someone be up for a PR?

DevS1993:
One way to monkey patch this for an immediate fix, would be to add an initializer (with any appropriately named file) and add this snippet :

```
Administrate::ResourceResolver.module_eval do
  def controller_path_parts
    path_parts = controller_path.split("/")[1..-1].map(&:singularize)
    # How you want to handle this.
  end
end
```

I didn't find any open PR's fixing this issue, @pablobm can i submit a PR?

pablobm:
@DevS1993 - Yes, I'll be happy to look at a PR for this.

rmachielse:
@DevS1993 did you end up submitting a PR? We are running into the same issue. If not, can I submit a PR instead @pablobm?

PR #2261, opened 2022-09-16
Avoid singularizing namespace in Administrate::ResourceResolver

rmachielse:
See #1931 for details.

### DIFF
--- a/lib/administrate/resource_resolver.rb
+++ b/lib/administrate/resource_resolver.rb
@@ -35,7 +35,8 @@ module Administrate
     end
 
     def controller_path_parts
-      controller_path.split("/")[1..-1].map(&:singularize)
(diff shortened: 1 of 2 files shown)


rmachielse:
@pablobm what do you think? Does this have a chance of getting merged?

pablobm:
@rmachielse - Would you be able to add an example to https://github.com/thoughtbot/administrate/blob/main/spec/lib/administrate/resource_resolver_spec.rb?

rmachielse:
@pablobm would it be okay if I adapt the existing examples to use a plural namespace, like `admin/libraries/books`?

pablobm:
@rmachielse - Wait, is there another way? :laughing: Go for it!

rmachielse:
@pablobm example added!

rmachielse:
Thank you!

Issue #1988, opened 2021-05-19
Filtering Field::HasMany


vijuarez:
* What would you like to be able to do? Can you provide some examples?

See, I've got a lot of users with different characteristics on my app. For example, they all have a `role`, which is an enum. Some models have references to users, but some of them have restrictions related to roles. For example, you can't select an user who's a `guess` to be the `editor` on a post. I've got all of that figured out via validations, but it would be nice to have it reflected on the interface. As if, when I start editing the `editors` attribute on a certain post, only users with the role `author` or `admin` appear as options.

* How could we go about implementing that?

Passing some kind of filter (maybe a function) to the field on `user_dashboard.rb` would be nice.

* Can you think of other approaches to the problem?

I could just filter it out during rendering (modifying the partial for `Field::HasMany`), but that's not nice to do if I want to do it on a lot of dashboards with many different conditions. A minimal example on how to do this would be extremely useful to me.

pablobm:
Yeah, I don't think there's a way to do this at the moment. Perhaps it would be a good idea to add a new option to the field that allows for this. It would probably have to modify `HasMany#candidate_resources` to use this new option. This is the concerned code:

https://github.com/thoughtbot/administrate/blob/b3bb0a3b7833deceff71cfe1687ac36cec12a6e8/lib/administrate/field/has_many.rb#L85-L92

vijuarez:
I prepared a very rudimentary implementation [here](https://github.com/vijuarez/administrate/tree/filter_has_many), but it's not that thought out. It would be probably better to pass scopes rather than plain `where` instructions, and I don't know how to filter with model methods. It's a start though.

https://github.com/vijuarez/administrate/blob/fd157d238f7f0049afac18edba1be22655c9cf3e/lib/administrate/field/has_many.rb#L85-L100

pablobm:
Yup, that's the sort of thing we would need. I think it could be done better if it accepted a `call`-able instead, like a lambda. This is what we already do for `Select` and `BelongsTo`. For example:

https://github.com/thoughtbot/administrate/blob/9071a7eff60cbbcaf0ddc1b7bfb6bfe7d22b0028/lib/administrate/field/select.rb#L16-L23

benoror:
Hi! Stumbled upon this issue while needing filtering on `HasMany` as well. 

[...]

PR #2266, opened 2022-09-21
HasMany: Allow :scope option

benoror:
Closes https://github.com/thoughtbot/administrate/issues/1988

### DIFF
--- a/lib/administrate/field/has_many.rb
+++ b/lib/administrate/field/has_many.rb
@@ -83,11 +83,13 @@ module Administrate
       end
 
       def candidate_resources
+        scope = options[:scope] ? options[:scope].call : associated_class
+        
         if options.key?(:includes)
           includes = options.fetch(:includes)
-          associated_class.includes(*includes).all
+          scope.includes(*includes).all
         else
-          associated_class.all
+          scope.all
         end
       end
 


pablobm:
Hello @benoror, thank you for this. Could you please clean the Hound warning and add an spec example, and I'll merge this?

For the spec, I suggest that you have a look at the one for `BelongsTo` for an idea of how to do it.

benoror:
@pablobm I ended using this extension, do you think it's worth it baking it into Administrate (maybe it is)? 
https://github.com/thoughtbot/administrate/issues/1988#issuecomment-1254477303

pablobm:
Initially, I prefer not baking the suggested plugin into Administrate. There are some improvements that can be made to fields, and adding blocks like this is one of them, but it needs thinking a bit more about what the arguments would be. For example, I would like the fields to know more about the current request (current user, etc) so that values can be customised per-request (and per-user, etc), so I would prefer not changing that just now.

pablobm:
If the plugin is enough for you, I'll close this. Happy to reopen and merge (after the Hound warning is resolved) if you still want to do it though.

benoror:
Hey @pablobm, the plugin has worked for me so far, all good closing this for now. Thanks!

PR #2269, opened 2022-09-26
Add SimpleCov support to the test suite

etagwerker:
Hi there,

This PR will add a new dependency to the test group and require SimpleCov if COVERAGE=true when running the test suite.

This will generate a code coverage report that will look like this: 

Screen Shot 2022-09-26 at 4 42 40 PM

Screen Shot 2022-09-26 at 4 42 47 PM

Please let me know what you think.

Thanks!


spec/spec_helper.rb
```diff
@@ -1,3 +1,12 @@
+if ENV["COVERAGE"] == "true"
+  puts "Calculating Code Coverage with SimpleCov"
+
+  require "simplecov"
+  SimpleCov.start "rails"
+
+  puts "Required SimpleCov"
```
pablobm, line 7: I think we can remove this second printout? :slightly_smiling_face:

etagwerker:
@pablobm Thanks for your review! I just addressed your comment. 😄

nickcharlton:
I'm not convinced that this is all that helpful for us. Do others feel a strong need for code coverage reports?

pablobm:
@nickcharlton - I personally like the idea, but feel that it risks becoming a "hidden" feature unless made more apparent somehow.

etagwerker:
@nickcharlton @pablobm Having support for `simplecov` means that you can quickly see which files are lacking coverage: 

According to the report I ran for this branch it looks like it is almost perfect: 

Screen Shot 2022-10-24 at 10 33 13 PM

If you wanted to, it would make it easier to enable something like `codecov` to quickly assess whether a PR is increasing/decreasing code coverage by the library's test suite. So it might save you time assessing whether a change/patch has enough test cases or not.

pablobm:
I think that adding some sort of warning in the GitHub checks (can you add warnings that don't block the merge?) as well as a coverage badge on the README can help raising awareness that we do want contributions to be well tested. Or perhaps it's a distraction? Thoughts, @nickcharlton?

etagwerker:
@nickcharlton A quick way to see SimpleCov's output in the console would be to combine these two things: 

1. Add `simplecov-console` to this PR -> https://github.com/chetan/simplecov-console
2. Add an environment variable in here: https://github.com/thoughtbot/administrate/blob/main/.circleci/config.yml (ie. `COVERAGE=true`)

Then every PR would output the code coverage percentage at the end of its execution. 

If you wanted to go the CodeCov route, you'd need to set up that service and then configure Circle CI to have the API key as a secret.

The comments would look like this: https://github.com/fastruby/skunk/pull/109#issuecomment-1464789962

PR #2272, opened 2022-09-29
update redirect path after delete resource

jubilee2:
keep UI in the same page when delete has_many records in show page

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -74,7 +74,7 @@ module Administrate
       else
         flash[:error] = requested_resource.errors.full_messages.join("<br/>")
       end
-      redirect_to after_resource_destroyed_path(requested_resource)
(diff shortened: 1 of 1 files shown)


jubilee2:
only one thing I am not sure, the "redirect_back" method add to Rails start from Rails 5

jubilee2:
we should guide people how to redirect back in after_resource_destroyed_path

https://github.com/rails/rails/blob/6-1-stable/actionpack/lib/action_controller/metal/redirecting.rb#L90

pablobm:
Why did you close the PR? Regarding compatibility, Rails 6 is the oldest currently supported.

jubilee2:
in the controller we have `after_resource_destroyed_path` to config the redirect path, in this PR I override the redirection order, make the `after_resource_destroyed_path` loss the meaning.

pablobm:
Oh, I see. Makes sense :+1: 

Do you have any alternatives? Or perhaps the redirection path hooks should work differently?

PR #2278, opened 2022-10-11
belongs_to resource can become a superclass first

drnic:
BelongsTo partial now allows a resource to become something else via `Admin::ApplicationController#becomes`

### DIFF
--- a/app/views/fields/belongs_to/_show.html.erb
+++ b/app/views/fields/belongs_to/_show.html.erb
@@ -19,7 +19,7 @@ By default, the relationship is rendered as a link to the associated object.
   <% if accessible_action?(field.data, :show) %>
     <%= link_to(
       field.display_associated_resource,
-      [namespace, field.data],
+      [namespace, becomes(field.data)],
     ) %>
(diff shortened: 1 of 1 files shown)


pablobm (changes requested):
Hello @drnic, thank you for your contribution to Administrate :slightly_smiling_face: 

Would you be able to explain what use cases you are targeting here?

Also, `becomes` is part of ActiveRecord, not ActionPack, so I think you mean `ActiveRecord::Persistence#becomes`, right? In any case, the specs are failing because `because` is not defined. Perhaps you defined it in your own `Admin::ApplicationController`?

(Re: specs failing, for some reason they didn't run here, but I ran them on my machine).

pablobm:
Thank you for the explanation. I get the impression that this would require changes in quite a few places, including a way to indicate what to "become", am I right? Would you be able to share how your `becomes` is implemented?

drnic:
I think it comes from Rails core

https://apidock.com/rails/ActiveRecord/Persistence/becomes

Issue #2014, opened 2021-06-30
Please don't titleize results of I18n calls


mvz:
* What were you trying to do?

Trying to get the email address column to be labeled "E-mailadres" (in Dutch)

* What did you end up with (logs, or, even better, example apps are great!)?

The title becomes "E Mailadres" because titleize destroys dashes and applies English capitalization rules.

* What versions are you running?
  - Rails: 6.1.4
  - administrate: 0.16.0

pablobm:
Fixed by #2280

PR #2280, opened 2022-10-24
Titleize the column not the user defined locale

jsntv200:
Solves #2014

Ran into this problem when I wanted to render a column with ID in the name. Assumed defining the label in a locale would resolve the issue but no joy.

Overriding a template to display a user defined locale seemed like overkill when it appears that `titleize` is incorrectly applied for the collection partial. Existing usage in other partials/templates is to `titleize` the column rather than the result of translate.

https://github.com/thoughtbot/administrate/blob/main/app/views/administrate/application/new.html.erb#L19
https://github.com/thoughtbot/administrate/blob/main/app/views/fields/has_one/_form.html.erb#L21
https://github.com/thoughtbot/administrate/blob/main/app/views/fields/has_one/_show.html.erb#L32

### DIFF
--- a/app/views/administrate/application/_collection.html.erb
+++ b/app/views/administrate/application/_collection.html.erb
@@ -33,8 +33,8 @@ to display a collection of resources in an HTML table.
         )) do %>
         <%= t(
           "helpers.label.#{collection_presenter.resource_name}.#{attr_name}",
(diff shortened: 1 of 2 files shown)


pablobm:
I agree that the `titleize` is incorrectly applied. Thank you for your contribution!

PR #2299, opened 2022-12-11
Update minimum supported Rails version on docs/getting_started.md

holamendi:
The guides still says that Rails 5.0 is supported, which is no longer the case. Rails 6.0 is the minimum required version.


pablobm (approved):
Good catch! Would you mind rebasing, now that the build has been fixed?

PR #2304, opened 2022-12-20
Set empty string param values to nil

mnort9:
When creating/updating a new resource, blank fields save as empty strings. This PR sets blank fields to nil instead. In most cases, nil is the desired value for the db over an empty string, so nil is the more sensible default. 

Ref #441

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -214,6 +214,8 @@ module Administrate
         end
       elsif data.is_a?(ActionController::Parameters)
         data.transform_values { |v| read_param_value(v) }
+      elsif data.is_a?(String) && data.blank?
+        nil
       else
         data
       end
--- a/spec/controllers/admin/customers_controller_spec.rb
+++ b/spec/controllers/admin/customers_controller_spec.rb
@@ -132,6 +132,19 @@ describe Admin::CustomersController, type: :controller do
         expect(page.resource).to be_a_new(Customer)
       end
     end
+
+    describe "with empty string param" do
+      it "sets empty string value to nil" do
+        empty_string_attributes = { country_code: "" }
+
+        locals = capture_view_locals do
+          post :create, params: { customer: empty_string_attributes }
+        end
+
+        page = locals[:page]
+        expect(page.resource.country_code).to be_nil
+      end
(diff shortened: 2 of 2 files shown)


pablobm:
I'm generally in favour, although this specific solution makes it difficult to override. For example, if somebody tries to override by overriding `read_param_value`, they must copy the whole thing instead of just providing for the string case and calling `super` for the rest.

Perhaps we can factor this case out into a `read_string_param` so that this is easy to override? What do you think?

mnort9:
@pablobm To override, wouldn't you be able to do:

```
def read_param_value(data)
  return '' if data.is_a?(String) && data.blank?
  super
end
```

Alternatively are you thinking of handling the case in `resource_params`? I'm thinking that may be more complex to override:

```
def resource_params
 params.require(resource_class.model_name.param_key).
   permit(dashboard.permitted_attributes).
   transform_values { |v| data.is_a?(String) && data.blank? ? '' : read_param_value(v) }
end

pablobm:
Er... 😓 Let's pretend I didn't say anything 😜

jubilee2:
@pablobm this is break update can we add override example into document?

pablobm:
@jubilee2 - Do you mean to provide an option? Would you be able to provide a PR with an idea of what you have in mind?

jubilee2:
after review the MR i think i was wrong. this ticket is fine for me

jonny5:
This was a breaking change for integration with [Shrine](https://shrinerb.com/) for file uploads. Shrine treats empty string differently than `nil` (ignores "" and deletes content if `nil`)

I used the override suggested, posting here in case anyone runs into this problem in the future.

pablobm:
I have come across another example where this change broke something. This webapp had one very customised dashboard (think something based on https://github.com/thoughtbot/administrate/blob/main/lib/administrate/custom_dashboard.rb). Here they had a date field, but it wasn't the usual `Field::Date`. Instead it was a bog-standard Rails `field.date_select`. Rails does break if it receives nil values on this one (works with empty strings).

Note that this was fixed by simply changing the field from `field.date_select` to `field.date_field` (which happens to be Administrate's implementation). So no big deal, but something to add to the list in case we need to rethink it.

PR #2308, opened 2023-01-03
Make overriding create resource easier

cyu:
Use `new_resource` in the `create` action.  This will make it easier for subclasses to further initialize the model before creating the model.


pablobm:
Thank you! :slightly_smiling_face:

PR #2324, opened 2023-01-27
Update to Rails 7

pablobm:
This description has been updated since the PR was first created. The first few comments in the thread reflect the conversation while trying to find a fix to the issue of lingering constants.

---

Administrate was already compatible with Rails 7, but its specs weren't. This PR includes all necessary fixes, including a workaround for a strange Ruby behaviour when undefining constants on Ruby 3.x.

Specifically, `spec/generators/dashboard_generator_spec.rb` creates a number of classes and then attempts to undefine them, but these appear to linger around, affecting results at `spec/generators/routes_generator_spec.rb`. From what I can tell, this affects the specs, but not real world usage.

After trying a few things, I ended up with a bit of a hack: in `spec/generators/dashboard_generator_spec.rb` I define those classes to be children of `TestRecord`. Then the routes generator explicitly filter those out.


pablobm:
@nickcharlton - Not sure what to do about this. I'm considering skipping the offending example under Ruby 3.x. Thoughts?

nickcharlton:
Hmm, interesting. On `spec/generators/routes_generator_spec.rb:123`, it does pass locally with Ruby 3.2. It looks like the failure is because the routes file contains `foos#index` but not `customers#index`, have we made any changes that might make that to be the case?

pablobm:
The specs do pass individually. The problem is running them all together, particularly when `spec/generators/dashboard_generator_spec.rb` runs before `spec/generators/routes_generator_spec.rb`.

1. The first one generates temporary models and dashboards, like `Foo` and `FooDashboard` at https://github.com/thoughtbot/administrate/blob/main/spec/generators/dashboard_generator_spec.rb#L28
2. These temporary classes are theoretically cleaned up in `ensure` blocks at the end of each example: https://github.com/thoughtbot/administrate/blob/main/spec/generators/dashboard_generator_spec.rb#L60
3. However they are still around when the spec for the routes generator runs, and they are returned by `ActiveRecord::Base.descendants` at https://github.com/thoughtbot/administrate/blob/main/lib/generators/administrate/routes/routes_generator.rb#L74.
4. These "zombie" models make the route generator think that those are models that need routes, creating them when they shouldn't exist.

pablobm:
No idea at the moment. I tried running a bisect through Rails's commit history, but there is a large section around 7.0.0.alpha (or something like that) where specs fail for different reasons (bug in Rails at the time, from what I gather). I may give it another whirl this week, but I'm going to timebox it as I don't want to spend too much longer on this.

pablobm:
I had another look. The other error that I mention appears for the first time with this Rails commit: https://github.com/rails/rails/commit/cb82f5f0a438c7be64ca1c0633556f6a535138b6 This is too esoteric for me, but it does sound related to our issue.

I've been playing with it a bit at https://github.com/thoughtbot/administrate/compare/main...pablobm:administrate:rails70-bisect. I may give it another go, but this is well beyond my expertise.

pablobm:
I appear to have found a fix for the issue of lingering constants (see PR description), so I have marked the PR as ready to review.

littleforest (approved):
🎉

PR #2331, opened 2023-02-20
Upgrade Ruby to 3.1.0

Joranhezon:
The previous ruby version, 2.7.5, is being deprecated soon. This upgrades Ruby to a version that is stable and is receiving support.


MatheusRich:
Are we dropping support for 2.7.x? If so, we should update gemspec too, and cut a new release.

littleforest:
closing due to #2367

PR #2347, opened 2023-04-05
Set _pundit_policy_scoped instance variable

nhippenmeyer:
Pundit exposes the [policy_scope](https://github.com/varvet/pundit#scopes) method which [internally](https://github.com/varvet/pundit/blob/6513071aa790f061dec72e676b99996fa007c9ff/lib/pundit/authorization.rb#L92-L95) sets the `@_pundit_policy_scoped` instance variable.

When using the `verify_policy_scoped` callback for [ensuring scopes are used](https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used), pundit checks whether this instance variable has been set in the context of a request. If it hasn't it raises the `Pundit::PolicyScopingNotPerformedError` error. Since administrate defines its own `policy_scope_admin` method, we also need to set this instance variable there to avoid the error being raised.

### DIFF
--- a/app/controllers/concerns/administrate/punditize.rb
+++ b/app/controllers/concerns/administrate/punditize.rb
@@ -24,6 +24,7 @@ module Administrate
       # to be overridden by 'resolve_admin' for a different index scope in Admin
       # controllers.
       def policy_scope_admin(scope)
+        @_pundit_policy_scoped = true
         ps = Pundit::PolicyFinder.new(scope).scope!.new(pundit_user, scope)
         if ps.respond_to? :resolve_admin
           ps.resolve_admin


pablobm:
Thank you for this @nhippenmeyer! Eventually I think we should deprecate `policy_scope_admin`, but until then this is a bug that should be addressed.

Would you be able to make the following changes, please?

First, I think this could use specs, or something that at least tests this with the current code. I gave it a try and I'm thinking perhaps  we can simply add the following. What do you think?

```diff
diff --git a/spec/example_app/app/controllers/admin/application_controller.rb b/spec/example_app/app/controllers/admin/application_controller.rb
index 1cb06541..65069e4d 100644
--- a/spec/example_app/app/controllers/admin/application_controller.rb
+++ b/spec/example_app/app/controllers/admin/application_controller.rb
@@ -23,6 +23,7 @@ module Admin
     end
 
     before_action :authenticate_admin
+    after_action :verify_policy_scoped, only: :index
 
     def authenticate_admin
       user_id = session[:user_id]
diff --git a/spec/example_app/app/controllers/admin/customers_controller.rb b/spec/example_app/app/controllers/admin/customers_controller.rb
index 4e06f6e4..c02e9ee0 100644
--- a/spec/example_app/app/controllers/admin/customers_controller.rb
+++ b/spec/example_app/app/controllers/admin/customers_controller.rb
@@ -13,7 +13,7 @@ module Admin
     private
 
     def scoped_resource
-      Customer.where(hidden: false)
+      policy_scope_admin Customer.where(hidden: false)
     end
   end
 end
```

Second, could you please rebase this on top of `main`? Also hopefully the specs will run in CI this time. I don't know what happened there.

Thank you again!

nickcharlton:
@nhippenmeyer, would you be able to take a look at this PR?

nhippenmeyer:
Apologies I haven't had time to respond to the feedback on this PR. I'm going to close it for now and can reopen it if I or someone else has the bandwidth to make the proposed changes.

Issue #1939, opened 2021-03-09
Enums  - Select tag dont keep saved option selected on edit page


DavidGeismarLtd:
* What were you trying to do?
create a select form an enum field.
In my model Program : 
`enum publication_status: { 'publié' => 'published', 'a venir' => 'upcoming', 'brouillon' => 'draft'  }, _prefix: :publication_status
`
Keys are the "human" readable values, values are the value to be stored in db 

In my administrate dashboard :

```
      ATTRIBUTE_TYPES = {
        publication_status: Field::Select.with_options(collection: Program.publication_statuses)
    }

```
This generates the following html : 

```
<select name="program[publication_status]" id="program_publication_status"><option value="published">publié</option>
<option value="upcoming">a venir</option>
<option value="draft">brouillon</option></select>
```

However when I then update the resource and go back on the edit page the saved value is not the option selected, instead it is always the first one.

Am I not following the right approach for using enums with administrate or is this a bug ?

* What did you end up with (logs, or, even better, example apps are great!)?
* 
* What versions are you running?
  - Rails 6.0.3
  - administrate 0.15.0

pablobm:
Hm, this might be a bug, but I'm not sure. I wonder if it's something that we specifically do not support, for some reason.

I noticed that there's a plugin to support enums: https://github.com/Valiot/administrate-field-enum, however I haven't tried it, so I don't know if it would solve your problem.

Perhaps this should be supported by Administrate? I'm not sure. It might be tricky for some reason. If you want to give it a try and provide a PR, we could have a look at it.

AFlowOfCode:
There's nothing special about an enum from the perspective of a select element. The enum values are simply the options available to choose from. 

The thing is enums are usually stored in the database as integers, so @DavidGeismarLtd's enum definition appears to be atypical and either incorrect or possibly using a gem that allows string-based enums. I'd expect to see this instead: 

`{ 'publié': 0, 'a venir': 1, 'brouillon': 2 }`

I didn't try to replicate this, so just taking a guess based on normal enum usage. If a value is actually getting saved (which can't be seen here so is not a certainty), it's probably being saved as 0 1 or 2 in the database. Any saved data associated with this field would not match the values being passed in, hence nothing will ever show up preselected.

pablobm:
I was able to reproduce this today. The issue stems from how ActiveRecord works with enums, which can be a bit too clever depending on your taste.

[...]

PR #2348, opened 2023-04-06
Field::Select to handle ActiveRecord enums correctly

pablobm:
Fixes https://github.com/thoughtbot/administrate/issues/1939

In the spirit of "make the change easy, then make the easy change", this PR includes a refactor of `Field::Select`, moving the logic out of the template and into the field class, where the actual functionality (and any future one) can be handled and tested more easily.

I have also written more exhaustive specs for the behaviour of this field, which helped me not breaking existing functionality while refactoring (or so I hope!).

### DIFF
--- a/app/views/fields/select/_form.html.erb
+++ b/app/views/fields/select/_form.html.erb
@@ -19,27 +19,14 @@ to be displayed on a resource's edit form page.
   <%= f.label field.attribute %>
 </div>
 <div class="field-unit__field">
-  <% if field.selectable_options.first&.is_a?(Array) %>
-    <%= f.select(
+  <%=
+    f.select(
       field.attribute,
-      options_from_collection_for_select(
+      options_for_select(
         field.selectable_options,
-        :last,
-        :first,
         field.data,
       ),
       include_blank: field.include_blank_option
-    ) %>
-  <% else %>
-    <%= f.select(
-      field.attribute,
-      options_from_collection_for_select(
-        field.selectable_options,
-        :to_s,
-        :to_s,
-        field.data,
-      ),
-      include_blank: field.include_blank_option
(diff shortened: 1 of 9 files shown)


littleforest (commented):
@pablobm these changes look good, but I'm not seeing that this PR solves the behavior noted in the ticket, mainly: 

> when I then update the resource and go back on the edit page the saved value is not the option selected

I tested in the UI with the Customer.kind. If I choose "vip" in the form and save, then "vip" is correctly getting saved as the customer kind, but when I go back to edit the record again, it shows "standard" as selected.

pablobm:
@littleforest - Oopsie, you are right. I had forgotten to remove the `collection` option from the Customers dashboard. I have just fixed it and added a feature spec. It should work now :crossed_fingers:

pablobm:
Hm. Something else has broken now. I see what it is, it's the product of a hack in a spec. I'm going to have lunch and hopefully solve it when I'm feeling less annoyed...

pablobm:
Fixed now! The required changes were:

- `spec/example_app/app/dashboards/customer_dashboard.rb` - Change the field declaration to not provide a `collection` and instead rely on the existing enum.
- `spec/features/form_spec.rb` - The existing examples relied on the field receiving options (via `#with_options`). This would turn the field into a `Field::Deferred` whose options can be modified. This is a bit of a hack... and so is my alternative, although I want to think that mine is less bad because it doesn't rely on private interfaces. Yeah, that's my excuse :stuck_out_tongue_closed_eyes: 

Ideally, fields should be proper object instead of being passed around as classes that are instantiated at some point. But that would require a larger refactor, so not for now.

littleforest (approved):
Looks great!

Issue #2066, opened 2021-11-23
Defining different attributes for `new`/`create` or `edit`/`update` actions without defining `FORM_ATTRIBUTES` makes it crash


pjmartorell:
* What were you trying to do?
Upgraded from `0.13` to `main` (9e462f7179f47151d1b91aab59291e6884870dab). I tried to define different attributes for new/create or edit/update actions as stated here: https://github.com/thoughtbot/administrate/blob/main/docs/customizing_dashboards.md#form-attributes

  When trying to `update` or `create` a record in the dashboard, there is a crash. This is because, when validating permitted attributes, it's trying to get these attributes from `self.class::FORM_ATTRIBUTES` but it should get them from `"FORM_ATTRIBUTES_#{action.upcase}"` as `FORM_ATTRIBUTES` is not defined anymore. `specific_form_attributes_for(action)` is returning `nil` because `action` is not being passed in `permitted_attributes`. So the solution is to pass the `action_name` when checking permitted attributes.

* What did you end up with (logs, or, even better, example apps are great!)?
The affected code is related with https://github.com/thoughtbot/administrate/pull/1991
**Logs:**
  <img width="1065" alt="Screenshot_2021-11-23_at_12_43_43" src="https://user-images.githubusercontent.com/345520/143021057-fbc9225d-3421-4b75-89e8-92a78b996401.png">

  **My workaround:**
  In my `UserDashboard` added the following overrides:

  ```ruby
  def permitted_attributes(action)
    action =
      case action
      when "update" then "edit"
      when "create" then "new"
      else action
      end

    form_attributes(action).map do |attr|
      attribute_types[attr].permitted_attribute(
        attr,
        resource_class: self.class.model
      )
    end.uniq
  end

  def form_attributes(action = nil)
    specific_form_attributes_for(action) || self.class::FORM_ATTRIBUTES
  end

  def specific_form_attributes_for(action)
    return unless action

    cname = "FORM_ATTRIBUTES_#{action.upcase}"

    self.class.const_get(cname) if self.class.const_defined?(cname)
  end
  ```
  In my `Admin::UsersController` add the following override:

  ```ruby
  def resource_params
    params.require(resource_class.model_name.param_key)
          .permit(dashboard.permitted_attributes(action_name))
          .transform_values { |v| read_param_value(v) }
  end
  ```

* What versions are you running?
  - Rails: 6.1.3.1
  - administrate (9e462f7179f47151d1b91aab59291e6884870dab)

[...]

PR #2356, opened 2023-04-17
make permitted_attributes support action

### LINKED ISSUE #2352: the permitted_attributes may need pass current action
https://github.com/thoughtbot/administrate/blob/062ea9f517b140ea04e2c6c4ff3a00a563bafcac/lib/generators/administrate/dashboard/templates/controller.rb.erb#L39

the permitted_attributes may need pass current action to support "FORM_ATTRIBUTES_NEW" and "FORM_ATTRIBUTES_EDIT". 

otherwise only go through "FORM_ATTRIBUTES" whatever edit or new action.

jubilee2:
fix #2352 and #2066

### DIFF
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -62,11 +62,18 @@ module Administrate
       self.class.const_get(cname) if self.class.const_defined?(cname)
     end
 
-    def permitted_attributes
-      form_attributes.map do |attr|
+    def permitted_attributes(action = nil)
+      action =
+        case action
+        when "update" then "edit"
+        when "create" then "new"
+        else action
+        end
+      form_attributes(action).map do |attr|
         attribute_types[attr].permitted_attribute(
           attr,
           resource_class: self.class.model,
+          action: action,
         )
       end.uniq
     end
--- a/lib/administrate/base_dashboard.rb
+++ b/lib/administrate/base_dashboard.rb
@@ -51,6 +51,12 @@ module Administrate
     end
 
     def form_attributes(action = nil)
+      action =
+        case action
+        when "update" then "edit"
+        when "create" then "new"
+        else action
+        end
       specific_form_attributes_for(action) || self.class::FORM_ATTRIBUTES
     end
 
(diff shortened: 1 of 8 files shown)


lib/administrate/base_dashboard.rb
```diff
@@ -62,11 +62,18 @@ def specific_form_attributes_for(action)
...
 
-    def permitted_attributes
-      form_attributes.map do |attr|
+    def permitted_attributes(action = nil)
+      action =
+        case action
+        when "update" then "edit"
+        when "create" then "new"
+        else action
+        end
```
pablobm, line 71: This code also exists in `lib/administrate/page/form.rb` where it's also used ahead of calling `form_attributes`. What do you think of removing it from both places and putting it at the start of `form_attributes` instead?

pablobm (commented):
Looks good, thank you! I have left one comment and I think this should be good to merge soon :+1:

lib/administrate/base_dashboard.rb
jubilee2: done

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -66,6 +66,11 @@ def default_sorting_direction
         expect(locals[:resources].map(&:name)).to eq sorted_customer_names
       end
     end
+
+    it "assigns the action name as a string" do
+      get :index
+      expect(controller.action_name).to be_a(String)
+    end
```
pablobm, line 73: Can you also remove this, please? This is standard Rails behaviour and unrelated to this feature, so it doesn't need to be tested. Sorry I missed this yesterday.

pablobm (commented):
Just one tiny thing left!

spec/controllers/admin/customers_controller_spec.rb
jubilee2: I want make sure action name is string not symbo, anyway I can remove it
```
      action =
        case action
        when "update" then "edit"
        when "create" then "new"
        else action
        end
```

pablobm (approved):

pablobm:
Thank you again!

PR #2367, opened 2023-04-21
Update to Ruby 3.2.2

sallyhall:
In order to deploy the demo app to a [heroku stack ](https://devcenter.heroku.com/articles/stack)that is still supported, we need to update ruby.

Looking at the [build logs](https://dashboard.heroku.com/apps/administrate-demo/activity/builds/3d80a706-5181-47ac-a537-e96778d646d4), it seems that although the ruby version on main is 2.7.5, heroku is deploying the app with ruby 2.6.4. [Heroku depends on the Gemfile to set a ruby version ](https://devcenter.heroku.com/articles/ruby-versions), so in order to update ruby for the demo app, we need to set it in the Gemfile.

However, setting the ruby version in the Gemfile breaks the tests that we run against several supported versions of ruby.

In order to still test the app against multiple ruby versions, but specify the version we want to use for the demo app, this uses the [CircleCI built in environment variable](https://circleci.com/docs/variables/#built-in-environment-variables) `CI`.


pablobm (approved):
Sounds good to me! :+1:

nickcharlton (approved):

Issue #2370, opened 2023-04-25
Sharing template of ""collection" between index page and has_many show  page


jubilee2:
https://administrate-demo.herokuapp.com/admin

jubilee2:
it's works in private mode

jubilee2:
@pablobm my session maybe broken
what I did is click the became button in customer show pages.
i believed that buttons should not be here
<img width="1401" alt="Screen Shot 2023-04-25 at 7 48 12 AM" src="https://user-images.githubusercontent.com/5558405/234281890-0edfe3e7-e62e-4329-bac8-9b5b352e428e.png">

pablobm:
You are right, something's off. Thank you for reporting.

jubilee2:
I think the problem is that the template uses the same resource as the parent for the has_many field. 
[    "collection",](https://github.com/thoughtbot/administrate/blob/b0fad736134bb1e47e5f5eb17f3a73640df010bf/app/views/fields/has_many/_show.html.erb#L25)

pablobm:
I have been looking into this. It's a tricky one, not just in terms of code but more importantly in terms of functionality. We need to put on our Product Manager hats and think of what we want to offer here.

We have a resource `Customer` for which we want to render a custom collection template. The bug here is that other collections for other resources in the `Customer` show page (present as `HasMany` fields) also use the custom collection template intended only for `Customer`.

From a functionality point of view, the "obvious" fix here is to ensure that each resource type should use its own collection type. Cool. But we have to think about the different variations of this pattern and make sure that our solution caters for all of them.

These are template variations that I foresee users wanting to customise:

- Resource collection on the resource's own index page (the basic one).
- Any "has many" collection on any resource's show page (a "has many" template different from the one used in index pages).
- The "has many" collection of a specific resource on any resource's show page (the "has_many" template of `Customer` in any show page)
- Any "has many" collection on the show page of a specific resource (all the "has_many" templates in a `Customer` show page)
- A specific "has many" collection on the show page of a specific resource (the "has many" template for `Order` when it appears in the `Customer` show page).
- Generalisation to other field types, possibly with some sort of inheritance mechanism. Eg: has_many_variation or nested_has_many to try a template of its own field type, then fall back on the has_many template, all while respecting the above variations!

## High-level view of the code
I found it useful at this point to see where Rails expects to find partial templates. For example, if instead of `collection_header_actions` I change the template to render `foo`, I get this error:
```
# ActionView::MissingTemplate:
[...]

PR #2374, opened 2023-04-28
Update has_many show pages

### LINKED ISSUE #2372: Authorization errors should not return 500
At the moment, when `Pundit::NotAuthorizedError` or `Administrate::NotAuthorizedError` is raised within Administrate, the app returns an HTTP 500 status.

To reproduce:
1. Visit the example app: https://administrate-demo.herokuapp.com/admin/
2. On the Customers dashboard, click to "Become" any customer.
3. Visit the payments dashboard at https://administrate-demo.herokuapp.com/admin/payments (it will have disappeared from the navigation as customers are not authorized).

Expected result: an appropriate HTTP 4xx status, probably 403, perhaps with an error message to match.
Actual result: HTTP 500 status and "We're sorry, but something went wrong" message.

Seems like these exceptions should be rescued by `Admin::ApplicationController`. More questions:

- Should we also show an error message? What should it look like? Rails doesn't provide (that I know) error pages for errors other than 404, 422 and 500, so we would have to provide something ourselves.
- If a user integrates their own authorization mechanism, the raised exceptions will be different (eg: they integrate with cancancan and raise [`CanCan::Error`](https://github.com/CanCanCommunity/cancancan/blob/develop/lib/cancan/exceptions.rb)). Should we provide a way to tell Administrate what these exceptions might be, so that they are rescued the same way?

jubilee2:
try to fix #2372, but not sure is it a perfect solve

### DIFF
--- a/app/views/fields/has_many/_show.html.erb
+++ b/app/views/fields/has_many/_show.html.erb
@@ -22,7 +22,7 @@ from the associated resource class's dashboard.
   <% order = field.order_from_params(params.fetch(field.name, {})) %>
   <% page_number = params.fetch(field.name, {}).fetch(:page, nil) %>
   <%= render(
-    "collection",
+    "administrate/application/collection",
     collection_presenter: field.associated_collection(order),
     collection_field_name: field.name,
     page: page,


nickcharlton:
Are you able to add a test for this?

I feel like even if it's not a perfect solution, but we've got a test around the regression then it'll do for now and we can revisit it in future if necessary!

jubilee2:
> Are you able to add a test for this?
> 
> I feel like even if it's not a perfect solution, but we've got a test around the regression then it'll do for now and we can revisit it in future if necessary!

after test this is not a soltion.

jubilee2:
After adding rspec, I realized that the original solution did not work. I started to think about a different way to fix the issue. We might need another namespace for the collection template of has_many fields.

pablobm:
Thank you @jubilee2! I have been looking into this and now I think that is a rather complex problem :exploding_head: I have been looking into it and published a more detailed study in the issue page (see https://github.com/thoughtbot/administrate/issues/2370#issuecomment-1635644394). I think the PR should close for now and discussions should happen in the issue page, until we have a better idea of how we want to tackle this.

jubilee2:
Thank you.

PR #2325, opened 2023-01-29
Enable ordering by HasOne fields

elrosa:
Currently, users are able to sort by BelongsTo fields (using ID or selected attribute) and HasMany fields.

This PR adds support for sorting by HasOne fields using both ID or selected attribute (passed via `Field::HasOne.with_options(order: attribute_name)`.

PR #2375, opened 2023-04-28
Update Field::HasOne options documentation

littleforest:
This updates the documentation for customizing the `Field::HasOne` with the `order` option, which was added in #2325. This also changes each of the field descriptions to an h4 tag for better accessibility and to provide links to those headers.


nickcharlton (approved):
Thanks!

Can we split out the heading change into another PR? I wonder if it's worth seeing if other pages have the same problem when doing that too. What do you think?

littleforest:
> Can we split out the heading change into another PR?

Sure, will do!

PR #2377, opened 2023-04-29
Add a script to generate a draft CHANGELOG update

nickcharlton:
The majority of the effort in making a new release is in putting together a good CHANGELOG. Back in #1968, we added a script to generate a list of templates that had changed since the last (provided) tag. This helped ensure we'd communicate any template changes, whilst requiring little effort to do so.

The next step in assembling a CHANGELOG was to put together a list of commits the release would include. By convention this has been:

    [KEY] [PR NUMBER] Commit Message

So far, this has required a lot of manual text manipulation, and when the commit wasn't introduced via a squash on GitHub, the PR reference needed to be tracked down (which could take some time).

This script attempts to automate the rest of this process by assembling a draft that needs much less effort to publish. By using the GitHub CLI, we're able to match commits back to the originating pull request and then automate much of the text manipulation which was needed before.

We then pull over the template warning checker from before, but in this implementation skip over the `spec` changes, as they shouldn't matter to end users.

We can also assume we want changes since the last tag, to remove the need to provide an argument.

An example run (trimmed):

    The following templates have changed since v0.18.0:

      app/views/administrate/application/_collection.html.erb
      app/views/administrate/application/_index_header.html.erb

    If your application overrides any of them, make sure to review your
    custom templates to ensure that they remain compatible.

    * [] [#2367] Update to Ruby 3.2.2
    * [] [#2371] Adapt to deprecations in the Faker API
    * [] [#2348] Field::Select to handle ActiveRecord enums correctly

https://cli.github.com/
https://git-scm.com/docs/pretty-formats#Documentation/pretty-formats.txt-trailersoptions https://stackoverflow.com/a/18558871
https://stackoverflow.com/a/30035045


bin/build-changelog
```diff
@@ -0,0 +1,38 @@
...
+last_release=$(git describe --tag --abbrev=0)
+
+template_changes=$(git diff --name-only "${last_release}" | grep html.erb)
+
+echo "The following templates have changed since ${last_release}:"
+echo
+echo "$template_changes" | while read -r line; do
+	if case $line in spec*) false;; esac; then
+		echo "  ${line}"
+fi
```
pablobm, line 19: Never sure with shellscript but, is this possible?

```suggestion
        fi
```

pablobm (approved):
Amazing!

bin/build-changelog
nickcharlton: Ooh yeah, you actually can!

nickcharlton:
It's been bugging me that I can't seem to work out how to split this line out:

```sh
commits_since=$(git log --author="^(?!dependabot).*$" --perl-regexp "${revision_range}" "${commit_format}")
```

Everything I've tried breaks the regex, but it's taken me too long to come up with anything, so I'm giving in and accepting it how it is. If anyone knows how to do such a thing in a POSIX shell compliant way, I'd love to know.

PR #2388, opened 2023-05-27
Add dartsass-sprockets shim to solve sassc dependency

abuisman:
There are quite some PRs and issues about problems with sassc.

For me it meant my docker builds took longer than I usable. 

I've seen a few PRs and forks that required some pretty big changes. The solution I found is just a 2-liner:

- I removed `sassc-rails` in favour of [`dartsass-sprockets`](https://github.com/tablecheck/dartsass-sprockets):

> This gem is a fork of [sass/sassc-rails](https://github.com/sass/sassc-rails) which maintains API compatibility but delegates to the [dartsass-ruby](https://github.com/tablecheck/sass-embedded-host-ruby) gem which uses Dart Sass instead of the libsass C implmentation.

- I require it in `lib/administrate/engine.rb` instead of `sassc-rails`

I believe this will at the very least be a good transition for many people, including myself.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -5,9 +5,9 @@ PATH
       actionpack (>= 5.0)
       actionview (>= 5.0)
       activerecord (>= 5.0)
+      dartsass-sprockets (~> 3.0)
       jquery-rails (>= 4.0)
       kaminari (>= 1.0)
-      sassc-rails (~> 2.1)
       selectize-rails (~> 0.6)
 
 GEM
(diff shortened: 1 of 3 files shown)


jubilee2:
this PR works for me on rails 7! Thank you

abuisman:
> this PR works for me to drop sassc on rails 7! Thank you

Glad I could help! Thanks for confirming it works.

jubilee2:
you can over rid the layout to exclude asset if you want

brunoprietog:
Why not dartsass-rails? This will doesn't work with Propshaft, the future asset pipeline for Rails in next versions

abuisman:
> Why not dartsass-rails? This will doesn't work with Propshaft, the future asset pipeline for Rails in next versions

I couldn’t get it to work. Still I think that this pr could be a step in between. Next up could be dartsass-rails or something else.

nickcharlton:
Thank you for contributing this and your patience whilst we figured out how to handle assets for now and in the future. I'm sure it's been very helpful for folks in getting their applications running.

I took the decision to go with #2397, which will mean in our next release we'll be bundling the assets required for Administrate in the gem itself. My expectation is that this will mean there's no longer any downstream problems caused by us when people change the way they go about bundling their assets, or Rails itself changes how they want to do things.

I'm going to start releasing some pre-release versions of the gem in the coming days, as I'm sure I'm about to break stuff for people, and so we can get feedback and make sure there's nice migration help sorted out.

PR #2390, opened 2023-06-02
Update documentation URL to correct Heroku URL

excid3:
https://administrate-prototype.herokuapp.com no longer works, but https://administrate-demo.herokuapp.com does. 

This fixes the broken URLs. 👍


pablobm:
Thank you!

PR #2394, opened 2023-06-27
Update Appraisals

nickcharlton:
Recent updates to the Ruby version and `selenium-webdriver` haven't been kept up with the Appraisal `gemfile`s.


nickcharlton:
I'm merging this with Bundle Audit failing, as that's going to be my next job!

Issue #2311, opened 2023-01-20
Asset Management


nickcharlton:
One of the biggest problems we have is handling assets. We've got a few CSS and JS requirements and as it stands, we assume you're using Sprockets. This isn't really the case any more, with lots of folks using Tailwind, Webpacker and new bundling approaches in their projects. All of these are reasonable and using Administrate shouldn't be as annoying or difficult as it currently might be.

There's been a lot of issues opened, but also not much traction in coming up with a solution that can work for all. This issue will pull together all existing issues, describe the current state of things and outline a plan to make everything that should work, work.

Existing issues:

* https://github.com/thoughtbot/administrate/issues/2151
* https://github.com/thoughtbot/administrate/issues/2130
* https://github.com/thoughtbot/administrate/issues/2091
* https://github.com/thoughtbot/administrate/issues/1617
* https://github.com/thoughtbot/administrate/issues/1469

Proposed solutions:

* https://github.com/thoughtbot/administrate/pull/2256
* https://github.com/thoughtbot/administrate/pull/2148
* https://github.com/thoughtbot/administrate/pull/2116
* https://github.com/thoughtbot/administrate/pull/1491
* https://github.com/thoughtbot/administrate/wiki/Rails-6-&-Webpacker
* https://github.com/thoughtbot/administrate/pull/2446
* https://github.com/thoughtbot/administrate/pull/2397

Current plan:

- [ ] Define the scope (i.e.: make sure we're considering everything everyone is using)
- [ ] Dig into the existing solutions and see what's missing
- [ ] ???

nickcharlton:
I'm taking a bit of a different approach to this problem, as it's been quite long-running, and had a few mis-starts and I think it really needs some dedicated time to dig into it and find the right solution.

I'm going to be taking a week to work on this in Q1 2023, whilst I can't guarantee we'll solve it all completely, I suspect this will be enough to unblock us.

My plan is to use this issue as a bit of a development diary, so others can follow along.

Some opening questions as I make sure I understand the full scope:

1. Is there anything asset related which isn't covered by those above open issues and PRs?
2. Are you using anything other than Sprockets, Webpacker, [CSS](https://github.com/rails/cssbundling-rails)/[JS](https://github.com/rails/jsbundling-rails) Bundling?

geniuskidkanyi:
You can use dart-sassc to fix the sassc dependency

jubilee2:
[...]

PR #2397, opened 2023-07-07
Attempt at publishing the gem with compiled assets

elias19r:
Issue https://github.com/thoughtbot/administrate/issues/2311

I've been looking at how [Avo handles the gem's assets] and decided to
try a similar solution for Administrate's issue with asset management.

In this PR, I'm using [jsbundling-rails] with `esbuild` and
[cssbundling-rails] with `saas` to build Administrate's asset files.
For that, I've removed the `jquery-rails`, `saasc-rails`, and
`selectize-rails` gems and replaced them with NPM packages, updating
imports and moving files as needed.

The goal is to be able to run `yarn run build` and `yarn run build:css`
to build Administrate's assets and then build the gem, including the
contents from `app/assets/builds` in the gem files.

So, the application using `administrate` won't need to compile those
assets because they are ready-to-use `.js` and `.css` files.
The application will be just adding them to its pipeline, generating
digests, and serving the files.

Let me know what you think about this approach!

I'm new to Rails Engine and Administrate, so I may have overlooked something.

[Avo handles the gem's assets]: https://avohq.io/blog/how-to-bundle-assets-in-a-rails-engine
[jsbundling-rails]: https://rubygems.org/gems/jsbundling-rails
[cssbundling-rails]: https://rubygems.org/gems/cssbundling-rails


nickcharlton:
This looks like a great start. I think shipping with compiled dependencies is likely my preferred approach, as any other involves a lot of extra work (when releasing, or for end users).

Could you take a start at documenting how this might be for end users? I'd love to see what it's like when integrating for the first time (after that, we'll want to make sure we have a migration guide in place, if it's needed.)

As you've been doing this, have you seen any cases where we can drop some of the JavaScript/CSS dependencies?

spec/example_app/app/assets/config/manifest.js
```diff
@@ -1,2 +1,6 @@
-//= link administrate/application.css
 //= link administrate/application.js
+//= link administrate/application.css
+//= link administrate/docs.css
```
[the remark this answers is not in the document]
elias19r: @nickcharlton 

Would you know why Hound is barking at .js and .rb files? My first time using it and I think we could ignore this one for `manifest.js`

elias19r:
Hi @nickcharlton

> I'd love to see what it's like when integrating for the first time (after that, we'll want to make sure we have a migration guide in place, if it's needed.)

I tested locally today with some new Rails apps and both `main` and this branch's version seem to work interchangeably, so I think we won't need a migration guide.

Other than that, there is generator called `bundle exec rails generate administrate:assets` that allows the user to copy the gem's original not-compiled assets over to the application's codebase so they can edit and override the gem's CSS and JavaScript. I haven't tested this yet, but I think as long as we keep the original paths for CSS and JavaScript files, this shouldn't be an issue -- assuming the user's application is properly set up to compile Administrate's assets, otherwise they wouldn't be customizing them in the first place.

I've changed the original CSS and JavaScript file paths in this branch to match the usage of jsbundling and cssbundling (main difference is `app/javascript/` instead of `app/assets/javascripts`). I'll work on undoing that.

> As you've been doing this, have you seen any cases where we can drop some of the JavaScript/CSS dependencies?

I haven't seen any cases yet. These are the current dependencies to compile Administrate's assets:

```
  "dependencies": {
    "esbuild": "^0.18.11",
    "jquery": "^3.7.0",
    "jquery-ujs": "^1.2.3",
    "sass": "^1.63.6",
    "selectize": "^0.12.6"
  },
```

spec/example_app/app/assets/config/manifest.js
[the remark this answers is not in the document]
nickcharlton: It's not understanding either the version of JavaScript it should be targetting (in the case of the `import` error) and here that it's a special Rails file. I'd not worry about it too much now, we can get this working and then come and have a look.

(Unless you're at that point already in which case … I have no idea how to fix this, but probably something to do with how Hound is configure is now broken.)

jordanstephens:
I'd be interested in helping get this over the finish line. If you need help, let me know!

jordanstephens:
Hey @elias19r ! Thanks for your work on this. I just wanted to report that I tried out your branch on our rails app and was able to confirm that everything worked as expected! 🎉  Let me know if there's anything else I can do to help push this effort forward.

chloerei:
I tested it on a demo project using propshaft and it works.

propshaft is a better alternative for cssbundling and jsbundling, hope administrator can push this improvement to adapt to propshaft.

danielmorrison:
To add a another voice, I'm testing this on the same setup as @chloerei and it works great.

pablobm:
Thank you for this work @elias19r, and apologies that it's taking so long to have it reviewed. I had a look today, and I really want to push this through the finish line.

I've been tinkering and I have my own version at https://github.com/thoughtbot/administrate/compare/main...pablobm:administrate:publish-gem-with-precompiled-assets-2--digests, which some changes. I still want to test this with some apps, as well as on production-like environments. It's promising so far!

Would you be able to rebase? There have been a few changes since July, so we should get up to date with that.

I changed the `package.json` scripts a bit. In part to clarify them, but also because the `--watch` in the CSS command was being applied only to the "internal docs" part, not the "application". This is because it's two commands separated by `&&` and the `--watch` is just appended at the end, so it only applies to the second command.

I also added `bin/dev_assets`. I used this while linking my local copy of Administrate from an app in my computer. I wanted to be able to regenerate the files without running the dev server, because my separate app had its own server.

.sample.env
```diff
@@ -1,3 +1,3 @@
-# http://ddollar.github.com/foreman/
-RACK_ENV=development
-SECRET_KEY_BASE=development_secret
+export RACK_ENV=development
+export SECRET_KEY_BASE=development_secret
+export PORT=3000
```
pablobm, line 3: Why this change? This file is intended for Foreman, for which the format was correct.

elias19r: I think I changed it so I can run `source .env` and run processes without foreman if I wanted. No big deal, happy to revert it

elias19r:
Thank you @pablobm! Sure, I'll rebase it soon

bin/build_gem
```diff
@@ -0,0 +1,8 @@
+# Exit if any subcommand fails
+set -e
+
+VERSION=$(ruby -r ./lib/administrate/version.rb -e "puts Administrate::VERSION")
+yarn run build
+yarn run build:css
+gem build administrate.gemspec
+
```
elias19r, line 8: @pablobm Having this in a separate script is great!

I noticed the `VERSION` variable is not being used here. Are we missing the following line in this script?

```sh
gem install administrate-${VERSION}.gem
```

.circleci/config.yml
```diff
@@ -44,6 +59,9 @@ commands:
       # Setup the database
       - run: bundle exec rake db:setup
 
+      # Build assets
+      - run: yarn run build-dev:js && yarn run build-dev:css
+
```
elias19r, line 64: @pablobm I like the `:js`, but I also think it is okay to leave it out because it seems to be a pattern that `build` builds js by default.

For example, when we generate a new rails app with some javascript and css setup:

```
cd /tmp/
rails new testing --javascript=esbuild --css=sass

cd testing
cat package.json
```

We get

```js
  // ...
  "scripts": {
    "build": "esbuild app/javascript/*.* --bundle --sourcemap --outdir=app/assets/builds --public-path=/assets",
    "build:css": "sass ./app/assets/stylesheets/application.sass.scss:./app/assets/builds/application.css --no-source-map --load-path=node_modules"
  }
```

package.json
```diff
@@ -0,0 +1,15 @@
...
+    "jquery-ujs": "^1.2.3",
+    "sass": "^1.63.6",
+    "selectize": "^0.12.6"
+  },
+  "scripts": {
+    "build-dev:js": "esbuild app/assets/javascripts/administrate/application.js --bundle --sourcemap --outdir=app/assets/builds/administrate --public-path=/assets/administrate",
+    "build:js": "yarn build-dev:js --minify",
+    "build:css": "sass ./app/assets/stylesheets/administrate/application.scss:./app/assets/builds/administrate/application.css ./app/assets/stylesheets/administrate-internal/docs.scss:./app/assets/builds/administrate-internal/docs.css --source-map --load-path=node_modules --style=compressed",
+    "build-dev:css": "yarn build:css --watch"
+  }
```
elias19r, line 14: @pablobm I love how you simplified this! I was thinking: we may be able to simplify this even further by removing `build-dev`s commands :thinking: 

We could publish the gem with un-minified JS and CSS and delegate the minification to app importing the gem. Because people's apps will include our compiled `.js` and `.css` files into their assets pipelines which would apply minification.

bin/dev_assets
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env sh
+
+"$(dirname "$0")"/dev -m js=1,css=1
```
elias19r, line 3: TIL
I didn't know about `foreman start`'s `-m` argument

.gitignore
```diff
@@ -7,17 +7,31 @@
...
+
 /spec/example_app/log/*
 /spec/example_app/tmp/*
 !/spec/example_app/log/.keep
 !/spec/example_app/tmp/.keep
 /spec/example_app/public/system
 /spec/example_app/public/assets
+
+/spec/example_app/app/assets/builds/*
+!/spec/example_app/app/assets/builds/.keep
```
elias19r, line 24: > This is the dir I get builds in. Am I missing something?

@pablobm Yes, this is the dir we get the builds in. I thought of not committing the compiled assets to Git because that doesn't seem necessary. Compiled assets will be in the gem when we build it.

No strong opinions. Let know if you'd like to commit them to Git

spec/example_app/app/assets/config/manifest.js
```diff
@@ -1,2 +1,4 @@
-//= link administrate/application.css
-//= link administrate/application.js
+//= link administrate-internal/docs.css
+
+//= link_tree ../images
+//= link_tree ../builds
```
elias19r, line 4: @pablobm  I've restored this line. I think it is necessary to link `../builds` here. I'll check again

Gemfile
```diff
@@ -14,6 +14,10 @@ gem "redcarpet"
 gem "sentry-raven"
 gem "unicorn"
 
+gem "cssbundling-rails", "~> 1.2"
+gem "jsbundling-rails", "~> 1.1"
+gem "sprockets-rails", "~> 3.4"
+
```
elias19r, line 20: > I appear to need sassc-rails to run the specs

@pablobm It shouldn't be necessary. I'll double-check! The goal is to use the `sass` command from node package and not rely on the gem anymore

elias19r (commented):
@pablobm I just merged your branch into mine and left comments about the things I didn't pick from you branch. Thank you!

elias19r:
I'll be removing comments from the Hound bot for now because they are noisy. We can rerun it later

.tool-versions
```diff
@@ -0,0 +1,3 @@
+ruby 3.2.2
+nodejs 18.17.0
+yarn 1.22.19
```
elias19r, line 3: @pablobm Oh I forgot to pick the removal of these files when merging your branch. I think we can specify a Node version in `package.json`

dorianmarie:
I'm testing this branch from a brand new rails app and I'm getting this error

Screenshot 2023-12-01 at 14 57 32

dorianmarie:
I used Rails 7.1.2 but I pointed to your branch in github instead of building the gem

.sample.env
```diff
@@ -1,3 +1,2 @@
-# http://ddollar.github.com/foreman/
```
pablobm, line 1: Silly thing, but I think let's bring this line back, just because it's not related to anything else, and it will  reduce the diff noise.

bin/build_gem
pablobm: Uh, I must have indeed :+1: Good catch.

.circleci/config.yml
pablobm: Oh right. Good point, let's stick to convention then :+1:

bin/dev_assets
pablobm: I didn't either. I spent some time looking up how to do this :stuck_out_tongue_closed_eyes:

.sample.env
elias19r: Makes sense, it'll restore it! We can update it in a separate PR. That link is currently responding with "Not Found"

elias19r:
@pablobm Yes, I had sass-related gems installed on my machine for the Ruby version I used when building the gem and the Rails app:

image

I just removed them and was still able to build and install the gem and the run Rails app https://github.com/thoughtbot/administrate/pull/2397#issuecomment-1836477337

elias19r:
When building the gem, we are using `sass` from the NPM package to compile Administrate's `.scss` --> `.css`

Then, the Rails app shouldn't need to do anything to Administrate's `.css` file because it is ready to be used :thinking:

elias19r:
> I used Rails 7.1.2 but I pointed to your branch in github instead of building the gem

@dorianmarie Just pointing to this GitHub branch in the Gemfile is not enough in this case because the repo doesn't contain the `.css` file but the `.scss` file.

When we build the gem, the `.scss` file is processed and turned into a `.css` that is included in the gem.

pablobm (commented):
Minor comments after having another look. When I have time next, I'll try run this in production-like environments in order to test "real world" scenarios.

bin/rails
```diff
@@ -1,5 +1,14 @@
...
+ENGINE_ROOT = File.expand_path("..", __dir__)
+ENGINE_PATH = File.expand_path("../lib/administrate/engine", __dir__)
+APP_PATH = File.expand_path("../spec/example_app/config/application", __dir__)
+
+# Set up gems listed in the Gemfile.
+ENV["BUNDLE_GEMFILE"] ||= File.expand_path("../Gemfile", __dir__)
+require "bundler/setup" if File.exist?(ENV["BUNDLE_GEMFILE"])
+
+require "rails/all"
+require "rails/engine/commands"
```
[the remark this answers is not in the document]
elias19r: I had updated it based on an example app I built with new `rails` version in https://github.com/thoughtbot/administrate/pull/2397/commits/548ccddcfa3e6466dc91e096f0237c3c3205393c
But I think that's not needed for this PR, I'll revert it.
We can update that in a separate PR if we need.

pablobm:
My review of this PR so far has been on a Linux machine. Here's a report running on a macOS machine. (Summary: bumpy but eventually positive).

First, Foreman default to port 5000, but apparently [macOS has called dibs on this port since Monterey](https://github.com/ddollar/foreman/issues/784). Running `./bin/dev` will result in `Address already in use - bind(2) for 0.0.0.0:5000 (Errno::EADDRINUSE)`.

We may want to change the default port. (Can we?). For now I'm running with `PORT=3000 ./bin/dev`.

Next problem is my fault, but I'm mentioning it here as others will see it too. We should document it.

Running Foreman in a different port, I was getting Unicorn to exit immediately with this output:

```
16:49:52 web.1  | I, [2023-12-17T16:49:52.281730 #6088]  INFO -- : listening on addr=0.0.0.0:3210 fd=12
16:49:52 web.1  | I, [2023-12-17T16:49:52.522187 #6088]  INFO -- : master process ready
16:49:52 web.1  | I, [2023-12-17T16:49:52.523042 #6088]  INFO -- : reaped #<Process::Status: pid 6307 exit 0> worker=0
16:49:52 web.1  | Unicorn master intercepting TERM and sending myself QUIT instead
16:49:52 web.1  | Unicorn master intercepting TERM and sending myself QUIT instead
16:49:52 web.1  | I, [2023-12-17T16:49:52.624685 #6088]  INFO -- : reaped #<Process::Status: pid 6309 exit 0> worker=2
16:49:52 web.1  | I, [2023-12-17T16:49:52.624750 #6088]  INFO -- : reaped #<Process::Status: pid 6308 exit 0> worker=1
16:49:52 web.1  | I, [2023-12-17T16:49:52.624878 #6088]  INFO -- : master complete
16:49:52 web.1  | exited with code 0
```

The following was also in the output, but initially I didn't notice it:

```
17:07:24 js.1   | error Command "esbuild" not found.
17:07:24 js.1   | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
17:07:24 js.1   | error Command failed with exit code 1.
17:07:24 js.1   | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
17:07:24 css.1  | error Command "sass" not found.
17:07:24 css.1  | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
17:07:24 css.1  | error Command failed with exit code 1.
17:07:24 css.1  | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
17:07:24 css.1  | exited with code 1
17:07:24 system | sending SIGTERM to all processes
17:07:24 js.1   | exited with code 1
```

I hadn't ran `yarn install`, so the `js` and `css` processes failed immediately, causing Foreman to `SIGTERM` the `web` process.

OK, so now I can run `PORT=3000 ./bin/dev`... but it segfaults 😭  I guess it's relevant to mention that I'm running Apple Silicon here. Specifically an Apple M1 Pro.

So I go with `./bin/rails s` and `./bin/dev_assets` in different terminals. I fire the browser and immediately I get `LoadError in DocsController#index - cannot load such file -- sassc`



So I add `sassc` to the `Gemfile`.  Finally this appears to work! 😅 

My next question is: how do I add custom CSS/JS to my Administrate setup? I believe that we don't document this, which is an oversight. But that's a different PR... Good news is that I was able to do it, but I deleted the file `spec/example_app/app/assets/stylesheets/application.sass.scss` as I think it's misleading. Although I can't be 100% sure.

The changes I needed to make this work can be seen at https://github.com/elias19r/administrate/compare/publish-gem-with-precompiled-assets-2--digests...pablobm:pablobm-precompiled-assets-review

Looks like this is coming along! We really need more people to review this though, both locally and in production-like environments.

On my part, I deployed the Administrate demo app to a Dokku instance (an open source clone of Heroku) and it worked as expected. Now I should try deploy other apps.

elias19r:
@pablobm Thank you for testing again!

I'm able to reproduce that `LoadError in DocsController#show - cannot load such file -- sassc` error.

It happens to me when:
- There aren't any processed `.js` and `.css` files in `app/assets/builds/`
- And I'm not running `yarn build --watch` or `yarn build:css --watch` to create those files

When you run `bin/dev_assets`, do you see the processed `.js` and `.css` files being created in `app/assets/builds/administrate` and `app/assets/builds/administrate-internal` folders?

pablobm:
That works indeed! Then I guess it was a product of not being able to run `bin/dev` on the Mac (due to no port 5000) and initially running `bin/rails s` only, without `bin/dev_assets`. We'll have to make sure to document it. Thank you for looking into it!

pablobm:
I have updated [my branch](https://github.com/elias19r/administrate/compare/publish-gem-with-precompiled-assets-2--digests...pablobm:pablobm-precompiled-assets-review) to:

- Remove the `sassc` reference.
- Make `bin/dev` run the webapp in port 3000, avoiding a clash with macOS having claimed 5000 for itself.

I have also started a separate PR to [document how to add custom CSS/JS to Administrate](https://github.com/thoughtbot/administrate/pull/2475). This is something that we'll need to double-check works with whichever solution we end up choosing for asset handling (I really hope it'll be this PR!).

.gitignore
```diff
@@ -7,17 +7,31 @@
...
+
+/spec/example_app/node_modules
 /spec/example_app/config/database.yml
+
+/node_modules
+/public
+
+/app/assets/builds/*
+!/app/assets/builds/.keep
```
pablobm, line 33: Doing more testing, I have found a reason to not gitignore this folder: if the compiled assets are not committed, you'll get the scss error if you pull the gem directly from the repo. Eg: `gem "administrate", github: "thoughtbot/administrate", branch: ...`.

nickcharlton:
I'm happy to say that I have time this week and next to spend getting this over the line, and also that this is broadly what I think is the right approach to take.

@elias19r, could you rebase it against current `main`? I've been trying on and off yesterday and today I'm not convinced I'm picking the right changes and it should be easier for you as you did most of it!

@pablobm, is there anything else from your branches you think should go into this?

.gitignore
elias19r: You're right, if we don't commit the compiled assets we lose the ability to point to a specific branch/commit directly in the Gemfile. I'll update to commit the compiled assets

bin/dev
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env sh
+
+if ! gem list foreman -i --silent; then
+  echo "Installing foreman..."
+  gem install foreman
+fi
+
```
pablobm, line 7: Since macOS uses port 5000, how about adding this line to make the server work on port 3000 instead?

```suggestion

export PORT=3000
```

pablobm:
@nickcharlton @elias19r - Last think from my branch, I have added a comment about something to add (`PORT=3000` in `bin/dev`).

I think this looks good. What I think is missing is:

1. Some documentation.
2. Additional testing in real applications that you may have at hand, particularly in a production-like environment.

nickcharlton:
Ah, excellent.

Yeah — docs and testing is broadly my plan to get this over the line. I've got some things I've got access to where I can try this out, and a few scenarios that I think we should test bobbing about in my head.

From there, I'll cut a pre-release version and then we can hopefully encourage others to try it out and report back.

.gitignore
elias19r: Updated in https://github.com/thoughtbot/administrate/pull/2397/commits/fb9beff9544445eef71f106881073e2a4a8214c6

That adds 10k lines from the compiled CSS and JS :smiley:

elias19r:
> @elias19r, could you rebase it against current `main`? I've been trying on and off yesterday and today I'm not convinced I'm picking the right changes and it should be easier for you as you did most of it!

@nickcharlton I've just merged main into my branch

bin/dev
elias19r: @pablobm Should we add this to the `.sample.env` as well?

pablobm: Uh, good shout. Perhaps that's the correct place to put it, instead of here. I just checked and Foreman does pick it up if it's in `.env`.

The downside would be that people are going to forget to provide a `.env` (or run `bin/setup` which does it for you) and will get this problem. I still think `.env` is the correct way to go as it allows people to select their desired port if they prefer it to be different, but we may need a troubleshooting section in the docs to warn about this.

pablobm:
@elias19r @nickcharlton - Would it be possible to tell Hound to ignore the `builds` folder?

nickcharlton:
I've just reworded the single commit to contain all the other changes we've done, cite both of you properly.

I then just added a commit to ignore `app/assets/builds`, too.

nickcharlton:
Okay, final thing before I merge this:

Do you need these?

```
spec/example_app/app/assets/javascripts/admin.js
spec/example_app/app/assets/stylesheets/admin.css
```

I think they were added to test everything works?

elias19r:
> Okay, final thing before I merge this:
> 
> Do you need these?
> 
> ```
> spec/example_app/app/assets/javascripts/admin.js
> spec/example_app/app/assets/stylesheets/admin.css
> ```
> 
> I think they were added to test everything works?

We don't need them. Those examples will be added by a separate PR https://github.com/thoughtbot/administrate/pull/2475

pablobm:
@nickcharlton, similarly we don't need `spec/example_app/config/initializers/administrate.rb`. That was part of the same test.

nickcharlton:
Ah, magic thanks!

nickcharlton:
Hm, whilst it's fine when running `scss-lint` locally, Hound doesn't seem to be following my configuration.

elias19r:
@nickcharlton I've just force-pushed amending your reworded single commit to remove these 2 lines:

```diff
diff --git a/spec/example_app/app/assets/config/manifest.js b/spec/example_app/app/assets/config/manifest.js
index 8f7a648..2589fe9 100644
--- a/spec/example_app/app/assets/config/manifest.js
+++ b/spec/example_app/app/assets/config/manifest.js
@@ -1,6 +1,4 @@
 //= link administrate-internal/docs.css
-//= link admin.css
-//= link admin.js
 
 //= link_tree ../images
 //= link_tree ../builds
```

nickcharlton:
I've been working on doing a final release before merging this in, so that the release _with_ this (which will be 1.0.0) won't also bring with it a bunch of new things.

I didn't get an answer when I asked Hound about the linting issue, so my plan is probably to replace the linting or similar so we can get this merged.

nickcharlton:
I've been working in #2492 to replace Hound with our `stylelint` configuration, but that's got me in that repo sorting that out. Two steps forward, one step back…

nickcharlton:
We're now linting CSS with `@thoughtbot/stylelint-config`. Hound is still going to complain, because it's actually picking up the organisation-level configuration and not this one. I didn't realise that's what might be happening until today.

Anyway, once the tests pass, I'll do a cursory walk over the diff and then I'll be merging it!

jayroh:
EXCITING! I can't wait to try this out!

Thank you all for all (@elias19r, @pablobm, @nickcharlton) for all of the hard work!

Issue #2048, opened 2021-10-13
Searching fields on has many associations returns duplicate results


tjvc:
### What were you trying to do?

Mark a has many field as `searchable` and set some `searchable_fields`. Perform a search, and the parent record will appear multiple times on the dashboard (when it has more than one associated child).

I'm not sure if this is intended to be supported (the search options don't seem to be documented for a has many field), but it at least partially works.

### What did you end up with (logs, or, even better, example apps are great!)?

Here's an example query from the logs (this is a has many through association, but the principle is the same):

```
SELECT "users".* FROM "users" LEFT OUTER JOIN "committee_customers" ON "committee_customers"."user_id" = "users"."id" LEFT OUTER JOIN "committees" ON "committees"."id" = "committee_customers"."committee_id" WHERE "users"."role" = $1 AND "users"."soft_deleted" = $2 AND (LOWER(CAST("users"."email" AS CHAR(256))) LIKE '%test%' OR LOWER(CAST("users"."first_name" AS CHAR(256))) LIKE '%test%' OR LOWER(CAST("users"."last_name" AS CHAR(256))) LIKE '%test%' OR LOWER(CAST("committees"."name" AS CHAR(256))) LIKE '%test%') LIMIT $3 OFFSET $4
```

We're left joining associations, so I think all that's required is a `DISTINCT`.

### What versions are you running?

`rails (6.0.4.1)`
`administrate (0.16.0)`

pablobm:
I agree that those duplicates should not appear. There must be some SQL magic we can apply there to fix this issue.

In the longer term, ideally I would like our current search replaced with something more abstract, with Administrate bundling only a basic implementation and alternatives provided by plugins or custom ad-hoc code. This way people could do things like integrating with Elasticsearch or enhancing the query format. But that won't happen soon... unless someone wants to take ownership of this task?

plabadille:
With the last version [0.18.0](https://github.com/thoughtbot/administrate/releases/tag/v0.18.0) there is an easy fix for this issue. Just need to slightly [customize the search](https://github.com/pablobm/administrate/blob/howto-search/docs/guides/customising_search.md#in-more-detail) :
```
def filter_resources(resources, search_term:)
  super(resources, search_term: search_term).distinct
end
```

I may miss some side effects but I think it should be included in the core method ? It won't work if your rails version is above 6.0 (this may be why it's not in the core yet).

pablobm:
[...]

PR #2402, opened 2023-07-21
Fix Searching fields on has many associations returns duplicate results #2048

plabadille:
Checkout https://github.com/thoughtbot/administrate/issues/2048#issuecomment-1625140924

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -84,7 +84,7 @@ module Administrate
         resources,
         dashboard,
         search_term,
-      ).run
+      ).run.distinct
     end
 
     def after_resource_destroyed_path(_requested_resource)


pablobm:
Hello @plabadille :slightly_smiling_face: Thank you for this PR and apologies for the delay in reviewing. There appears to be a failing spec, related to SQL going wrong (oof...). Would you be able to have a look and see if it can be fixed?

Also, I wonder if it would be possible to create a spec with an example of the specific behaviour we want?

plabadille:
Hello @pablobm 

So there's an impact adding the `.distinct`. In fact my fix proposal only work if there's no need to order by some col of a joined table.

The error pretty clear : `PG::InvalidColumnReference: ERROR:  for SELECT DISTINCT, ORDER BY expressions must appear in select list`

In your demo used in test the customer dashboard contain : `orders: Field::HasMany.with_options(limit: 2, sort_by: :id),` which generate the folowing sql query :
```
SELECT DISTINCT customers.*
FROM customers
LEFT OUTER JOIN orders ON orders.customer_id = customers.id
WHERE customers.hidden = FALSE
GROUP BY customers.id
ORDER BY COUNT(orders.id) asc LIMIT 20 OFFSET 0
```

I'm sorry but I have no clue on how to make it work for now and I do not have the time to dig deeper. My solution #2048 work well only without sort_by defined on an association.

It sadly can't be generalized without working on this generated query (on the distinct, the select or the order_by clauses).

pablobm:
Argh, this is annoying. I feel you. Coincidentally I came a cross something very similar recently, so it's sort of fresh in my mind.

I'm going to close this for now. Perhaps we can come up with some other solution.

PR #2405, opened 2023-07-29
New feature: field hints for forms inputs

blocknotes:
Hey!
Thank you for the nice gem! 😄 
I use it from quite some time and now I'm here to propose a small feature: adding optional hints (as small text lines) below form input fields.
I'm opening the PR in draft as proof of concept, if you like the idea I could complete it (with tests / anything needed).

I'm using this feature in a couple of projects, I think it can help offering context information to the users.

The changes should be minimal, I just added a div below `div.field-unit__field` and set basic styles for the new element.
I also had to add `flex-wrap: wrap;` to `.field-unit` in order to place the new div on a new line.
In another version I was passing the hint via field options but now I prefer the translations way because in any case the string would be stored there for my projects.

Sample locale:

```yml
en:
  administrate:
    field_hints:
      book:
        content: Brief summary of the book
```

The effect is like the following:

image

Please let me know your opinion 🙏

### DIFF
--- a/app/assets/stylesheets/administrate/components/_field-unit.scss
+++ b/app/assets/stylesheets/administrate/components/_field-unit.scss
@@ -2,6 +2,7 @@
   @include administrate-clearfix;
   align-items: center;
   display: flex;
+  flex-wrap: wrap;
   margin-bottom: $base-spacing;
   position: relative;
   width: 100%;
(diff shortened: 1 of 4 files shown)


pablobm:
Nice! I like the idea and I think using i18n is the way to go :+1: 

As well as a test, would you be able to put together a bit of documentation? Perhaps under https://github.com/thoughtbot/administrate/blob/main/docs/customizing_dashboards.md

Thank you @blocknotes!

blocknotes:
Hey @pablobm,
I should have set up the required changes; I put the PR in Ready for review.
Please let me know if you spot something.

nickcharlton:
Thanks! This is a nice little feature I've wanted in the past, but never got around to do anything about it!

I just rebased it, and I'll merge it shortly.

Issue #2404, opened 2023-07-28
Example app: don't use float types for money


pablobm:
I just realised that the example app uses a float type for prices. For example:

```
  create_table "line_items", id: :serial, force: :cascade do |t|
    t.integer "order_id"
    t.integer "product_id"
    t.float "unit_price"
    t.integer "quantity"
    t.datetime "created_at", null: false
    t.datetime "updated_at", null: false
    t.index ["order_id"], name: "index_line_items_on_order_id"
    t.index ["product_id"], name: "index_line_items_on_product_id"
  end
```

Bad bad app! Floats are inexact and not suitable for use with money, regardless of precision. Instead we should use `decimal` (which would store fractions correctly) or `integer` (moving to store money as fractional amounts such as cents: `unit_price_in_cents`).

This doesn't affect Administrate, but it sets a terrible example to developers out there who may see it and not know better. Let's fix it!

viniciusueharaweb:
Hey there! I'd love to tackle this issue! Thanks

pablobm:
@viniciusueharaweb - Go for it! :slightly_smiling_face: Happy to provide more guidance if you need any.

davqasd:
This issue had been open for 5 days, so, I decided to try :slightly_smiling_face:

PR #2408, opened 2023-08-02
Example app: change float column types to decimal

davqasd:
This commit fixes https://github.com/thoughtbot/administrate/issues/2404.


pablobm:
Looks good! Thank you :slightly_smiling_face: 

By the way, I had a look at why those datetimes lost their `precision`. This seems to be the reason: https://github.com/rails/rails/pull/44358

I haven't checked, but this must be the first time we add a migration since moving to Rails 7, and consequently the first `schema:dump`. Therefore this change is where this behaviour manifested first.

PR #2411, opened 2023-08-04
Add RichText field type

littleforest:
This adds a RichText field type, completing the work started in #1660.

Screenshot 2024-03-22 at 2 14 15 PM

Screenshot 2024-03-22 at 2 47 59 PM


littleforest:
Will look into the failing tests, some of them are not failing for me locally :(

app/views/fields/rich_text/_show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+## Local variables:
+
+- `field`:
+  An instance of [Administrate::Field::RichText][1].
+  A wrapper around the RichText object pulled from the database.
+
+[1]: http://www.rubydoc.info/gems/administrate/Administrate/Field/RichText
+%>
+
+<%= field.data.body&.html_safe %>
```
seanpdoyle, line 18: Could we rely on Action Text to render itself as HTML safe?

```suggestion
<%= field.data %>
```

lib/administrate/field/rich_text.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module Administrate
+  module Field
+    class RichText < Administrate::Field::Base
+      def to_s
+        data&.to_plain_text
+      end
+
+      def truncate
+        to_s[0...truncation_length]
```
seanpdoyle, line 11: Is there an advantage to using `#[]` access over [String#truncate](https://edgeapi.rubyonrails.org/classes/String.html#method-i-truncate)?

```suggestion
        to_s.truncate(truncation_length)
```

lib/administrate/base_dashboard.rb
```diff
@@ -7,6 +7,7 @@
 require "administrate/field/has_one"
 require "administrate/field/number"
 require "administrate/field/polymorphic"
+require "administrate/field/rich_text"
```
seanpdoyle, line 10: As an alternative to eager-loading the file, have you considered utilizing Rails' [Engine load hooks](https://guides.rubyonrails.org/engines.html#available-load-hooks)

If we relied on the host application depending on the `action_text` gem and loading `action_text/engine`, then we could defer loading the `Administrate::Field::RichText`.

I've tried this out locally with the following changes:

```diff
diff --git a/Gemfile.lock b/Gemfile.lock
index c740df4..a890f52 100644
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,6 @@ PATH
   specs:
     administrate (0.19.0)
       actionpack (>= 5.0)
-      actiontext (>= 6.0)
       actionview (>= 5.0)
       activerecord (>= 5.0)
       jquery-rails (>= 4.0)
diff --git a/administrate.gemspec b/administrate.gemspec
index b6dbad7..dc704d1 100644
--- a/administrate.gemspec
+++ b/administrate.gemspec
@@ -16,7 +16,6 @@ Gem::Specification.new do |s|
   s.add_dependency "actionpack", ">= 5.0"
   s.add_dependency "actionview", ">= 5.0"
   s.add_dependency "activerecord", ">= 5.0"
-  s.add_dependency "actiontext", ">= 6.0"
 
   s.add_dependency "jquery-rails", ">= 4.0"
   s.add_dependency "kaminari", ">= 1.0"
diff --git a/lib/administrate/engine.rb b/lib/administrate/engine.rb
index 3e6dc7b..00158a2 100644
--- a/lib/administrate/engine.rb
+++ b/lib/administrate/engine.rb
@@ -1,4 +1,3 @@
-require "action_text/engine"
 require "jquery-rails"
 require "kaminari"
 require "sassc-rails"
@@ -30,6 +29,12 @@ module Administrate
       ]
     end
 
+    initializer "administrate.action_text" do
+      ActiveSupport.on_load :action_text_rich_text do
+        require "administrate/field/rich_text"
+      end
+    end
+
     def self.add_javascript(script)
       @@javascripts << script
     end
diff --git a/spec/example_app/app/controllers/application_controller.rb b/spec/example_app/app/controllers/application_controller.rb
index 2fc8543..c17b696 100644
--- a/spec/example_app/app/controllers/application_controller.rb
+++ b/spec/example_app/app/controllers/application_controller.rb
@@ -1,5 +1,3 @@
-require "action_text"
-
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
diff --git a/spec/example_app/config/application.rb b/spec/example_app/config/application.rb
index 82bbe4f..ebb45c7 100644
--- a/spec/example_app/config/application.rb
+++ b/spec/example_app/config/application.rb
@@ -5,6 +5,7 @@ require "active_record/railtie"
 require "action_controller/railtie"
 require "action_view/railtie"
 require "sprockets/railtie"
+require "action_text/engine"
 
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
```

app/assets/javascripts/trix_custom.js
seanpdoyle, line 1: As of https://github.com/rails/rails/commit/9c73d4aeab112a921798f8e1acd595af498ed64a, `actiontext.js`, `trix.js`, and `trix.css` are available from the asset pipeline. I wonder if there's a way to utilize those built-in asset dependencies.

santosh-1987:
Are we planning to release this anytime soon - Need this badly for actiontext

nickcharlton:
@littleforest, are you able to pick up from @seanpdoyle's comments here?

(If not, I can do so!)

littleforest:
@nickcharlton I will see if I can find time to work on this second week of January. Might go faster if I can find time to pair with you or @seanpdoyle .

nickcharlton:
@littleforest, sorry, only getting back to this one now. How did you get on? Feel free to drop some time on my calendar if you've like to pair on this.

seanpdoyle:
@littleforest the asset changes merged as part of https://github.com/thoughtbot/administrate/pull/2397 have introduced merge conflicts.

If you're interested, I'm willing to give resolving those conflicts a try and force-pushing to this branch.

If not, and you'd rather handle that yourself, that's great too!

lib/administrate/base_dashboard.rb
littleforest: @seanpdoyle I tried to implement this, but then I am back to the `undefined method `rails_direct_uploads_url'` error.

littleforest: But not when I run the specs individually, hmmm.

littleforest: Still getting these errors when running the test suite as a whole, but not when run individually, even after all of our additions :(

littleforest: Oh, forgot about [this comment](https://github.com/thoughtbot/administrate/pull/2411#issuecomment-1693583352)

littleforest: Fixed in 7304e0b47c3fa6e97c63bc685892a6d16c23c358

littleforest: @seanpdoyle ultimately I can't get it to work when I remove this `require "administrate/field/rich_text"` (with the changes in `lib/administrate/engine.rb`). Any other ideas, or should I just remove the changes in `lib/administrate/engine.rb`?

littleforest:
@seanpdoyle @nickcharlton @pablobm this is ready for re-review.

Gemfile
```diff
@@ -7,6 +7,7 @@ gem "administrate-field-image"
 gem "faker"
 gem "front_matter_parser"
 gem "globalid"
+gem "image_processing"
```
nickcharlton, line 10: Does this mean we'll now have a system dependency on `ImageMagick` or `libvips`?

I guess if you're doing this with Rails that's going to happen anyway, so maybe it's not a concern for us?

spec/support/generator_spec_helpers.rb
```diff
@@ -37,7 +37,7 @@ def each_file_in(path)
 
   def reset_routes
     Rails.application.routes.clear!
-    load "spec/example_app/config/routes.rb"
+    Rails.application.routes_reloader.execute
```
nickcharlton, line 40: Ooh, that's a nice change from just loading the file!

littleforest: I needed this since there were some active storage direct upload URLs that didn't get reset without it.

nickcharlton: Ahh, interesting!

nickcharlton:
I've just rebased, and I'll let the tests run before merging this in.

For those following along, it'll go out with our next v1.0 beta release, which I hope to do soon, but I'm behind things at the moment so it might be a while.

nickcharlton:
And merged. Thanks @sedubois for the original work, and @littleforest for consistently pushing this through!

Gemfile
littleforest: I was going to see if it made sense to not require this here, and then the Rails app that is using Administrate would require it instead? So then if you are using Administrate but not ActionTest/ActiveStorage you won't be unnecessarily requiring this. Would that work?

nickcharlton: Ah, hm! Yeah, I'd marked it as resolved on the assumption it'd be okay but maybe it's not?

If you can, I think this is worth exploring.

kaka-ruto:
> it'll go out with our next v1.0 beta release

Hi @nickcharlton . Thanks for the work in this gem. Is the above still the case? As I don't see anywhere in the v1.0 releasementioning the addition of the RichText field type

nickcharlton:
It's waiting on a second beta release, where it'll be included then. I'm going to try and do this today, so keep an eye out on the releases.

PR #2412, opened 2023-08-04
Upgrade selenium-webdriver gem and remove webdrivers

littleforest:
Upgrade selenium-webdriver gem and remove webdrivers gem. The webdrivers gem is no longer necessary as Selenium Manager now handles its own installations of the Chrome Driver.

https://github.com/rails/rails/pull/48847


littleforest:
Doesn't work with Ruby < 3

PR #2410, opened 2023-08-04
Another year, another change to how to deal with webdrivers

pablobm:
Webdrivers breaking again (at least locally). Here's one of the fixes proposed at https://github.com/titusfortner/webdrivers/issues/247

PR #2415, opened 2023-08-09
hasMany collection columns

jubilee2:
idea come from https://github.com/nickcharlton/administrate-field-nested_has_many/pull/54 but I use positive list not only skip the column but also can add/remove/order the columns.


pablobm (commented):
Looks good! :slightly_smiling_face: Would you be able to add a test? I'm thinking a simple one would be to change the order dashboards to display the line items like this:

```ruby
    line_items: Field::HasMany.with_options(collection_attributes: [:product, :quantity, :unit_price, :total_price]),
```

Then you can check that the "Order" column doesn't appear. What do you think?

jubilee2:
> Looks good! 🙂 Would you be able to add a test? I'm thinking a simple one would be to change the order dashboards to display the line items like this:
> 
> ```ruby
>     line_items: Field::HasMany.with_options(collection_attributes: [:product, :quantity, :unit_price, :total_price]),
> ```
> 
> Then you can check that the "Order" column doesn't appear. What do you think?

added

pablobm:
Oof, build issue 😫 I think the issue is related to https://github.com/thoughtbot/administrate/pull/2410. Would you be able to rebase on top of the latest `main` and re-push?

jubilee2:
> Oof, build issue 😫 I think the issue is related to #2410. Would you be able to rebase on top of the latest `main` and re-push?

merged main

pablobm:
OK, let's go! 🚀 

Thank you for the contribution 🙂

PR #2418, opened 2023-08-18
Build SQL expressions with Arel instead of string interpolation

thoughtbot-summer:
There, but for the grace of `column_exist?`, go we.

In `Administrate::Order`, there are several places where we are building SQL expressions using string interpolation, _e.g._ `order = "#{relation.table_name}.#{attribute} #{direction}"`. Doing this can be dangerous, because it's easy to introduce a SQL injection vulnerability. Thankfully, it seems like this class is carefully designed in a way that prevents such a vulnerability. Nevertheless, it feels precarious; if someone makes changes, they have to be very careful to keep the code safe.

This change replaces all string-interpolated SQL expressions with Arel expressions. The database adapter will then automatically quote & sanitize the components of these expressions.

### DIFF
--- a/spec/support/matchers/to_sql.rb
+++ b/spec/support/matchers/to_sql.rb
@@ -1,0 +1,5 @@
+RSpec::Matchers.define :to_sql do |sql|
+  match do |actual|
+    actual.to_sql == sql
+  end
+end
--- a/lib/administrate/order.rb
+++ b/lib/administrate/order.rb
@@ -10,9 +10,9 @@ module Administrate
       return order_by_association(relation) unless
         reflect_association(relation).nil?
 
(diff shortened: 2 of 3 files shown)


spec/support/matchers/to_sql.rb
```diff
@@ -0,0 +1,5 @@
+RSpec::Matchers.define :to_sql do |sql|
+  match do |actual|
+    actual.to_sql == sql
```
nickcharlton, line 3: Oh, nice.

nickcharlton (approved):
An excellent idea, thanks!

nickcharlton:
I just rebased (only actual change I made was to add the contents of your PR description to the commit itself) and I'll merge this in a bit.

PR #2419, opened 2023-08-18
Update `selenium-webdriver` and remove `webdrivers`

thoughtbot-summer:
I was unable to run tests locally, because I have Chromium v116 installed, but the tests lock `chromedriver` to v114. Per [`webdrivers`' recommendation](https://github.com/titusfortner/webdrivers/tree/v5.3.1#update-future-of-this-project), this change updates [`selenium-webdriver`](https://github.com/SeleniumHQ/selenium/tree/trunk/rb) and removes [`webdriver`](https://github.com/titusfortner/webdrivers), allowing us to remove that version constraint.


thoughtbot-summer:
CI is failing on this PR because [`selenium-webdriver` dropped support for Ruby 2.7](https://github.com/SeleniumHQ/selenium/pull/11966) (because Ruby 2.7 is EOL). Should we also drop support for it? This would only affect CI for now, since we don't specify a [`required_ruby_version`](https://guides.rubygems.org/specification-reference/#required_ruby_version) in [our gemspec](https://github.com/thoughtbot/administrate/blob/817d79973a6e4d7fcbd82281a334640667bd0bc1/administrate.gemspec), and [`selenium-webdriver` is a test-only dependency](https://github.com/thoughtbot/administrate/blob/817d79973a6e4d7fcbd82281a334640667bd0bc1/Gemfile#L34).

pablobm:
@nickcharlton wasn't sure about dropping support for Ruby 2.7 recently, so pinging him here :slightly_smiling_face: (I'm ok with dropping it).

nickcharlton:
I don't think we should drop it yet, even though it's EOL.

I've not been able to keep up with the recent Selenium changes, is stopping running any tests against it breaking that completely? Is there something else we can pin for older Ruby versions?

pablobm:
I thought that webdrivers had the ability to download the correct (or in this case, pinned) version automatically? I guess this is not happening here. Could that be something that can be fixed? Am I mixing this up with something else?

thoughtbot-summer:
`webdrivers` has the ability to download the pinned version of the **driver**, but unfortunately it does not have the ability to download the corresponding version of the **browser**. Doing so is definitely possible ([Puppeteer does it](https://github.com/puppeteer/puppeteer/tree/puppeteer-v21.1.1#installation)), but I'm not sure if there's a Ruby gem out there that provides that functionality.

nickcharlton:
Where are we on fixing Selenium/Webdrivers/etc? What should we do here?

The initial change seems to have settled down now and everything adjusted.

pablobm:
I was just fixing my local setup yet again, and got it working with pretty much the same changes as in this PR.

@nickcharlton - Any thoughts on when it would be ok to drop Ruby 2.7?

nickcharlton:
I just opened #2469, which is the Ruby 2.7 removal. I figured that since we need that to do this, we should go about 2.7 now.

I cherry-picked @thoughtbot-summer's commit, so the change doesn't get lost when we built the CHANGELOG.

PR #2440, opened 2023-09-27
Reduce iterations and allocations for Namespace#routes

enviable:
Hello! I noticed that the Namespace#routes method was called during most/(all?) requests and could be optimized to reduce iterations and memory allocations.

### DIFF
--- a/lib/administrate/namespace.rb
+++ b/lib/administrate/namespace.rb
@@ -11,10 +11,16 @@ module Administrate
     end
 
     def routes
-      @routes ||= all_routes.select do |controller, _action|
-        controller.starts_with?("#{namespace}/")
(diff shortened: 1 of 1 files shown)


nickcharlton:
I rebased this as the tests didn't seem to have run. I'd like to test this out on a few applications first just to check there's no odd regressions in the situations I have access to, but it might take me a while.

pablobm:
@nickcharlton - Were you able to test this? By the way, the CI failed, but it was some time ago so I can't see the logs. Would someone be able to rebase, please?

pablobm:
I have been testing this a bit and I think it's a good change. Merging.

Thank you @enviable!

PR #2447, opened 2023-10-11
Implement client-side with Stimulus

seanpdoyle:
Depend on [@hotwired/stimulus][] on the client-side. Add the client-side
dependency for `@hotwired/stimulus` to the `package.json`, , then
scaffold out the changes to the `application.js` and
`controllers/index.js` modules.

Next, migrate the jQuery-powered code into controllers: the `select`
controller to integrate with `selectize.js` and the `table` controller
to manage `<table>` elements.

Finally, render each necessary field (`belongs_to`, `has_many`,
`polymorphic`, `select`) with the `[data-controller="select"]` attribute
(powered by a new `Field#html_controller` method.

[@hotwired/stimulus]: https://stimulus.hotwired.dev


nickcharlton (commented):
I think this looks good!

Can you foresee any upgrade issues? We'll be able to catch the template changes and warn of those — although I wonder if we've got enough changes we want to explore an upgrade tool? I'm not sure yet.

nickcharlton:
Thanks for keeping this up to date! I just did a quick click around in the demo to double-check everything, but I'll merge this once the suite passes again.

PR #2448, opened 2023-10-11
Replace `jquery-ujs` with `@hotwired/turbo`

seanpdoyle:
Depend on [@hotwired/stimulus][] on the client-side. Add the client-side
dependency for `@hotwired/stimulus` to the `package.json`, , then
scaffold out the changes to the `application.js` and
`controllers/index.js` modules.

Next, migrate the jQuery-powered code into controllers: the `select`
controller to integrate with `selectize.js` and the `table` controller
to manage `<table>` elements.

Finally, render each necessary field (`belongs_to`, `has_many`,
`polymorphic`, `select`) with the `[data-controller="select"]` attribute
(powered by a new `Field#html_controller` method.

[@hotwired/stimulus]: https://stimulus.hotwired.dev

Replace `jquery-ujs` with `@hotwired/turbo`
===

Replace our client-side dependency on `jquery-ujs` with
[@hotwired/turbo][].

In addition to adding an entry to the `package.json` file, this commit
replaces `[data-confirm]` attributes with `[data-turbo-confirm]`, and
`[data-method]` attributes with `[data-turbo-method]`.

In an effort to minimize the diff, this commit *does not* replace the
`link_to` calls made with `method: :delete` with [button_to][] calls.
Ideally, they'd utilize `button_to` to render a `<form>` element for the
destructive action.

In the future, once that change is complete, the `data: {turbo_method:
:delete}` attributes can once again be passed as `method: :delete`
options.

[@hotwired/turbo]: https://turbo.hotwired.dev
[turbo-rails]: https://github.com/hotwired/turbo-rails
[link_to]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to
[button_to]: https://edgeapi.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-button_to


gillisd:
@seanpdoyle Thank you for doing this. What are your thoughts on replacing this `window.location =` with `Turbo.visit` as well?

https://github.com/thoughtbot/administrate/blob/024296f1a4e010229b8f5353beaa7648e593a1de/app/assets/javascripts/administrate/components/table.js#L16

spec/example_app/app/views/admin/customers/_collection_item_actions.html.erb
```diff
@@ -12,7 +12,7 @@
     [namespace, resource],
     class: "text-color-red",
     method: :delete,
-    data: { confirm: t("administrate.actions.confirm") }
+    data: { turbo_confirm: t("administrate.actions.confirm") }
```
mvz, line 15: I think the specs are failing because the `method: :delete` bit should also be changed to a `turbo_method: :delete` data item.

seanpdoyle:
Thank you for pushing this forward!

Just to highlight some troubles: this PR seems to introduce some test flakiness with regard to the search field. I'm not sure about the cause, since the https://github.com/thoughtbot/administrate/pull/2447 PR was the one responsible for modifying how those test drive the search behavior.

It's something to monitor in the future.

nickcharlton:
Ah, interesting, thanks! I'll keep an eye on it and open an issue if it keeps happening.

I know on a certain client project we had some trouble with the way we're using to writing feature/system specs and test flakiness and I might need some help digging into it.

PR #2452, opened 2023-11-01
Updated Views::IndexGenerator to copy all partials

goosys:
I believe this is an oversight from when the templates were split in a previous version.
Please review.

### DIFF
--- a/lib/generators/administrate/views/index_generator.rb
+++ b/lib/generators/administrate/views/index_generator.rb
@@ -9,7 +9,11 @@ module Administrate
         def copy_template
           copy_resource_template("index")
           copy_resource_template("_collection")
+          copy_resource_template("_collection_header_actions")
+          copy_resource_template("_collection_item_actions")
           copy_resource_template("_index_header")
+          copy_resource_template("_pagination")
+          copy_resource_template("_search")
         end
       end
     end


nickcharlton:
I'm not sure if we do actually want to do this — are you finding you do override stuff like pagination, or the headers? The code here hasn't changed much in years.

One other option would be to add a new generator for overriding the specifics, or we could also just do nothing and let users figure it out!

goosys:
@nickcharlton 
Thank you for the review.

These templates are not generated by any generator. Shouldn't we enable them to be generated by the generator as a guide for users who want to customize the templates?

> One other option would be to add a new generator for overriding the specifics

I'm prepared to create a new generator and submit a pull request. In that case, what do you think would be an appropriate name for the generator?

pablobm:
Initially I wasn't keen into incorporating these partials on the generator, but I'm less sure now. I think a user wanting to customise the index views will want them, so it makes sense to provide them.

nickcharlton:
Yeah, I was thinking the same, but I think it's worthwile having now.

@goosys, would you like to revisit this? I think it needs a rebase and that's probably it.

goosys:
Thank you for checking. I rebased.

pablobm (approved):
Thank you!

Issue #2363, opened 2023-04-20
Provide more context to fields


pablobm:
A common issue for users of Administrate is that fields do not receive enough context. The most typical use case is that they cannot modify the configuration of a field depending on the current user: hiding/showing the field, showing different select options, etc.

These are some relevant issues, marked https://github.com/thoughtbot/administrate/labels/fields-context. Some of these refer directly to the lack of context, others are a bit more indirect but still related:

- https://github.com/thoughtbot/administrate/issues/1342
- https://github.com/thoughtbot/administrate/issues/1586
- https://github.com/thoughtbot/administrate/issues/1862
- https://github.com/thoughtbot/administrate/issues/1988
- https://github.com/thoughtbot/administrate/issues/2060

My initial intuition would be to pass the current controller down to fields, so that it can be used as provider of context. This can be passed in turn to the call-able that some options take, as well to new options such as `hide_if` or something like that.

Issue #2060, opened 2021-11-14
valid_actions? doesn't work for has_many relations


getaaron:
* What were you trying to do?

I have a handful of `Role`s in my db like `Administrator`, `Author`, etc. A `User` `has_many` `Role`s.

I don't want users of the admin page to be able to destroy or edit roles, just view them. So I followed the instructions on https://github.com/thoughtbot/administrate/blob/main/docs/customizing_controller_actions.md#customizing-actions and added this to the roles controller:

```
  class RolesController < Admin::ApplicationController
    def valid_action?(name, resource = resource_class)
      %w[edit destroy].exclude?(name.to_s) && super
    end
```

I expected this to remove the edit and destroy buttons.

* What did you end up with (logs, or, even better, example apps are great!)?

Indeed, on the Roles page, the edit button is now gone.

However, on the Users dashboard page, I can still see Edit and Destroy buttons:

<img width="1022" alt="image" src="https://user-images.githubusercontent.com/789577/141662693-b725e38f-8407-4b43-b504-a89212a3f1a2.png">

And they work, which unfortunately I think means that both:

1. the view code needs to be updated for the has_many table
2. the controller is not validating `valid_action?` — it's a frontend check only, which is a security concern

* What versions are you running?
  - Rails 5.0.7.2
  - administrate 9e462f7179f47151d1b91aab59291e6884870dab (latest on `main`, also reproduces with 0.16.0 release)

DucMirack:
Hello !

I allow myself to reopen the subject as the problem still exists here.

We are in a similar case as @getaaron , we have a `Field::HasMany` in a show view which allow users to do action they are not supposed to be allowed to.

The possibility for an object to realize an action is conditioned by `#authorized_action?` which is a defined in a controller. But we need to handle 2 different resource types in the `#authorized_action?` of 1 controller.

The second resource (the one of the `Field::HasMany`) should have it's authorized actions handled in it's controller and not on it's parent one.

To put this in code, we have : 
```
class UserDashboard < Administrate::DashboardBase
  ATTRIBUTES_TYPES = {
     ...
     roles: Field::HasMany,
     ...
  }
end

class UsersController
  ...
  def authorized_action?(resource, action)
    if resource.is_a?(User)
      ...
    elsif resource.is_a?(Role)
      ...
    end
  end
end
```

Is there any workaround for this behavior ? 

Let them know if this isn't clear enough, have a nice day :)

pablobm:
[...]

Issue #1862, opened 2021-01-09
Restricting specific fields visible in show/edit


sedubois:
* What were you trying to do?

The authorization/pundit system allows to restrict which resources an admin can view/edit, but how to authorize them to only view/edit specific fields within specific records? The [authorization doc](https://administrate-demo.herokuapp.com/authorization) does not seem to cover this.

* What did you end up with (logs, or, even better, example apps are great!)?

I did not find a way to do this.

* What versions are you running?
  - Rails 6..1.1
  - administrate 0.14.0

pablobm:
Have a look at the documentation for [Pundit](https://github.com/varvet/pundit). A policy object typically includes the current user and the resource to be authorized. In Administrate, we have an `ApplicationPolicy` superclass, which stores both as `@user` and `@record`:

https://github.com/thoughtbot/administrate/blob/c40c317678ca3e0b140d09deb81842948e498443/spec/example_app/app/policies/application_policy.rb#L4-L7

Then these can be used in the policy methods. Here's an example from the example app:

https://github.com/thoughtbot/administrate/blob/c40c317678ca3e0b140d09deb81842948e498443/spec/example_app/app/policies/order_policy.rb#L16-L18

Does this answer your question? Perhaps we could improve the documentation on this point.

pablobm:
Oh, wait. You say fields, not records! Apologies, I see now...

I don't think there's a way to do it at the moment, no :thinking:

pablobm:
Thinking a bit about this, perhaps a way to achieve it would be to use different dashboards for different users. Say that you have a `User` model, and only admins can access some information. The `UserDashboard` could have limited fields, whereas another ActiveRecord model, `AdministrableUser` (or something like that) would inherit from `User` and have its own dashboard, while only accessible by admins.

I haven't tried this though, and it would quickly get out of hand if there are many fields to authorize across many models. However, I think we should try out these solutions first before complicating Administrate's internal APIs in order to add more options that will rarely be used. Ultimately we may implement new APIs, but we'll benefit from trying other things first.

Would this solution work for you?

sedubois:
Thanks @pablobm for sharing your thoughts on this. In practice I probably would not be willing to make several copies of the same dashboard as it likely would quickly get out of hand and lead to more issues than it solves.

Here are some use cases I had in mind:
- an editor should be able to view/modify most fields in a blog post, but only an admin should be able to change its publication status
[...]

PR #2457, opened 2023-11-10
Add context to Page

goosys:
I have some proposal regarding Issue #2363.
I believe this PR can still be improved, so if anyone has any opinions, please review and let me know.

---

## Administrate::Page#context

By keeping the context in the Page as follows,

```ruby
page = Administrate::Page::Form.new(dashboard, requested_resource)
page.context = self # -> current controller
```

We can now define it as follows in the Dashboard.

```ruby
ATTRIBUTE_TYPES = {
  catetory: Field::BelongsTo.with_options(
      scope: -> (field) { field.page.context.current_user.organization.categories }
  ),
```

With this, we can now differentiate the BelongsTo options based on the current user's permissions.

Also, by using context in form_attributes like this, we can now differentiate form elements based on the current user's permissions.

```ruby
def form_attributes(action = nil, context = nil)
  if ["new", "create"].include?(action.to_s) && context.try(:pundit_user).try(:admin?)
    super
  else
    super - [:customer]
  end
end
```

## Administrate::ApplicationController#contextualize_resource

I've prepared a hook point to add context to the resource.
This allows us to automatically set the current_user(pundit_user) to the customer form element that was omitted in the previous section.

```ruby
def contextualize_resource(resource)
  if ["new", "create"].include?(action_name) && !pundit_user.admin?
    resource.customer = pundit_user
  end
end
```

Also, we can change the context here for models where the content of the validation changes under certain conditions.

```ruby
class User < ApplicationRecord
  attr_accessor :inviting_by_admin
  validate :validate_invitation_code_matching, on: :create, unless: -> { inviting_by_admin.present? }
end

def contextualize_resource(resource)
  resource.inviting_by_admin = true if action_name == "create"
end
```

---

How do you think about?


goosys:
Is it better to create Administrate::Page::Context because the contents of the context are too free and it's concerning?

app/controllers/administrate/application_controller.rb
```diff
@@ -283,6 +300,19 @@ def authorize_resource(resource)
...
+        authorize_resource(resource)
+      end
+    end
+    helper_method :authorized_new_resource
+
+    # Override this if you want to contextualize the resource differently.
+    #
+    # @param resource [ActiveRecord::Base] A resource to be contextualized.
+    # @return nothing
+    def contextualize_resource(resource); end
```
pablobm, line 314: I'm not sure that I understand this method. It's intended to make changes on the resource, but the context seems better suited to live in the `Page` object (which has access to the resource anyway). I'm not sure I would want to alter the model in this way.

Would you be able to explain this a bit more? 🙂

app/controllers/administrate/application_controller.rb
```diff
@@ -5,11 +5,13 @@ class ApplicationController < ActionController::Base
...
       authorize_resource(resource_class)
       search_term = params[:search].to_s.strip
-      resources = filter_resources(scoped_resource, search_term: search_term)
+      authorized_scope = authorize_scope(scoped_resource)
+      resources = filter_resources(authorized_scope, search_term: search_term)
       resources = apply_collection_includes(resources)
       resources = order.apply(resources)
       resources = paginate_resources(resources)
       page = Administrate::Page::Collection.new(dashboard, order: order)
+      page.context = self
```
pablobm, line 14: Following on my comment about `contextualize_resource`, I wonder if this is what it should be doing instead? Acting on the page instead of the resource here:

```suggestion
      contextualize_page(page)
```

pablobm (commented):
Apologies, it's been 4 months 😓 I'm starting to look into this now. I had a first quick look today, but couldn't quite get everything yet. I thought perhaps I can start asking questions and that way I'll force myself to be on top of this, if you are still available?

Thank you for the PR, by the way!

app/controllers/administrate/application_controller.rb
goosys: Thank you for your review!! I have been waiting for your response for a long time.

I would like to contextualize the `Page` object, but I also intend to contextualize the resource in the Controller. By contextualizing the resource, we can use it to branch validations or trigger callbacks.
I believe that `contextualize_resource` should be separate from the `Page` 's context, as it serves a similar function to `scoped_resources` and `authorize_resource` .

app/controllers/administrate/application_controller.rb
```diff
@@ -62,8 +72,9 @@ def update
           notice: translate_with_resource("update.success"),
         )
       else
+        page = Administrate::Page::Form.new(dashboard, requested_resource)
```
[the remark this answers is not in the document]
goosys: Ah, you're right. Sorry.

app/controllers/administrate/application_controller.rb
goosys: I think the Controller object is almost essential as a Context for the Page, so how about passing it to the initializer? Like the following:

```ruby
page = Administrate::Page::Collection.new(dashboard, context: self, order: order)
or
page = Administrate::Page::Collection.new(dashboard, controller: self, order: order)
```

app/controllers/administrate/application_controller.rb
```diff
@@ -186,7 +197,11 @@ def requested_resource
     end
 
     def find_resource(param)
-      scoped_resource.find(param)
+      authorize_scope(scoped_resource).find(param)
+    end
+
+    def authorize_scope(scope)
```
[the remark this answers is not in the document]
goosys: Yes, you're right, this change is off-topic. I appreciate your accurate understanding of my intention. 
However, I thought this change was also necessary for effective use of the `contextualize_resource` method, so I included it in the same PR. If necessary, I can split the PR.

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -232,5 +232,65 @@ def send_request(order:)
...
+        expect { get :new }.not_to raise_error
+      end
+    end
+
+    describe "POST create" do
+      def send_request
+        post(
+          :create,
+          params: {
+            order: attributes_for(:order, customer: nil, customer_id: user.id),
```
pablobm, line 284: Why not do this with the customer object? I feel it would be more natural. I had to read this a couple of times and check the factory to see what was going on:

```suggestion
            order: attributes_for(:order, customer: user),
```

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -232,5 +232,65 @@ def send_request(order:)
...
+          },
+        )
+      end
```
pablobm, line 287: I would inline the method into the example, as it's only used once.

app/controllers/administrate/application_controller.rb
pablobm: OK, I think I see now what is going on 👍 

When contextualizing, we can hide fields and then put default values on them. In your example you do this to use the `customer` field from non-admins, then set its value to the current user on new/create.

I'm not convinced about the following:

- Doing this as part of `authorize_resource`, which is a separate concern.
- Doing this only when `resource.is_a?(ActiveRecord::Base)`, which seems artificial. OK, it's not like we have currently support for anything other than ActiveRecord, but this explicit check seems strange to me.
- Having this new API instead of the existing `new_resource` and `requested_resource` instead. It's two separate steps, but it uses existing, known APIs. I have tried locally and it appears to work (I'm having issues running specs so I can't be 100% sure).

pablobm (commented):
I had another look 🙂 While we are doing this, would you be able to rebase this in top of master, please? It will help me run the specs, as the way webdrivers are loaded has changed since the PR was created.

app/controllers/administrate/application_controller.rb
goosys: Thank you for your review.

I have separated `contextize_resource` from `authorize_resource`. As a result, the resource no longer passes as a model class from the index, which also resolves the `ActiveRecord::Base` issue.

I believe the current form of `requested_resource` is ideal. What do you think? I would like to introduce the same form for `new` and `create`. Do you have any good suggestions on how to do this?

For example, what do you think about this form?

```ruby
def built_resource(params: {})
  @built_resource = new_resource(params).tap do |resource|
    authorize_resource(resource)
    contextualize_resource(resource)
  end
end
```

spec/controllers/admin/orders_controller_spec.rb
goosys: Since this is a request parameter, I was unable to format it as `customer: user`.
However, this spec had many mistakes overall, so I have made corrections.

app/controllers/administrate/application_controller.rb
pablobm: Initially that makes sense, but then the developers wouldn't be able to alter the context easily with a hook like `contextualize_page`. Having said that, perhaps there isn't a use case for that...? I have no idea yet. I'm experimenting now to see the possibilities of your PR.

pablobm:
As mentioned in comments above, I have created a branch that uses your code and I'm experimenting with it. It's at https://github.com/thoughtbot/administrate/compare/main...pablobm:use-cases-for-contextualize. My goals are:

- Work with your proposal so see how comfortable I am with the decisions.
- See how it can address the issues listed at https://github.com/thoughtbot/administrate/issues/2363, and what changes would be necessary in order to tackle all of them.

app/controllers/administrate/application_controller.rb
goosys: Shifting slightly to a different topic, I feel that the current action methods in the Administrate's ApplicationController are quite fat, which I believe limits the customizability for us, the application developers.
Therefore, it would be beneficial if these methods were broken down further, allowing us to override specific methods during application development.

I believe it would be ideal if we could proceed with method division and achieve a more unified form as shown below.
(It might be beneficial to also divide `page = Administrate::Page::Form.new` and the part where context is injected into `Page` into separate methods.)

```ruby
    def new
      page = Administrate::Page::Form.new(dashboard, built_resource)
      render ...

    def edit
      page = Administrate::Page::Form.new(dashboard, requested_resource)
      render ...

    def create
      if built_resource(resource_params).save
      ...

    def update
      if requested_resource.update(resource_params)
      ...

    def built_resource(params: {})
      @built_resource ||= new_resource(params).tap do |resource|
        authorize_resource(resource)
        contextualize_resource(resource)
      end
    end

    def requested_resource
      @requested_resource ||= find_resource(params[:id]).tap do |resource|
        authorize_resource(resource)
        contextualize_resource(resource)
      end
    end
```

 What do you think?

goosys: I think the other two commits are good ideas. Thank you.

goosys: The reason I prepared built_resource in addition to new_resource is because I thought it might affect the following process that is used as a helper_method from the view. Do you foresee any issues with this?
https://github.com/search?q=repo%3Athoughtbot%2Fadministrate+new_resource+language%3AHTML%2BERB&type=code&l=HTML%2BERB

pablobm:
I think this is the way to go, but there are some things still to polish. Let's see if I can make sense.

The main issue is https://github.com/thoughtbot/administrate/issues/2363. This lists several other issues that require context. Therefore when we look for a solution we should find something that can tackle all the presented use cases, or at least enable a path towards tackling them in the future. Let's look at them one by one.

### https://github.com/thoughtbot/administrate/issues/1342 How to scope dropdown options to current_user in forms

I think this is addressed here, but the issue covers only part of the story. The proposed change tackles `BelongsTo`, but this should also extend to `HasMany` and `HasOne`. In fact, now I'm thinking it should extend to all field types, as I can envision similar issues with `Select` and possibly others.

### https://github.com/thoughtbot/administrate/issues/1586 Support for virtual fields (regression)

I'm not sure this one is actually related :sweat_smile: In any case, if all fields can get a context, then this one will benefit too.

### https://github.com/thoughtbot/administrate/issues/1862 Restricting specific fields visible in show/edit

There are several things here:
1. Show different attributes depending on context: this works for forms, but I think not for `show` and `index`. Let's add that too.
2. Ability to set certain fields depending on the context: I think this changes enables this behaviour by passing the context to `permitted_attributes`.
3. Same as 1, but for associations. I'm having a look and that might be tricky though. Not sure.

### https://github.com/thoughtbot/administrate/issues/1988 Filtering Field::HasMany

Already mentioned.

### https://github.com/thoughtbot/administrate/issues/2060 valid_actions? doesn't work for has_many relations

I think the context on `permitted_attributes` would help here too...?

Please let me know what you think!

goosys:
> **#1342 How to scope dropdown options to current_user in forms**
> **#1988 Filtering Field::HasMany**

While it is possible to use the `scope` in Lambda format that [`administrate-field-scoped_belongs_to`](https://github.com/XPBytes/administrate-field-scoped_belongs_to) and [`administrate-field-scoped_has_many`](https://github.com/XPBytes/administrate-field-scoped_has_many) already introduce, it would be great to have the `scope` option available in `Field::BelongsTo` and `Field::HasMany` as well, considering the high demand for this feature.
`Field::Select` already has a `collection` option in Lambda format, so I believe it can still be used as-is even with the introduction of context.
Personally, I would be happy to see similar support for [`administrate-field-nested_has_many`](https://github.com/nickcharlton/administrate-field-nested_has_many) as well.

> **#1862 Restricting specific fields visible in show/edit**

I have another suggestion regarding this.

In the edit screen, there are cases where users don't have permission to edit certain data but still need to see it.
For example, a BelongsTo field might be set by higher-level administrators and cannot be edited by regular administrators. However, not displaying this field in the edit screen would negatively affect usability. In the demo application, if we set the `Customer` field in the `/orders/:id/edit` screen to be hidden for regular administrators, and it disappears from the edit screen, users will not understand what they are trying to edit.

Do you have any ideas for mixing the `_form` partial and the `_show` partial in the edit screen? Of course, I would like to utilize context for this as well.

> **#2060 valid_actions? doesn't work for has_many relations**

Isn't this the same issue as specific fields restrictions?
I think it would be sufficient to just add context to `permitted_attributes`, but it would be nice to have a more concise way to describe it.
Could Pundit's permitted_attributes_for_#{action} be a useful reference?
https://github.com/varvet/pundit/blob/main/lib/pundit/authorization.rb#L128

pablobm:
I think we are in agreement here:
- Yes to adding `scope` to all associative fields, as well as context to `collection` in `Select`.
- Very good point about showing read-only fields in forms.
    - I wonder if this could be inferred from the `permit_attributes_for_#{action}` that you propose.
    - So for example: `Order#customer` is listed in `FORM_ATTRIBUTES`, but not allowed in `permit_attributes_for_edit`, so Administrate knows to show it read-only. Would that work? Would that be too complex or clever? I'll let you explore :-)
- And indeed, `permit_attributes_for_#{action}` might be good for us here. I'll let you explore this one too!

goosys:
@pablobm Is there anything else I can help with moving forward? If the scope of this PR is decided, I'm ready to commit to it until the end.

pablobm:
@goosys - I think you are good to go, thank you! :+1: If it helps, you can let me know as you go and I can keep try keep an eye on your pushes to see how you are doing.

goosys:
@pablobm I've created a PR for virtual fields. Please review it when you have time.

- #2658

goosys:
@pablobm 
Apologies, I misunderstood in my previous comment. I noticed that `HasMany` also didn't have a callable scope option, so I added it.
For now, I copied the same implementation from other associative fields, but I feel that this is insufficient.

Since field is passed to the scope, the amount of code required to reach the context-aware resource is quite large.
For example, like this: 

```ruby
scope: -> (field) { field.page.resource.published_entries }
```

I would like to simplify it to something like the following: 

```ruby
scope: :published_entries
```

What do you think?

goosys:
I’ve been testing this branch in my project, and I’m starting to feel that using only the Page context isn’t enough.
There were many cases where I wanted to adjust behavior based on the context in both the Dashboard and the Field. So, I’ve made an additional commit as an experiment.
https://github.com/thoughtbot/administrate/pull/2457/commits/753e2244aca0c16b38e2702abd3ef4cf2039326b

What do you think?

goosys:
I rebased after #2658 was merged.

goosys:
I have rebased again.  

@pablobm  
Since some time has passed, I'd like to take a moment to review and organize this PR.  
First, here’s what this PR accomplishes:  

- Page, Dashboard, and Field contextualization  
  - Page  
    - Allows for conditional validation and similar logic  
  - Dashboard  
    - Enables dynamic FORM_ATTRIBUTES based on the user  
  - Field  
    - Allows dropdowns and other elements to be dynamic  

As a result, I believe the status of the issues in #2363 is as follows.  
Does this align with your understanding?  

- #1342  
  - ◎ Resolved  
- #1586  
  - ◎ Resolved in a separate PR #2658  
- #1862  
  - △ Partially resolved  
- #1988  
  - ◎ Resolved  
- #2060  
  - -- Not yet resolved  

For #1862 and #2060, we came up with new ideas during discussions.  
Would it make sense to address them in the next PR instead?  
Can we consider this PR complete once the contextualization is fully in place?

goosys:
After this, I still have a few tasks I’d like to work on:  

- ApplicationController refactoring  
  - Standardizing slight variations in formatting across different actions  
    - https://github.com/thoughtbot/administrate/pull/2457#discussion_r1594329750  
- Supporting non-HTML responses in ApplicationController  
  - Adjusting actions to work seamlessly for JSON API responses and CSV downloads  
- Introducing callbacks at appropriate points in ApplicationController  
  - Previously, a hook was added only for successful create actions, but I found that approach inconsistent.  
  - I’d like to implement a unified callback system across all actions.

goosys:
@nickcharlton 
Could you also share the milestone for this? It's very important for my application as well. If there's anything else needed, I'm happy to help.

app/controllers/administrate/application_controller.rb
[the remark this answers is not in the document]
pablobm: Unresolving this as I think it's not settled. Currently I think this should be separated into a different PR. Perhaps in the context of your work on https://github.com/thoughtbot/administrate/issues/2814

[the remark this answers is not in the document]
goosys: I’ve removed `authorize_scope` from this PR.  
I’m planning to revisit it either in a separate PR or during a future refactor.

pablobm:
@goosys - Would you be able to rebase?

goosys:
@pablobm 
Rebased.

The following pages can be verified using `bin/dev`:

- Access restrictions for viewing data under a regular user role
  - `/admin/customers`
    - admin: all customers are visible
    - user: only the current user is visible
  - `/admin/orders`
    - admin: all orders are visible
    - user: only the current user’s orders are visible
- Role-based form behavior
  - `/admin/orders/new`
    - admin: can select a customer when creating a new order
    - user: cannot select a customer when creating a new order; the current user is automatically assigned on save

app/controllers/administrate/application_controller.rb
```diff
@@ -261,6 +289,13 @@ def new_resource(params = {})
     end
     helper_method :new_resource
 
+    # Override this if you want to authorize the resource differently.
+    # This will be used to authorize the resource for the all actions without `index`.
+    # In the case of `index`, it is used to authorize the resource class.
+    #
+    # @param resource [ActiveRecord::Base]
+    # @return [ActiveRecord::Base]
+    # @raise [Administrate::NotAuthorizedError] if the resource is not authorized.
```
pablobm:
```suggestion
    # Override to add authorization logic.
    # On the `index` action, this will receive the resource class.
    # On any other action, it will receive the resource instance.
    #
    # @param resource [ActiveRecord::Base, Class]
    # @return [ActiveRecord::Base, Class]
    # @raise [Administrate::NotAuthorizedError] If the resource is not authorized.
```

pablobm (commented):
I'm playing with this when I have time, trying to use it in applications and see how it feels. Today I came across an example that I'm not sure how to best handle.

Say that `Machine.has_many :pieces`. But some pieces are `restricted` and only authorized users can add them to machines. How would you handle that? Both for `create` and `update`.

pablobm (commented):
Apologies, what I said earlier is not really about this PR 😓 I was testing things and ended up going an unrelated rabbit hole.

The inline comments are still valid though.

goosys:
@pablobm 
Thanks for taking a look!

I don’t think we currently support applying restrictions to individual `pieces` within a `has_many` association.

At the moment, here’s what we can do:
- Filter `pieces` using scopes for viewing index/show/edit
- Control whether `pieces` are shown in the new/edit forms
- Prevent editing of unauthorized `pieces` by raising validation errors using a validation context

pablobm:
I've been kicking the tyres of this a bit more. As usual, thank you so much for your patience. Here's what I have now.

I still think we shouldn't have `contextualize_resource`. If we want to add the current user to the resource, I think that should happen in `new_resource`/`requested_resource`. If we want it for authorisation reasons, I think it should be done as part of `authorize_resource`. I think adding a new method `contextualize_resource` add confusion, at least at the moment, when there are existing methods that are good for this already.

I'm not sure that we need `Page#dashboard_context`. What's it for? I think it's not being used. When I remove it, specs are green: https://gist.github.com/pablobm/185e9db2fe212ea27dc7e31c9911cf06

I wonder what other options should receive the context. For now I think it's fine if it's only those that already could take a `Proc`. In the future, we may want to involve `:order` and things like that.
- The one place where I think should be added now is in `lib/administrate/field/polymorphic.rb`, for the option `:classes`. It currently can handle a callable, but it's not passing anything in. Could you please add it there?

Regarding things like changing which attributes appear in the index/show/form pages, or whether they appear in field or "show" form, I think that is out of scope here. We need a separate effort to define a new way to declare dashboards, which can give the necessary information. Something more powerful than the current `FORM_ATTRIBUTES`/etc.
- Right now, technically it could be done by overriding templates...? It's a lot of effort though. Let's see if we can do something better. But not in this PR anyway.

Finally, I think this doesn't work well with virtual fields. Specifically, I tried with a `getter` and it doesn't work because the context is added after the field is initialised. By then, the value has already been read.
- I was able to put together some code to delay when data is read, which fixes the issue. See https://gist.github.com/pablobm/3c01817202ae10c707150d1f35b17d62. It actually sounds like something we have should have done long time ago. I wonder if @nickcharlton has any thoughts on this?

After you make any changes, would you be able to rebase? It helps making sure we are not introducing incompatibilities.

goosys:
Thank you as well for taking such a detailed look -- it’s really helping refine this PR.

I still think we need something in addition to `authorize_resource`, because `authorize_resource` alone is not enough.

For example, there are cases where we need to add or skip special handling only when an action is performed through the admin interface, like this.

```ruby
new_resource.tap do |resource|
  authorize_resource(resource)
  resource.skip_user_invitation_notification!
  resource.set_audit_log_context(self)
  resource.last_updated_from = request.remote_ip
  resource.last_updated_by = current_user
  # etc...
end
```

That said, I’m also starting to wonder whether naming it `contextualize_` is too restrictive and may not be quite right either.

What about making it a callback instead?
That way, application developers could add as many custom-named hooks as they want. And if `authorize_resource` is not needed, they could also choose to remove it freely with `skip_callback`.

```ruby
module Administrate
  class ApplicationController < ActionController::Base
    define_callbacks :new_resource

    set_callback :new_resource, :after, :authorize_resource

    def new
      @resource = resource = run_callbacks :new_resource do
        new_resource
      end
    end
  end
end

class ApplicationController < Administrate::ActionController
  set_callback :new_resource, :after, :contextualize_resource
  set_callback :new_resource, :after, :auditize_resource
  # skip_callback :new_resource, :after, :authorize_resource # if they do not need it

  def contextualize_resource(resource)
    # Implement this however they like
  end

  def auditize_resource(resource)
    # Implement this however they like
  end
end
```

What do you think? Do you not prefer using callbacks? @pablobm @nickcharlton

goosys:
> I'm not sure that we need Page#dashboard_context.  
> Something more powerful than the current FORM_ATTRIBUTES/etc.

That makes sense. I’ll remove the method `dashboard_context=`.

> I wonder what other options should receive the context.

I haven’t thought very deeply about that yet. Are there any others you have in mind?

> polymorphic, classes

OK, I’ll add that.

> virtual fields

Thank you! I’ll take a look.

goosys:
@pablobm 

> I wonder what other options should receive the context.

Would it make sense to handle option fetching in a unified way, so that any option can support callables?

```ruby
module Administrate
  module Field
    class Base
      def fetch_option(name, default = nil)
        val = options.fetch(name, default)
        val.respond_to?(:call) ? val.call(self) : val
      end
```

lib/administrate/field/has_many.rb
```diff
@@ -106,12 +108,11 @@ def includes
...
 
       def candidate_resources
-        if options.key?(:includes)
-          includes = options.fetch(:includes)
-          associated_class.includes(*includes).all
-        else
-          associated_class.all
-        end
+        scope = options[:scope] ? options[:scope].call(self) : associated_class.all
+        scope = scope.includes(*options.fetch(:includes)) if options.key?(:includes)
```
pablobm: I was fiddling with this and I think we can make this work without a conditional:

```suggestion
        scope = scope.includes(options.fetch(:includes, []))
```

spec/controllers/admin/customers_controller_spec.rb
pablobm, line 1: Why the changes to this file? They seem unrelated.

spec/controllers/admin/application_controller_spec.rb
pablobm, line 1: Not sure about this, it feels very brittle. I'd rather that we tested external behaviour when possible.

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -47,6 +47,22 @@
       expect(locals[:resources].map(&:id)).to eq customers.map(&:id).sort
     end
 
+    context "when the user is an admin" do
+      controller(Admin::CustomersController) do
+        def authenticate_admin
+          @current_user = Customer.last
```
pablobm: This is unreliable to get the "last" customer, which you expect to be the `customer = create(:customer)` but won't always be because order is not guaranteed by SQL unless there's an explicit `ORDER BY`. This test may fail randomly.

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -34,8 +34,8 @@ def pundit_user
     end
 
     describe "GET new" do
-      it "raises a Pundit error" do
-        expect { get :new }.to raise_error(Pundit::NotAuthorizedError)
+      it "allows me to new my records" do
```
pablobm:
```suggestion
      it "allows access to /new" do
```

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -34,8 +34,8 @@ def pundit_user
...
+        expect { get :new }.not_to raise_error
```
pablobm, line 38: Something that I'm not sure about with this is that it removes a place where this type of authorization error is tested. Although the specs are a mess anyway already so things like this are difficult to avoid.

Perhaps a case for `Pundit::NotAuthorizedError` could be created somewhere else? For example test `payments/new`.

Also, the context setup at the top (`context "with namespaced Punditize concern" do`) can now lose the `include` and the `def policy_namespace` as I don't think they affect the test any more. (I think this was the case with the `include` already).

lib/administrate/page/base.rb
```diff
@@ -27,11 +27,19 @@ def item_associations
...
       def attribute_field(dashboard, resource, attribute_name, page)
         field = dashboard.attribute_type_for(attribute_name)
-        field.new(attribute_name, nil, page, resource: resource)
+        field.new(attribute_name, nil, page, resource: resource).tap do |f|
+          f.context = context
+        end
+      end
+
+      def get_attribute_value(resource, attribute_name)
+        resource.public_send(attribute_name)
```
pablobm: Is this method used for anything?

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -106,5 +106,34 @@ def send_request(order:)
...
+    controller(Admin::OrdersController) do
+      def pundit_user
+        Customer.find_by(name: "Current User")
+      end
+    end
+
+    let!(:user) { create(:customer, name: "Current User") }
+
+    describe "GET new" do
+      it "allows me to new records" do
```
pablobm:
```suggestion
      it "allows access to /new" do
```

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -106,5 +106,34 @@ def send_request(order:)
...
+        expect { get :new }.not_to raise_error
+      end
+    end
+
+    describe "POST create" do
+      it "allows me to create my records with current customer" do
```
pablobm:
```suggestion
      it "allows creating records with the current customer" do
```

spec/dashboards/order_dashboard_spec.rb
```diff
@@ -8,10 +8,74 @@
...
+          dashboard = OrderDashboard.new
+          dashboard.context = ctx_with_admin_user
+          expect(
+            dashboard.permitted_attributes("edit")
+          ).not_to include("customer_id")
+          expect(
+            dashboard.permitted_attributes("update")
+          ).not_to include("customer_id")
+        end
+      end
```
pablobm: The nesting makes it a bit more difficult to follow. I think we can simplify into a flat structure:

```suggestion
    let(:ctx_with_non_admin_user) do
      Struct.new(:pundit_user).new(Struct.new(:admin?).new(false))
    end

    let(:ctx_with_admin_user) do
      Struct.new(:pundit_user).new(Struct.new(:admin?).new(true))
    end

    it "does not allow non-admins to change customer_id" do
      dashboard = OrderDashboard.new
      dashboard.context = ctx_with_non_admin_user
      expect(
        dashboard.permitted_attributes("new")
      ).not_to include("customer_id")
      expect(
        dashboard.permitted_attributes("create")
      ).not_to include("customer_id")
      expect(
        dashboard.permitted_attributes("edit")
      ).not_to include("customer_id")
      expect(
        dashboard.permitted_attributes("update")
      ).not_to include("customer_id")
    end

    it "allow admins to change customer_id" do
      dashboard = OrderDashboard.new
      dashboard.context = ctx_with_admin_user
      expect(
        dashboard.permitted_attributes("new")
      ).to include("customer_id")
      expect(
        dashboard.permitted_attributes("create")
      ).to include("customer_id")
      expect(
        dashboard.permitted_attributes("edit")
      ).not_to include("customer_id")
      expect(
        dashboard.permitted_attributes("update")
      ).not_to include("customer_id")
    end
```

spec/example_app/app/dashboards/order_dashboard.rb
```diff
@@ -70,4 +70,12 @@ class OrderDashboard < Administrate::BaseDashboard
       "details" => %i[line_items total_price shipped_at payments]
     )
     .freeze
+
+  def form_attributes(action = nil)
+    if %w[new create].include?(action.to_s) && context.try(:pundit_user).try(:admin?)
+      super
+    else
+      super.dup.tap { |attrs| attrs[""] = attrs[""] - %i[customer] }.delete_if { |_k, v| v.blank? }
```
pablobm: I had to read a few times to understand this. No need to make it so clever. How about just this?


```suggestion
      attrs = super.dup
      attrs.delete("")
      attrs
```

Or if we are going to make it more generic (which isn't really needed at the moment), we can do this:


```suggestion
      super.dup
        .transform_values { |v| v.without(:customer) }
        .delete_if { |_k, v| v.blank? }
```

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -260,7 +260,7 @@
         create_list(:customer, 3)
         options = {
           order: "name",
-          scope: -> { Customer.order(name: :desc) }
+          scope: ->(_field) { Customer.order(name: :desc) }
```
pablobm: I think these instances of adding an argument to the block are not necessary? Which is fortunate because it won't cause back-compatibility issues for anyone who already uses blocks without arguments.



This applies to all other instances, in this file and others.
```suggestion
          scope: -> { Customer.order(name: :desc) }
```

lib/administrate/field/has_many.rb
```diff
@@ -106,12 +108,11 @@ def includes
...
+
+        order = options.delete(:order)
+        order ? scope.reorder(order) : scope
```
pablobm: Good call with `reorder`, but I would put it in a different PR, to simplify this one.

By the way: do we need to `delete`, or can we just read the value?


```suggestion
        order = options[:order]
        order ? scope.reorder(order) : scope
```

spec/lib/fields/belongs_to_spec.rb
```diff
@@ -280,7 +280,7 @@
         order = build(:order)
 
         1.upto(3) { |i| create :customer, name: "customer-#{i}" }
-        scope = -> { Customer.order(name: :desc).limit(2) }
+        scope = ->(_field) { Customer.order(name: :desc).limit(2) }
```
pablobm, line 283: Oh my, I just realised that this example and the previous one are exactly the same... But not a problem of this PR, so don't worry about it 😓

pablobm (commented):
Re: `contextualize_resource` - I still don't understand the need. The example you show could still be implemented as part of `authorize_resource`. Also this PR doesn't put `contextualize_resource` everywhere: why `new`/`create` but not `edit`/`update`?

What you say about callbacks might better as it allows more flexibility. But I would do that as a different PR.

> I wonder what other options should receive the context.

Don't worry for now. The ones that you have covered are good, and more can be added later.

Also I think don't worry about unified code to call the callables. I think it won't buy us much at this point. We can revisit later as we add more.

spec/controllers/admin/orders_controller_spec.rb
goosys: Added

spec/controllers/admin/customers_controller_spec.rb
goosys: Thanks for the review.
Previously, this spec used a policy scope that did not narrow records, so the index results were effectively unchanged in the “all records” case.
Since this PR updates the customer policy scope, I used multiple customer records so any scope-related change in index results becomes visible.

spec/controllers/admin/customers_controller_spec.rb
goosys: Fixed

spec/dashboards/order_dashboard_spec.rb
goosys: I flattened the nesting in this spec

goosys:
@pablobm 
Thank you as always. I really appreciate your patience and all the time you’ve spent reviewing this.
I addressed the review feedback.
I removed contextualize_resource in this PR, but I still need that capability, so I plan to open a follow-up PR with a callback-based approach.

spec/lib/fields/belongs_to_spec.rb
pablobm: In fact, there's even less need to change this line because `belongs_to.rb` didn't change at all.

docs/customizing_controller_actions.md
```diff
@@ -31,8 +31,8 @@ class Admin::FoosController < Admin::ApplicationController
   #   Foo.find_by!(slug: param)
   # end
 
-  # Override this if you have certain roles that require a subset
-  # this will be used to set the records shown on the `index` action.
+  # Override this if you have certain roles that require a subset.
+  # This will be used in all actions except for the `new` and `create` actions
```
pablobm: Same as https://github.com/thoughtbot/administrate/pull/2457#discussion_r2983426417. Create a separate PR, let me know, and I'll merge it straightaway.

spec/controllers/admin/customers_controller_spec.rb
```diff
@@ -47,6 +47,22 @@
       expect(locals[:resources].map(&:id)).to eq customers.map(&:id).sort
     end
 
+    context "when the user is an admin" do
```
pablobm:
```suggestion
    context "when the user is not an admin" do
```

spec/controllers/admin/payments_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
+require "rails_helper"
+
+# Ensure we keep explicit coverage for authorization errors on `new` actions.
+# standard:disable Lint/ConstantDefinitionInBlock
```
pablobm:
```suggestion
```

Leftover from a previous implementation, perhaps before using `controller`?

spec/controllers/admin/payments_controller_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  let!(:user) { create(:customer, name: "Current User") }
+
+  describe "GET new" do
+    it "raises a Pundit error" do
+      expect { get :new }.to raise_error(Pundit::NotAuthorizedError)
+    end
+  end
+end
+# standard:enable Lint/ConstantDefinitionInBlock
```
pablobm:
```suggestion
```

Same as above.

spec/controllers/admin/orders_controller_spec.rb
pablobm: The other problem here is that these changes remove the intent of the context block: testing namespaced policies. The `policy_namespace` doesn't do anything any more in these examples because the default policy (`OrderPolicy`) now behaves in a way much closer to how the namespaced policy (`Own::OrderPolicy`) does.

So either this context needs a new example that tests the different behaviour of the namespaced policy, or a new context testing namespaced policies has to be added elsewhere (eg: payments controller, like it was done with the previous comment).

pablobm: Another problem from this: the new context you added below ("when the user is not an admin") doesn't do anything that this one couldn't do. Eg: the "POST create" block below could work up here too.

spec/dashboards/order_dashboard_spec.rb
pablobm: Thank you. You even corrected a mistake I made in my suggestion :sweat_smile: where the description implied that admins can change the `customer_id` when they can't.

spec/features/orders_form_spec.rb
```diff
@@ -21,22 +21,23 @@
...
     it "displays translated label when translation for the attribute is available" do
       order = create(:order)
-      custom_attribute_name = "Client"
+      custom_attribute_name = "Lines"
       translations = {
         activerecord: {
           attributes: {
             order: {
-              customer: custom_attribute_name
+              line_items: custom_attribute_name
```
pablobm, line 40: Why this change?

pablobm (commented):
Thank you for your patience in turn 🙏 Here's a new set of comments.

pablobm:
Forgot to mention: this also needs adding some documentation.

spec/features/orders_form_spec.rb
goosys: <s>Sorry, I do not remember clearly.

I think this is likely a leftover from one of the changes that had originally been split across several PRs.

If I remember the exact reason, I will open a follow-up PR. </s>

(edited)
I’m very sorry.

I changed this test to use the `line_items` column because whether the `customer` column is shown depends on admin permissions, so it is not guaranteed to be present. The `line_items` column is always present, which makes it a more reliable target for the translation check.

docs/customizing_controller_actions.md
goosys: reverted

spec/controllers/admin/customers_controller_spec.rb
goosys: fixed

spec/controllers/admin/payments_controller_spec.rb
goosys: removed

spec/controllers/admin/payments_controller_spec.rb
goosys: removed

goosys:
@pablobm 
Thanks for the review.
I’ve made the requested changes.
I also added some documentation. How does this look?

pablobm (commented):
Just two things now!

First, another rebase please as there's a merge conflict (plus that issue that breaks rails80 and is solved in main).

Second: do you remember when you had this idea and I said it wasn't necessary? ([Comment](https://github.com/thoughtbot/administrate/pull/2457#issuecomment-4161941103))

> Would it make sense to handle option fetching in a unified way, so that any option can support callables?

So now I'm realising it would be very helpful. I'm worried that the options that receive the context will become backwards-incompatible. So if for example you had this:
```ruby
tags: Field::HasMany.with_options(scope: -> { something })
```
Then now that's going to break and people will need to change it to this:
```ruby
tags: Field::HasMany.with_options(scope: ->(_) { something })
```
Just because there's now a parameter that they are not using.

So I'm thinking that perhaps we should have a unified way to call these so that these calls don't break. Something like this:
```ruby
def fetch_option_value(option_name, *args, &block)
  return block.call unless options.key?(option_name)

  option_value = options.fetch(option_name)
  return option_value unless option_value.respond_to?(:call)

  return option_value.call unless option_value.arity.positive?

  option_value.call(*args)
end
```
This can also be used in the couple of places where we check for `#arity` for the same reason.

goosys:
@pablobm 

Thanks, I agree with this direction.

Since the `fetch_option_value` abstraction would affect multiple option-handling paths and broaden the scope of this PR, I think it would be better to introduce it in a follow-up PR. What do you think?

For this PR, I made the minimum change needed for `HasMany` backward compatibility: scopes with a positive arity receive the field, while scopes that do not accept arguments are called without one. This preserves existing usages such as `scope: -> { ... }` while still supporting context-aware scopes.

I also added a regression spec for argument-less scopes. The rebase is already complete.

pablobm:
I'm ok with doing a broader cleanup in a separate PR. However I think this PR should also do something in `lib/administrate/field/polymorphic.rb`, which is the other place that goes from `call()` to `call(self)` and can break people's code.

goosys:
Sorry for missing that. I've fixed `Polymorphic` as well.
Apologies for the extra round of review, and thank you for taking another look.

pablobm (approved):

pablobm:
OK, so it has taken nearly 3 years! But I think it's time to merge this. Thank you so much for your patience, @goosys :heart:

PR #2473, opened 2023-12-12
Remove CircleCI

nickcharlton:
In #1932, we migrated to GitHub Actions which is generally faster and has a simpler configuration. This completes that work by removing Circle CI.

This commit will fail checks for Circle CI, which after it's merged we can disable in the UI.


pablobm (approved):

sarahraqueld (approved):

nickcharlton:
I'm confident that this is now working after opening a bunch of PRs. I've disabled CircleCI on the project so it doesn't get used for new PRs, but I'm pretty sure this will break existing PRs (which can't have GitHub Actions without recreating) which is annoying.

But so is waiting for CircleCI to finish so that's the way has to be.

PR #2475, opened 2023-12-22
Document custom assets

pablobm:
We are not documenting how to add custom assets to complement or override Administrate's own. In fact I recently came across a client project where this was done the wrong way, causing breakage.

So here's some documentation on it, as well as adds examples to the example app. The examples (particularly the CSS one) are a bit artificial, but I couldn't think of anything else. Suggestions welcome.

### DIFF
--- a/docs/customizing_page_views.md
+++ b/docs/customizing_page_views.md
@@ -119,3 +119,25 @@ For example, you can add a button in the middle of the header as follows:
 
 <%= render template: 'administrate/application/_index_header', locals: local_assigns %>
 ```
+
+## Adding custom CSS and JS
+
+You can add custom CSS and JS to Administrate. Put the files in the
+appropriate folders (typically under `assets`) and point Administrate to them
+using the following API, preferably in an initializer. For example, if your
+files are called `admin.css` and `admin.js`:
+
+```
+/// config/initializers/administrate.rb
+Administrate::Engine.add_stylesheet("admin")
+Administrate::Engine.add_javascript("admin")
+```
+
+Then make sure to list them in your manifest file (Rails will helpfully remind
+you of this step if you miss it):
+
+```
+/// app/assets/config/manifest.js
+//= link admin.css
+//= link admin.js
+```
--- a/spec/example_app/app/assets/stylesheets/admin.css
+++ b/spec/example_app/app/assets/stylesheets/admin.css
@@ -1,0 +1,3 @@
+/*
+*= require_tree ./admin
+*/
--- a/spec/example_app/app/assets/config/manifest.js
+++ b/spec/example_app/app/assets/config/manifest.js
@@ -2,3 +2,5 @@
 
 //= link_tree ../images
 //= link_tree ../builds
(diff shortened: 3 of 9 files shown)


docs/customizing_page_views.md
```diff
@@ -94,3 +94,25 @@ rails generate administrate:views:layout
...
+```
+
+Then make sure to list them in your manifest file (Rails will helpfully remind
+you of this step if you miss it):
+
+```
+/// app/assets/config/manifest.js
+//= link admin.css
+//= link admin.js
+```
```
elias19r, line 118: Looks good, thank you!

@pablobm @nickcharlton I wonder what we should to do with the following generators:
- `lib/generators/administrate/assets/assets_generator.rb`
- `lib/generators/administrate/assets/javascripts_generator.rb`
- `lib/generators/administrate/assets/stylesheets_generator.rb`

They are another way of customizing assets. As far as I understand, by running those generators, Administrate's non-compiled assets are copied to the users' app. So users can edit a non-compiled asset and _compile it themselves_ if their app is set up for that. And then their assets will take precedence over the ones from the Administrate gem.

I lean toward removing those generators and making users customize assets by appending to Administrate's CSS/JS, just like the example in this PR, instead of completely replacing them with their own.

elias19r (approved):

spec/example_app/app/assets/stylesheets/admin.css
```diff
@@ -0,0 +1 @@
+//= require_tree ./admin
```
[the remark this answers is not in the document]
elias19r: Hound's got a point. Should this be

```css
/*
*= require_tree ./admin
*/
```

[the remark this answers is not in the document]
pablobm: Good point. To be honest, as long as it works and Hound shuts up, I'm in :+1:

docs/customizing_page_views.md
pablobm: I agree. I think that this way of customising assets is more likely to cause issues than to solve them. Adding additional, separate assets to add new behaviour (or override CSS) feels like a better solution to me. Then if anyone wants to fully override Administrate's own assets, they can copy them over manually and do that.

nickcharlton: Yeah, that makes sense to me 👍

nickcharlton (approved):
I think it'd be good to remove those generators, but in another PR (I'll get around to it eventually, but if someone else wanted to, go ahead!). And then we can merge this once #2397 goes in.

pablobm:
Forgot about this one after the merge of #2397. I have rebased and made a change (converted SCSS to CSS). Nor merging just yet as I'm not sure if it'll need to wait for https://github.com/thoughtbot/administrate/pull/2673 now.

pablobm:
And related PR to remove old asset generators at https://github.com/thoughtbot/administrate/pull/2674, as discussed above.

Blackskar875 (approved):

PR #2478, opened 2023-12-26
Add resource instance variables and audit log sample

goosys:
I have made changes so that instance variables are stored in both the index and create actions, in order to capture controller operation logs and perform other post-processing tasks.
Please review.


app/controllers/administrate/application_controller.rb
```diff
@@ -9,6 +9,7 @@ def index
       resources = apply_collection_includes(resources)
       resources = order.apply(resources)
       resources = paginate_resources(resources)
+      @resources = resources
```
nickcharlton, line 12: This should be available through the locals (see lines 14-19). What use case do you have where that's not enough?

goosys: For consistency. Since I made @requested_resource and @new_resource as instance variables, I made it so that @resources can also be referenced from instance variables.

nickcharlton:
Ah yeah, I see. Yeah, consistency is important here.

goosys:
rebased

spec/example_app/app/controllers/admin/application_controller.rb
```diff
@@ -32,5 +32,22 @@ def authenticate_admin
     def pundit_user
       @current_user
     end
+
+    after_action :audit_log, only: %i[create update destroy]
+
+    def audit_log
+      if (resource = @requested_resource || @new_resource)
+        Rails.logger.info(
+          sprintf(
```
nickcharlton, line 41: Hah! It's so rare to see the use of a `sprintf`, fun!

spec/example_app/app/controllers/admin/application_controller.rb
```diff
@@ -32,5 +32,22 @@ def authenticate_admin
     def pundit_user
       @current_user
     end
+
+    after_action :audit_log, only: %i[create update destroy]
```
nickcharlton, line 36: Where should I expect this to log? I've run it locally and tried a few things, but I didn't see any log output

goosys: I’m not sure either. For some reason, this demo application isn’t showing any logs in the console. I believe if you specify the log level and output destination in the config, the logs should appear.

nickcharlton: Oh yeah, I see. You don't get logs in `bin/dev`, but you do in `rails s`. That's a pain.

app/controllers/administrate/application_controller.rb
```diff
@@ -26,7 +27,7 @@ def show
     end
 
     def new
-      resource = new_resource
+      @new_resource = resource = new_resource
```
nickcharlton, line 30: @pablobm, where you thinking that this should be `@resource` and not `@new_resource`? I'm not sure if we made any changes here after that comment

pablobm: That's right. Unless there's a reason for it, I think this should be `@resource` for consistency.

app/controllers/administrate/application_controller.rb
```diff
@@ -40,7 +41,7 @@ def edit
     end
 
     def create
-      resource = new_resource(resource_params)
+      @new_resource = resource = new_resource(resource_params)
```
pablobm, line 44: And same thing here:

```suggestion
      @resource = resource = new_resource(resource_params)
```

app/controllers/administrate/application_controller.rb
pablobm: Actually... now I'm torn. I forgot that we have `@requested_resource`, so having `@resource` might be confusing :thinking:

goosys: https://github.com/goosys/administrate/blob/20cf3a7ab273e17312033ca1d3ed2e1c1503623f/app/controllers/concerns/administrate/resource_manager.rb#L24

I was reorganizing `ApplicationController` for a separate task, and I think `@built_resource` would be a good choice.  

It conveys that the resource is not yet saved, making it clearer than simply using `@resource`.

pablobm: This makes me uneasy... :thinking: Increasing the number of instance variables increases complexity. In your `audit_log` for example, having to do `if (resource = @requested_resource || @new_resource)` is a smell. I would prefer to do `if @resource`. The action can be obtained from `action` if further filtering is required.

As mentioned, the presence of `@requested_resource` complicates things a bit... I'm looking at it now and I think that's the problem: a noun-method that should be a verb-method. I think instead we should do this in `update`, `edit`, `destroy`:

```
    def update
      @resource = find_resource(resource_params)
      if @resource
        redirect_to(
          after_resource_updated_path(@resource),
          notice: translate_with_resource("update.success"),
          status: :see_other
        )
      else
        render :edit, locals: {
          page: Administrate::Page::Form.new(dashboard, @resource)
        }, status: :unprocessable_entity
      end
    end
```

Which now relies on verb-method (also requires renaming the pre-existing `find_resource`):

```
    def find_resource
      find_resource_in_scope(params[:id]).tap do |resource|
        authorize_resource(resource)
      end
    end

    def find_resource_in_scope(param)
      scoped_resource.find(param)
    end
```

This is now consistent with `index`,`new`, and `create`, which retrieve the resource with a helper method and put it in an instance variable.

app/controllers/administrate/application_controller.rb
```diff
@@ -9,6 +9,7 @@ def index
       resources = apply_collection_includes(resources)
       resources = order.apply(resources)
       resources = paginate_resources(resources)
+      @resources = resources
       page = Administrate::Page::Collection.new(dashboard, order: order)
```
goosys, line 13: @nickcharlton @pablobm 
If it's for the audit log, I feel like having @page would be nice too. Thoughts?

pablobm: Given how prevalent it is (it appears in all actions that render a page), I think it may make sense, yes :+1: Go for it.

pablobm:
@goosys - Thoughts on this one? Particularly what I say at https://github.com/thoughtbot/administrate/pull/2478#discussion_r2012091707

goosys:
@pablobm 
I agree with the idea of standardizing everything to `@resource`!

nickcharlton:
@goosys, I might be wrong, but it looks like we just need to settle on `@resource` here and then we can look at merging this. Could you take a look?

goosys:
@pablobm @nickcharlton 
I’ve unified the instance variables to `@resource`. How does this look?

The goal here is to make the resource available for post-processing in each action, so I think it’s reasonable to stop at this point for now.

We could consider renaming `requested_resource` separately in another step.

nickcharlton:
Ah yeah, nice. I think that's it! Thanks!

PR #2479, opened 2023-12-28
Use a dedicated ActiveSupport::Deprecation instance

maciek-rr:
#### Description
Rails 7.1 started deprecating the direct usage of `ActiveSupport::Deprecation` methods. Instead, the preferred way is to instantiate the class and use the custom instance.

This PR adds the deprecator instance and switches the warn calls to use it. 
Compatible with Rails 4.0+

#### Deprecation example
Each deprecation originating from Administrate is accompanied by a [Rails 7.1 deprecation](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/deprecation/instance_delegator.rb#L37).
```
DEPRECATION WARNING: The method `valid_action?` is deprecated. Please use `accessible_action?` instead, or see the documentation for other options. (called from index at /backend/app/controllers/concerns/admin_search.rb:13)
DEPRECATION WARNING: Calling warn on ActiveSupport::Deprecation is deprecated and will be removed from Rails (use your own Deprecation object instead) (called from index at /backend/app/controllers/concerns/admin_search.rb:13)
```

### DIFF
--- a/app/controllers/concerns/administrate/punditize.rb
+++ b/app/controllers/concerns/administrate/punditize.rb
@@ -42,7 +42,7 @@ module Administrate
         end
 
         if policy_scope.respond_to? :resolve_admin
-          ActiveSupport::Deprecation.warn(
+          Administrate.deprecator.warn(
             "Pundit policy scope `resolve_admin` method is deprecated. " +
             "Please use a namespaced pundit policy instead.",
           )
--- a/lib/administrate.rb
+++ b/lib/administrate.rb
@@ -2,7 +2,7 @@ require "administrate/engine"
 
 module Administrate
   def self.warn_of_missing_resource_class
-    ActiveSupport::Deprecation.warn(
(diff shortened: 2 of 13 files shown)


nickcharlton:
Nice, thanks!

maciek-rr:
Hey @nickcharlton thanks a ton, I've missed it, will take a better look at it during the weekend.

maciek-rr:
Hey @nickcharlton, I think I've addressed all the remaining deprecation warnings. One type was coming from RSpec and the usage of a deprecated file exists matcher, another one was about using an old Rails cache format, I've modified the test.rb of the example app to use the new format in Rails 7+. Please take a look when you have a moment, thanks.

nickcharlton:
Excellent, thanks! This looks good to me now and I'm not seeing any deprecation warnings now so I'm going to merge it.

PR #2484, opened 2024-01-13
Yield created resource if block is given

trevorrjohn:
Creates an interface to allow you to trigger actions after a resource is created (for example run a background job).

Right now you can do this in the `after_resource_created_path` but it doesn't quite feel right to have a side affect in a method that is supposed to be determining the redirection path.

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -44,6 +44,7 @@ module Administrate
       authorize_resource(resource)
 
       if resource.save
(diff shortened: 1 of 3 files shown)


nickcharlton (commented):
Thanks!

It'd be great if we could add a mention of this to the docs somewhere (it's possible there's not an obvious place), if you could give that a go?

trevorrjohn:
@nickcharlton added a note in the docs.

nickcharlton:
Magic, thanks!

PR #2491, opened 2024-01-22
fix: set require "administrate/version"

JohnnyKei:
We met `uninitialized constant Administrate::VERSION` error after v0.20.0.

**How to reproduce**
Remove `require "administrate/version"` from administrate.gemspec.

```
bundle exec rspec spec/controllers/admin/application_controller_spec.rb:123
```


dorianmarie (approved):
Looks good to me, we are not requiring that file except in gemspec https://github.com/search?q=repo%3Athoughtbot%2Fadministrate%20version&type=code

nickcharlton:
Which version did you upgrade from? Are you doing anything with the `VERSION` constant?

FWIW, that line has been in there for 9 years. I'd not expect that to be a problem because of that version because of that: https://github.com/thoughtbot/administrate/blame/main/administrate.gemspec#L3

sojan-official:
We are also facing this on this pr: https://github.com/chatwoot/chatwoot/pull/8741 

here is the trace 

Screenshot 2024-01-22 at 4 27 53 PM

I think it throws this error when `administrate` renders the deprecation warnings.

nickcharlton:
Ah hah, yes, that all adds up. I think this is the right way to go then.

I can't think of a good way to test this, at least with what structure we have already, so I'm not going to let that hold us back.

nickcharlton:
I've just pushed [v0.20.1](https://github.com/thoughtbot/administrate/releases/tag/v0.20.1) which includes this fix, thanks for reporting and providing a solution!

PR #2492, opened 2024-01-22
Switch to using @thoughtbot/stylelint-config

nickcharlton:
In #2397, we're having trouble with ignoring directories using the existing `scss-lint` configuration on Hound. `scss-lint` has been deprecated for a while and we have another configuration based on thoughtbot guides, so this is a good fit.

We use BEM style selectors in our CSS. So we add a custom rule, based on some prior work by @a-tokyo, but adjusted to fit in our JSON configuration file.

In addition, we disable these rules as they don't make sense on Administrate:

* Disable scss/dollar-variable-empty-line-before on variables
* Disable function-no-unknown/no-global function names

https://github.com/thoughtbot/stylelint-config
https://github.com/sds/scss-lint/commit/4ac76f9575464028447cbc4042d92eb1089d3899
https://github.com/humanmade/coding-standards/issues/193#issuecomment-1405099508
https://github.com/simonsmith/stylelint-selector-bem-pattern/issues/23


nickcharlton:
I need to fix GHA, but also when it does run, there's some linting failures as the rules seem to be slightly different and/or maybe broken. I can't tell.

nickcharlton:
This has been a ten (10) day side quest and I can finally think about sorting out the rules which are now failing.

nickcharlton:
I published `stylelint-config` [v4.0.0](https://github.com/thoughtbot/stylelint-config/releases/tag/v4.0.0), then updated to that here.

Now this is ready to go. Hound keeps trying to lint, ~~even though it's not configured~~ it's configured at the org level and they mustn't merge the configurations, which … explains why it wasn't working before. That's very confusing.

PR #2495, opened 2024-01-25
Fix generator path for namespaced models

sdwolfz:
Running a generator with a model that is namespaced, for example `Foo::Bar` puts the generated files in:
- `app/controllers/admin/bars_controller.rb`
- `app/dashboards/bar_dashboard.rb`

The correct location is:
- `app/controllers/admin/foo/bars_controller.rb`
- `app/dashboards/foo/bar_dashboard.rb`

This change ensures files are generated in the proper location. Without it the directory needs to be created and files need to be moved manually.

I've also added a `docker-compose.yml` file for easy running of specs on local.

### DIFF
--- a/lib/generators/administrate/dashboard/dashboard_generator.rb
+++ b/lib/generators/administrate/dashboard/dashboard_generator.rb
@@ -37,15 +37,18 @@ module Administrate
       source_root File.expand_path("../templates", __FILE__)
 
       def create_dashboard_definition
+        scope = regular_class_path.join("/")
+
         template(
           "dashboard.rb.erb",
-          Rails.root.join("app/dashboards/#{file_name}_dashboard.rb")
+          Rails.root.join("app/dashboards/#{scope}/#{file_name}_dashboard.rb")
         )
       end
 
       def create_resource_controller
+        scope = "#{namespace}/#{regular_class_path.join("/")}"
         destination = Rails.root.join(
-          "app/controllers/#{namespace}/#{file_name.pluralize}_controller.rb"
+          "app/controllers/#{scope}/#{file_name.pluralize}_controller.rb"
         )
 
         template("controller.rb.erb", destination)
--- a/spec/generators/dashboard_generator_spec.rb
+++ b/spec/generators/dashboard_generator_spec.rb
@@ -431,5 +431,28 @@ describe Administrate::Generators::DashboardGenerator, :generator do
       remove_constants :Foo
       Manager.send(:remove_const, :FoosController)
     end
+
+    it "uses the given model namespace to create controllers" do
+      ActiveRecord::Schema.define { create_table :foo_bars }
+      module Foo
+        def self.table_name_prefix
+          "foo_"
+        end
+
+        class Bar < Administrate::Generators::TestRecord; end
+      end
+
+      run_generator ["Foo::Bar"]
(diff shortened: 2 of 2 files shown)


sdwolfz:
@nickcharlton I was hoping CI would kick in and run the spec I added as I don't have a compatible local setup but I guess you need to trigger that manually?

docker-compose.yml
```diff
@@ -0,0 +1,57 @@
+---
```
dorianmarie, line 1: why is there a docker-compose.yml file added?

nickcharlton: > I don't have a compatible local setup

> Edit: added a docker-compose.yml file for easy running of specs on local.

We can worry about it later.

nickcharlton:
@sdwolfz, it's [because the `push` event doesn't support running on forks where the person who's opened the PR doesn't have write access](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target). Thanks for pointing it out, I'd forgotten about that when I sorted out GitHub Actions a few weeks back and will need to try some things out.

dorianmarie:
I think you would be open to bitcoin-mining attacks for instance if you open the build to any user

nickcharlton:
@dorianmarie, the biggest concern — [as outlined in the blog post mentioned on the link I posted above](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) — is really secrets exfiltration. If you were running self-hosted runners (especially so if it were reused VMs) you could likely break out of the running environment and walk up the directory too, but that's perhaps more theoretical. There are better ways to catch/dissuade mining attacks.

@sdwolfz, I rebased your PR once I figured out the GHA event changes, and you should be able to push and have CI run now. Could you take a look at the failures when you get a chance?

docker-compose.yml
sdwolfz: I can drop the commit if you don't need it and just keep that as a snippet for myself for future contributions.

nickcharlton: If you could drop it, that'd be great (otherwise we'd come back to it when ready to merge). It's not something we generally use and maintaining it becomes difficult in that case.

sdwolfz: done

nickcharlton:
Ah, hah, good stuff.

I think breaking out the modules would be a good idea, but probably not in this PR. If you'd be up to it, could you do that in another one?

I just approved the GHA run (it shouldn't have taken me this long, but apparently 3 weeks passed!).

sdwolfz:
Rebased with main, bundler audit failure was because of nokogiri, unrelated to changes.

>I think breaking out the modules would be a good idea, but probably not in this PR. If you'd be up to it, could you do that in another one?

Yes but I can't commit to an ETA.

> it shouldn't have taken me this long, but apparently 3 weeks passed!

I'v done worse...

nickcharlton:
Yeah, I wouldn't expect you to commit to anything, just get to it when you can!

I happened to merge in #2502 the other day, which is also generator related. Does this help at all with your testing?

sdwolfz:
I've never generated views before so I never encountered that.

Will fix minting soon

sdwolfz:
Tests are failing because of https://github.com/thoughtbot/administrate/issues/2523 which is unrelated.

Issue #2503, opened 2024-01-31
Doubtful usage of `<fieldset>` in dashboard show pages


Uaitt:
# Description
With Administrate v0.20.0, now it's possible to visually group together fields of the same model (introduced in [this PR](https://github.com/thoughtbot/administrate/pull/2422)).

While I like the feature, I think that there is a small conceptual problem.

In both the `app/views/administrate/application/_form.html.erb` and `app/view/administrate/application/show.html.erb` a `<fieldset>` element was introduced (indeed, in order to group fields together). 
However, most documentation ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset), [W3School](https://www.w3schools.com/tags/tag_fieldset.asp), [W3](https://www.w3.org/TR/2011/WD-html5-20110405/forms.html#the-fieldset-element) says that this element "groups together element in a form"). With that being said, I really don't think this element should be used in the `show` view (while I totally agree with using it on the `_form` partial). 

I am writing this also because I have noticed a strange problem on one of my current projects after this update to the `<fieldset>` in the `show` view, that occurs when one of the fields contain an horizontal scrollbar.

Before the change:
![Screenshot 2024-01-31 at 14 38 23](https://github.com/thoughtbot/administrate/assets/97030518/78abbe08-9899-4b45-ac5d-a0a8686da4f9)

After the change:
![Screenshot 2024-01-31 at 14 35 59](https://github.com/thoughtbot/administrate/assets/97030518/945b215c-16a8-4d18-af97-6bb525ebbe06)
Layout is completely broken (I had to zoom out a lot in order to see the actual fields). The layout is fixed if the `<fieldset>` element is replaced with a generic container, like a `<div>` or `<section>`.

I would thus propose to replace the `<fieldset>` element in `app/view/administrate/application/show.html.erb` with a simple container (either `<div>` or `<section>`) and style it similar to a `<fieldset>`.

# Versions
- Administrate v0.21.0
- Rails 7.1.3
- Ruby 3.2.2

nickcharlton:
Oh! Yes, I totally agree.

I do think a `<section>` is the right thing semantically. Would be able to contribute a PR to fix this?

Uaitt:
@nickcharlton Yep, I will work on it 😃.

blocknotes:
Hey 👋 
Any update about this issue?

[...]

PR #2504, opened 2024-02-01
Replace <fieldset> with <section> on show pages

Uaitt:
This PR fixes issue https://github.com/thoughtbot/administrate/issues/2503.

Basically, the `<fieldset>`s and `<legend>`s in 
- `app/views/administrate/application/show.html.erb` 
- `app/views/fields/has_one/_show.html.erb` 

have been replaced by a combination of `<section>`s, `<h1>`s and `<nav>`s. Those new elements have then been styled so that they look exactly like the previous two.

A few screenshots down here 👇 

**BEFORE THIS CHANGE**:

Show page of an object with some grouped and ungrouped attributes.
Screenshot 2024-02-01 at 16 57 28

Show page of an object with an `has_one` field and some other attributes.
Screenshot 2024-02-01 at 16 57 35

**AFTER THIS CHANGE**:

Show page of an object with some grouped and ungrouped attributes. (Note the difference in width of the grouped block).
Screenshot 2024-02-01 at 16 55 51

Show page of an object with an `has_one` field and some other attributes. (Note that a bug has been fixed too, there's no more overlapping).
Screenshot 2024-02-01 at 16 55 40

Any feedback is welcomed 😃.


app/views/administrate/application/show.html.erb
```diff
@@ -43,9 +43,13 @@ as well as a link to its edit page.
 <section class="main-content__body">
   <dl>
     <% page.attributes.each do |title, attributes| %>
-      <fieldset class="<%= "field-unit--nested" if title.present? %>">
+      <section class="<%= "field-unit--nested" if title.present? %>">
```
seanpdoyle, line 46: Semantically, `<fieldset>` elements typically wrap interactive form controls. Since the `show` templates present information without the opportunity to edit that information, this change improves the semantics of what these elements are communicating.

Having acknowledged that, rendering multiple `<h1>` elements [poses some accessibility risks](https://www.w3.org/WAI/tutorials/page-structure/headings/), and nesting a `<nav>` element (a role that comes with its own set of semantics and requirements) inside an `<h1>` poses additional risks.

Are there other structural changes we could make to achieve a similar outcome without the risks?

Uaitt: What element (or elements) would you recommend to replace my `<h1>` and `<nav>` combination? Do you recommend something in particular? 🤔

PR #2505, opened 2024-02-02
Stop saying Administrate doesn't have a DSL

### LINKED ISSUE #2267: Misleading claim on readme: administrate does have a DSL.
> No DSLs (domain-specific languages)

A DSL is more than class methods invoked in class definitions.

Are we really declaring that the following is not a DSL:
```ruby
ATTRIBUTE_TYPES = {
  status: Field::Select.with_options(
    searchable: false,
    collection: %w[draft review approved] },
  )
}
```

and the following would be a DSL?

```ruby
attribute_type :status, Field::Select.with_options(
  searchable: false,
  collection: %w[draft review approved] },
)
```

What is so great about using constants vs class methods? That is still a set of predefined names that the user has to know and follow their expected value structure in order to work, so IMO there's absolutely no difference.

I believe that the greatest disadvantage of ActiveAdmin was that they went too far with the DSL: they created a DSL even for defining forms when you could just write a form erb file using the Rails helpers with not many more lines of code.

But to me, having to define things in a declarative way via constants with specific names in Administrate is not different at all from calling class methods on the class definition such as ActiveAdmin does for declaring basic things such as columns available for the index page.

nickcharlton:
The original intent of the line in the README was to emphasise that it's not fully DSL-driven, instead as close to standard Rails as possible.

This rephrases to avoid saying "we don't have a DSL", but try and provide a more useful way to describe the projects' vision.

Closes #2267

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,10 @@ and to be easier for developers to customize.
 
 To accomplish these goals, Administrate follows a few guiding principles:
 
-- No DSLs (domain-specific languages)
+- Stay as close to standard Rails as possible, keeping the
+  Administrate-specific code as small as practical,
 - Support the simplest use cases, and let the user override defaults with
-  standard tools such as plain Rails controllers and views.
+  standard tools such as plain Rails controllers and views,
 - Break up the library into core components and plugins,
   so each component stays small and easy to maintain.
 


nickcharlton:
We also want to update `administrate.gemspec`, as it has the same words in it.

jutonz (approved):
looks great! I like this a lot :)

README.md
```diff
@@ -12,28 +12,22 @@ A framework for creating flexible, powerful admin dashboards in Rails.
...
-
-[Rails Admin]: https://github.com/sferik/rails_admin
-[ActiveAdmin]: http://activeadmin.info/
+## What's Administrate?
+
+Administrate is a library for Rails that generates admin dashboards.
+Administrate's admin dashboards give users clean interfaces that allow them to
+create, edit, search, and delete records for any model in the application. It
+aims to provide the best user experience, and doing as much work as possible
+for you, whilst also being easy to customise.
```
slickepinne, line 21: Instead of easy, you could maybe use "flexible" or just "customisable"?

slickepinne (approved):
Love it.

README.md
nickcharlton: Ooh yeah, "flexible" does actually fit quite well.

PR #2516, opened 2024-02-13
Unsafe Actions: Replace `link_to` calls with `button_to`

seanpdoyle:
The `<a>` element is suitable for [Safe HTTP Methods][] (like `GET`) to drive page navigations. Unsafe HTTP Methods (like `POST`, `PUT`, and `DELETE`) are better initiated by `<form>` submissions.

This commit replaces generated calls to `link_to` with calls to `button_to`.

The rest of the styling changes aim to preserve design decisions made about preserving the appearance of elements that were once presented as `<a>` elements that are now presented as `<input type="submit">` elements nested within `<form>` elements.

While the design changes preserved backwards compatibility, it's worth re-considering the choice to present them as "navigation" links instead of "action" buttons.

[Safe HTTP Methods]: https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP


nickcharlton:
Ah, excellent, thanks!

PR #2532, opened 2024-02-25
Explicit module nestings

sdwolfz:
Initial draft of explicit module nesting


nickcharlton (commented):
The need to add a lot of `<%= indent %>`, isn't sitting well with me, but I can't think of an alternative. What else did you try?

sdwolfz:
The way I tried:

First I split the template into 3:
1. Top: containing just the module and class definition
2. Middle: the content of the class as is right now
3. End: just properly indented end statements

Then I used ERB, passing in `binding` as argument so it gets rendered with the current context.

This did not work. The `template` method does a lot more than just ERB rendering and writing to file, there were many things missing and I would get nil errors.

Unfortunately I did not find another way to generate 3 templates and then stitch them together.

nickcharlton:
Interesting, thanks! I'd like to play around with it a bit, I think.

@pablobm, do you have any thoughts on this implementation? I'm really not sure what to suggest here.

pablobm:
Yeah, the internals of Rails templates are not easy to work with... :sweat: Something that might work though: how about generating the three files (head, body, foot) and then editing and merging them using normal `File` functions? My understanding is that the `template` call puts the file in place, suggesting that we might be able to take it from there.

sdwolfz:
@pablobm `how about generating the three files (head, body, foot) ` this is what's failing. Generating the 3 files using ERB is not the same as doing it with `template`, even if I pass binding in:

```ruby
tpl = ERB.new(File.read('whatever.erb'))
result = template.result(binding)
```

This will fail with errors related to missing methods. Binding does not have the same access that `template` has to the Generator's context. It does more than just erb rendering.

Having the indents added in every line of the template is the only thing I could do to make it work.

goosys:
@sdwolfz @nickcharlton @pablobm 
I found optimize_indentation. Wouldn't this be usable?
https://garrettdimon.com/journal/posts/erb-whitespace-and-indentation-in-rails-generator-templates
https://apidock.com/rails/Rails/Generators/Actions/optimize_indentation

pablobm:
That looks useful. Thank you @goosys :clap:

@sdwolfz, would you have time to look into it?

And if that doesn't work... perhaps the crazy idea below will?

```ruby
Standard::Cli.new(["--autocorrect", target_file]).run
```

Issue #2523, opened 2024-02-16
Flakey spec: Search admin search with a filter with arguments


nickcharlton:
Flakey, but consistently flakey since we merged in #2448.

```
Failures:

  1) Search admin searches with a filter with arguments
     Failure/Error: expect(page).not_to have_content(standard_match.email)
       expected not to find text "me@kind.com" in "Back to app\nCustomers\nLog Entries\nOrders\nPages\nProducts\nProduct Meta Tags\nPayments\nSeries\nBlog Posts\nBlog Tags\nStats\nCustomers\nYou are logged in as Admin.\nNew customer\nEmail Email Subscriber Lifetime Value Name Orders Log Entries Kind Territory Password\nvip@kind.com - $0.00 Customer 157 0 orders 0 log entries vip Country 201 Edit\nBecome\nme@kind.com - $0.00 Customer 158 0 orders 0 log entries standard Country 202 Edit\nBecome\nstandard@kind.com - $0.00 Customer 159 0 orders 0 log entries vip Country 203 Edit\nBecome"
     # ./spec/features/search_spec.rb:105:in `block (2 levels) in <top (required)>'
     # /Users/nickcharlton/.gem/ruby/3.2.2/gems/webmock-3.20.0/lib/webmock/rspec.rb:39:in `block (2 levels) in <top (required)>'

Finished in 14.98 seconds (files took 2.49 seconds to load)
579 examples, 1 failure

Failed examples:

rspec ./spec/features/search_spec.rb:88 # Search admin searches with a filter with arguments

Randomized with seed 33089
```

mike-burns:
https://stackoverflow.com/questions/11980109/cucumber-capybara-check-that-a-page-does-not-have-content#comment134688856_35640662

I wonder (also don't have time to work on this right now), what if we swapped lines 106 and 105. Would that cause Capybara to wait correctly?

PR #2550, opened 2024-03-14
Call dynamic-readme reusable workflow

stefannibrasil:
We want to have a way to edit our README footer at one place and have the changes from there be propagated to our repos.

By adding this snippet in the README, we call this reusable workflow: https://github.com/thoughtbot/templates/blob/main/.github/workflows/dynamic-readme.yaml that renders the README footer dynamically.

This workflow has been updated to accommodate repositories that have main branch rules. To generate the README dynamically, you’ll need to merge a PR with the README changes. That will trigger the GitHub workflow.
The PR won’t be generated if the footer does not have any differences. It will only be open if there’s an update to the footer snippet.

### DIFF
--- a/.github/workflows/dynamic-readme.yml
+++ b/.github/workflows/dynamic-readme.yml
@@ -1,0 +1,19 @@
+name: update-templates
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - README.md
+  workflow_dispatch:
+
+jobs:
+  update-templates:
+    permissions:
+      contents: write
+      pull-requests: write
+      pages: write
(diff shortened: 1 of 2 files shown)


nickcharlton:
I fixed the CVE in #2540 (sorry, I didn't see your PR until too late!).

This looks fine to me, other than a comment which applies here too that I made on `stylelint_config`: https://github.com/thoughtbot/stylelint-config/pull/57#discussion_r1539021053

stefannibrasil:
branch rebased and comment addressed 👍

nickcharlton:
Thanks for keeping this going! The test failures are #2523, so I'm ignoring them for now.

PR #2556, opened 2024-03-22
Update contributing guide with command to start server

littleforest:
This updates `CONTRIBUTING.md` to use the `bin/dev` command instead of `rails s`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -38,11 +38,12 @@ coverage and (where necessary) have internationalisation support.
 Administrate's demo application can be run like any Rails application:
 
 ```sh
-bundle exec rails s
+bin/dev
 ```
 
 This will start the application defined in `spec/example_app`.
-You can view the `example_app` in the browser by navigating to `/admin`
+You can view the `example_app` in the browser by navigating to
+http://localhost:5000/admin.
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -38,11 +38,12 @@ coverage and (where necessary) have internationalisation support.
...
 
 ```sh
-bundle exec rails s
+bin/dev
 ```
 
 This will start the application defined in `spec/example_app`.
-You can view the `example_app` in the browser by navigating to `/admin`
+You can view the `example_app` in the browser by navigating to
+http://localhost:5000/admin.
```
pablobm, line 46: Problem here 😕 After running `./bin/setup` the first time, the `.env` file provided by default will set `PORT=3000`.

We had to do this because macOS has been using port 5000 for something else since v12 (Monterey), which we discovered while working on https://github.com/thoughtbot/administrate/pull/2397

littleforest: Ah okay, I'll revert that part then.

littleforest: Fixed in c0391563952edb75f7ec0a0ceea894c7114de51b

pablobm (approved):

PR #2557, opened 2024-03-22
Update _field-unit.scss

### LINKED ISSUE #2555: the has_one looks different in edit page
I do not sure is it expect or bug. the inputbox next row of label.
it was in the same row.

https://administrate-demo.herokuapp.com/admin/orders/121361/edit

currently look like
Screenshot 2024-03-22 at 12 57 50 PM

it was looks like i think
Screenshot 2024-03-22 at 12 58 54 PM

jubilee2:
fix #2555 I do not test if works for hit, but I think should be fine


jubilee2:
reopen in #2558 that is based on v0.20.1

PR #2558, opened 2024-03-25
fix style

jubilee2:
fix https://github.com/thoughtbot/administrate/issues/2555 I do not test if works for hint, but I think should be fine


nickcharlton:
I did a quick test of it, and it seems fine. Thanks!

Issue #1468, opened 2019-11-14
Fields without corresponding model methods and fancier behavior for `Page::Base#get_attribute_value`


dylnclrk:
First off, I'm having a great time using Administrate on a project. Thank you for maintaining such a helpful gem!

### What would you like to be able to do? Can you provide some examples?

I have a User model and am hitting an external API to gather some additional information about the user (e.g., to get "badges" assigned to a user). I'd love to show the results of this API call on the admin user show page. 

One way to do this is to add a method `User#badges` that hits our badge API and returns the results. However, I think it would be preferable not to add this method to `User`, as it won't be used outside of the admin pages. Is there a way to do this?

I'd love to be able to somehow modify the way [`Page::Base#get_attribute_value`](https://github.com/thoughtbot/administrate/blob/ebf9f3502a6d91b235085b3d7731a8fb185d59c1/lib/administrate/page/base.rb#L34) works for my User page. Such that, when the attribute in question is `:badges`, it doesn't end up calling `resource.public_send(:badges)`. 

Does this already exist? Am I missing an obvious workaround?

### How could we go about implementing that?

Provide a hook in the dashboard like `ATTRIBUTE_GETTERS`, and change `get_attribute_value` to first look in `ATTRIBUTE_GETTERS` before calling `resource.public_send(:badges)`. The values in `ATTRIBUTE_GETTERS` might be methods with the signature `some_getter(resource)`.

### Can you think of other approaches to the problem?

* I could modify the show page for User, and render the badges in the show template manually, and not treat them as fields. 
* I could also create an Admin::User class that decorates `User` with the additional fields it needs. But I haven't seen a good pattern for doing this.
* I could also perhaps modify the User Page, so that it has a custom implementation of `#get_attribute_value` that knows about the special case for `:badges`. But I'm not exactly sure how to do this. 

----------------

Also, FWIW, I'd be happy to take a stab at the `ATTRIBUTE_GETTERS` implementation if that's something that feels like it's in the spirit of the project, and assuming I haven't missed something obvious (I probably have). But it would probably take me a while.

Thanks again!

nickcharlton:
I'm wondering if a presenter would be the right way to do this, do you think that'd solve your problem? I'd take a look at #1773 and the linked PRs.

januszm:
[...]

PR #2569, opened 2024-04-25
Support custom fields without corresponding model attributes

shouichi:
The current implementation assumes fields have corresponding model
attributes. But sometimes, we just want to display a button or a link
that is not backed by an attribute.

This change simply relaxes the assumption and allows custom fields that
are not tied to model attributes.

See also https://github.com/thoughtbot/administrate/issues/1468.


nickcharlton:
Thanks for opening this!

How does this compare to #2658?

shouichi:
This PR is simpler than #2658 (only one line change in `lib` and other changes are just for tests). Each application can implement its virtual attributes and administrate doesn't need to know anything.

goosys:
I tried creating a similar `NonAttributeField` in the #2658  branch, and it worked perfectly.
Unless there’s a specific intention to raise an exception when accessing an attribute that doesn’t exist in the model, I believe this approach is sufficient.

nickcharlton:
Thanks, that's helpful.

I think I conceptually prefer this, as the code change is much smaller compared to the other. But I need to think about it!

PR #2581, opened 2024-05-09
Fix "search with a filter with arguments" spec

CaleCrawford:
Hi. This PR will solve the failing flakey test listed in https://github.com/thoughtbot/administrate/issues/2523. While debugging the issue I found a few things. 

The flake would happen when the second search of `"kind:vip"` was not applied to the url. I was able to run the test in headed chrome and see the failure happen. The url param was never set and was therefore blank. Adding a `sleep(0.1)` after the `clear_search` seemed to work as one pathway. Capybara

I went down the path of modifying the existing test to pass with a single param by the suggestions of a few at the RailsConf hackathon. Happy to add an additional scenario testing `"kind:standard"` if that would be preferred.

Thanks!

### DIFF
--- a/spec/features/search_spec.rb
+++ b/spec/features/search_spec.rb
@@ -86,24 +86,17 @@ feature "Search" do
   end
 
   scenario "admin searches with a filter with arguments", :js do
-    kind_match = create(:customer, kind: "vip", email: "vip@kind.com")
-    standard_match = create(:customer, kind: "standard", email: "me@kind.com")
-    no_match = create(:customer, kind: "vip", email: "standard@kind.com")
+    kind_match = create(:customer, kind: "vip", email: "standard@kind.com")
+    total_mismatch = create(:customer, kind: "standard", email: "me@kind.com")
+    kind_mismatch = create(:customer, kind: "standard", email: "vip@kind.com")
 
     visit admin_customers_path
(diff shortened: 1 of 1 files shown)


littleforest (commented):
This looks great, thanks for fixing! I don't think the second filter test is actually necessary. Would you mind fixing the linting?

CaleCrawford:
@littleforest sounds good thank you! Should be fixed now

nickcharlton:
Wonderful! Thank you so much for taking the time to look at this, it's been a real thorn-in-my-side for the last weeks, and I'm _terrible_ when it comes to flakey specs!

I'm really glad the RailsConf hackathon turned out so well, too!

CaleCrawford:
@nickcharlton glad it worked out! haha I hear you, same. Yeah really enjoyed the hackathon. 

I let Steffani know that I run the open source guild at my company and we have a few people who are looking for places to contribute. Would definitely love to stay in contact if you ever need a few contributors to take on similar things

nickcharlton:
Oh! That's cool to hear! We've started catching up as maintainers internally, so we'll have a chat about how we can make it a bit clearer on how to jump in. In the case of Administrate, it's really just picking something that interests you — but I appreciate that can be quite hard of a suggestion!

PR #2582, opened 2024-05-10
Copy the index_header partial when generating index views to customize

littleforest:
When customizing the index page of a resource, currently
the `index_header` partial is not copied for the user to
customize. The `index_header` partial contains the search
bar and the New resource button, either of which may need
to be customized. Currently this file must be manually 
created if someone wishes to edit it. 

This change adds the `index_header` file to the list of generated
`index` files.


nickcharlton (approved):
Wonderful, thanks!

PR #2584, opened 2024-05-13
Fix ordering HasOne fields without explicit order

mikz:
This fixes a bug in awesome https://github.com/thoughtbot/administrate/pull/2325.

Ordering by a primary key of the HasOne field wasn't working
and resulted in invalid sql, because it did not join the associated table.

This unifies the code paths to use the same logic to add the ordering
from another table and use reflection to get relevant table information.

### DIFF
--- a/lib/administrate/order.rb
+++ b/lib/administrate/order.rb
@@ -75,7 +75,7 @@ module Administrate
 
     def order_by_belongs_to(relation)
       if ordering_by_association_column?(relation)
-        order_by_attribute(relation)
+        order_by_association_attribute(relation)
       else
         order_by_id(relation)
       end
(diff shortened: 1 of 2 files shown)


mikz:
@nickcharlton looks like even after https://github.com/thoughtbot/administrate/pull/2498 approval is needed to run the workflow in a fork. Could I get approval from someone?

shouichi:
I also encountered this problem. I tried this PR locally and it worked as expected :+1:

nickcharlton:
> @nickcharlton looks like even after https://github.com/thoughtbot/administrate/pull/2498 approval is needed to run the workflow in a fork

Yeah, this is, unfortunately, expected behaviour. That PR allowed running _at all_, but then you need a maintainer (or equivalent, I don't remember the default roles off the top of my head) to approve them to run. The intention is to protect against data exfil, I think, or using lots of compute, maybe? It doesn't really apply to this project, so it's just another hurdle to have to jump over.

mikz:
@nickcharlton yeah, I understand. But can you approve it, so we can get a build and merge this fix?

nickcharlton (approved):
I've tried to finish reviewing this 4 four times, but kept getting pulled away 🤦.

Anyway, this looks good, thanks!

mikz:
@nickcharlton Thank you very much! 🎉

PR #2544, opened 2024-03-07
Url field data is not truncated in index view as per documentation

thebravoman:
Documentation says Field::URL accepts option for truncate https://administrate-demo.herokuapp.com/customizing_dashboards but code is different. 

Showing truncated data for urls. 

Could probably be downported to previous versions.

PR #2597, opened 2024-05-25
Documentation says Field::URL accepts option for truncate but code is different

thebravoman:
Documentation at https://administrate-demo.herokuapp.com/customizing_dashboards 
Code was not calling truncate.
With current change truncate is called.

This is a clean PR along with a spec. Original PR is at 
https://github.com/thoughtbot/administrate/pull/2544
but it was mostly to get some feedback.

### DIFF
--- a/app/views/fields/url/_index.html.erb
+++ b/app/views/fields/url/_index.html.erb
@@ -16,5 +16,5 @@ By default, the value is rendered as an `a` element.
 %>
 
 <%= content_tag :a, href: field.data, **field.html_options do %>
-  <%= field.data %>
+  <%= field.truncate %>
 <% end %>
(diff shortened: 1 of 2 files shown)


pablobm:
Thank you for this, and apologies for the delay to look at it. One problem is that the spec doesn't test the functionality as it's all stubbed. I understand it still needed updating to avoid it breaking.

`Field::String` and `Field::Text` are also affected. Would you be up for fixing those and see if you can provide specs that do test for regressions?

thebravoman:
I was thinking about this and tried staying with the current way tests are written and have a minimal change. Calling the method with less stubbing would mean moving a little further than what is currently in the specs. It will require probably a new type of spec. I could think about it.

pablobm:
@thebravoman - If you could give it a go, that would be great :slightly_smiling_face: If after some trying, you find it's too complex or otherwise have difficulty, let us know here and we can rethink it.

PR #2598, opened 2024-05-29
Fixes problem in table controller

anthony0030:
If there was any tag inside the link to edit or delete like a strong tag or an icon tag, the link would not take you to the correct page.

### DIFF
--- a/app/assets/javascripts/administrate/controllers/table_controller.js
+++ b/app/assets/javascripts/administrate/controllers/table_controller.js
@@ -9,7 +9,7 @@ export default class extends Controller {
         event.keyCode == keycodes.space ||
         event.keyCode == keycodes.enter) {
 
-      if (event.target.href) {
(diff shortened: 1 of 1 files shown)


nickcharlton:
I think it'd be a good idea to add a test around this, so that we don't break it in future. Is that something you could do?

anthony0030:
@nickcharlton I took a look and could not find any test file for table_controller. Please point me in the right direction, to be able to see if I am able to add some tests in.

@excid3, is this something that would be easy for you to do?

nickcharlton:
It's possible we just don't have any at all. A good starting point might be one of the [feature specs](https://github.com/thoughtbot/administrate/tree/main/spec/features), if there's one that fits this functionality best.

anthony0030:
I looked into writing a test for this. I have never written tests in this format. At this time I won't be able to assist further.

PR #2609, opened 2024-06-07
Add diff-check to catch missing Appraisal runs

nickcharlton:
If we forget to run `bundle exec appraisal`, which is not uncommon, it means we miss when the appraisal files change. This should catch it and fail the build, reporting back that they changed and hinting at what should be done.

https://github.com/nickcharlton/diff-check
https://nickcharlton.net/posts/diff-check-github-action

### DIFF
--- a/.github/workflows/diff-check.yml
+++ b/.github/workflows/diff-check.yml
@@ -1,0 +1,16 @@
+---
+name: diff-check
+on: [push]
+
+jobs:
+  appraisal:
+    runs-on: ubuntu-latest
+    env:
+      DIFF_CHECK_APPRAISAL: "true"
+    steps:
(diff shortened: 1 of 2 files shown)


nickcharlton:
This puts an action I wrote a little while ago in, as I was waiting for a time it actually would fail correctly. I also found a bug in `diff-check` itself, which I've fixed on a branch and waiting to hear back if it solves things elsewhere too.

That said, we're seeing the `Gemfile.lock` change and I don't really understand why yet.

- [ ] Change "demo" to "appraisal" in the workflow
- [ ] Change `diff-check` branch away from `fix-path`
- [ ] Figure out if `Gemfile.lock` changing is expected, or we need to ignore certain changes

PR #2612, opened 2024-06-18
for multi database connection

stadia:


nickcharlton:
Ah, nice thanks!

I think if we're going to introduce multi-database support (which we should, as it's a Rails default option now), we're going to need to think about testing for that situation, it might not just be here that's a problem (although you'd presumably tried this out and not hit anything else!) but we also might accidentally break it in future.

Do you have any thoughts on how best we might do that?

The first idea which comes to mind is another GitHub Action to spin up multiple databases and fetch records from different ones, but that's probably quite a bit of work.

PR #2637, opened 2024-07-30
Update rexml to fix a security vulnerability

stefannibrasil:
CI is failing due to this security vulnerability
reported two weeks ago:

https://github.com/ruby/rexml/security/advisories/GHSA-4xqq-m2hx-25v8

Bumping its version to use a patched version to keep the library secure.

PS. The other changes are just a matter of moving the gem around, let me know if you'd prefer to not include them here.


nickcharlton (approved):
Thanks!

Issue #2459, opened 2023-11-17
Scope select options to a particular resource?


wengzilla:
I would like to scope a BelongsTo field to a particular set of things based on the resource, but I'm not sure how to achieve this...

Right now, I have the following relationship: Companies have many Departments which have many Employees. Or, in other words, Employee belongs to a Department which belongs to a Company.

I want to have a field on my EmployeeDashboard that looks as follows:

```
ATTRIBUTE_TYPES = {
    departments:
      Field::BelongsTo.with_options(scope: ->(resource) { resource.company.departments }),
  }.freeze
```

I could monkey patch the scope call to pass in resource, but is there a more elegant way to do this?

Nitr:
@wengzilla I have the same issue, you can monkey patch [BelongsTo](https://github.com/thoughtbot/administrate/blob/main/lib/administrate/field/belongs_to.rb#L43) like this:

```ruby
def candidate_resources
  scope = if options[:scope]
    options[:scope].arity.positive? ? options[:scope].call(self) : options[:scope].call
  else
    associated_class.all
  end

  order = options.delete(:order)
  order ? scope.reorder(order) : scope
end
```
then in your dashboard:
```ruby
ATTRIBUTE_TYPES = {
    departments:
      Field::BelongsTo.with_options(scope: ->(field) { field.resource.company.departments }),
  }.freeze
```
Another solution is to use Select:
```ruby
ATTRIBUTE_TYPES = {
    departments:
      Field::Select.with_options(collection: ->(field) { field.resource.company.departments.pluck(:name, :id) }),
  }.freeze
```

benoror:
Nice @Nitr , this should be by default

PR #2652, opened 2024-08-27
Allow scope to accept field as argument for Field::BelongsTo

Nitr:
https://github.com/thoughtbot/administrate/issues/2459

Sometimes need to do a query dependent on resource's state, field and etc in forms.

I used to use combination of `Field::Select` for form page and `Field::BelongsTo` for collection page like this:
```ruby
ATTRIBUTE_TYPES = {
  customer_id: Field::Select.with_option(collection: ->(field) { Customer.some_scope.where(column: field.resource.column) }), 
  customer: Field::BelongsTo
}
COLLECTION_ATTRIBUTES = [:customer]
FORM_ATTRIBUTES = [:customer_id]
```
But I've noticed that many people are confused by this and that it's more common to have only one field `customer` like this:
```ruby
ATTRIBUTE_TYPES = { 
  customer: Field::BelongsTo.with_option(scope: ->(field) { Customer.some_scope.where(column: field.resource.column) }), 
}
COLLECTION_ATTRIBUTES = [:customer]
FORM_ATTRIBUTES = [:customer]
```

I've just found this [gem](https://github.com/XPBytes/administrate-field-scoped_belongs_to) but it doesn't check arity.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -82,7 +82,8 @@ the table views and in the dropdown menu on the record forms.
 You can set multiple columns as well with direction. E.g.: `"name, email DESC"`.
 
 `:scope` - Specifies a custom scope inside a callable. Useful for preloading.
-Example: `.with_options(scope: -> { MyModel.includes(:rel).limit(5) })`
+Example #1: `.with_options(scope: -> { MyModel.includes(:rel).limit(5) })`
+Example #2: `.with_options(scope: -> (field) { field.resource.my_models.includes(:rel).limit(5) })`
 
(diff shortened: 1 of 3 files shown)


nickcharlton:
I just let the test suite run, could you take a look at those failures when you get a chance?

Nitr:
> I just let the test suite run, could you take a look at those failures when you get a chance?

Yeah! It's done! Could you please run it again

nickcharlton:
I was surprised it wasn't caught already by Dependabot, but I've not looked into it. I'm generally comfortable ignoring that check as long as we know why it's failing, though so it won't hold this back.

Issue #1586, opened 2020-03-26
Support for virtual fields (regression)


pablobm:
When https://github.com/thoughtbot/administrate/pull/920 was merged, it broke apps that depended on the old behaviour.

These are two reports on this issue:

### Report by @wkirby in https://github.com/thoughtbot/administrate/pull/920#issuecomment-599722598

This PR appears to break any support for virtual fields by enforcing the the resource _must_ define a method to respond to all keys in the `Dashboard` class. Maybe this is the desired behavior, but the following use case was possible in `0.12.0`, and was very useful:

```ruby
class ReportDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    id: Field::Number,
    filename: Field::String,
    mimetype: Field::String,
    s3_key: Field::String,
    download: S3DownloadField.with_options({ download_path: :download_lab_report_path }),
    page_count: Field::Number,
    created_at: Field::DateTime,
    updated_at: Field::DateTime
  }.freeze

  COLLECTION_ATTRIBUTES = [
    :id,
    :filename,
    :download,
    :mimetype,
    :created_at
  ].freeze

  SHOW_PAGE_ATTRIBUTES = [
    :id,
    :download,
    :filename,
    :mimetype,
    :s3_key,
    :page_count,
    :created_at,
    :updated_at
  ].freeze

  FORM_ATTRIBUTES = [
    :lab,
    :filename,
    :mimetype,
    :s3_key,
    :page_count
  ].freeze
end
```

```ruby
class S3DownloadField < Administrate::Field::Base
  def to_s
    data
  end

  def downloadable?
    options.key?(:download_path)
  end

  def destroyable?
    options.key?(:destroy_path)
  end

  def destroy_path(field, attachment)
    path_helper(:destroy_path, field, attachment)
  end

  def download_path(field, attachment)
    path_helper(:download_path, field, attachment)
  end

  private

  def path_helper(key, field, _attachment)
    path_helper = options.fetch(key)
    Rails.application.routes.url_helpers.send(path_helper, field.resource)
  end
end
```

This allowed us to inject a download link for our S3 resource _without_ editing the underlying views, or modifying the display for any of the actual attributes.

Is there any advice on how to fix this _without_ defining useless methods on our models?

### Report by @Timmitry in https://github.com/thoughtbot/administrate/issues/1570#issuecomment-600472047

[...]

Issue #1314, opened 2019-04-12
Possibility to set field option as not sortable


privorotskii:
It is possible to set field as searchable, but there is no field option to set it as sortable or not sortable. It could be useful for some cases - not each column at collection page has to be sortable by default.
May I create a PR for adding such option? It would be working like this:

```
  ...
  ATTRIBUTE_TYPES = {
    ...
    comment: Field::Text.with_options(sortable: false),
    ...
  }.freeze
  ...
```

So the heading of the 'comment' column at the collection page would not be clickable.
Or, may be, there is another way to do it I don't know about?

nickcharlton:
I know it's been a long time — I think your PR fell aside somewhere — but if you're still interested in opening that PR that'd be great!

pablobm:
Closing due to lack of activity. Still happy to see a PR about this.

PR #2658, opened 2024-09-10
Support for virtual fields

goosys:
Support for virtual fields, based on #1586.

# Usage

- It is possible to use a name as attribute_name that doesn't exist in the DB column or as a method in the model.

image

```ruby
class CustomerDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    name: Field::String, # real field
    nickname: Field::String.with_options( # virtual field
      # You can specify a symbol if Customer#generate_nickname is defined
      getter: :generate_nickname,
      # Or, You can directly specify a function
      getter: ->(field) { "Mr. #{field.resource.name}man" if field.resource.name.present? },
      # Virtual fields cannot be searchable, so 'false' must be specified
      searchable: false 
    ),
```

---

(Note: The Sortable implementation has been split into #2659.)

---

What do you think? I'll make any necessary changes.
Please review it.


pablobm:
Thank you for this. I hope I can remember things clearly, let's see what we can do here... :eyes:

### `sortable` option
First, you are also introducing the `sortable` option, which is a separate concern. That would be best served in a separate PR.

By all means, it's really good that you are implementing  `sortable` (it's been requested before, see https://github.com/thoughtbot/administrate/issues/1314), but separate, smaller PRs are easier to review and reason about. I'll be happy to review that.

### The virtual field
The solution you propose is different from the one that was lost when https://github.com/thoughtbot/administrate/pull/920 was merged, resulting in issue https://github.com/thoughtbot/administrate/issues/1586.

You propose an option `:getter` to define how a field will be accessed. In contrast, the lost functionality allowed the creation of fields that didn't need special options or interfaces.

So for example, it was possible to have this:

```ruby
## field/foobar.rb
require_relative "base"

module Administrate
  module Field
    class Foobar < Field::Base
      def self.searchable?
        false
      end

      def foobar
        "This is a Foobar: #{resource.inspect}"
      end
    end
  end
end

## views/fields/foobar/_collection.html.erb
<%= field.foobar %>
```

In my view, the broken functionality was simpler to use and I would prefer that to return.

Also, I see that your code would only work when when `@data.nil?`. This could lead to unexpected behaviour, as it's perfectly normal for a non-virtual field to be `nil` instead of having a value. In my mind, it goes "if it's nil, perhaps it's a virtual field".

In comparison, the original functionality established distinction between virtual and non-virtual fields. In my mind, it goes "it is virtual or it is not, but there's no confusion".

I hope that makes sense.

What do you think? Would you be able to separate the `sortable` code into a different PR, and change this one to behave the way I describe?

goosys:
@pablobm 
Apologies for the confusion. With the implementation of virtual fields, the issue with the sort buttons on the index screen became more noticeable, so I initially included both in the same PR.
I've now split the PRs. I'll address the points you mentioned later, and I'd appreciate your review again at that time.

goosys:
@pablobm 

> Also, I see that your code would only work when when @data.nil?. This could lead to unexpected behaviour, as it's perfectly normal for a non-virtual field to be nil instead of having a value. In my mind, it goes "if it's nil, perhaps it's a virtual field".

This is for compatibility reasons.
We can fetch data exclusively from within the field using `read_value`. While the application will work perfectly in this state, many RSpec tests fail because they insert data using the `data` argument.

```ruby
module Administrate
  module Field
    class Base
      def initialize(attribute, _data, page, options = {})
        @attribute = attribute
        @page = page
        @resource = options.delete(:resource)
        @options = options
        @data = read_value
      end
```

image

Should I fix all the RSpec tests?
In that case, the `data` argument in `initialize` method would become unnecessary, so it would be better to either remove it or make it optional. However, this might cause compatibility issues with third-party Fields.

goosys:
@pablobm 

> You propose an option :getter to define how a field will be accessed. In contrast, the lost functionality allowed the creation of fields that didn't need special options or interfaces.
> In my view, the broken functionality was simpler to use and I would prefer that to return.

I believe this case should not be an issue. For custom fields, you can override read_value to fetch data and format it using any method you prefer.
I've written a sample and committed it, so please check it.

https://github.com/thoughtbot/administrate/pull/2658/commits/a83e2ff6ae76b7f289b393cbff6a6d03cac34a0c

lib/administrate/field/base.rb
```diff
@@ -52,6 +52,22 @@ def name
...
+          else
+            resource&.public_send(options[:getter])
+          end
+        else
+          if data.nil?
+            resource&.public_send(attribute)
+          else
+            data
+          end
+        end
```
pablobm, line 68: Unfortunately, the example I mentioned with `Foobar` doesn't work: `public_send` fails because the method doesn't exist in the resource. Note that the safe navigation operator (`&.`) doesn't help here because it's useful only when the resource is nil.

However, I think it can be fixed with a `respond_to?` instead, as below:

```suggestion
        elsif data.nil?
          resource.respond_to?(attribute) ? resource.public_send(attribute) : nil
        else
          data
        end
      end
```

This should also fix the linter error asking for `elsif` to be used.

spec/example_app/app/dashboards/customer_dashboard.rb
```diff
@@ -8,6 +8,10 @@ class CustomerDashboard < Administrate::BaseDashboard
     email_subscriber: Field::Boolean,
     lifetime_value: Field::Number.with_options(prefix: "$", decimals: 2),
     name: Field::String,
+    nickname: Field::String.with_options(
+      getter: ->(field) { "Mr. #{field.resource.name}man" if field.resource.name.present? },
+      searchable: false
+    ),
```
pablobm, line 14: Thinking of an example that doesn't feel too artificial: how about a `full_address` field for `Order`? Something like this:

```ruby
    full_address: Field::String.with_options(
      getter: ->(field) {
        r = field.resource
        [
          r.address_line_one,
          r.address_line_two,
          r.address_city,
          r.address_state,
          r.address_zip
        ].compact.join("\n")
      },
      searchable: false
    ),
```

Then its `SHOW_PAGE_ATTRIBUTES` can be something like this:

```ruby
  SHOW_PAGE_ATTRIBUTES = FORM_ATTRIBUTES
    .except("address")
    .merge(
      "" => %i[customer full_address created_at updated_at],
      "details" => %i[line_items total_price shipped_at payments]
    )
    .freeze
```

spec/lib/fields/virtual_field_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
+require "administrate/field/base"
+
+module Administrate
+  module Field
+    class VirtualField < Field::Base
```
pablobm, line 6: I wonder if we can find an example in the dashboard where we can define this type of virtual field without having to drop it here?

pablobm (commented):
OK, I'm more onboard now. Thank you for the changes and explanation :slightly_smiling_face: 

Regarding the `data` argument, for some time I have been thinking that it should disappear, but let's not do that now. This PR can be an intermediate step with back-compatibility, towards a future where fields calculate their value instead of being calculated at the `Page` level.

I wasn't sure about the `:getter`, but after playing with it a bit I'm liking it better. I think it can stay.

spec/example_app/app/dashboards/customer_dashboard.rb
goosys: Thanks. I added.

lib/administrate/field/base.rb
goosys: Apologies for the oversight. I’ve made the fixes and checked with standardrb.
I reviewed the your reference code and felt that `try` would be sufficient, so I rewrote it using `try`. Do you have any concerns about this approach?

spec/lib/fields/virtual_field_spec.rb
goosys: What else is needed here? Documentation, maybe?
Or would an approach using an option like `virtual_field: true` to explicitly indicate virtual fields work better?

spec/example_app/app/dashboards/customer_dashboard.rb
Nitr: Looks like `searchable: false` is excess. If getter has been provided it makes field as unsearchable.

Nitr: Have you thought about using decorators(like draper) for virtual fields?

spec/lib/fields/base_spec.rb
```diff
@@ -167,4 +167,34 @@
...
+      expect(resource).to receive(:custom_getter)
+      expect(field.read_value).to eq("value")
+    end
+
+    it "returns nil if the resource is nil" do
+      field = field_class.new(:attribute, :date, :page, resource: nil)
+
+      expect(field.read_value).to eq(nil)
+    end
+  end
```
pablobm, line 199: I see `read_value` as internal (perhaps it should be `private`?), and in general I think it's better to write specs about external behaviour, which in this case is `data`. How about a spec for `data` instead?

```suggestion
  describe "#data" do
    context "when given nil data" do
      it "reads the value from the resource" do
        resource = double(attribute: "resource value")
        field = field_class.new(:attribute, nil, :page, resource: resource)

        expect(field.data).to eq("resource value")
      end
    end

    context "when given non-nil data" do
      it "uses the given data" do
        resource = double(attribute: "resource value")
        field = field_class.new(:attribute, "given value", :page, resource: resource)

        expect(field.data).to eq("given value")
      end
    end

    context "when given a :getter value" do
      it "reads the attribute with the name of the value" do
        resource = double(custom_getter: "custom value")
        field = field_class.new(:attribute, :date, :page, resource: resource, getter: :custom_getter)

        expect(field.data).to eq("custom value")
      end
    end

    context "when given a :getter block" do
      it "uses it to produce a value" do
        resource = double("Model", custom_getter: "custom value")
        field = field_class.new(:attribute, :date, :page, resource: resource, getter: ->(f) { f.resource.custom_getter + " from block" })

        expect(field.data).to eq("custom value from block")
      end
    end

    context "when given a :getter block" do
      it "returns nil if the resource is nil" do
        field = field_class.new(:attribute, nil, :page, resource: nil)

        expect(field.data).to eq(nil)
      end
    end
  end
```

This also is better at explaining what happens with the `data` parameter for the initializer, which is important. I'm hoping to remove it in the future as discussed, but for now it's part of the interface and important to specify.

Also I have removed the `expect(...).to receive` because we can just test for the returned value and that's more reliable.

spec/lib/fields/virtual_field_spec.rb
pablobm: Answering on the main thread :slightly_smiling_face:

pablobm (commented):
Coming along nicely! Let's see...

Could you please add documentation about the new `:getter` option?

Then about the `spec/lib/fields/virtual_field_spec.rb` spec. I think best to remove it completely and a good example like the following:

```diff
+require "administrate/field/receipt_link"
 require "administrate/base_dashboard"
 
 class PaymentDashboard < Administrate::BaseDashboard
   ATTRIBUTE_TYPES = {
     id: Field::Number,
+    receipt: Field::ReceiptLink,
     created_at: Field::DateTime,
     updated_at: Field::DateTime,
     order: Field::BelongsTo
   }
 
   COLLECTION_ATTRIBUTES = [
-    :id
+    :id,
+    :receipt
   ]
 
   SHOW_PAGE_ATTRIBUTES = ATTRIBUTE_TYPES.keys
diff --git a/spec/example_app/app/views/fields/receipt_link/_index.html.erb b/spec/example_app/app/views/fields/receipt_link/_index.html.erb
new file mode 100644
index 00000000..9f755667
--- /dev/null
+++ b/spec/example_app/app/views/fields/receipt_link/_index.html.erb
@@ -0,0 +1 @@
+<%= link_to field.filename, field.data %>
diff --git a/spec/example_app/lib/administrate/field/receipt_link.rb b/spec/example_app/lib/administrate/field/receipt_link.rb
new file mode 100644
index 00000000..0756b259
--- /dev/null
+++ b/spec/example_app/lib/administrate/field/receipt_link.rb
@@ -0,0 +1,15 @@
+require "administrate/field/base"
+
+module Administrate
+  module Field
+    class ReceiptLink < Base
+      def data
+        "/files/receipts/#{filename}"
+      end
+
+      def filename
+        "receipt-#{resource.id}.pdf"
+      end
+    end
+  end
+end
```

Still needs a `_show` template and a simple spec (doesn't need to be big). Could you please add that?

spec/example_app/app/dashboards/customer_dashboard.rb
goosys: @Nitr @pablobm 
Thank you for the advice. I have made the changes. https://github.com/thoughtbot/administrate/pull/2658/commits/cab60966f9eb8edca26c5bfd18b75962c59691eb
I'm not sure if we should always set `searchable` to `false` when a `getter` is present, but since I can't think of any use cases at the moment, this seems appropriate.

goosys:
Next, I would like to add documentation.
However, it seems like this feature has become more complex and capable than I initially expected, so I plan to organize it a bit before documenting it. If possible, I would appreciate your continued assistance.

pablobm:
@goosys - Sounds good :slightly_smiling_face:

goosys:
@pablobm 
I've created documentation for all the use cases I can think of at the moment. Do you think there are any other use cases we should include?
https://github.com/thoughtbot/administrate/pull/2658/commits/d1fd3a9b08eebe9c78f355a5d40c5935648ba41c

Also, if there are any mistakes in the English within the documentation, I’d appreciate it if you could correct them. Thank you!

Nitr:
@goosys 
`For all field types` - I guess it won't work properly for associative types. It will try to includes association by `key` to avoid N+1 and it will fail.

goosys:
@Nitr 
Thanks for the valuable feedback!
I checked, and it seems to be working for now. Although multiple options besides the `:getter` need to be specified, nothing seems to be broken. (Also, it appears that the deprecated `:class_name` is still required.)
Regarding the N+1 issue, it can be avoided with the existing `:includes` option.

```ruby
class CustomerDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    orders: Field::HasMany.with_options(limit: 2, sort_by: :id),
    recent_orders: Field::HasMany.with_options(
      getter: ->(field){ field.resource.orders.where(created_at: [3.days.ago...]).order(created_at: :desc).limit(2) },
      class_name: "Order",
      foreign_key: :customer_id,
      includes: :orders
    ),
```

It’s quite strange at this point, but I think we can improve it going forward. What do you think?

docs/customizing_dashboards.md
```diff
@@ -436,3 +436,133 @@ en:
...
+  end
+end
+```
+
+```erb
+<%# app/views/fields/custom_action_buttons/_index.html.erb %>
+<% if field.data.present? %>
+  <%= button_to "some action 1", [:some_action_1, namespace, field.resource] %>
+  <%= button_to "some action 2", [:some_action_2, namespace, field.resource] %>
+  <%= button_to "some action 3", [:some_action_3, namespace, field.resource] %>
```
pablobm, line 566: This hadn't even occurred to me :smile: Great stuff.

spec/lib/fields/base_spec.rb
```diff
@@ -167,4 +167,50 @@
...
+    context "when given a :getter block" do
+      it "uses it to produce a value" do
+        resource = double("Model", custom_getter: "custom value")
+        field = field_class.new(:attribute, :date, :page, resource: resource, getter: ->(f) { f.resource.custom_getter + " from block" })
+
+        expect(field.data).to eq("custom value from block")
+      end
+    end
+
+    context "when given a :getter block" do
```
pablobm, line 208: This bit can be deleted as the two contexts are the same.

```suggestion
    end

    context "when given a :getter block" do
```

Was this part of my suggestion the other day? :flushed:

pablobm:
I think this is it? (Apart from the two small suggestions now). Do you think there's anything else to cover or should we go and merge?

Tangentially: I need to un-deprecate the `:class_name` option as we have come across legitimate use cases.

goosys:
@pablobm 

> I think this is it? (Apart from the two small suggestions now). Do you think there's anything else to cover or should we go and merge?

I believe there are still plenty of ways we could use this, depending on new ideas. But for now, I think it’s fine to go ahead and merge it.

pablobm:
One last thing. Sorry! :sweat: 

1. I saw a couple of `read_value` that I don't think are used anywhere and removed them. Also removed the default value `data = nil` from the normal `read_value.
2. I added some nicety to the receipt downloader.

Changes are at https://github.com/thoughtbot/administrate/compare/main...pablobm:administrate:goosys-issue-1586-virtual-fields. Does this make sense? If so, could you please merge my two commits into your branch?

goosys:
@pablobm 
Thank you for checking.

It seems the `Deferred#read_value` isn’t necessary. I wasn’t entirely sure, so I added it just in case. Sorry about that.
Also, thanks for the other additions! I’ve merged the changes.

One thing I wanted to mention—while pushing to this PR several times, I noticed that the `payment_index_spec` was unstable, failing and passing inconsistently only with Ruby 3.0. I addressed it since it was happening quite often.
Could it be that updating `current_path` is sometimes delayed? It’s strange because no other tests had this issue, just this one file. After modifying it to use have_current_path, it stopped failing locally.

pablobm:
Re: `current_path`, no idea :sweat_smile: I wasn't seeing the problem so I was not aware. But if your change appears to fix things, then that's good :+1: 

Merging at last! :rocket:

PR #2659, opened 2024-09-13
Add sorting_column to Field

goosys:
I've split the Sortable from #2658.

# Usage

- It is possible to use a name as attribute_name that doesn't exist in the DB column but is defined in the model.

image

```ruby

class OrderDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    full_address: Field::String.with_options(
      getter: ->(field) {
        r = field.resource
        [
          r.address_line_one,
          r.address_line_two,
          r.address_city,
          r.address_state,
          r.address_zip
        ].compact.join("\n")
      },
      sorting_column: :address_zip
    ),
```

- Attributes that cannot be sorted will not have a sort link in the table header.

```ruby
module Administrate
  module Field
    class Password < Field::Base
      def self.sortable?
        false
      end
```

```ruby
class CustomerDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    # Since the aggregation is done on the Ruby side and sorting is not possible, it is set to not be sortable.
    lifetime_value: Field::Number.with_options(prefix: "$", decimals: 2, sortable: false),
```

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -141,16 +141,18 @@ module Administrate
       @order ||= Administrate::Order.new(
         sorting_attribute,
         sorting_direction,
-        association_attribute: order_by_field(
+        sorting_column: sorting_column(
           dashboard_attribute(sorting_attribute)
         )
       )
     end
 
-    def order_by_field(dashboard)
-      return unless dashboard.try(:options)
+    def sorting_column(dashboard_attribute)
+      return unless dashboard_attribute.try(:options)
 
-      dashboard.options.fetch(:order, nil)
+      dashboard_attribute.options.fetch(:sorting_column) {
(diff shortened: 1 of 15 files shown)


goosys:
I rebased after #2658 was merged and made a few additional fixes.

goosys:
I feel that we need to discuss the option name.
We already have `sort_by` and `order` options related to sorting, but the newly added option doesn’t fully serve the same purpose as either of these, though there is some overlap.
For now, I’ve named it `sorting_column` to avoid conflicts, but it doesn’t feel very elegant. I’d appreciate any ideas you might have.

nickcharlton:
Hmm, yeah, the column names aren't great. At least `sorting_column` fits in with the rest of them.

Could you add some docs to it?

goosys:
@nickcharlton I have added the documentation. Please take a look.

PR #2665, opened 2024-09-27
Allow configuring text field input options

balvig:
## What

Allows configuring the rendered `<textarea>` for `Field::Text` fields, e.g.:

```rb
ATTRIBUTE_TYPES = {
  body: Field::Text.with_options(input_options: { rows: 20 }),
}.freeze
```

Result in:

```html
<textarea rows="20" name="post[body]" id="post_body">...</textarea>
```

## Why

The main use case for this was being able to scale up text areas when needed, f.ex:
|Before|After|
|-|-|
|Edit CoachingContent #3 - Moment Web2024-09-27 at 11 53 31@2x|Edit CoachingContent #3 - Moment Web2024-09-27 at 11 53 12@2x|

## How

This provides a generic `input_options` to access all the options provided by `text_area`, but if that feels like overkill, it could also be hardcoded to only support `rows`?

### DIFF
--- a/spec/administrate/views/fields/text/_form_spec.rb
+++ b/spec/administrate/views/fields/text/_form_spec.rb
@@ -1,0 +1,56 @@
+require "rails_helper"
+require "administrate/field/text"
+
+describe "fields/text/_form", type: :view do
+  it "allows configuring input options" do
+    textarea = instance_double(
+      "Administrate::Field::Text",
+      attribute: :name,
+      data: nil,
+      options: { input_options: { rows: 50 } }
+    )
+
+    render_partial(textarea)
+
+    expect(rendered).to have_css(%(textarea[rows=50]))
+  end
+
+  it "doesn't require input options" do
+    textarea = instance_double(
+      "Administrate::Field::Text",
+      attribute: :name,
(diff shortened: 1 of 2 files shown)


spec/administrate/views/fields/text/_form_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+require "administrate/field/text"
+
+describe "fields/text/_form", type: :view do
```
nickcharlton, line 4: Minor thing really, but I generally prefer to avoid the mixin approach for RSpec (and we also don't need to specify the type):

```suggestion
RSpec.describe "fields/text/_form" do
```

nickcharlton (commented):
This is a lovely improvement, thanks!

I just triggered the CI run, too.

spec/administrate/views/fields/text/_form_spec.rb
balvig: Cheers, thank you @nickcharlton!

nickcharlton (approved):
Thanks for doing this! And finding out that we do need `type: view`, too!

PR #2668, opened 2024-09-29
Embed sources in CSS sourcemaps

ignisf:
* What were you trying to do?
Trying to debug UI issues with the new Administrate beta version (turns out they are resolved in `main`?).

Whenever I open Firefox/Chrome dev tools, they try to load the SCSS source files from the Rails server, which fails, as expected, with 404.

  - Rails 8.0.0.beta1
  - administrate 1.0.0.beta1

I suggest you add the `--embed-sources` switch to the `build:css` `yarn` incantation.

Tested locally:
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "build": "esbuild app/assets/javascripts/administrate/application.js --bundle --sourcemap --outdir=app/assets/builds/administrate --public-path=/assets/administrate",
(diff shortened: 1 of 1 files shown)


nickcharlton:
That's a lovely idea, thanks.

(I'll be cutting a new release candidate release soonish so hopefully you'll also have less annoying bugs to track down!)

ignisf:
Thank you, I am eagerly spamming `bundle update` ;)

PR #2673, opened 2024-10-04
Accumulated build files since #2411

pablobm:
Not sure what to make of this one. I was getting these build files in a branch I was working on. Eventually I realised this was happening in main too. Seems to be the case since merging #2411, and I think changes might have been accumulating over time?

So... should these be committed then?

I can't add @littleforest as a reviewer for some reason, so pinging here instead.


goosys:
I think it might be necessary to add the `app/assets/builds` directory to the `.gitignore` .

nickcharlton:
Hm, the intent is that we'd update `builds` once we cut a release. Which hasn't been a while.

How much of a problem is this creating when developing between releases?

goosys:
By updating modules with Dependabot, running `bin/dev` on the latest main branch causes changes in the `builds` output, resulting in differences.
I have to manually delete these changes each time to avoid including them in commits, which is a bit inconvenient.

nickcharlton:
Ah yeah, that'd do it. 

This feels like something we can automate in CI so that we catch it when working on the PR. I'll give that some thought.

nickcharlton:
That was easier than I thought. I wondered if I'd need to make some modifications to `diff-check` to do this, but it worked first time.

pablobm:
Another issue with not checking in the assets is that gems included from Git (eg: `gem "administrate", git: ...`) do not get the correct CSS. For example, I just built a quick Administrate app pulling from `main`, and it doesn't include the changes introduced at https://github.com/thoughtbot/administrate/pull/2558

nickcharlton:
I'm going to close this, but we need to make sure we start running `yarn build; yarn build:css`, so hopefully I can fix that action shortly.

PR #2674, opened 2024-10-04
Not the recommended way to provide custom assets

pablobm:
As discussed at https://github.com/thoughtbot/administrate/pull/2475, these generators no longer represent the recommended way to provide custom assets. Therefore they are deprecated and can be removed.


pablobm:
@elias19r - Does this look correct to you? (If you can remember :smile:)

elias19r (approved):
Hi! Trying to remember. But reading comments from the linked PR, I agree with our past selves :D

> I lean toward removing those generators and making users customize assets by appending to Administrate's CSS/JS, just like the example in this PR, instead of completely replacing them with their own.

Issue #2561, opened 2024-03-29
In title on object creation form name of object showing in plural form


nooncat:
https://administrate-demo.herokuapp.com/admin/product_meta_tags/new

![image](https://github.com/thoughtbot/administrate/assets/4406725/cbd11936-a1dd-470a-b945-efcd0ba28d17)

  - Rails 7.1.3
  - administrate 0.20.1
  
  Can you fix it please?

PR #2684, opened 2024-10-15
Singularize object creation form resource name.

Tresor11:
Currently, the resource name on the object creation form is pluralized

This commit adds a `singular: true' parameter
to the display_resource_name method in the
object creation view to explicitly show the
resource name in singular form.

Before:

318019918-cbd11936-a1dd-470a-b945-efcd0ba28d17

After:
Screenshot 2024-10-15 at 16 14 45

issue #2561


pablobm:
I think this is good, and I'm generally suspicious of `titleize` so that's a nice plus :+1: Thank you!

pablobm (approved):

PR #2691, opened 2024-10-25
Add asdf tool-versions

baldarn:
asdf is a multiple langs/tool version manager

having multiple langs here, is useful to enable asdf usage

see [asdf](https://asdf-vm.com/)


nickcharlton (approved):
Looks good!

We mentioned during Open Summit about adding yarn, I don't use asdf personally (although we do at thoughtbot in general), so that might be worth adding.

nickcharlton:
I'm using 1.22.22, but I think it'd be okay to also set the latest version.

Thrizian:
I use asdf too. Happy to see this one coming thanks. I had initially added it to my pr but decided to remove it, Happy i did. cause it´d be a conflict with this.

PR #2692, opened 2024-10-25
Add docker compose for services

baldarn:
Forcing developers to use services on them local machine is imho a bad practice

this enable bot the usage of a local postgres or a docker one, for the peace of mind of everyone XD

### DIFF
--- a/.sample.env
+++ b/.sample.env
@@ -2,3 +2,9 @@
 RACK_ENV=development
 SECRET_KEY_BASE=development_secret
 PORT=3000
+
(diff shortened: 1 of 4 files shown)


nickcharlton:
I much prefer to just run it locally instead of maintaining extra levels of abstraction, then rely on CI to provide an isolated environment.

I'll leave this open for today, as you've been using it during Open Summit, but I won't merge this in (it's been similarly added a few times over the years too and not been merged).

PR #2693, opened 2024-10-25
Generate `fields/rich_text/form` with documentation

seanpdoyle:
All form partials include guiding documentation atop the template files. Those comments are valuable additions, since view template overrides installed via generators will include them.

This commit updates the new `fields/rich_text/form` partial to include documentation.

On the test side, this commit also includes changes to view specs to generate Form Builders using *real* Action View form builders and view helpers.

Removing code that mimicked that builder instantiation uncovered an error in a how the Rich Text form partial coverage expected the `<trix-editor>` element's corresponding `<input type="hidden">` element generated its `[id]`.

### DIFF
--- a/app/views/fields/rich_text/_form.html.erb
+++ b/app/views/fields/rich_text/_form.html.erb
@@ -1,3 +1,19 @@
+<%#
+# Rich Text Form Partial
+
+This partial renders a trix-editor element for ActionText::RichText attributes.
+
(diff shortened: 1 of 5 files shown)


pablobm (approved):
Great stuff! :rocket: Those view helpers are always tricky and a bit of unknown territory. Glad to see them coming back to convention.

Issue #2546, opened 2024-03-11
How to type an attribute that returns a model but doesn't use a has_one association


mpvosseller:
My `Foo` model has a method called `bar` that returns a `Bar` model instance (or nil).

It is technically not a `has_one` association though because there is not foreign key. Rather the `bar` method uses some business logic to determine which `Bar` to return.

In the `Foo` dashboard, what "type" should I use for the `bar` attribute?

I found that I can use the "Field::HasOne" with the `class_name` option but that option is deprecated:
```ruby
  bar: Field::HasOne.with_options(class_name: "Bar"),
```

If I omit the `class_name` option Administrate crashes because (I think) it can't determine the type via reflection.
```ruby
  bar: Field::HasOne,
```

Thanks.

nickcharlton:
I'm not really sure what a good suggestion for this might be, aside from writing a custom field (which, perhaps is too much in this case?).

I will note #2384, though, where we've found other cases where `class_name` is necessary.

pablobm:
A custom field is probably a good option, particularly since https://github.com/thoughtbot/administrate/pull/2658

However, I think it's reasonable to do with a `HasOne` too :thinking:

nickcharlton:
@mpvosseller, how did you get on? Did a custom field work out here?

mpvosseller:
@nickcharlton I just kept it as a `Field::HasOne` for now and been hoping that the `class_name` option gets  undeprecated. 🤞

nickcharlton:
Okay cool! We merged in the undeprectation, so it'll be here to stay!

Issue #2384, opened 2023-05-24
"The option :class_name is deprecated." but I think it is still needed. 


gregmatthewcrossley:
* What were you trying to do?
I am defining some `ATTRIBUTE_TYPES` in a dashboard. One of my attributes (`lead_practitioners`) is a `belongs_to` relation where its class name doesn't match the relation name. My model defines this `belongs_to` relation like so:
```ruby
class Practice < ApplicationRecord
  has_many :practitioner_practices, inverse_of: :practice, dependent: :destroy
  has_many :practitioners, through: :practitioner_practices
  has_many :lead_practitioners, -> { where(practitioner_practices: {lead: true}) }, through: :practitioner_practices, source: :practitioner
  ...
```

* What did you end up with (logs, or, even better, example apps are great!)?
I am using the `class_name` option when I define the attribute in the dashboard file, like this:
```ruby
ATTRIBUTE_TYPES = {
   ...
    lead_practitioner: Field::BelongsTo.with_options(class_name: "Practitioner"),
    ...
 }
```
which works, but warns me with this:
```shell
DEPRECATION WARNING: The option :class_name is deprecated. Administrate should detect it automatically. Please file an issue at https://github.com/thoughtbot/administrate/issues if you think otherwise.
```

If I remove the `class_name` option and let the Administrate gem try to determine the class name on its own, I get this error:
```shell
ActionView::Template::Error:
       undefined method `klass' for nil:NilClass
     
               reflection(resource_class, attr).klass
 ```
 
 Even if I explicitly add the class name to the model relation, like so:
 ```
   has_many :lead_practitioners, -> { where(practitioner_practices: {lead: true}) }, through: :practitioner_practices, source: :practitioner, class_name: "Practitioner"
```
I get the same error as above.
* What versions are you running?
  - Rails: **7.0.4**
  - administrate: **0.18.0**

gregmatthewcrossley:
@jamie, I think this depreciation warning is from [your commit](https://github.com/thoughtbot/administrate/commit/2b6813035bbdfd5ab04a07b6bc5ba653da88010c). Just wondering if you agree with my assessment here.

jamie:
You're seeing `warn_of_deprecated_option`, not method - probably coming from https://github.com/thoughtbot/administrate/blob/44ad6ce1072fbd897f1faaf7231b3b5a0b92f535/lib/administrate/field/associative.rb#L40. You may want to try `resource_class` rather than `class_name`.

[...]

PR #2697, opened 2024-10-26
Undeprecate :class_name

pablobm:
If we are going towards v1.0.0, perhaps we should undeprecate `:class_name` :smile:

At https://github.com/thoughtbot/administrate/issues/2384 and https://github.com/thoughtbot/administrate/issues/2546, we have seen legitimate use cases for this option, so perhaps it should be back on the menu after all.

I'm not clear yet about the other deprecations that came with it though: `:foreign_key` and `:primary_key`. I think those might be ok? Has anyone come across use cases for them? Perhaps @littleforest?

While updating the documentation, I realised there's an example that uses `:class_name` and doesn't need it, so I have addressed it.

### DIFF
--- a/docs/customizing_dashboards.md
+++ b/docs/customizing_dashboards.md
@@ -106,11 +106,11 @@ For example:
 with this, you will be able to search through the column `name` from the
 association `belongs_to :country`, from your model.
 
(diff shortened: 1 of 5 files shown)


nickcharlton (approved):
Looks good!

PR #2700, opened 2024-11-01
Add feature-test coverage for Rich Text Field

seanpdoyle:
Follow-up to [#2411][]

Add Capybara-driven feature test coverage for filling in and rendering the contents of a `<trix-editor>` element.

[#2411]: https://github.com/thoughtbot/administrate/pull/2411

### DIFF
--- a/spec/features/products_form_spec.rb
+++ b/spec/features/products_form_spec.rb
@@ -3,7 +3,7 @@ require "rails_helper"
 describe "product form has_one relationship" do
   include ActiveSupport::Testing::TimeHelpers
 
-  it "saves product and meta tag data correctly" do
+  it "saves product and meta tag data correctly", js: true do
     visit new_admin_product_path
 
     fill_in "Name", with: "Example"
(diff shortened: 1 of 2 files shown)


spec/features/products_form_spec.rb
```diff
@@ -53,13 +61,29 @@
...
+      <div>A banner with a <a href="https://example.com">link</a>.</div>
+    HTML
 
     visit edit_admin_product_path(product)
 
+    within(:rich_text_area, "Banner") do
+      within("div", text: "A banner with a link") do
+        expect(page).to have_link("link", href: "https://example.com")
+      end
+    end
```
pablobm, line 75: Ah, perhaps the previous double-within initially came to be as a copy of this one? (Which does need to be double in this case).

spec/support/features/action_text.rb
```diff
@@ -0,0 +1,23 @@
+module Features
+  if Rails.version >= "6.1"
```
pablobm, line 2: Ah, just 0.1 requirements of a version away from doing away with this part :smile: I wonder if we should drop support for 6.* now that they have been EOL'd (https://rubyonrails.org/2024/10/15/new-maintenance-policy-and-eol-annouments). Thoughts, @nickcharlton?

But since this is here and solved for now, I think it's not a blocker. We can decide about Rails requirement separately.

pablobm (approved):
Just a minor comment.

spec/features/products_form_spec.rb
```diff
@@ -53,13 +59,29 @@
...
+    fill_in_rich_text_area "Banner", with: <<~HTML
+      <div>A changed banner without a link.</div>
+    HTML
     click_on "Update Product"
 
+    within(".trix-content") do
+      within("div", text: "A changed banner without a link.") do
+        expect(page).to have_no_link(href: "https://example.com")
+      end
+    end
```
seanpdoyle, line 84: This can be collapsed as well:

```suggestion
    within(".trix-content div", text: "A changed banner without a link.") do
      expect(page).to have_no_link(href: "https://example.com")
    end
```

spec/support/features/action_text.rb
nickcharlton: This is perhaps a separate discussion, but I wonder if we can drop support post-1.0 (1.1?).

I don't want 1.0 to be so big that it means people avoid upgrading (which is a good argument to call it in the next few weeks?), then drop it on the next release. It's a bit similar to how Apple does it with macOS.

PR #2735, opened 2024-12-31
fix(form): Display hint for has_one nested fields

mizoR:
Hi, thank you for maintaining this useful gem.

## What I'm having trouble with

In my use case, certain fields are used only as nested fields. However, the field_hints feature in administrate does not work for nested fields, making it impossible to add hints for these fields. This limitation impacts usability, as users cannot see helpful hints for nested fields to understand their purpose or usage.

## What I changed in this PR

Extend the functionality of field_hints to work with nested fields as well. With this change, hints can be displayed for nested fields, improving the consistency and usability of the UI.

```yaml
en:
  administrate:
    field_hints:
      product_meta_tag:
        meta_title: Input meta title here
```

image

- - -

Thank you for reviewing this pull request.


spec/administrate/views/fields/has_one/_form_spec.rb
```diff
@@ -2,21 +2,60 @@
 require "administrate/field/has_one"
 
 describe "fields/has_one/_form", type: :view do
-  it "displays the resource name" do
-    has_one = instance_double(
+  let!(:has_one) {
```
nickcharlton, line 5: [I would typically try and avoid using a `let`, let alone a `let!` as that might tell us something about the order dependency in the tests](https://thoughtbot.com/blog/lets-not).

nickcharlton (commented):
Can you also add a reference to field hints to the documentation?

Issue #2732, opened 2024-12-29
Dark mode


edwardloveall:
I was wondering if there was any interest a dark theme for Administrate. I couldn't find any other threads about this but point me there if I missed it.

I'm picturing that it would be a different set of colors, and a theme switcher.

nickcharlton:
I definitely think this is something we should have. I think the closest related issue is #128, which is about making the template responsive.

I think there are probably a couple of stages to doing this, first moving colours out to CSS variables to make it easier to do, then adding new themes. I would also be on the eye out for things we can improve around the CSS in general too (new features we can use, things we can now drop because browser support is so good, etc.)

If you'd be interested in starting working on something like this, that'd be great!

madhums:
Can't wait for this! 😍

PR #2737, opened 2025-01-01
Convert internal stylesheets to use css vars

edwardloveall:
Supporting #2732

This is in support of a future where we can easily set colors based on media queries. For example: dark mode.


edwardloveall:
You got it.

PR #2742, opened 2025-01-04
Ensure danger buttons have red hover state

edwardloveall:
074cb1a6 converted links that submit forms to buttons. To keep these now-buttons looking like links it uses the `link` class. CSS looks for buttons that are `not(.link)` to continue to style them appropriately.

However any `.button--danger` buttons did not get all the style overrides. In particular, `:hover` and `:disabled:hover` states fell back to the normal button background colors (blue).

This fixes those buttons to have the appropriate red hover states.

In addition, `.link--danger` elements also had a blue hover state. I'm not sure when this was broken, or if it was always supposed to be like that. Either way, I changed it so they now also appear slightly darker when hovered to match the buttons.


nickcharlton:
> +3,250 −3,342

lol

Could you rebase this when you get a chance? Otherwise it looks good!

edwardloveall:
Look, I just love removing 3.3k lines

PR #2705, opened 2024-11-08
Add support for Rails 8.0

nickcharlton:
We needed to:

* Bring over quite a few things from `rails app:update`, which hopefully should make future upgrades easier,
* Fix a change to the `enum` `ActiveModel` signature, which changed between Rails 6.0 and 7.0,
* Add a version check from 7.0 for `raise_on_missing_callback_actions`,
* Make a few changes to CI (on top of #2712) to adjust the Ruby versions we run CI against.

https://rubyonrails.org/2024/11/7/rails-8-no-paas-required

Closes #2703

PR #2744, opened 2025-01-06
Switch to Puma for Rack compatibility

### LINKED ISSUE #2730: Prerelease app is broken because of a Rack/Unicorn incompatibility
We're getting the following error log when trying to load `/admin` on the prerelease demo app:

```
2024-12-24T14:40:15.544423+00:00 app[web.1]: E, [2024-12-24T14:40:15.544391 #5] ERROR -- : app error: undefined method `=~' for ["_administrate-prototype_session=aW1pVHNZcmF0WG9zSHJ0UEFJQWI5cUxKRjVWS2JuUWZxZktMSENRclhEQTZlZzY2VVQzTEc3d1VVR21wU3RaRGU4cWtWazVOM29UdCt4VzFMU0lPMG5MZ1hXakc3RHpWV1ExcEdwZHdBWUdyQXozT2ZMQ3Y2RU1PVUhwcGZLWG9DT1pid05tSUJhdGJoSVpyVlovNUljYkpWMzRBK1hhUFdzQTdMRFNnb05vUTMvaTRXQzlCOUFQaE4wYjdCblNWLS02YldIejUrMzZnS3RwcWZBcDJrMEFnPT0%3D--3e7eabdc3aa263fc3c2c45256b73a7638b067b4e; path=/; secure; httponly"]:Array (NoMethodError)
```

[It sounds like it's an incompatibility with Rack 3 and Unicorn](https://stackoverflow.com/a/78594913), the last release of which was 6.1.0 on 2021-12-25.

Options:

1. Switch to Puma
2. Monkey-patch Rack 3 support in whilst we wait for a new Unicorn release

nickcharlton:
We're now using Rack 3, since upgrading to Rails 8.0 in #2705. There's an incompatiblity between Unicorn and Rack 3, which whilst it has a commit to fix it, hasn't been released yet.

Unfortunately it wasn't possible to bundle Unicorn directly from the repo, and building the gem locally didn't work despite a few attempts.

Instead, let's switch to Puma.

Fixes #2730.

https://stackoverflow.com/a/78594913
https://yhbt.net/unicorn.git/9d7bab0bc2211b20806d4d0289a7ea992e49a8a1/s/


nickcharlton:
There's no easy way to test compatibility with older Rails versions, and this is something that should only affect the demo app and documentation, so I'm not that concerned about it.

It's running now on Heroku for the prerelease and seems to be working acceptably.

PR #2755, opened 2025-01-12
Convert stylesheets to use css custom properties

edwardloveall:
Supporting https://github.com/thoughtbot/administrate/issues/2732

This converts all sass color variables (and a couple other variables) to use native css variables instead.

~There's a lot here! If you'd prefer, I can figure out how to break this up into smaller parts; maybe one per var or something. It also doesn't get rid of _all_ sass variables. I could go even farther and do that if we want. Just let me know.~

Almost all variables are split into their own commit for easier review.

I also made sure you can still override variables and I'm happy to report that I followed the [existing steps](https://administrate-demo-prerelease.herokuapp.com/customizing_page_views) and it worked great.

A related consideration is I know administrate is in its 1.0-beta phase. Is this a breaking change worth making this late in the process? It will surely break current sass variable overrides as those no longer exist to be overridden.

Finally, this is also a very large step in the direction away from sass. While there would still be some work to do to fully remove sass, it's a lot more possible if you'd want to do that (in a separate PR).


app/assets/stylesheets/administrate/components/_flashes.scss
```diff
@@ -1,19 +1,19 @@
 @each $flash-type, $color in $flashes {
   .flash-#{$flash-type} {
     background-color: $color;
-    color: mix($black, $color, 60%);
+    color: color-mix(var(--black) 60%, $color);
```
edwardloveall, line 4: These are technically the only css vars left. CSS (as far as I know) does not have any type of map equivalent. Even so, these get replaced with normal CSS vars in the end.

edwardloveall:
> Thanks! Yeah, it's a bit big!

Cool, I'll work on breaking it up.

> …have you seen a path to removing Sass entirely? That'd be fun.

I _think_ so. I haven't done a full audit or anything but css has come a long way. I think the blocking factor would be which old browsers does administrate need to support.

edwardloveall:
Alright here's a start with a much smaller scope. I see two ways forward:

* I ship just this PR and follow up with about 20 (give or take 10) of a similar size converting a couple vars at a time.
* I keep going on this branch so it's one var per commit until we've done em all, but at least you can review one commit at a time.

Preference?

nickcharlton:
One branch is better, I think, especially if it's easier for you.

app/assets/stylesheets/administrate/base/_typography.scss
```diff
@@ -30,15 +30,15 @@ a,
 ) {
   color: $action-color;
   text-decoration-skip-ink: auto;
-  transition: color $base-duration $base-timing;
+  transition: color var(--base-duration) var(--base-timing);
 
   &:hover {
-    color: mix($black, $action-color, 25%);
+    color: var(--action-color-active);
```
edwardloveall, line 36: This changed ever so slightly. There were two selectors that used `$action-color` as their base. This one mixes in 25% `$black` and `&:not(.link):hover` in `buttons.scss` mixes in 20% `$black`. I went with the 20% in both places so we only had one "it's like `$action-color` but tinted" variable because button backgrounds are more prominant than text hovers.

app/assets/stylesheets/administrate/components/_flashes.scss
edwardloveall, line 1: As noted in the commit message, this is _quite_ a bit bigger than it's scss counterpart. I was seeing if I could really remove all sass variables, and I did! Also, in about a year it's probably pretty safe to use `color-mix` and this gets _way_ smaller (as seen in the comment at the start of the file).

edwardloveall:
Alright, this is ready. I tried to call out a couple places that are a bit messy or surprising. However, this replaces all sass variables with css variables. It's slightly more verbose because sass can still do more than css at this point, but give it about a year and `color-mix` will be considered [baseline widely available](https://developer.mozilla.org/en-US/docs/Glossary/Baseline/Compatibility#baseline_badges) and it's probably pretty safe to switch.

nickcharlton:
Nice, this looks great!

I'm going to sit on it for a little bit, as I work out where it fits inside the release.

What else would we want to do for removing Sass?

madhums:
Is this on track with version 1.0.0 release?

madhums:
Can we merge this? :)

nickcharlton:
It looks like we've got some conflicts to fix, which @edwardloveall might be best placed to do.

Then as long as we check off the last few things on the list above, we'll be in a good place to merge it.

edwardloveall:
Rebased. 

> Then as long as we check off the last few things on the list above, we'll be in a good place to merge it.

What do you think about merging this as-is? [The list above](https://github.com/thoughtbot/administrate/pull/2755#issuecomment-2740187183) was for completely removing sass. I _could_ take it on in this PR, but there's a lot here already and all the changes are similar to each other.

nickcharlton:
Thank you!

Oh! Right, yes. I misread. Excellent. Yeah, we shouldn't do that in this PR for sure. I'll read over this again and then I think we can merge it.

nickcharlton:
Dammit, I left this too long and now it's got conflicts. @edwardloveall, could you do the honours?

---

I've also since come across [BaseWatch](https://basewatch.dev), and I'd made a note to come back to this in particular once I'd figured out a way to track when things are baseline available. I think for this PR we're good to go, but if we want to try out new features we can try out BaseWatch to be alerted when it's safe to be merged.

edwardloveall:
Sure thing @nickcharlton. Hopefully I got it to you in time for this week. If not, let me know if you want a reminder at some point.

edwardloveall:
The failing tests don't _look_ related, but I'm not too familiar with that part of the code. Something about using a Fiber for i18n in the example app?

nickcharlton:
Thanks! You did!

Mmm, I don't think they're related enough to worry about it. I'm going to merge this so it's (finally) done. Thanks for your hard work on this.

nickcharlton:
I left myself last week unsure how to merge this, but I've gone with not squashing to keep the whole history. It'll be a pain when I go to release it, but it's much more valuable long-term.

edwardloveall:
Yeah, it was annoying when I did it last week. FWIW, I've been using [jujutsu](https://www.jj-vcs.dev/latest/) and I handled the merge conflicts automatically using a little shell script while `bin/dev` was running:

```sh
while true; do
  jj next --edit && echo && sleep 1 && jj status
done
```

Issue #2750, opened 2025-01-09
Make collection filters discoverable


jared-thoughtbot:
### Current state

You can add collection filters to a dashboard and then search using them using the search bar on the index page.

![image](https://github.com/user-attachments/assets/eaaf0216-e1b3-4c05-8875-d317e2dba0f2)

### Problem

These collection filters are not discoverable. You have to know about them already to know how to use them.

### Potential solution

Add a tooltip next to the search box when collection filters are provided. Probably with an option to disable the tooltip in case some people do not want it.

We could dynamically populate the content based on the collection filters. Eg.

![image](https://github.com/user-attachments/assets/f56e065c-d84c-42ff-a7e7-0728bd86c7de)

I appreciate we might not want to introduce tooltip code and assets just for this one thing. I don't know if there are other issues / future features on the roadmap that could also benefit from tooltips?

### Alternatives

- Using the input placeholder or somehow inlining the filter options - but these don't scale very well if you have a few filters
- Linking to a new page which lists the filter options
- Something else?

kevinkimball:
Big +1 to this feature request. This is one of the biggest things I feel is missing, and not straightforward to implement yourself. 

Additional features that would make this even better:
1. Option to provide dropdown options for filters similar to `Field::Select.with_options(collection: [...])` 
2. Option to provide query for autocomplete for filters

nickcharlton:
I think this is an excellent idea.

As I started reading, I was thinking about the placeholder being a good solution, but when you've got several it's not so nice. But I wonder if we can randomly show a different filter in there, or something!

[Tooltips are a bit easier to achieve now](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover), so I think writing our own here would be the right approach to take. We could then use this to provide tooltips in other areas of the project where it makes sense.

A good place to start, if someone wanted to pick this up, would be to see if we've got the right sort of data accessible in the view to populate a tooltip, and then start exploring a tooltip implementation we could ship as part of Administrate (it shouldn't involve any dependencies, use modern standards and as little JS as we can get away with.)

@kevinkimball, thanks! Those are good ideas too. Could you create different issues to track those? Especially so if you could do a mockup like @jared-thoughtbot did.

PR #2758, opened 2025-01-17
Add tooltip to search

jared-thoughtbot:
You can add collection filters to a dashboard and then use them in the
search bar on the index page. eg. `email:jared@example.com`

However, these collection filters are not discoverable. You have to know
about them already to know how to use them.

This change introduces a tooltip which lists the available collection
filters, allowing all users to discover what is available.

The tooltip icon is `help-circle` from [Feather icons] (MIT licensed)

We use [anchor positioning] with a [polyfill] to position the [popover] by
the tooltip.

Some basic JavaScript is introduced to show/hide the popover on hover to
ensure the tooltip behaviour works as expected for a tooltip.

I've had a first pass at the design, borrowing from existing styles. But
I don't feel strongly about it if we want to go in a different direction.

[Feather icons]: https://feathericons.com
[anchor positioning]: https://developer.mozilla.org/en-US/docs/Web/CSS/position-anchor
[polyfill]: https://github.com/oddbird/css-anchor-positioning
[popover]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover

Closes https://github.com/thoughtbot/administrate/issues/2750

search-tooltip

search-tooltip-modal


nickcharlton:
Nice! Thanks for opening this!

I think to avoid custom JavaScript for the anchor positioning, we should [use a polyfill instead](https://github.com/oddbird/css-anchor-positioning).

app/assets/javascripts/administrate/controllers/index.js
```diff
@@ -5,3 +5,13 @@ import TableController from "./table_controller";
 
 application.register("select", SelectController);
 application.register("table", TableController);
+
+// Bug: Visit another page and come back to this page, the mouesover popover will not work
```
jared-thoughtbot, line 9: I'm not sure why this doesn't work after navigating to another page and back. Is it something to do with Turbo page loads?

app/assets/stylesheets/administrate/components/_buttons.scss
```diff
@@ -61,3 +61,15 @@ button,
...
+
+.button--tooltip {
+  background: none;
+	border: none;
+	color: inherit;
+	cursor: pointer;
+  padding: 0;
+
+  &:hover {
+    background-color: unset !important;
```
jared-thoughtbot, line 73: The linter doesn't like `!important` here. This whole class is used to effectively unset all the default styles on buttons because a [popover](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/popover) need to use a button from what I understand

app/assets/stylesheets/administrate/components/_search.scss
```diff
@@ -44,3 +44,38 @@ $search-icon-size: 1rem;
...
+  position-anchor: --tooltip-anchor;
+  top: anchor(bottom);
+  transform: translateX(-50%);
+  
+  background-color: $blue;
+  border-color: $blue;
+  border-radius: $base-border-radius;
+  color: $white;
+  padding: 2rem;
+  width: max-content;
```
jared-thoughtbot, line 76: The first set of styles is all to do with positioning the popover, the second set (after the gap) is all to do with the visual style of the popover. The linter doesn't like the gap and the non alphabetical between them. It feels useful to separate the styles as I have. Shall I created 2 separate classes?

app/views/administrate/application/_index_header.html.erb
```diff
@@ -2,6 +2,16 @@
...
 <% end %>
 
+<%#
+  TODO: Move where this polyfill lives. Where?
+  Also, do we want to import the polyfill js directly instead of linking to unpkg.com?
+ %>
+<script type="module">
+  if (!("anchorName" in document.documentElement.style)) {
+    import("https://unpkg.com/@oddbird/css-anchor-positioning");
+  }
```
jared-thoughtbot, line 12: - Where should this polyfill live?
- Do we want to import the polyfill js directly instead of linking to unpkg.com?

app/controllers/administrate/application_controller.rb
```diff
@@ -10,12 +10,14 @@ def index
...
       page = Administrate::Page::Collection.new(dashboard, order: order)
+      filters = Administrate::Search.new(scoped_resource, dashboard, search_term).valid_filters
 
       render locals: {
         resources: resources,
         search_term: search_term,
         page: page,
-        show_search_bar: show_search_bar?
+        show_search_bar: show_search_bar?,
+        filters: filters
```
jared-thoughtbot, line 20: Do we want an option to disable the tooltip even if there are collection
filters?

app/assets/javascripts/administrate/controllers/index.js
nickcharlton: Yeah, most likely. I can never remember the precise thing, but you'll want to look up certain page hooks we'll need to implement for it to behave right

app/views/administrate/application/_index_header.html.erb
nickcharlton: Once upon a time, there used to be a top-level `vendor` directory. We could either do that, or put it in a vendor directory beside the other JS.

We definitely should bundle it, rather than pull from an external source.

app/assets/stylesheets/administrate/components/_search.scss
nickcharlton: _Potentially_, worst case we can merge them.

app/controllers/administrate/application_controller.rb
nickcharlton: We've generally tried to go for sensible defaults, rather than provide too many options.

But if you felt it warranted it still, I'd think of a way to provide a method in one of the generated controllers which would let you disable it.

nickcharlton:
It looks like the test failure is a deprecation change:

```
1) Admin::OrdersController with deprecated Punditize concern GET index shows only your own orders
     Failure/Error:
       expect(Administrate.deprecator).to have_received(:warn)
         .with(/`resolve_admin` method is deprecated/)

       (#<ActiveSupport::Deprecation:0x00007fd0a6aa6968 @gem_name="Administrate", @deprecation_horizon="1.0.0.beta3", @silenced=false, @debug=false, @silence_counter=#<Concurrent::ThreadLocalVar:0x00007fd0a91b56d0 @default_block=nil, @default=0, @index=27>, @explicitly_allowed_warnings=#<Concurrent::ThreadLocalVar:0x00007fd0a91b52e8 @default_block=nil, @default=nil, @index=28>>).warn(/`resolve_admin` method is deprecated/)
           expected: 1 time with arguments: (/`resolve_admin` method is deprecated/)
           received: 2 times with arguments: (/`resolve_admin` method is deprecated/)
     # ./spec/controllers/admin/orders_controller_spec.rb:151:in 'block (4 levels) in <top (required)>'
```

app/assets/javascripts/administrate/controllers/index.js
jared-thoughtbot: Solved this by using a Stimulus controller https://github.com/thoughtbot/administrate/pull/2758/files#diff-2a17b4666ae46ef340c4562870ce1860aaac86acc059ed074a9576df7d361026

app/assets/stylesheets/administrate/components/_search.scss
jared-thoughtbot: I split them https://github.com/thoughtbot/administrate/pull/2758/files#diff-3d467ae3b5c2ca48c6546b64cd6c7461f380c4673cf4ee7f2d9dbec4bb439ad6

app/views/administrate/application/_index_header.html.erb
jared-thoughtbot: Popped it in `app/assets/javascripts/administrate/vendor/css-anchor-positioning.js` https://github.com/thoughtbot/administrate/pull/2758/files#diff-87bc008071aee86e06b4ba36761965aaf136742a932fadf004b6b4067ec1fe5c

app/controllers/administrate/application_controller.rb
jared-thoughtbot: I'm thinking perhaps it doesn't need an option. If there are no collection filters or search is disabled then it doesn't show anyway. If someone really wants it we can react to that.

app/assets/stylesheets/administrate/components/_buttons.scss
```diff
@@ -65,3 +65,16 @@ button,
...
 }
+
+.button--tooltip {
+  background: none;
+	border: none;
+	color: inherit;
+	cursor: pointer;
+  padding: 0;
+
+  /* stylelint-disable declaration-no-important */
```
jared-thoughtbot, line 76: Unsure if this is ok. This whole class is used to effectively unset all the default styles on buttons because a popover needs to use a button

spec/controllers/admin/orders_controller_spec.rb
```diff
@@ -150,6 +150,7 @@ def pundit_user
         expect(locals[:resources]).to contain_exactly(order1, order3)
         expect(Administrate.deprecator).to have_received(:warn)
           .with(/`resolve_admin` method is deprecated/)
+          .at_least(:once)
```
jared-thoughtbot, line 153: I solved the [failing test](https://github.com/thoughtbot/administrate/pull/2758#issuecomment-2663462642) by doing this

.github/workflows/codeql-analysis.yml
```diff
@@ -34,6 +34,7 @@ jobs:
           config: |
             paths-ignore:
               - 'app/assets/builds'
+              - 'app/assets/javascripts/administrate/vendor'
```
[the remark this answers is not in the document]
nickcharlton: I think that's valid here. It's going to be the same cause as the other problem.

nickcharlton (commented):
This looks good!

I had thought that perhaps using `<>` to surround the value to filter was the wrong choice based on what I'm used to seeing on terminal applications. Then I looked and realise I'm wrong; it's `[]` when the value is optional.

My only other thought was that the blue is quite …bold, but also maybe it's fine. I don't feel strongly about it.

I'm going to make a change to our default icon set, then we can merge this in!

app/views/administrate/application/_icons.html.erb
```diff
@@ -10,4 +10,10 @@
   <symbol id="icon-up-caret" viewBox="0 0 48 48">
     <path d="M2.988 33.02c-1.66 0-1.943-.81-.618-1.824l20-15.28c.878-.672 2.31-.67 3.188 0l20.075 15.288c1.316 1.003 1.048 1.816-.62 1.816H2.987z" />
   </symbol>
+
+  <symbol id="icon-question-mark" viewBox="0 0 24 24" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
```
nickcharlton, line 14:
```suggestion
<symbol id="icon-question-mark" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
```

I just noticed that we're duplicating the `viewBoxz

jared-thoughtbot: Doh! Thanks

jared-thoughtbot: Do you prefer that I squash this or will you just do when you merge it in?

nickcharlton:
Thanks for keeping this updated!

PR #2760, opened 2025-01-21
Update Bundler to v2.6.3

nickcharlton:
https://github.com/rubygems/rubygems/releases/tag/bundler-v2.6.3


nickcharlton:
We can't jump to this version of Bundler yet because it doesn't support every version of Ruby/Rails that we do.

(I did this [through a script I have](https://github.com/nickcharlton/dotfiles/blob/main/bin/update-bundler) originally, because I thought a problem I was facing might be bundler related. It wasn't.)

PR #2777, opened 2025-02-13
Add guide for switching templates with the :variant option and add some samples

### LINKED ISSUE #2578: view variants support
* What were you trying to do?
add a view variant, example:

```
$ ls -lah app/views/admin/courses
drwxr-xr-x  7 casa  staff   224B May  4 02:17 .
drwxr-xr-x  4 casa  staff   128B May  4 01:00 ..
-rw-r--r--  1 casa  staff   2.0K May  4 02:38 show+student.html.erb
-rw-r--r--  1 casa  staff   2.0K May  4 02:38 show.html.erb
```

currently in the main controller, cant pass a variant, **seems variant: :value** not works
```ruby
# main controller
    def show
      render locals: {
        page: Administrate::Page::Show.new(dashboard, requested_resource)
      }
    end
```
so I implemented something like this in my generated controller
```ruby
# generated controller
    def show 
      variant = ""
      if current_user.is_student
        variant = "+student"
      end
      page = Administrate::Page::Show
      render "show#{variant}", locals: {
          page: Administrate::Page::Show.new(dashboard, requested_resource)
      }
    end
```
I guess we can add some like

```ruby
# main controller
    def show
     variant = ""
     if !request.variant.nil? 
       variant = "+#{request.variant.to_s}"
     end
      render render "show#{variant}"locals: {
        page: Administrate::Page::Show.new(dashboard, requested_resource)
      }
    end
# generated controller
    def show
      request.variant = :student
      super
    end
```

goosys:
- resolve #2578 
- refs https://guides.rubyonrails.org/layouts_and_rendering.html#the-variants-option

I have added a guide on how to switch templates using the variant option.

http://localhost:3000/guides/switching_template_with_the_variant_option.md

Additionally, I have added actual usage examples to the sample application.
I created two examples: one for the template used in the Dashboard and another for non-Dashboard screens. The display changes depending on whether the user is logged in as an Admin or another user.

http://localhost:3000/admin/customers
http://localhost:3000/admin/stats

### DIFF
--- a/docs/guides.md
+++ b/docs/guides.md
@@ -5,3 +5,4 @@ title: Guides
 - [Hiding Dashboards from the Sidebar](./guides/hiding_dashboards_from_sidebar)
 - [Customising the search](./guides/customising_search)
 - [Scoping HasMany Relations](./guides/scoping_has_many_relations.md)
+- [Switching templates with view variants](./guides/switching_templates_with_view_variants.md)
--- a/docs/guides/switching_templates_with_view_variants.md
+++ b/docs/guides/switching_templates_with_view_variants.md
@@ -1,0 +1,45 @@
+---
+title: Switching templates with view variants
+---
+
+You can switch to different templates using Rails' support for [view
+variants][], which can be used to override the template used inside the
+controller.
+
+By setting the `request.variant` option to any value (in this case, `:admin`)
+at any point in any controller, you can add a new variant to Rails' template
+lookup tree (in this case, `.html+admin.erb`).
+
+For example, to add a button to become the admin to view certain functionality:
+
+```ruby
+class CustomersController < Admin::ApplicationController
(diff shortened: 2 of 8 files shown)


nickcharlton:
I made a few minor rewording changes, and some style preferences, as I thought I'd be able to get this merged today.

But, testing this out just before, I realised that we don't have an obvious way to _not_ be the admin in the demo app to use this functionality. Any ideas on how we could do that?

goosys:
Thank you.

To test it, you can switch to a non-admin user by clicking the "Become" button in each row on the customer list page. 
After switching, you should be able to test it.

nickcharlton:
Hah! I totally missed that.

I'm glad someone knows the application better than I do after maintaining it for not far off a decade 😂

PR #2781, opened 2025-02-19
Add administrate default format

goosys:
- refs #2702 
- The only way to change the time format was to modify the Rails default setting. However, changing it would also affect other parts of the application outside the admin panel. Therefore, I added a dedicated format specifically for Administrate.
  - `administrate_date_default` `administrate_datetime_default` `administrate_time_default`
- Currently, the Time format is fixed as `"%I:%M%p"`, but in Japanese, we would prefer the order `"%p %I:%M"`. I believe other languages may also have their own preferred orders. I have made it possible to change this for each language using `administrate_time_default` .

### DIFF
--- a/lib/administrate/field/date.rb
+++ b/lib/administrate/field/date.rb
@@ -6,14 +6,23 @@ module Administrate
       def date
         I18n.localize(
           data.in_time_zone(timezone).to_date,
-          format: format
+          format: format,
+          default: fallback_format
         )
       end
 
       private
 
       def format
-        options.fetch(:format, :default)
+        options.fetch(:format, default_format)
+      end
+
+      def default_format
+        :administrate_date_default
+      end
+
+      def fallback_format
+        I18n.t("date.formats.default", default: "%Y-%m-%d")
       end
 
       def timezone
--- a/lib/administrate/field/date_time.rb
+++ b/lib/administrate/field/date_time.rb
@@ -6,21 +6,35 @@ module Administrate
       def date
         I18n.localize(
           data.in_time_zone(timezone).to_date,
-          format: format
+          format: format(type: :date),
+          default: fallback_format(type: :date)
         )
       end
 
       def datetime
         I18n.localize(
           data.in_time_zone(timezone),
-          format: format
+          format: format(type: :datetime),
+          default: fallback_format(type: :datetime)
         )
       end
 
       private
 
-      def format
-        options.fetch(:format, :default)
+      def format(type: :date)
+        options.fetch(:format, default_format(type: type))
+      end
+
+      def default_format(type: :date)
+        :"administrate_#{type}_default"
+      end
+
+      def fallback_format(type: :date)
+        if type == :date
+          I18n.t("date.formats.default", default: "%Y-%m-%d")
+        else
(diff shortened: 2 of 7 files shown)


goosys:
At this point, there is a backward compatibility issue.  

Since `administrate_*_default` does not exist in everyone's Rails setup, some initial configuration is required to start using this feature. I feel bad about forcing this initial setup on everyone, so I’d like to discuss if there’s a better way.  

1. Write the configuration values for each language in all `config/locales/administrate.*.yml` files.  
2. Create `config/locales/date.yml` and include only the default (en) settings.  
3. Use `I18n.localize` with `default: [:"time.formats.default"]` to enable fallback.  

Option 1 is very tedious. Option 3 would likely work, but it introduces unnecessary processing if the initial configuration is already set up, so I’d prefer to avoid it if possible.
What do you think?

pablobm:
I think we can start looking into merging this. I have a couple of questions.

First: should the format keys live under `administrate` instead of the root translations file? So `administrate.date.formats.default` instead of `date.formats.administrate_date_default`.

Although I see that this will cause `spec/i18n_spec.rb` to fail until the key is present in all our translation files... not sure what to do about that, but I think that's an issue with how we handle translations, instead of this specific PR.

Second: of the options you give, my preferred is option 3, as it offers easy back-compatibility. When you say it involves unnecessary processing, do you mean in terms of computing performance? If that's so, I think it's not a concern here, as it would be negligible.

But perhaps both problems can be solved by sourcing the format strings from somewhere. Is there's a project out there (with a compatible license) where we can respectfully "steal" them from?

goosys:
@pablobm 
> First: ...

Yes.

`date.formats.administrate_date_default` is the correct place for it.

This is because the format is used by `I18n.l`, not `I18n.t`. Since `I18n.l` looks up formats under `date.formats`, the translation needs to be defined there.

goosys:
> do you mean in terms of computing performance? 

Yes

> If that's so, I think it's not a concern here, as it would be negligible.

OK

> Is there's a project out there (with a compatible license) where we can respectfully "steal" them from?

It looks like we could fall back to defaults provided by `rails-i18n` for `date` and `datetime`.

For `time`, `rails-i18n` does not seem to provide a default format without a date component. However, it should be possible to generate one mechanically from `ruby-cldr`.

https://github.com/svenfuchs/rails-i18n/blob/master/rails/locale/de.yml#L227
https://github.com/ruby-i18n/ruby-cldr

---

Would it be okay if I proceed with the remaining work as follows?

1. Do not add translations for `date.formats.administrate_*_default` in this PR.
2. Add the `default` option to `I18n.l` so that `administrate_*_default` can be used when available.

goosys:
@pablobm 
I've rebased and made some additional changes. Please take another look.

For the fallback format, I set a simple format that doesn't rely on things like `%a: date.abbr_day_names`. This is only intended to prevent errors on the screen, and in normal cases it shouldn't be used if `rails-i18n` is installed. If the format here feels off, I'm happy to adjust it.

pablobm:
I think this is good. Can you rebase, please?

goosys:
Rebased!

pablobm:
Thank you! 🙌

PR #2787, opened 2025-02-22
Fix Become button on demo app

goosys:
Fix Become button appearing in `HasMany` Field on `customers.show` demo app.

Before:

image

After:

image


nickcharlton:
Thanks!

Issue #2545, opened 2024-03-10
Selectize removes the blank options from selects


thebravoman:
I have the model 

```
class RssEntry < ApplicationRecord
  belongs_to :narrative, optional: true
```

In the dashboard I have 
```
  ATTRIBUTE_TYPES = {
    ...
    narrative: Field::BelongsTo.with_options(include_blank: true),
    ...
  }.freeze
```

When the html page is delivered there is blank options (looking at the network tab and the response from the server)
```
<div class="field-unit__field">
  <select data-controller="select" name="rss_entry[narrative_id]" id="rss_entry_narrative_id">
    <option value="" label=" "></option>
    <option selected="selected" value="1">Narrative #1</option></select>
</div>
```

But then selectize is called in 
select_controller.js as 
```
import { Controller } from "@hotwired/stimulus";
import $ from "jquery";

export default class extends Controller {
  connect() {
    $(this.element).selectize({});
  }
};
```

this changes the html element to 
```
<form class="form" id="edit_rss_entry_2" action="/disi_client/rss_entries/2" accept-charset="UTF-8" method="post">
  <input type="hidden" name="_method" value="patch" autocomplete="off">
  <input type="hidden" name="authenticity_token" value="..." autocomplete="off">
  <fieldset class="">
    <div class="field-unit field-unit--belongs-to field-unit--optional">
      <div class="field-unit__label">
        <label for="rss_entry_narrative_id-selectized">Narrative</label>
      </div>
      <div class="field-unit__field">
        <select data-controller="select" name="rss_entry[narrative_id]" id="rss_entry_narrative_id" tabindex="-1" class="selectized" style="display: none;">
          <option value="1" selected="selected">Narrative #1-USR</option>
        </select>
        <div class="selectize-control single">
          <div class="selectize-input items full has-options has-items">
            <div class="item" data-value="1">Narrative #1-USR</div>
            <input type="select-one" autocomplete="off" tabindex="" id="rss_entry_narrative_id-selectized" style="width: 4px; opacity: 0; position: absolute; left: -10000px;">
          </div>
          <div class="selectize-dropdown single" style="display: none; visibility: visible; width: 704px; top: 36px; left: 0px;">
            <div class="selectize-dropdown-content">
              <div class="option selected" data-selectable="" data-value="1">Narrative #1-USR</div>
            </div>
          </div>
        </div>
      </div>
    </div>
  </fieldset>
  <div class="form-actions">
[...]

PR #2790, opened 2025-02-25
Add allow empty option

goosys:
I investigated the issue mentioned in #2545.

Strictly speaking, include_blank is not completely broken—it does exist as a feature and works correctly, but it is not visually reflected, making it unintuitive.

In a typical select dropdown, you would select the empty option at the top to cancel a selected option. However, in Selectize, the expected way to remove a selection is by pressing "Backspace."
While this behavior is correct once you know about it, it is not very intuitive.

---

To make it more intuitive, I added `allowEmptyOption` to Selectize.  
I tested it in several cases, and it seems safe to keep this set to `true` at all times. For selects without `include_blank` (e.g., `customer.kind`), the blank option did not appear, so it should be fine.  

However, the blank option's height was too narrow, making it difficult to click. To address this, I adjusted the CSS.  
(Since there was no dedicated CSS for Selectize, I added the styles to the form. Please let me know if there is a more appropriate place for them.)  

Next, I made some changes to allow verification in the demo app.  
Since `product.release_year` was already defined in the Model to allow blank values, I added `include_blank` in the Dashboard as well.  

image

For `customer.territory`, which is a `BelongsTo` association, I made some modifications to verify that it behaves the same as a regular select.  
Since we have already confirmed that `true` works for other selects, I specified a string for `prompt` to check how it behaves when passing a string value.  

image

---

I consider this fix to be temporary.  
In the latest version of Selectize (`@selectize/selectize`), new options related to empty selections have been introduced, and CSS for blank options also seems to be prepared.  

Therefore, rather than making extensive modifications, I believe it is sufficient to adjust the appearance with minimal changes.


nickcharlton:
> In the latest version of Selectize (@selectize/selectize), new options related to empty selections have been introduced, and CSS for blank options also seems to be prepared.

Would we be better off in upgrading Selectize to use these new options?

goosys:
I think it's better to update to the latest version, but I saw a comment in another issue suggesting there might be a reason not to update. I assumed there might be some specific circumstances.  

If there are no issues, I’d like to go ahead with the update.

nickcharlton:
Let's update first and we can see what breaks. Thanks!

Issue #1994, opened 2021-06-02
Adding I18n translations for polymorphic fields


DavidGeismarLtd:
* What were you trying to do?
Translations for a polymorphic field item. here's an example. Item can be a program or a session and both are translated in the locale.yml file
I have a `GroupItemDashboard` : 
```
  ATTRIBUTE_TYPES = {
    # cannot be made searchable ...
    item: Field::Polymorphic.with_options(
      classes: [Program, Session],
      order: "lms_data ->'name' ASC"
    ),
    group: Field::BelongsTo.with_options(
      searchable: false
    ),
}
  FORM_ATTRIBUTES = %i[
    item
    group
  ].freeze

In my locale.yml file I have : 
```
 activerecord:
          models:
              item:
                one: Session Ou Parcours
                other: Sessions ou Parcours
              program:
                  one: Parcours
                  other: Parcours
            session:
                  one: Session
                  other: Sessions
              group:
                one: Groupe
                other: Groupes
```
However, those translations for polymorphic fields are never taken into account. Neither for the input label or within the select options (see here label is "item", and select options refer to "programs" and not "parcours") : 
<img width="1059" alt="Screenshot 2021-06-02 at 09 27 04" src="https://user-images.githubusercontent.com/68914479/120441332-b97cdf00-c384-11eb-8f87-d6858a5368a3.png">


Is there any way to add a translation for a polymorphic field ?
```

* What did you end up with (logs, or, even better, example apps are great!)?
* What versions are you running?
  - Rails 6.0.3
  - administrate 0.15.0

pablobm:
Hm, on first inspection, I think the problem stems from us not applying i18n in `Field::Polymorphic#associated_resource_grouped_options` (specifically the `klass.to_s` in line 12):

https://github.com/thoughtbot/administrate/blob/b3bb0a3b7833deceff71cfe1687ac36cec12a6e8/lib/administrate/field/polymorphic.rb#L10-L16

Perhaps we should be using `#model_name` instead of of `#to_s`. This would break if for some reason the field doesn't use `ActiveModel` or something compatible though... but perhaps that's not a concern here.

PR #2797, opened 2025-03-06
Allow translations of active record model names in polymorphic dashboard

heyvaleria:
Addresses https://github.com/thoughtbot/administrate/issues/1994

In polymorphic associations we were not displaying the translations even when present.
By replacing `to_s` with `model_name.human` we preserve the reference to the activerecord objects, and we are able to show the translated values.

### Screenshots
model_name-human
en-yml

If I remove translations for recipe, we get the model name `Recipe` where above you can see `Ricetta`
no-translation

I'm gonna see if I can add a spec.

Thanks @pablobm for the recommendation on the method to apply.

### DIFF
--- a/lib/administrate/field/polymorphic.rb
+++ b/lib/administrate/field/polymorphic.rb
@@ -9,7 +9,7 @@ module Administrate
 
       def associated_resource_grouped_options
         classes.map do |klass|
-          [klass.to_s, candidate_resources_for(klass).map do |resource|
+          [klass.model_name.human, candidate_resources_for(klass).map do |resource|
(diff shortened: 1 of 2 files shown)


pablobm:
Grazie! Do you think you can add that spec then?

heyvaleria:
Hi @pablobm 
I was on PTO and am just back. 
I have not forgotten about this. 😄 
I'll send an update ASAP.

heyvaleria:
I'll look at the failures in the morning

nickcharlton:
Thanks!

PR #2798, opened 2025-03-07
Add quickstart script for Codespaces environment

goosys:
How about recommending the use of Codespaces to make contributions easier?  

With this script, we can launch a clean environment for any branch in just three minutes and discard it when it's no longer needed.  

To set up the environment, run the following command:  

```sh
source bin/setup-codespaces
```

What do you think?  
Let me know if there are any other necessary tasks, such as updating the README.


nickcharlton:
It's funny that you add this today, as that's what I've been working on for much of this week.

We're trying to see if we can use Codespaces for certain client projects to reduce the onboarding burden, and if it works well enough it makes sense to support it here too.

I'm still deep in the middle of figuring out what the ideal setup might be like, but I'll come back to this soon.

nickcharlton:
Ooh, that's really good to know, thanks!

I had some comments of previous problems, but no one's been clear enough with me what they actually struggled with.

goosys:
One small issue remains. Currently, since only Foreman loads Dotenv, any command other than `bin/dev` requires manually exporting the ENV variable to access the database:

```sh
export DATABASE_URL=postgresql://postgres:@localhost
```

This needs to be set for each terminal session, which is a bit inconvenient, but once we get used to it, it’s not a big deal.

Do you have any ideas?

nickcharlton:
Oh, interesting. Are you configuring your own [dotfiles](https://docs.github.com/codespaces/setting-your-user-preferences/personalizing-github-codespaces-for-your-account#dotfiles)? One option would be to use something like [`direnv`](https://direnv.net)

bin/setup-codespaces
```diff
@@ -0,0 +1,21 @@
...
+
+rvm install $(cat .ruby-version)
+rvm --default use $(cat .ruby-version)
+ruby -v
+
+docker run -d -i --name postgres -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust postgres
+export DATABASE_URL=postgresql://postgres:@localhost
+
+./bin/setup
+./bin/setup
```
pablobm, line 18: Is `setup` running twice by accident or on purpose?

pablobm: Oh, of course. Thank you. How about a comment to remove the confusion?

goosys: Added comments!

goosys:
> Are you configuring your own dotfiles?

No, not really.

The simplest way is to add the `export` to `.bashrc`.  
Normally, it's not ideal to modify this file, but since this is a temporary environment, keeping it simple might be fine.  

```sh
echo "" >> ~/.bashrc && echo "export DATABASE_URL=postgresql://postgres:@localhost" >> ~/.bashrc
```

nickcharlton:
Ah, interesting.

I've so far gone down the path of figuring out how to setup a `.devcontainer` (which is also going to need a `docker-compose.yml`) and that does at least mostly feel like a better approach, but the journey there has been generally ...unpleasant. I should have something to share soon.

goosys:
Just a quick update:

The devcontainer image has changed again since my earlier comments.  
With the latest default image (universal:2), Chrome and the WebDriver are no longer included.

https://github.com/devcontainers/images/tree/main/src/universal/.devcontainer

On the positive side, Codespaces now supports multiple devcontainer configurations.

Given these changes, it might be a good time to revisit this topic again.

nickcharlton:
Oh interesting. It's still something I'd wanted to revisit, but the conclusion I came to last was that the best approach was to build a custom image because otherwise spinning up an environment is just too slow.

(Relatedly, does anyone know of a testing tool for containers with a good set of examples? I am incapable of publishing anything which doesn't have comprehensive tests and eight months ago I figured I'd have to write one.)

goosys:
Codespaces’ `universal` image is a 20GB “everything included” image, but it starts up very quickly when running inside Codespaces.

What’s missing for Administrate is only a browser (Chrome) and the Ruby version specified in `.ruby-version`.  
If we base a Dockerfile on the `universal` image and install just those missing pieces, we can keep the startup time minimal.

(The only caveat is that we should clearly communicate that contributors should not use this image locally.  
Pulling a 20GB image to a local machine would be far too heavy.😅)

goosys:
This is my latest setup script.

https://gist.github.com/goosys/2872b39df6727603a64155743253723b

goosys:
I’ve pushed the devcontainer version as well.

The base image is `universal:4`, which is already cached on Codespaces, so it requires no pull time at all. On top of that, it installs the Ruby version specified in `.ruby-version` and also installs a browser.

These additional installations take a bit of time, but once the environment is up, `rspec` runs immediately, so I think it provides a convenient workflow.

Issue #2291, opened 2022-11-25
Multiple "looks" per Field


ammancilla:
### What would you like to be able to do? Can you provide some examples?

> As a developer I can define multiple looks for a `Field` and determine which one to use when using the field.

For example, I have a dashboard with three `Field::HasMany` attributes. For one of them I want to use the "default look" (the set of partials provided by `administrate`) but for the other two I want to use two different "looks", defined by me.

### How could we go about implementing that?

Introduce the concept of **looks (or variants?)**.  A **look** is a group of partials that determine how a `Field` is displayed in different pages. There is a `default` look provided by `administrate` (the existing partials for each field) and developers can create, within their applications, as many custom looks as needed.

```bash
app/
├─ views/
│  ├─ fields/
│  │  ├─ belongs_to/
│  │  │  ├─ looks/
│  │  │  │  ├─ default/
│  │  │  │  │  ├─ _form.html.erb
│  │  │  │  │  ├─ _show.html.erb
│  │  │  │  │  ├─ _index.html.erb
│  │  ├─ has_many/
│  │  │  ├─ looks/
│  │  │  │  ├─ default/

...
```

**Creating custom looks**
A custom look is created by adding a new folder (named after the look) under `app/views/fields/<field_name>/looks/` with the respective partials to render a Field. A custom look might define all or a subset of the three partials needed for a Field. In case of missing partials, the ones defined in the default look are used as fallback.

```bash
app/
├─ views/
│  ├─ fields/
│  │  ├─ belongs_to/
│  │  │  ├─ looks/
│  │  │  │  ├─ simple/
│  │  │  │  │  ├─ _index.html.erb
│  │  │  │  ├─ extended/
│  │  │  │  │  ├─ _show.html.erb
```

**Using a custom look**
To use a custom look,  its name is passed as an option to a Field. If the option is not given, the default look is used.

```ruby
  # app/dashboards/book_dashboard.rb

  ATTRIBUTE_TYPES = {
    pages: Field::HasMany,
    authors: Field::HasMany.with_options(look: :simple),
    publishers: Field::HasMany.with_options(look: :extended)
  }
```

```ruby
  # administrate/lib/administrate/field/base.rb

  def to_partial_path
    variant = options.fetch(:variant, :default)
    
    "/fields/#{self.class.field_type}/#{variant}/#{page}"
  end
```

### Can you think of other approaches to the problem?

[...]

Issue #2354, opened 2023-04-13
Allow grouping of options in Field::Select


pablobm:
At the moment, there appears to be no way to group options with `<optgroup>` elements in `Field::Select`.

Currently, `Field::Select` uses `options_from_collection_for_select` to render, which doesn't allow option groups. In order to allow them, we would have to either detect groups in the `collection` argument, or provide a different argument (eg: `grouped_options`) that tells `Field::Select` to render groups.

PR #2803, opened 2025-03-11
Add custom look

goosys:
Introduces a custom look implementation based on the discussion in #2291.

### Changes:
- Added `look` option to `FieldBase`
  - Added RSpec tests for `FieldBase`
  - Updated RSpec tests for each `Field`
- Added `look` option to generators
  - Field generator
  - Custom Field generator
- Updated documentation
  - Custom Dashboard
  - Field generator
  - Custom Field generator

Please review and let me know if any changes are needed! 🚀

### Usage:

```ruby
class PageDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    product: Field::BelongsTo.with_options(look: :product_card)
  }
end
class LineItemDashboard < Administrate::BaseDashboard
  ATTRIBUTE_TYPES = {
    product: Field::BelongsTo.with_options(look: :product_card)
  }
end
```

image

image


goosys:
I added a sample since #2354 might also be easily resolved with looks. https://github.com/thoughtbot/administrate/commit/f280d504a55d283e3989487c39c22e88fd8acd04

`/admin/products/:id/edit`

image

goosys:
@pablobm @nickcharlton 

I've been playing around with this for the past few days, and it's incredibly fun! 🎉  
I feel like this significantly expands the possibilities for customizing Administrate.  
Some simple third-party custom field plugins could already be replaced with the `looks` feature.  
Would you like to give it a try?

Also, could you update the beta tag to `beta4` once `looks` is merged?  
With `virtual field`, `partial prefixes`, and `looks` as the foundation, plugin development should become easier than before. Having an updated tag would be really helpful.

nickcharlton:
The code change is small, but it's a big (positive) step in functionality which makes a nice highlight for a next release. I'm just trying to see if I can draw a line and get v1 done, as I've dragged my feet on it much longer than I was hoping I'd do.

I appreciate this makes things annoying for you and your application though 😭

goosys:
Thank you! I really appreciate it. I’m looking forward to the completion of v1. 😊

docs/customizing_attribute_partials.md
```diff
@@ -54,3 +54,17 @@ Changing numbers to display to three decimal places might look like this:
 
 If you only want to change how an attribute appears
 on a single page (e.g. `index`), you may delete the unnecessary templates.
+
+## Using a Custom Look
+
+If you want to generate a different look, run the following:
+
+```bash
+rails generate administrate:views:field number --look custom
```
pablobm, line 63: Does this actually work? I've been trying today and it didn't work, and only now I'm realising perhaps this generator isn't supposed to run on its own.

docs/adding_custom_field_types.md
```diff
@@ -58,6 +58,21 @@ Since we want to display an image, we can change it to:
...
+
+If you want to generate a different look, run the following:
+
+```bash
+rails generate administrate:field gravatar --look custom
+```
+
+This will generate three files:
+
+- `app/fields/gravatar_field.rb`
```
pablobm, line 71:
```suggestion
- `app/fields/gravatar_field.rb` (assuming it doesn't exist already)
```

lib/administrate/field/base.rb
```diff
@@ -36,17 +47,22 @@ def self.permitted_attribute(attr, _options = nil)
...
           else
-            local_partial_prefixes
+            local_partial_prefixes(look: look)
           end
       end
 
-      def self.local_partial_prefixes
-        ["fields/#{field_type}"]
+      def self.local_partial_prefixes(look: :default)
+        fallback = ["fields/#{field_type}/looks/default", "fields/#{field_type}"]
```
pablobm, line 60: Do we need to allow for `default`? I think it might get confusing and affect maintenance in the long run. There's already the normal option (no look), so not sure we need this.

pablobm (commented):
Looking great! :slightly_smiling_face: Let some minor comments and after that I think this is a good first feature for the post-v1 release.

lib/administrate/field/base.rb
goosys: I don't think having both would cause significant confusion, but which approach do you prefer?  

I felt that plugin developers might prefer having everything neatly organized within `looks` dir, so I designed it this way to allow both options.

goosys:
@pablobm 
Rebased.
Some time has passed since then, so I may have forgotten if there were any remaining issues. Please let me know if you notice anything.

pablobm (commented):
I think this is good, just the minor comments. The one thing that I think is missing is simple tests to check that the example looks are working. I would write something like `spec/example_app/spec/features/field_looks_spec.rb` and examples checking some simple HTML: the `` tags and the `<optgroup>` tag should do the trick.

Note that I would put these under `spec/example_app/spec` instead of under `spec` because they are things that I think users of Administrate should be testing when they customise their dashboards (it's their own code!), as opposed to things that Administrate should guarantee.

goosys:
@pablobm 
Thank you for the review. I've made the requested changes and pushed a new commit.

(The failing tests are due to a separate Ruby 4–related issue.)

pablobm:
Finally!

pablobm:
@goosys - One tiny issue I just realised: the "group" look doesn't set the attributes of the form field correctly. As a result it doesn't use selectize.

I was playing with it and seems to be fixed with https://github.com/thoughtbot/administrate/compare/main...pablobm:administrate:select-group-fix Also added some styling, but that's less of a problem. What do you think?

goosys:
@pablobm 
Thank you.

It felt a bit odd to have styles for the demo view included on the library side, so I ended up finishing it without adding styling.

That said, I think it would be great to provide styles on the library side assuming grouped selects will be commonly used!

PR #2804, opened 2025-03-11
Display field hints on nested forms

sergioisidoro:
Currently the filed hints work on the `application/_form.html.erb`, but when a nested form is used, they are not displayed. 

This is basically copying those lines from the form partial, ~~with the difference that we're using `f.object_name`instead.~~

### DIFF
--- a/app/views/fields/has_one/_form.html.erb
+++ b/app/views/fields/has_one/_form.html.erb
@@ -30,6 +30,12 @@ The form will be rendered as nested_from to parent relationship.
           <div class="field-unit field-unit--<%= attribute.html_class %>">
             <%= render_field attribute, f: has_one_f %>
           </div>
+          <% hint_key = "administrate.field_hints.#{field.name}.#{attribute.name}" %>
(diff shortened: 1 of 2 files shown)


nickcharlton:
Could you look into adding a test to cover the change?

sergioisidoro:
@nickcharlton I would appreciate some help in setting up the tests.

I was trying to add them to `spec/administrate/views/fields/has_one/_form_spec.rb` but that spec is really limited, it only tests that the has one renders the title. I'm unsure how to expand that example. Also `nested_form` seems to be an instance of Page::View instead of Page::Form.

Another alternative would be to add it to `spec/features/form_spec.rb` but the test dashboard does not have any `HasOne` field. I wonder if I should expand the test coverage there instead?

Some guidance?

nickcharlton:
Oh, nice one! Yeah, you've got it here. I'd expect that we'd end up with a feature test for this, as otherwise as you notice we end up mocking out loads and that's not very nice.

PR #2811, opened 2025-03-17
Fix Stimulus Selectize reinitialization issue

goosys:
## Issue  

Previously, Selectize was initialized every time the Stimulus controller reconnected, leading to multiple instances being created when navigating back in the browser. Additionally, since `disconnect()` was not implemented, Selectize instances were never properly cleaned up, which likely caused unexpected behavior and potential memory leaks.  

image

## Fix  
- Prevents Selectize from losing selected values and settings on reinitialization  
- Stores initial Selectize options (`settings`) and restores them on reconnection  
- Implements `disconnect()` to properly destroy Selectize and prevent multiple instances  
- Retains the selected values between disconnections and reconnections  

This ensures that Selectize behaves correctly across page navigations and Stimulus lifecycle events.


nickcharlton:
Thank you!

PR #2812, opened 2025-03-17
Add requireness to HasOne form

goosys:
The nested form for HasOne didn't have the requiredness mark, so I added it.

Before:

image

After:

image

### DIFF
--- a/app/views/fields/has_one/_form.html.erb
+++ b/app/views/fields/has_one/_form.html.erb
@@ -27,7 +27,7 @@ The form will be rendered as nested_from to parent relationship.
         <% end %>
 
         <% attributes.each do |attribute| %>
-          <div class="field-unit field-unit--<%= attribute.html_class %>">
+          <div class="field-unit field-unit--<%= attribute.html_class %> field-unit--<%= requireness(attribute) %>">
             <%= render_field attribute, f: has_one_f %>
           </div>
         <% end %>
--- a/spec/features/form_spec.rb
+++ b/spec/features/form_spec.rb
@@ -36,7 +36,9 @@ describe "edit form" do
       Product.human_attribute_name(:name),
       Product.human_attribute_name(:description),
       Product.human_attribute_name(:price),
-      Product.human_attribute_name(:image_url)
+      Product.human_attribute_name(:image_url),
+      ProductMetaTag.human_attribute_name(:meta_title),
+      ProductMetaTag.human_attribute_name(:meta_description)
     ]
 
     required_field_labels = find_all(".field-unit--required").map(&:text)


goosys:
I've added tests.  

By the way, `ProductoMetaTag` is the only place in the demo app that uses `Field::HasOne`, but it has too few columns, both of which are `String` type and required. This makes it less suitable for comprehensive testing.  

Would you consider adding more columns?

goosys:
Also, regarding this helper name -- should it be "requiredness" instead of "requireness"?

goosys:
Rebased!

app/views/fields/has_one/_form.html.erb
```diff
@@ -27,7 +27,7 @@ The form will be rendered as nested_from to parent relationship.
         <% end %>
 
         <% attributes.each do |attribute| %>
-          <div class="field-unit field-unit--<%= attribute.html_class %>">
+          <div class="field-unit field-unit--<%= attribute.html_class %> field-unit--<%= requireness(attribute) %>">
```
nickcharlton, line 30:
```suggestion
           <div class="field-unit field-unit--<%= attribute.html_class %> field-unit--<%= requiredness(attribute) %>">
```

I think we should probably make the change to "requiredness", as I do think that's more correct than "requireness".

goosys: Should this be changed in this PR?
Or would it be okay if I open a new PR?

goosys:
Rebased.

pablobm:
> We should add more if it's needed for testing. The demo app has grown organically and is neither consistent nor coherent in how it works. I've long wanted to switch the demo app to follow a particular pattern. Maybe a shop or something.

For a while now I've been thinking that ideally we should have several example apps, or one with several admin interfaces handling unrelated models (or perhaps related). I feel that trying to get every single use case with a single app gets out of hand pretty quickly.

pablobm:
I think this is good :shipit:

goosys:
What do you think about moving this into a separate dedicated repository?

Something like `administrate-kitchensink` or `administrate-storybook`. Ideally, I’d like it to have all the community plugins installed, and include many examples using `looks`.

For example, we could provide multiple implementations for `Field::Date`, such as a react-daterange-picker version, a stimulus-flatpickr version, and a jquery-date-range-picker version, each implemented using `looks`. We could also add various examples for other Fields as well.

I think this would help users get a better sense of how things can be used in practice, and also serve as a reference for implementation.

In an ideal world, it could become a place where we can easily try out Fields built on top of minor gems just to see how they feel.

Just a thought.

PR #2818, opened 2025-03-24
Add cell styles for action buttons

goosys:


spec/example_app/app/views/admin/customers/_collection_header_actions.html.erb
```diff
@@ -2,7 +2,7 @@
   [
     existing_action?(collection_presenter.resource_name, :edit),
     existing_action?(collection_presenter.resource_name, :destroy),
-    resources.class == Customer, # "Become" action
+    resources.klass == Customer, # "Become" action
```
goosys, line 5: This was a mistake from my previous work. Sorry about that.

app/assets/stylesheets/administrate/components/_buttons.scss
```diff
@@ -42,6 +42,14 @@ button,
   }
 }
 
+
+form.button_to { /* stylelint-disable selector-no-qualifying-type, selector-class-pattern */
```
goosys, line 46: This is a class automatically generated by Rails' `button_to`, so it didn't conform to BEM rules.

spec/example_app/app/views/admin/customers/_collection_header_actions.html.erb
nickcharlton: In that case, should we fix this in another PR?

goosys: Removed! Sorry about that!

PR #2824, opened 2025-04-03
Add color input

coorasse:
Adds a `Field::Color` to render a color picker. It uses the HTML color picker input in the form and shows a small square with the color in the index and show.

## Form
CleanShot 2025-04-03 at 21 30 16@2x

## Show
CleanShot 2025-04-03 at 21 30 28@2x

## Index
CleanShot 2025-04-03 at 21 30 58@2x


app/assets/stylesheets/administrate/base/_forms.scss
```diff
@@ -92,6 +92,23 @@ textarea {
...
+
+  span {
+    width: 1em;
+    height: 1em;
+    line-height: 1.5em;
+    display: inline-block;
+    margin-right: 0.5em;
+  }
+}
+
```
coorasse, line 111: I am not sure this fits in this CSS file to be honest

spec/lib/fields/color_spec.rb
coorasse, line 1: I probably could test more....but I added a field in the demo app 😄

PR #2825, opened 2025-04-04
Replace form_for with form_with

goosys:
Since `form_for` has been deprecated in Rails 5.1, I replaced it with `form_with`.
The change required minimal adjustments thanks to good compatibility.

In the demo app, we need to set `config.action_view.form_with_generates_ids = true` to automatically generate `id` and `for` attributes.  
However, this is already `true` by default in `load_defaults` from Rails 6.0 onward, so it shouldn’t cause problems in most applications.


nickcharlton:
Thank you!

PR #2826, opened 2025-04-04
Fix become button header

goosys:
- fix #2787 

I made a mistake during the work on #2787 — the condition wasn’t evaluated correctly, so the table header for the Become button wasn't being rendered.  

Before:

image

After:

image


nickcharlton:
Oh, thanks! I seem to not notice when it's `class` vs. `klass`, even when I tested it manually 😬

Issue #2815, opened 2025-03-21
Remove deprecated APIs


pablobm:
Some APIs have been marked as deprecated over time, have supported alternatives, and should be removed entirely to avoid confusion and unnecessary code.

These should be easy to spot by looking at the `warn_of_deprecated_*` methods at https://github.com/thoughtbot/administrate/blob/main/lib/administrate.rb

PR #2832, opened 2025-04-10
Remove deprecated code

pablobm:
**DO NOT MERGE YET.** This probably should be merged closer to v1.0.

Fixes https://github.com/thoughtbot/administrate/issues/2815

Before going into v1.0, we should shed deprecated APIs that we have supported for back-compatibility.


pablobm:
@nickcharlton - Pinging you for a look even if this probably shouldn't be merged yet. In my view, this should be done closer to releasing v1.0. Perhaps as the last item.

pablobm:
Thinking again: perhaps this should be merged a bit earlier. Gems versioned 1.x-pre probably should have the API expected of v1.0.0, which includes removing these deprecated APIs.

nickcharlton:
If you think it's ready to go now, we can merge it now.

The thing that's holding me back is running through some long and tedious regression testing and assuming we don't find anything (I suspect we won't), we're good to cut v1.

Issue #2821, opened 2025-03-25
Field::Polymorphic no longer renders like a Field::BelongsTo in version 1.0.0.beta3


mpvosseller:
In `1.0.0.beta3` a `Field::Polymorphic` field no longer renders similar to a `Field::BelongsTo` field as it did in `0.20.1`. 

Specifically it no longer has a text field that a user can type into to filter the list. It almost looks like `Field::Polymorphic` is now rendering as a browser built in select widget rather than the nice custom Administrate widget it previously used (and which `Field::BelongsTo` still uses).

You can see this by running the `example_app` and looking at a `LogEntry` `logeable` and comparing version `0.20.1` to `1.0.0.beta3`.

Below are screenshots comparing `0.20.1` as deployed [on heroku](http://administrate-demo.herokuapp.com) and `1.0.0.beta3` running locally on my machine: 

<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/133c585c-2ace-411b-9a18-efd63406e092" />
<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/be523931-283c-425b-ad43-020ebfa41594" />
<img width="1392" alt="Image" src="https://github.com/user-attachments/assets/0d1b03e6-8a14-4b5c-babf-53fa36d97c7c" />

nickcharlton:
It sounds like we've broken Selectize here, have you seen the same anywhere else in the demo app?

mpvosseller:
@nickcharlton I have not. Thanks for looking into this!

mpvosseller:
@nickcharlton Any chance you all could release a `1.0.0.beta4`?

PR #2839, opened 2025-04-28
Selectize polymorphic select field

heyvaleria:
Addresses https://github.com/thoughtbot/administrate/issues/2821

When building the select field in the polymorphic form, we omitted to
pass `nil` to the choices argument, which resulted in an erroneous
argument order, and a `{}` was passed to it instead.
This caused the selectize gem to fail applying the `.selectize` method
under the hood.

Passing `nil` will now restore its intended behavior and  allow selectize
to perform the dynamic search in the select field again.

Adding a feature spec for the /new form where we type in the input field to
make use of the selectize function.

### References
Where it started: https://github.com/thoughtbot/administrate/pull/2447
Where I looked for answers: https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/select
And here: https://github.com/thoughtbot/administrate/pull/2447/files#diff-75dadcafc8eab3ca77123a071efca9c25c06adf2db9a3f8c4827419f23e698d7R26

### Screenshots
Screenshot 2025-04-28 at 12 51 34
Screenshot 2025-04-28 at 12 51 40
Screenshot 2025-04-28 at 12 51 51

### DIFF
--- a/app/views/fields/polymorphic/_form.html.erb
+++ b/app/views/fields/polymorphic/_form.html.erb
@@ -22,7 +22,7 @@ This partial renders an input element for polymorphic relationships.
 
   <div class="field-unit__field">
     <%= pf.hidden_field(:type, value: field.class.name) %>
-    <%= pf.select(:value, {}, data: {controller: field.html_controller}) do %>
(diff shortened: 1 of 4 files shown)


nickcharlton:
This is great, thanks for digging into this and figuring it out!

PR #2842, opened 2025-05-01
Conventional location and name for HasManyVariant in demo app

pablobm:
Yesterday I had some confusion with the `administrate:field` generator, until I realised that it uses a convention different from the one used by the custom field `HasManyVariant`, as present in the demo app.

After thinking about which of the two styles is right or wrong, I think the generator is doing the right thing. This PR moves `HasManyVariant` and renames it `HasManyVariantField`, in order to follow the convention set out by the generator (which is also the one shown in the docs).

Notes:
- The field class is generated under `app/fields`. The location means it's autoloaded and we don't need a `require`.
- The `Field` suffix may seem annoying, but it's not different from controllers, helpers, etc, and removes the need for a namespace (eg: `Field::HasManyVariant`).
- This establishes a difference between the official field types (eg: `Administrate::Field::String`) and custom ones (eg: `MyStringField`), but I think this is fine.

Tangentially, I also noticed that, if for any reason a user puts the template files in the wrong place, this is difficult to debug as Administrate will just fail silently, rendering empty templates. This is due to the changes introduced at https://github.com/thoughtbot/administrate/pull/2467. This should not be an issue if using the generator, but might be if creating the field manually.

### DIFF
--- a/spec/example_app/app/dashboards/customer_dashboard.rb
+++ b/spec/example_app/app/dashboards/customer_dashboard.rb
@@ -1,4 +1,3 @@
-require "administrate/field/has_many_variant"
 require "administrate/base_dashboard"
 
 class CustomerDashboard < Administrate::BaseDashboard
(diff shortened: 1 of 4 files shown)


nickcharlton (approved):
oh yeah nice. Thanks for doing that!

PR #2844, opened 2025-05-02
Removed the link role from the index table

pangolingo:
Each row of the collection table had a `role=link` attribute, which was removing the default "row" role. This caused screen readers to ignore all of the table's rows. Removing the attribute resolves this issue.


enatario (approved):
Nice!

pangolingo:
@pablobm I recommend merging this PR as sooner because it's make the table literally unusable to some people. Are you or @nickcharlton aware of a reason that removing this role would break anything? I don't really have any context on the history.

PR #2870, opened 2025-08-11
Fix unprocessable entity deprecation warning

floehopper:
I'm seeing this deprecation when using newer versions of `rack` with request specs that use [the `have_http_status` matcher][1] from newer versions (v8?) of `rspec-rails`.

    Status code :unprocessable_entity is deprecated and will be removed in a future version of Rack.
    Please use :unprocessable_content instead.

The `:unprocessable_entity` symbol was deprecated in `rack` in [this commit][2] which was released in [`rack` v3.1.0][3]. It has been replaced by `:unprocessable_content`.

[1]: https://rspec.info/features/8-0/rspec-rails/matchers/have-http-status-matcher/
[2]: https://github.com/rack/rack/pull/2137
[3]: https://github.com/rack/rack/blob/ee7ac5a1db5bc5c65e4b83342b8f4df88ef3c075/CHANGELOG.md#310---2024-06-11


nickcharlton:
Ah, nice, thanks! Do you think you could fix the conflicts? (I tried to do it myself, but I couldn't write to the upstream.)

floehopper:
@nickcharlton I'll give it a go 👍

floehopper:
@nickcharlton:

> Ah, nice, thanks! Do you think you could fix the conflicts? (I tried to do it myself, but I couldn't write to the upstream.)

Done ✅

floehopper:
It's worth highlighting that this will only work with rack v3.1.0 or higher...

nickcharlton:
Ah yes. I was hoping the tests would give us the answer to what wouldn't be supported, and I imagine that's it.

Whilst we still have support for Rails 6.0 and 6.1 we'll leave this until it's dropped. Unless I'm remembering wrong, the plan would be to drop support after we release v1.0. Which is now planned for the 31st October.

floehopper:
@nickcharlton I imagine those test failures are likely to be because you want to support Rails v6.0 & Rails v6.1 and I imagine they depend on an earlier version of Rack. Can you confirm that's the case and if so, do I need to make the fix work in those cases too?

floehopper:
> Ah yes. I was hoping the tests would give us the answer to what wouldn't be supported, and I imagine that's it.
> 
> Whilst we still have support for Rails 6.0 and 6.1 we'll leave this until it's dropped. Unless I'm remembering wrong, the plan would be to drop support after we release v1.0. Which is now planned for the 31st October.

OK 👍

nickcharlton:
Yeah — let's just wait until we drop support. I already set myself up for a date we've got to hit and it's close enough to avoid making a whole lot of extra work.

PR #2874, opened 2025-08-20
Remove unused association attribute from Order

rossme:
This PR removes an unused attribute from `AdministrateOrder`. The original change was done in this commit https://github.com/thoughtbot/administrate/commit/b1bf9523

The attribute seems to be unused and it can be safely removed.


nickcharlton:
Thank you!

PR #2886, opened 2025-09-10
Add primary key tiebreaker for stable sorting

goosys:
This pull request improves the stability of sorting in Administrate by introducing a tiebreaker using the primary key.

When ordering by a column, the primary key is now used as a secondary sort key to ensure stable and predictable results.
If the relation does not have a primary key, the tiebreaker is skipped.

Changes:

- Add the primary key as a tiebreaker when ordering by a column
- Skip the tiebreaker if the relation has no primary key
- Update and add specs to cover these scenarios

Motivation:

Without a tiebreaker, sorting by a non-unique column can result in unpredictable order for records with the same value.
This can cause issues such as the order changing on every page load, or the same record appearing on multiple pages when paginating.
By using the primary key as a tiebreaker, we guarantee a stable sort order and prevent these problems.

Please review!

### DIFF
--- a/docs/using_administrate.md
+++ b/docs/using_administrate.md
@@ -1,0 +1,5 @@
+---
+title: Using Administrate
+---
+
+- [Stable sorting](./using_administrate/stable_sorting.md)
--- a/docs/using_administrate/stable_sorting.md
+++ b/docs/using_administrate/stable_sorting.md
@@ -1,0 +1,26 @@
+---
+title: Stable Sorting
+---
+
+The display order of `index` pages and `HasMany` fields is controlled by `Administrate::Order`.
+By default, both are set to `nil`, so the model's default sort order is applied.
+
+You can toggle sorting by clicking on a table header attribute. When sorting by a specific attribute, a tiebreaker is used to ensure stable sorting.
+
+The tiebreaker uses the table's primary key.
+The generated SQL looks like this:
+
+```sql
+-- When toggling the name attribute
+SELECT * FROM users ORDER BY name DESC, id DESC;
+
+-- When toggling the name attribute again
+SELECT * FROM users ORDER BY name ASC, id ASC;
+```
+
+If there is no primary key (such as in a join table), the tiebreaker is not used.
+
+```sql
+-- When toggling the name attribute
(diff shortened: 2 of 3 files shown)


nickcharlton:
Oh, fun! Could you add something to the docs on this as well?

goosys:
I've added the docs! How does this look?

goosys:
By the way, do you think the tiebreak logic should live inside `apply`, or would it be better to split it out into its own method?

nickcharlton:
@pablobm, do you have any thoughts on this one? I think you might've spent more time in this part of the code base than I have.

pablobm:
Always tricky with these things, but... this looks good to me :slightly_smiling_face: 

The one thing I'm not sure is if the guides is the right place to put this documentation. In my mind the guides is a place to put "recipes" that require configuration or code, while here we are just explaining how the UI works. Having said that, I don't know what a good place would be :thinking: Thoughts?

pablobm:
Crazy idea, re: where to put this documentation - Perhaps it should be a comment in [the controller template](https://github.com/thoughtbot/administrate/blob/main/lib/generators/administrate/dashboard/templates/controller.rb.erb). Similar to the existing ones. Perhaps we should add comments for `default_sorting_attribute`/`default_sorting_direction` and add this documentation there.

Just a thought! Not 100% convinced myself.

goosys:
Thank you for your comment!
As you mentioned, I was also wondering if it would be appropriate to put this content in the `guides/`.
Writing it in the controller's template also seems like a good idea. Hmm.

I've previously read comments from everyone wanting to enrich the documentation, so I'm keen to write more in the `docs/`.
With the current hierarchical structure, there isn't a place to put it, but perhaps we could create a hierarchical level specifically for this kind of documentation.
Everyone would want documentation explaining the implicit aspects of Administrate.

pablobm:
@goosys - Sounds good :slightly_smiling_face: I think it makes sense to create a new doc under `docs/` to be linked from the main docs sidebar. Something like "Using Administrate" or similar.

goosys:
Added new item to sidebar.

image

pablobm (approved):
Let's go with this.

PR #2926, opened 2025-11-12
Use Field::RichText type for ActionText has_rich_text associations

### LINKED ISSUE #2925: Dashboard Generator doesn't recognise ActionText Rich Text fields
I installed administrate v1 and ran the generator and was intrigued to find that I immediately had errors loading the dashboard. Some of those errors were due to lack of ActiveStorage support (and I've seen that there's already other issues for this). But others were related to ActionText rich text fields.

I've seen that support for ActionText Rich Text fields has been added with the Field::RichText type, but the generator did not automatically recognise rich text fields (are actually associations) and use the built-in type accordingly. I believe that the generator should do this.

I'll try submit a PR for this shortly

smartygus:
- detects ActionText rich text associations based on class name and attribution name and if it's a match, returns the built-in Field::RichText type
Closes #2925

### DIFF
--- a/lib/generators/administrate/dashboard/dashboard_generator.rb
+++ b/lib/generators/administrate/dashboard/dashboard_generator.rb
@@ -128,7 +128,10 @@ module Administrate
 
       def association_type(attribute)
         relationship = klass.reflections[attribute.to_s]
-        if relationship.has_one?
+        if relationship.name.start_with?("rich_text_") &&
+            relationship.options[:class_name]&.start_with?("ActionText::")
+          "Field::RichText"
+        elsif relationship.has_one?
           "Field::HasOne"
         elsif relationship.collection?
           "Field::HasMany"


nickcharlton:
Thank you!

(I tested it locally, and it'll take a `rich_text_content` — [like what's in the Rails guides][1] — and set to use the right field.)

[1]: https://guides.rubyonrails.org/action_text_overview.html

smartygus:
> Thank you!
> 
> (I tested it locally, and it'll take a `rich_text_content` — [like what's in the Rails guides](https://guides.rubyonrails.org/action_text_overview.html) — and set to use the right field.)

Cool, yeah that was my intention 👍 
It was basically a hiccup in my first-run experience with administrate, cos my app was using Action Text a fair bit, and it seemed like a relatively small fix, so I figured I may as well contribute!

Thanks for the review, the tweaks, and the merge 😃

PR #2936, opened 2025-11-21
Fix tooltip cutoff

jaredlt:
We introduced a tooltip for search in https://github.com/thoughtbot/administrate/pull/2758 This works fine with the default settings (displaying the "New" button), however, if you have disabled "New" routes the tooltip is nearer the right edge of the screen and the tooltip is cutoff.

We fix this by aligning the tooltip to the right, such that even if there is no "New" button the full tooltip is still displayed.

<details>
  <summary>Before</summary>

  
administrate-tooltip-cutoff

</details>

<details>
  <summary>After</summary>

  
tooltip-with-new-button
tooltip-without-new-button

</details>


nickcharlton:
Thanks!

PR #2948, opened 2025-12-02
Add devcontainer for Codespaces

goosys:
- closes #2798 

Regarding #2798, I'd like to propose an alternative approach.

Codespaces now supports using multiple devcontainer configurations, which means we can keep additional settings without causing conflicts. Based on this, I added an initialization script to the devcontainer to help users get started with this repository more easily.

If you select the **“For Codespaces”** configuration when launching a Codespace and wait for a moment, the environment will come up with both `bin/dev` and `bin/rspec` immediately available.

What do you think?


goosys:
By creating this Prebuild, I can now spin up any branch of this repository in about 1–2 minutes.

Prebuild:
image

Start Container:
image

nickcharlton:
I started digging back into this last week, because I had a bit of time at last.

I'd already played around with Codespaces on a generic Rails app, which is where I found it to be too slow, so I wanted to explore building a custom image. But what worked 9 months ago no longer does, so I ended up opening a support ticket for it on Monday. I've not heard back yet.

But, prebuilds might solve a good chunk of the problem, it's good to hear it's down to 1-2 minutes!

I'll come back to this as soon as I can, but it's going to be a bit slow because it's something that I'm keen to properly understand myself so it can be reused elsewhere.

.devcontainer/codespaces/onCreateCommand.sh
```diff
@@ -0,0 +1,20 @@
+RUBY_VERSION=3.4.6
+rvm list
```
pablobm: Is this line necessary?

.devcontainer/codespaces/postCreateCommand.sh
```diff
@@ -0,0 +1,13 @@
...
+until docker info >/dev/null 2>&1; do sleep 1; done
+docker rm postgres >/dev/null 2>&1
+docker run -d -i --name postgres -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust postgres
+until docker exec postgres pg_isready -U postgres; do sleep 1; done
+
+./bin/setup
+./bin/setup
+RAILS_ENV=test bundle exec rake db:setup dev:prime
+
+echo "" >> .env && echo "DATABASE_URL=postgresql://postgres:@localhost" >> .env
```
pablobm: What's this envvar for? Also at `.devcontainer/codespaces/devcontainer.json`

pablobm (commented):
I've been playing with devcontainers lately and I'm still very new to them. This review here is more me trying to understand them better and get your opinions.

At another project, I've been trying to get devcontainers set up too (see https://github.com/openstreetmap/openstreetmap-website/pull/6424). I started from the template provided with Rails (`bin/rails devcontainer`) and iterated from there. The Rails one is based on the image `ghcr.io/rails/devcontainer/images/ruby` and uses a separate container for Selenium. I also set up Postgres in a separate container.

My intention was to follow The Rails Way as closely as possible. However it's not been smooth sailing and I have had to work around some issues. The setup you offer here appears to just work? Have you experimented with other ways of setting devcontainers in Rails? Do you have any opinions as to how to best do this?

.devcontainer/codespaces/postCreateCommand.sh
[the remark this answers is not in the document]
goosys: Sorry for taking your time on this. 🙇 
After taking another look, it didn’t seem necessary after all, so I removed it.

Previously, if this wasn’t added here, I would sometimes run into an error where the database couldn’t be found when running certain commands (I don’t remember whether it was `bundle exec` or something under `bin/*`). Because of that, I’d been adding it out of habit.

It seems that the `ENV` settings in `devcontainer.json` are applied globally, and after testing again, everything appears to work fine without it.

.devcontainer/codespaces/onCreateCommand.sh
goosys: This was necessary.

Because the `.ruby-version` requirement is very strict, Bundler fails even when the Ruby version in the container image is *newer* than the version specified in `.ruby-version`, which prevents `bundle install` from running.  
(Is there a way to control this via a Bundler-related `ENV`, by any chance?)

Also, the Ruby version in the container image is updated to the latest one over time, so we need to explicitly specify the version here to keep it aligned with `.ruby-version`.  
(At the `onCreate` stage, the source code isn’t available yet, so we can’t read the version from `.ruby-version` there.)

goosys:
@pablobm 
We could certainly create a devContainer tuned specifically for Rails development, but this setup is intended purely to make it possible to spin up Codespaces quickly and reliably at any time.

In my mind, this is primarily for reviewers: it enables a workflow where they can launch a per-PR environment directly on GitHub, verify the changes, and approve them without touching their local setup.

For that reason, I’m not planning to build a development-optimized configuration. Instead, the goal is to use a generic image (`universal:4`) and minimize startup time as much as possible.

What do you think?

.devcontainer/codespaces/onCreateCommand.sh
pablobm: No idea about Bundler envvars 😓 But I don't understand, what does `rvm list` exactly do here? Perhaps it fails if `RUBY_VERSION` is not available or something like that?

goosys: Ah, sorry 🙇 
`rvm list` was just a command for my own verification, so it’s not needed.

nickcharlton:
I thought I'd just publish what I'd already started working on, even though it's not working: https://github.com/thoughtbot/codespaces-demo/tree/custom-config

This is what I was doing at the start of last year. I came back to it towards the end of last year, where I was just trying to confirm everything I'd already tried was working with the plan of _then_ building out something custom for Rails which is a lot faster and less confusing to get up and running with.

However, this is no longer works (even though nothing is meaningfully different from what did work previously). I opened the support ticket around that, but I haven't heard anything in the last two weeks.

goosys:
@nickcharlton 

It looks like a new version of `universal:linux` was released toward the end of last year.
https://mcr.microsoft.com/en-us/artifact/mar/devcontainers/universal/tags

Since things tend to break as soon as the version changes, I think it might be safer to pin the image using a specific version tag.

.devcontainer/codespaces/onCreateCommand.sh
```diff
@@ -0,0 +1,33 @@
+# Workaround for install Ruby 4.0.0 with RVM
+rvm list
```
goosys: This isn’t required for the installation itself, but when it fails, the cause is often unclear, so I’m outputting version information and similar details to the build logs.

.devcontainer/codespaces/onCreateCommand.sh
```diff
@@ -0,0 +1,33 @@
...
+curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
+curl -sSL https://rvm.io/pkuczynski.asc | gpg2 --import -
+rvm get master
+rvm list known
+
+# Install Ruby
+rvm install 3.4.6
+rvm install 4.0.0
```
goosys: Some older PR branches still use older Ruby versions, so I install several versions to allow switching easily via RVM when needed.

goosys (commented):
Updated to support Ruby 4.0.

nickcharlton:
After initially waiting for weeks, and then a lot of back and forth with GitHub Support, I finally got my [baseline configuration](https://github.com/thoughtbot/codespaces-demo/commit/b380d6e2dc4f02a6538007d1ed7374bae5177b3d) completed. I'm going to look at making a custom image next, but I thought it'd be good to share my parallel progress as it's going along.

.devcontainer/codespaces/devcontainer.json
```diff
@@ -0,0 +1,18 @@
+// For format details, see https://aka.ms/devcontainer.json. For config options, see the
```
pablobm, line 1:
```suggestion
// For format details, see https://containers.dev/implementors/json_reference/. For config options, see the
```

The domain aka.ms appears to be used by Microsoft as a link shortener (https://en.wikipedia.org/wiki/.ms#Usage_examples). It looks weird if you don't know this though, so I suggest using the real URL

.devcontainer/codespaces/onCreateCommand.sh
pablobm: Cool. Can you clarify that in the comment, please? The current comment doesn't explain anything. Perhaps it should be:

```sh
# Print debug information in case something breaks
rvm list
```

pablobm (commented):
I just tried this out and, after running `bundle install`, it complains that:

> Your Ruby version is 3.4.7, but your Gemfile specified 4.0.0

I suspect that `.devcontainer/codespaces/onCreateCommand.sh` is not running. I'm also getting this when I try to sort the issue manually:

```
$ rvm get master
Downloading https://bitbucket.org/mpapis/rvm/raw/master/binscripts/rvm-installer.asc
Verifying /usr/local/rvm/archives/rvm-installer.asc
gpg: Signature made Tue Jul 23 21:59:45 2019 UTC
gpg:                using RSA key 7D2BAF1CF37B13E2069D6956105BD0E739499BDB
gpg: Can't check signature: No public key
GPG signature verification failed for '/usr/local/rvm/archives/rvm-installer' - 'https://bitbucket.org/mpapis/rvm/raw/master/binscripts/rvm-installer.asc'! Try to install GPG v2 and then fetch the public key:

    gpg2 --keyserver hkp://pool.sks-keyservers.net --recv-keys 409B6B1796C275462A1703113804BB82D39DC0E3 7D2BAF1CF37B13E2069D6956105BD0E739499BDB

or if it fails:

    command curl -sSL https://rvm.io/mpapis.asc | gpg2 --import -
    command curl -sSL https://rvm.io/pkuczynski.asc | gpg2 --import -

In case of further problems with validation please refer to https://rvm.io/rvm/security
```

Same with the other commands. I don't see a `config/database.yml` for example.

Generally speaking: have you had much luck with Dev Containers? I provided a config for https://github.com/openstreetmap/openstreetmap-website and it gives a lot of trouble.

goosys:
I tried it again for the first time in a while, and it seemed to work without any issues.

That error may be happening because no configuration was selected, so the Dev Container itself might not have started.

> Have you had much luck with Dev Containers?

I agree with this. I don’t find them particularly easy to use either.

That said, this repository requires quite a bit of environment setup before RSpec can be run, which feels like a high barrier to getting started as a contributor.

Even if contributors don’t use it as a Dev Container, having the environment setup steps documented in JSON and shell scripts would still be helpful. I’d appreciate it if you would consider this.

goosys:
TODO:

* Update to support the latest version, since Universal has evolved from 4 to 5 and now 6.
* Universal 4 → 5 appears to involve a major change from Ubuntu 20.04 to 24.04, so start with a clean configuration.
* RVM doesn’t seem to have a 4.x release yet, so using `get master` still appears to be necessary.
* Switch from RVM to rv instead?

nickcharlton:
I started looking at this again the other week (which is actually four weeks ago now, which is a little horrifying as I've been spending way more time on this than I ever expected it'd need), as I'm still keen to get this working as best we can.

I've managed to put together a custom image (including build automation), and I've been fighting with building a feature for [asdf](https://github.com/asdf-vm/asdf) the last week or so. I'm trying to build something that's close to thoughtbot standards as I can (even if I personally do something different; I don't use asdf) so it's useful across a lot of places.

I think I've managed to achieve what I wanted because having a custom image _is_ much faster to start, but it's been so painful to get to this stage. Maybe I'll be able to share something by the end of the week, if I can resolve my current problem with the asdf feature!

goosys:
I updated the base image to `universal:6` and changed the setup to install Ruby using `rv`.

Since `rv` doesn’t build Ruby from source, it made installing multiple Ruby versions very fast and convenient.

Issue #2823, opened 2025-03-26
Split Templates for Improved Customizability


goosys:
## What would you like to be able to do? Can you provide some examples?  

I would like the templates to be more finely divided to improve customizability.  
By splitting them, we can always call the common parts like `Label + render_field`, allowing for a custom design while keeping the core functionality consistent.  

## How could we go about implementing that?  

I tried splitting the templates with a simple approach.  
By using `render` with `collection` and `layout`, we can write it more concisely without `each`.  
https://github.com/goosys/administrate/commit/31a27e9d42e6ce048770e8136625059f0f30c469

With this change, for example, it becomes easier to structure a custom design like this:  

```ruby
# Helper
def render_item_unit(attribute_field)
  render(
    partial: 'resource_item',
    layout: 'resource_item_unit',
    locals: local_assigns.merge(attribute: attribute_field)
  )
end
```

```erb
# admin/users/show.html.erb
<%= content_for(:main) %>
   <section class="main-content__body">
     <figure><%= render_field(page.attributes[:icon]) %></figure>
     <div>
       <%= render_item_unit(page.attributes[:name]) %>
       <%= render_item_unit(page.attributes[:email]) %>
    </div>
   </section
<% end %>
```

## Can you think of other approaches to the problem?  

I believe we could make this even more convenient by leveraging `Renderable` from Rails 6.1.  
However, regardless of that, I think this level of granularity in template splitting is necessary.  

Additionally, it would be beneficial to split not only `show` but also `form`, `collection`, and `search_bar`.  

Would you be open to considering this?  
If you have any specific partial file names in mind, please let me know—I’d be happy to work on this and submit a PR!

nickcharlton:
This sounds like a good idea. What do you think the impact would be on users upgrading? Would they need to make a lot of changes given we currently encourage people to generate the views for themselves and modify them?

goosys:
For this sample, it remains compatible with existing partials, so it won’t affect users' applications.
However, when working on splitting the collection, there might be some non-backward-compatible changes. In that case, developers may need to make some adjustments on their end.

goosys:
I have spent a lot of time adapting to the latest templates with each Administrate update. 🙂
There will be some additional work needed after the template spliting, but I truly believe future updates will require much less effort.

goosys:
I tried implementing it for the `show` templates. #2954 
I think the logic and styling are now nicely separated.

[...]

PR #2954, opened 2025-12-04
split show template

goosys:
- refs #2823

### DIFF
--- a/app/views/administrate/application/_resource_group.html.erb
+++ b/app/views/administrate/application/_resource_group.html.erb
@@ -1,0 +1,9 @@
+<% attributes = resource_group.second %>
+
+<%= render(
+  partial: 'resource_item',
+  layout: 'resource_item_layout',
+  collection: attributes,
+  as: :attribute,
+  locals: local_assigns
+) %>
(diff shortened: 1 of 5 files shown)


goosys:
**Memo (just thinking out loud)**

- `local_assigns` feels too loose — maybe we should pass only the required `locals`?
- The partial file names probably need another look.
  - We'll be doing similar work in `form/index`, so it might be better to align the naming.

nickcharlton:
Now some time has passed, do you have any more toughts on the `local_assigns` and partials naming?

goosys:
It’s been a while, and I’m a bit fuzzy on what the original concern was, but looking at it again, both options seem fine to me.

If there are no concerns about the approach of splitting `show`, I can go ahead and split `index` and `form` in the same way and put together a draft PR. Then we can review those and make a final decision based on that -- what do you think?

nickcharlton:
Yeah, sounds like a good plan.

Re: `local_assigns`, I think we just want to ensure we don't inadvertently leak lots of context into a page, but we also don't want to make people's job overriding stuff harder unnecessarily.

PR #2959, opened 2025-12-19
feature: add Administrate.deprecator to the host app

kpcirclemedical:
This auto-adds `Administrate.deprecator` to the host app, helping with tools like [next_rails deprecation tracker](https://github.com/fastruby/next_rails?tab=readme-ov-file#deprecation-tracking)

Without this change, app needs to add it in initializer, e.g.

```ruby
# config/initializers/administrate.rb
Rails.application.deprecators[:administrate] = Administrate.deprecator
```

This code is based on [how the devise gem does it](https://github.com/heartcombo/devise/blob/c8a64b549c8b37e494eaca7be2def136a7e1b236/lib/devise/rails.rb#L20).

### DIFF
--- a/lib/administrate/engine.rb
+++ b/lib/administrate/engine.rb
@@ -30,6 +30,10 @@ module Administrate
       end
     end
 
+    initializer "administrate.deprecator" do |app|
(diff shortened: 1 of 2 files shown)


nickcharlton:
Thanks!

Issue #2965, opened 2026-01-08
DEPRECATION WARNING: String#mb_chars in Rails 8.1+


samuelLkennedy:
When using Administrate with Rails 8.1, a deprecation warning is triggered for String#mb_chars and ActiveSupport::Multibyte::Chars. The suggested fix is to use normal string methods instead.

Example location was found in search:
https://github.com/thoughtbot/administrate/blob/main/lib/administrate/search.rb#L114
`["%#{term.mb_chars.downcase}%"] * fields_count`

Suggesting in prep of future Rails updates. Thanks for checking this out and lmk if my issue is whack because I'm new to writing them 👍

nickcharlton:
Thanks! Yes, this looks like something we should solve sooner rather than later.

Would you like to try opening a PR to make the change? It sounds like you know where to make it, and hopefully the normal string methods just drop in. The tests should help out with any edge cases (hopefully!).

mpvosseller:
Thanks. Ran into this today too and created a quick PR: https://github.com/thoughtbot/administrate/pull/2967

PR #2967, opened 2026-01-09
Remove use of deprecated method String#mb_chars

mpvosseller:
Removes use of the deprecated method `String#mb_chars`.

`String#mb_chars` was needed historically because Ruby's `String#downcase` didn't handle Unicode properly. Since Ruby 2.4+, String#downcase (and upcase, capitalize, etc.) handle Unicode correctly by default.

Fixes https://github.com/thoughtbot/administrate/issues/2965

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -111,7 +111,7 @@ module Administrate
       fields_count = search_attributes.sum do |attr|
         searchable_fields(attr).count
       end
-      ["%#{term.mb_chars.downcase}%"] * fields_count
+      ["%#{term.downcase}%"] * fields_count
     end
 
     def search_attributes


mpvosseller:
@nickcharlton any chance we can get this merged?

nickcharlton:
Thanks! Yeah, just took a while to get the other thing I wanted in done! 😅

bzf:
Hi 👋🏽

Ran into this today and found this issue but noticed it's not available on RubyGems. Seems to be a few things that haven't been released yet, so could be nice to get them out (https://github.com/thoughtbot/administrate/compare/v1.0.0...main) 😄 

Thanks for all your hard work! 💖

PR #2977, opened 2026-01-23
Fix Sentry configuration

nickcharlton:
We've long used Sentry for exception monitoring, but it's not been used for a long time. This re-follows the getting started guide, and introduces an initialiser, and the HTML addition.

If they're set, `SENTRY_DSN` and `SENTRY_ENV` will be used from the environment. The environment is set to be "prerelease" and "release" for each Heroku app.


pablobm:
Is this ready to go? Who has access to the Sentry environment?

nickcharlton:
Oh! Yeah it is. I should just merge it. Everything else is ready to go.

PR #2998, opened 2026-03-08
Add support for page-specific look

goosys:
I’ve added support for changing the `look` per page.

For example, like in the sample app, you can use a `time_ago_in_words` look only on the index page.

```ruby
ATTRIBUTE_TYPES = {
  created_at: Field::DateTime.with_options(look: {index: :relative})
}
```

| index | show |
| -- | -- |
| `relative` look | `default` look |
| image | image |

What do you think?


goosys:
@pablobm @nickcharlton 
Rebased. Please take a look when you have a chance.

PR #3014, opened 2026-03-24
Add support for custom validation contexts in resource creation and update

goosys:
- Separated from #2457 

---

This pull request introduces support for customizable validation contexts when creating or updating resources in the `Administrate` controllers. This allows developers to specify different validation scenarios based on user roles or other conditions, providing more flexibility for resource validation logic. Documentation and tests have been updated to reflect and verify this new functionality.

**Controller enhancements for validation contexts:**

* Added `validation_contexts_on_create` and `validation_contexts_on_update` methods to `Administrate::ApplicationController`, allowing developers to override and specify custom validation contexts for resource creation and update. The default context is `:create` or `:update` based on whether the record is new. (`app/controllers/administrate/application_controller.rb`)

**Documentation updates:**

* Added a new section in `docs/customizing_controller_actions.md`.

**Testing improvements:**

* Added controller specs in `spec/controllers/admin/application_controller_spec.rb`.

---

**Usage:**

```ruby
# Model
validates :name, presence: true, on: :save_by_regular_admin

# Controller
def validation_contexts_on_create(resource)
  if current_user.super_admin?
    super + [:save_by_super_admin]
  else
    super + [:save_by_regular_admin]
  end
end

def validation_contexts_on_update(resource)
  if current_user.super_admin?
    super + [:save_by_super_admin]
  else
    super + [:save_by_regular_admin]
  end
end
```

### DIFF
--- a/app/controllers/administrate/application_controller.rb
+++ b/app/controllers/administrate/application_controller.rb
@@ -58,7 +58,7 @@ module Administrate
         authorize_resource(resource)
       end
 
-      if resource.save
+      if resource.save(context: validation_contexts_on_create(resource))
         yield(resource) if block_given?
         redirect_to(
           after_resource_created_path(resource),
@@ -75,7 +75,9 @@ module Administrate
 
     def update
       @resource = resource = requested_resource
-      if resource.update(resource_params)
+      resource.assign_attributes(resource_params)
+
+      if resource.save(context: validation_contexts_on_update(resource))
         redirect_to(
           after_resource_updated_path(resource),
           notice: translate_with_resource("update.success"),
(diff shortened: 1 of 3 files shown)


goosys:
@pablobm 
I rebased the branch and pushed the updated changes. Could you please review it again?

pablobm:
Could this be done differently, in a more flexible way? For example, we could have a `resource_save` hook (or perhaps `resource_save_on_create`/`resource_save_on_update`). Then users can add these options or others like `validate: false` and `touch: false`, as well as run additional changes and checks before/after the save itself. Or even save differently (not with `save`). What do you think?

goosys:
I’m working on refactoring the controllers, and I expect that the save operation itself will likely be split into methods like this in the future:

https://github.com/goosys/administrate/blob/e9aa93f5eea7d0fdebc35a95c55b87966387db40/app/controllers/concerns/administrate/resource_manager.rb#L59-L61

This would allow users to customize the `save` method to perform additional work before or after saving, or even to change how the resource is saved.

Since `validation_context` is a natural Active Record feature, I think it would still be reasonable to expose it through a dedicated hook or method, even if the save operation is later split into separate methods.

What do you think?

pablobm:
The initial problem is probably with me: I don't think I have ever used validation contexts in the real world :sweat_smile: They sound great, but for one reason or the other I haven't ended up using them. Perhaps they are more common in the wider Rails community than they are in my own bubble, so I can entertain the option.

But then I think a bit more and remember that `#save` takes other options. There are valid reasons why users may want to do `validation: false` or `touch: false` and we aren't adding additional hooks for them. I think a hook specific for the `resource.save` (like the one you have in that other branch) is simpler, more useful, and more obvious to users. If that hook is coming up (in your separate proposal), I think it's not worth complicating it more by adding more sub-hooks that are tiny and I feel not as common.

PR #3026, opened 2026-04-01
Fix destroy button click navigating to show page on index

### LINKED ISSUE #2978: Deleting resource (from index collection table) redirects to resource page
Clicking on "Destroy" button from index collection table causes a redirect to resource page. This happens after the confirmation browser alert is closed, both when confirming delete and cancelling delete.

This happens when using default administrate templates. In addition the same bug would be visible when using custom collection view partial that adds additional actions that are rendered not as links with `href` tag.

Stack to reproduce it: administrate v1.0.0, Rails 8.1 (I'm using assets bundled with administrate with propshaft).

### Why it happens
[Stimulus `click->table#visitDataUrl` action attached to `_collection` partial](https://github.com/thoughtbot/administrate/blob/v1.0.0/app/views/administrate/application/_collection.html.erb#L21) only considers [target with `href` attribute](https://github.com/thoughtbot/administrate/blob/v1.0.0/app/assets/javascripts/administrate/controllers/table_controller.js#L12-L14) as a condition when the url visit should not apply. This conflicts with [`_collection_items_actions` partial](https://github.com/thoughtbot/administrate/blob/v1.0.0/app/views/administrate/application/_collection_item_actions.html.erb#L10-L16) that renders destroy action as `button_to` (which doesn't have `href` attribute).

### Possible solutions

**idea 1**: change `_collection_items_actions` partial to use `link_to` instead of `button_to` for destroy action
- this can be used as a temporary workaround in an application, as the partial can be customized easily
- drawback: delete is now a link instead of form submit
```erb
<% if existing_action?(collection_presenter.resource_name, :edit) %>
  <td class="cell-label--action-button"><%= link_to(
    t("administrate.actions.edit"),
    [:edit, namespace, resource],
    class: "action-edit",
  ) if accessible_action?(resource, :edit) %></td>
<% end %>

<% if existing_action?(collection_presenter.resource_name, :destroy) %>
  <td class="cell-label--action-button"><%= link_to(
    t("administrate.actions.destroy"),
    [namespace, resource],
    class: "link link--danger",
    data: { turbo_method: :delete, turbo_confirm: t("administrate.actions.confirm") }
  ) if accessible_action?(resource, :destroy) %></td>
<% end %>
```

**idea 2:** change `visitDataUrl` action in `table_controller` to exclude targets that are buttons or allow configuration for targets
- `event.target.tagName == 'BUTTON'` or `event.target.type == 'submit'` could be used as an OR condition to the existing early return
- even better would be to extend the existing early return condition check with some target configuration (data attribute or some kind of Stimulus property) that would disable the url visit, e.g. consider a custom tag that triggers modal opening (added in a custom `_collection_items_actions` partial)
```html
<span data-bs-toggle="modal" data-bs-target="my-resource-modal" data-disable-url-visit="true">
  Open my modal
</span>
```

55728:
## What

Fixes #2978

Clicking the "Destroy" button on the index page collection table
was navigating to the resource's show page instead of triggering
the delete action.

## Why

The `visitDataUrl` action in `table_controller.js` checked
`event.target.href` to decide whether to skip row navigation.
Since `button_to` generates a `<form>` with a `<button>` (which
has no `href` attribute), the check was bypassed, causing
`Turbo.visit(dataUrl)` to fire.

## How

Replaced `event.target.href` with
`event.target.closest("a[href], button, input[type='submit']")`
to properly detect interactive elements that should not trigger
row navigation. This also handles:

- `<button>` elements (the reported bug)
- `<input type="submit">` elements (future safety)
- Child elements inside links/buttons (e.g. icons)

Added a feature spec that verifies dismissing the destroy
confirmation keeps the user on the index page.

## Changed files

- `app/assets/javascripts/administrate/controllers/table_controller.js` (1 line)
- `spec/features/orders_index_spec.rb` (+11 lines)

### DIFF
--- a/app/assets/javascripts/administrate/controllers/table_controller.js
+++ b/app/assets/javascripts/administrate/controllers/table_controller.js
@@ -9,7 +9,7 @@ export default class extends Controller {
         event.keyCode == keycodes.space ||
         event.keyCode == keycodes.enter) {
 
-      if (event.target.href) {
+      if (event.target.closest("a[href], button, input[type='submit']")) {
         return;
       }
 
--- a/spec/features/orders_index_spec.rb
+++ b/spec/features/orders_index_spec.rb
@@ -57,6 +57,17 @@ feature "order index page" do
     )
   end
 
+  scenario "clicking destroy does not navigate to show page", js: true do
+    create(:order)
+
+    visit admin_orders_path
+    dismiss_confirm do
+      click_on t("administrate.actions.destroy")
+    end
+
+    expect(page).to have_current_path(admin_orders_path)
+  end
+
   scenario "cannot delete because associated payment", js: true do
     create(:payment, order: create(:order))
 


55728:
Closing this as stale. Happy to reopen if needed. Thanks!

pablobm:
Thank you for your contribution and apologies for the delay. As mentioned at https://github.com/thoughtbot/administrate/issues/2978, I'm having trouble to reproduce the issue. Also I have tried the feature example you provide here, and it doesn't trigger the issue. Is there a circumstance for you where this would fail if the fix was not in place?

As also mentioned on the issue report, there's the annoyance of the Turbo loading bar (when cancelling a destroy), so I would accept a PR for that.

PR #3029, opened 2026-04-25
Fix routes template indentation

tanakakc:
## Problem

The routes template generates misaligned code in `config/routes.rb`:

```ruby
namespace :admin do
    resources :users

    root to: "users#index"
  end
```

This happens because the template includes extra indentation that compounds with Rails' `route()` method, which automatically adds indentation when injecting routes.

## Fix

Adjusted the indentation in `routes.rb.erb` so that the generated output is properly aligned:

```ruby
namespace :admin do
  resources :users

  root to: "users#index"
end
```

## Tests

All existing route generator specs pass.


nickcharlton:
This is one of those really annoying things I've been meaning to fix but never around to! Thanks!

PR #3030, opened 2026-05-14
Yield resource in update action

graywh:
Any reason not to add this?


nickcharlton:
Oh this seems like a good idea. Could you update the docs as well so people know it's there?

PR #3041, opened 2026-07-10
Scope Bundler Audit to Dependabot Bundler PRs

goosys:
I’m not sure we need to run this audit on every PR.
Would it make sense to limit it to Dependabot PRs only?

Also, since this workflow checks Bundler dependencies specifically, I additionally scoped it to Dependabot Bundler PRs (dependabot/bundler/*).

What do you think?

### DIFF
--- a/.github/workflows/bundle-audit.yml
+++ b/.github/workflows/bundle-audit.yml
@@ -1,14 +1,12 @@
 ---
 name: Bundler Audit
 on:
-  push:
-    branches:
-      - 'main'
   pull_request:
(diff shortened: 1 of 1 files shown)


pablobm:
I think we should disable bundler-audit altogether. This days fails more often than it succeeds, and seeing a failure won't alert us to anything any more. We will continue to merge dependabot PRs at the usual rate, so there's nothing to be gained.

nickcharlton:
Hm yeah, I agree. I've opened #3059 to just remove it. I'll close this.

PR #3049, opened 2026-07-17
Delegate Search connection to `@scoped_resource`

seanpdoyle:
When an `Administrate::Search::Query` instance is constructed for a resource class that connects to a different database than the primary, connections scoped by `ActiveRecord::Base` fail to connect to the database specified by the class that descends from `ActiveRecord::Base` and declares its own `connect_to`.

This commit modifies the `Administrate::Search::Query` class to access the connection through the `@scoped_resources` Active Record Relation instance when possible, falling back to `ActiveRecord::Base` when not possible.

### DIFF
--- a/lib/administrate/search.rb
+++ b/lib/administrate/search.rb
@@ -137,15 +137,15 @@ module Administrate
           else
             @scoped_resource.reflect_on_association(attr).klass.table_name
           end
-        ActiveRecord::Base.connection.quote_table_name(unquoted_table_name)
+        connection.quote_table_name(unquoted_table_name)
       else
-        ActiveRecord::Base.connection
+        connection
           .quote_table_name(@scoped_resource.table_name)
       end
     end
 
     def column_to_query(attr)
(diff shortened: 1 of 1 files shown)


lib/administrate/search.rb
```diff
@@ -162,6 +162,14 @@ def term
       query.terms
     end
 
+    def connection
+      if @scoped_resource.respond_to?(:connection)
+        @scoped_resource.connection
+      else
+        ActiveRecord::Base
+      end
```
seanpdoyle, line 170: Executing the test suite with the following change still fully passes:

```diff
      if @scoped_resource.respond_to?(:connection)
        @scoped_resource.connection
      else
-       ActiveRecord::Base
+       fail
      end
```

@nickcharlton @pablobm to me, that signals that the `@scoped_resource` will *always* be an ActiveRecord::Relation or ActiveRecord::Base-descended class. Conceptually, is that true?

If so, I'm happy to remove the `else` branch of this conditional and inline the `#connection` calls back into the original call sites.

PR #3062, opened 2026-08-19
Fix Appraisals requiring ruby-vips

nickcharlton:
This started being required, but wasn't included for these Appraisals which meant they all started failing.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -97,13 +97,12 @@ jobs:
       - run: cp spec/example_app/config/database.yml.sample spec/example_app/config/database.yml
       - name: Prepare main database schema
         run: bin/generate-database-schema
-
-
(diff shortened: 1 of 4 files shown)


nickcharlton:
```
Bundler::GemRequireError:
  There was an error while trying to load the gem 'ruby-vips'.
  Gem Load Error is: Could not open library 'vips.so.42': vips.so.42: cannot open shared object file: No such file or directory.
  Could not open library 'libvips.so.42': libvips.so.42: cannot open shared object file: No such file or directory.
```

Sigh. Looks like we need it in the container as well.

nickcharlton:
Installing `libvips-dev` into the container worked, but now there's an extra step only for Appraisals. These have a habit of hanging around, but we'll just need to rely on eventually discovering it's not needed.

PR #3066, opened 2026-09-15
WIP: Refactor ApplicationController into composable concerns

### LINKED ISSUE #2814: ApplicationController refactoring
I *fully* understand that this might be overly presumptuous, and I truly don’t mean to overstep, but I’d really like to refactor the `ApplicationController`.  

Here are some reasons why:  

- **Improved maintainability**  
  - The current `ApplicationController` has many methods, making it difficult to read and maintain.  
  - The order of methods is not well-structured, making it hard to identify which method belongs to which functionality.  
  - Many action methods are too fat, requiring developers to copy entire methods for partial customization, which leads to inconsistencies during version upgrades.  
  - I'd like to remove methods that are unrelated to CRUD (e.g., view-related methods).  
- **Better customization**  
  - Methods should be split into single-responsibility functions so developers can override only the necessary parts.  
  - Built-in features like Kaminari, search, and filtering should be easily replaceable with alternatives like Pagy or Ransack.  
  - Developers should be able to pick and include only the necessary modules from concerns to create a custom `ApplicationController`.  
- **Enhanced functionality**  
  - Support multiple response formats using `respond_to`, allowing responses like JSON or `render turbo_stream:` for Hotwire.  
  - Make it possible to use Administrate’s advanced CRUD functionalities for building API controllers.  

I’ve done a simple experiment, and ideally, it should look something like this:  https://github.com/goosys/administrate/commit/20cf3a7ab273e17312033ca1d3ed2e1c1503623f 
https://github.com/goosys/administrate/blob/labo/application_controller_refactoring/app/controllers/administrate/application_controller.rb
https://github.com/goosys/administrate/tree/labo/application_controller_refactoring/app/controllers/concerns/administrate

With this level of simplicity (around 7 methods, ~70 lines), even if developers create their own custom `ApplicationController`, they won’t have major conflicts with the latest codebase, or at least the necessary adjustments will be minimal.  

Again, I *genuinely* don’t mean to be intrusive, but I really believe this would be beneficial. I’d love to hear your thoughts! 🙏

goosys:
- resolves #2814

## Summary

Refactor `Administrate::ApplicationController` by splitting its responsibilities into focused concerns.

The default controller includes all concerns required for the existing HTML CRUD behavior, so this change preserves the current behavior while making customization more explicit.

With this structure, custom controllers can compose only the functionality they need. For example:

- A controller that does not render HTML can omit `HtmlRenderer`
- Custom controllers can include only the actions they need, such as `index`, or `new` and `create`.
- Kaminari pagination can be replaced with a custom pagination concern
- The default search implementation can be replaced with Ransack, Elasticsearch, or another solution

I understand the concern that concerns can hide methods that still belong to the controller. In this case, however, the goal is not only to organize methods, but to make each responsibility independently composable and replaceable. The included concerns are explicitly listed in `ApplicationController`, and custom controllers can choose their own composition.

Existing HTML CRUD behavior and customization points are preserved.

## WIP

- Validation context support is included, but has not been merged yet.
- Unit tests for each concern still need to be added.
- Documentation still needs to be updated.
- Discussion about the granularity of method extraction and module naming.


goosys:
@pablobm @nickcharlton 

I’d like to reopen this discussion. Do you have any thoughts?

Issue #1193, opened 2018-07-20
Global Date/Time Field Timezone


mnort9:
Is there any way to set a global date field display timezone other than changing the Rails config setting? 

I'd rather not make an app-wide timezone change for displaying dates in an admin.

If not, I can create a PR. Any suggestions on where it should be set?

ACPK:
@nickcharlton - Any thoughts on this? It's currently resolving to the application time zone (https://github.com/thoughtbot/administrate/pull/1070).

dgmstuart:
I think there is a way: [generate attribute partials](https://administrate-prototype.herokuapp.com/customizing_attribute_partials) for the Date and DateTime fields and use the Rails helpers to convert the dates into the desired timezone. It's not ideal since you need to duplicate it in up to 6 partials, but it's an option.

nickcharlton:
I'm wondering if the right solution here is to use the timezone of the user who's currently using the dashboard.

Are you providing this as an option? I think that's where I'd start and then from there having Administrate know how to pick that up would be ideal.

yourivdlans:
> I think there is a way: [generate attribute partials](https://administrate-prototype.herokuapp.com/customizing_attribute_partials) for the Date and DateTime fields and use the Rails helpers to convert the dates into the desired timezone. It's not ideal since you need to duplicate it in up to 6 partials, but it's an option.

I think this is the right approach. Personally I don't mind the form being in a different format as the user can pick a date and time here. So that would leave me with 4 partials. If you use a localization file for your date and time formats you could use one of those formats or create your own.

`en.yml`
```yml
en:
  time:
    formats:
      ...
      long: "%B %d, %Y %H:%M"
      ...
```

e.g. `date_time/_index.html.erb`
```erb
<% if field.data %>
  <%= I18n.l(field.data, format: :long) %>
<% end %>
```

PR #3067, opened 2026-09-18
Add guide for date/datetime/time default format

goosys:
I created this documentation because #2781 did not include any.

Please let me know if you have any suggestions for changes to its location, title, or wording.


goosys:
I just came across #1193, and it appears to be resolved by this change. What do you think?

